PR #68, opened 2024-07-17
Allow enlisting bucket that is already in pending state

amritghimire:
Instead of raising PendingIndexing, we can now wait for the indexing to
be over. Once it is indexed or marked as stale, we will continue
accordingly.

Tried simply without raising exception, but we get error without
partials ans well.

Todo: 
- [x] Test cases
- [x] optimize it further 

Closes https://github.com/iterative/dvcx/issues/1528


amritghimire:
Looks like we get few issues like 
```
E             +     ')',
E             +     'CREATE TABLE dqlapp_idgenerator (',
E             +     'DETAIL:  Key (typname, typnamespace)=(dqlapp_idgenerator, 2200) already '
E             +     'exists.',
E             +     'DETAIL:  Key (typname, typnamespace)=(dqlapp_idgenerator, 2200) already '
E             +     'exists.',
E             +     'The above exception was the direct cause of the following exception:',
E             +     'Traceback (most recent call last):',
E             +     'Traceback (most recent call last):',
E             +     '[SQL: ',
E             +     ']',
E             +     'psycopg2.errors.UniqueViolation: duplicate key value violates unique '
E             +     'constraint "pg_type_typname_nsp_index"',
E             +     'sqlalchemy.exc.IntegrityError: (psycopg2.errors.UniqueViolation) '
E             +     'duplicate key value violates unique constraint '
E             +     '"pg_type_typname_nsp_index"
``` 

with database constraints and other clashes. I think we need to revert back to waiting for it to be over. 

We can set some timeout for it to be over before raising error or other safeguards. But essentially, the implementation at 0443cf089584203d5d6fe6adad08201f2985a442 . cc. @shcheklein

shcheklein:
@amritghimire let's pause this then. I don't think we need to invest time into band-aids much tbh. It doesn't solve the issue to my mind. I would rather discuss and see if we can get rid of partials and properly drop these locks. I don't see a good reason for them to exist.

amritghimire:
> @amritghimire let's pause this then. I don't think we need to invest time into band-aids much tbh. It doesn't solve the issue to my mind. I would rather discuss and see if we can get rid of partials and properly drop these locks. I don't see a good reason for them to exist.

Closing this for now.

PR #71, opened 2024-07-17
llm evaluation tutorial

tibor-mach:


mnrozhkov:
Good work @tibor-mach! I fixed a few typos. Also, there are several questions 
1. Is it possible to set the `device` as an argument to run the example on the CPU?
2. Would it be better to rename "pdf_chunks" to something more descriptive, such as "process_pdf"?
3. We have a defined `Chunk` class, but we use an additional `record = {}` object to yield processed chunks. It would be better to yield the `Chunk` object. I couldn't find an appropriate example, though.
4. Let's discuss how we could improve the "Evaluation" section's conclusions. Currently, it only contains a table with metrics. Can we use these metrics to solve a particular practical problem? For example, finding a more appropriate paper or paragraph.

volkfox:
Let's infer the direction from the end goal.
What is it?

Users don't want to code, they want to solve their problems. If we define the problem as finding the right combination of RAG moving parts – e.g. the chunker, the embedder, the LLM, the answer evaluator – that's fine.

However, it is not enough to say 'we have you versioning so be happy' because we need to drive to the result – the RAG accuracy evaluator. There are multiple libraries which do exactly that – like RAGAS. Unless we plan to recreate them, we need to find way to integrate.

tibor-mach:
@mnrozhkov Thanks for the suggestions, I think they are all valid, so I tried to incorporate all of them.

I expanded the evaluation a bit to make it a bit more interesting (while avoiding the use of additional tools). It could be made more robust but I opted for a "where to next" section at the end where I describe what you would do in a more realistic scenario to avoid making the tutorial too heavy (maybe I am worried about it too much, but I think it should be relatively simple).

@volkfox I get your point ... I think we might really be thinking of two different types of examples. The one I have in mind are mostly tutorials - stuff that shows how DataChain works and what you can do with it. FWICT you seem to be mostly interested in creating a recipe - something people can just take and mostly re-use directly with slight modifications to solve a specific use-case.

I think it makes sense to have both....so perhaps I can rename this and we can merge this as a tutorial while we iterate more on the recipe one. For tutorials I would argue that the fewer external tools we use and especially fewer external APIs (particularly paid ones!) we use, the better. For recipes, that is not necessarily the case and I agree that people will start asking things like "how do I use RAGAS with DataChain?".

tibor-mach:
@volkfox @dberenbaum @shcheklein @mnrozhkov I think we can merge this and include it as a part of the readme (and then perhaps update it as we go).

Let me know if you agree (I am waiting for a review). 

Also it seems that we are moving the tutorials outside of this repo, so maybe I should move this PR ... somewhere else ... instead?

shcheklein:
> Also it seems that we are moving the tutorials outside of this repo, so maybe I should move this PR ... somewhere else ... instead?

yes, I think @mnrozhkov should know the new place for it

tibor-mach:
Closing in favour of moving examples to a [dedicated repo](https://github.com/iterative/datachain-notebooks). Added there with [this](https://github.com/iterative/datachain-notebooks/commit/62d84748bf10daf6c1cce98341e92401b9e5836b) commit.

dberenbaum:
@tibor-mach @mnrozhkov It seems we now have both https://github.com/iterative/datachain-notebooks and https://github.com/iterative/datachain-examples. Can we agree on one so we can start moving the notebooks?

volkfox:
it would make sense to have one "examples" repo private to store things related to customers, and one public but this is not how it looks for now.

In any case, I am using 'datachain-examples' for now as it is what we discussed in the call today.

PR #108, opened 2024-07-20
storage: index as a dir if no glob

shcheklein:
Related to:

https://github.com/iterative/dvcx/issues/1488#issuecomment-2139318338
https://github.com/iterative/dvcx/pull/1485#issuecomment-2133400527

(see links and discussion in those threads as well).

Trying to make things like:

```
datachain ls neurips
```

and 

```
datachain ls neurips/
```

behave the same way and avoid indexing the parent directory. I think is the most expected thing from the user perspective. Otherwise a query like `datachain ls neurips` now can index an unpredictable amount of all sorts of files in the parent directory, while still producing the exact same result at the end as `datachain ls neurips/`.

Same goes for `DataChain.from_storage("gs://datachain-demo/50k-laion-files/")` should be the same as `gs://datachain-demo/50k-laion-files` (no slash at the end).

## TODO:

- [x] Check tests
- [x] Add more tests
- [ ] Even with this fix a query like `from_storage("path-to-a-file")` leads to indexing the parent dir. It affects `from_parquet`, etc. Not optimal. Can be fixed in a followup.
- [ ] Review `cp` command semantics, `no_glob` etc
- [ ] Create followup to remove glob support from from_storage. Keep it in `cp`, etc

### DIFF
--- a/src/datachain/catalog/catalog.py
+++ b/src/datachain/catalog/catalog.py
@@ -1,4 +1,5 @@
 import ast
+import glob
 import io
 import json
 import logging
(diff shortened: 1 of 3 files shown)


ilongin (approved):
LGTM

Issue #110, opened 2024-07-20
Schema and optimization handling improvements in from_csv()


volkfox:
### Description

There is a number of improvements that we need to implement at some point:

1. currently, the code does not validate the data against the automatically created schema.  This means the data is lazily loaded onto the chain and blows up later in chain operations when picked with collect() that triggers Pydantic checks. 
For example, check the last example in json-csv-reader.py which looks great until one uses iterate() or collect() on it.

3. currently, the code automatically creates the schema in "strict" format. We need to default to "lax" format marking the fields "optional" instead of mandatory, and give user an option to change this behavior.

4. currently, the code does not allow for passing the static schema. We should permit that because the users may choose to enforce which CSV columns they want and what types those should have.  It is also related to future capabilities of ser/de-ser hierarchical Pydantic objects to and from CSV using the "object__field" column name conventions.

also see: "spec" argument in from_json()

5. currently, the code does not allow for limiting the number of entries generated. This prevents optimizations like "from_csv(blah).limit(N)" to work effectively.

also see: "nrows" argument in from_json()

PR #126, opened 2024-07-22
Arrow improvements

dberenbaum:
Niceties related to #110 that make the example there work better:

- Adds `nrows` to `from_csv/parquet()` to quickly parse only the first few rows
- Validates both input pyarrow schema and output pydantic schema during parsing
- Handles null types so they work with methods like `collect()`


tests/unit/lib/test_datachain.py
```diff
@@ -791,12 +792,26 @@ def test_from_csv_no_header_output_list(tmp_dir, catalog):
...
+    df.to_csv(path, sep="\t", index=False)
     dc = DataChain.from_csv(path.as_uri(), delimiter="\t")
     df1 = dc.select("first_name", "age", "city").to_pandas()
     assert df1.equals(df)
 
 
+def test_from_csv_null_collect(tmp_dir, catalog):
+    # Clickhouse requires setting type to Nullable(Type).
+    # See https://github.com/xzkostyan/clickhouse-sqlalchemy/issues/189.
+    skip_if_not_sqlite()
```
dberenbaum, line 804: I'm not sure how committed we are to clickhouse, so I think it's best to skip this for now in clickhouse and create an issue in studio. It would be a bit painful to implement support since we would have to wrap sql types with `clickhouse_sqlalchemy.types.Nullable` but only in clickhouse and only if the type may contain null values. Since all the types live in the open-source codebase, it would be painful to introduce clickhouse-specific logic.

src/datachain/lib/arrow.py
```diff
@@ -25,24 +28,32 @@ def __init__(
...
     def process(self, file: File):
         path = file.get_path()
-        ds = dataset(path, filesystem=file.get_fs(), schema=self.schema, **self.kwargs)
+        ds = dataset(
+            path, filesystem=file.get_fs(), schema=self.input_schema, **self.kwargs
+        )
         index = 0
         with tqdm(desc="Parsed by pyarrow", unit=" rows") as pbar:
-            for record_batch in ds.to_batches():
+            for record_batch in ds.to_batches(use_threads=False):
```
dberenbaum, line 49: Fixes a bug where this can hang when using threads. Tested that this doesn't hurt performance when reading the 3 million row dataset from `gs://datachain-demo/laion-aesthetics-csv` in full. Thanks to @mattseddon for catching this.

dberenbaum:
@rlamy @ilongin Could one of you take a look when you have a chance?

rlamy (approved):

PR #146, opened 2024-07-23
removed stale examples for now

volkfox:
Dropping all examples except the latest JSON/CSV, Dave's CLIP multimodal and the LLM eval

We will gradually add examples back after re-working them.


mnrozhkov:
@volkfox also the  `fashion_product_images` is actively used to work with customers. Please, provide another CV example if you delete this one.

shcheklein:
Closing this until we sync with @mattseddon @dberenbaum who AFAIU have been improving those examples (and probably had some discussion on which one of those to keep and why). @volkfox if you have specific details on what is considered stale, how good example should look like - please share this, it will help to set the context.

PR #152, opened 2024-07-23
Error handling for type errors in `.mutate()` and tests

ilongin:
When we create new column using `.mutate()` that column type is inferred from the expression itself. The problem is that with some expression, like subtraction or division, this cannot be inferred automatically so `NullType` is set for that column which causes errors along the way (when saving dataset etc.).  

Example:
```python
from sqlalchemy import column
print((column("x") + 5).type)   # Integer()
print((column("x") * 5).type)   # Integer()
print((column("x") - 5).type)   # NullType()
print((column("x") / 5).type)   # NullType()
```

To overcome this, user must cast the type explicitly with `cast(...)` function. 
Example:
```python
from datachain.lib.dc import C, DataChain, cast                                 
from datachain.sql.types import Int

(
    .mutate(size_diff = cast(C("file__size") - 2, Int))
    .save("example")
)
```

This PR added:

- `cast` function to `DataChain` which is just `sqlalchemy.cast`
- better error handling when type failed to infer im `.mutate()`
- test cases with mutate and types

### DIFF
--- a/src/datachain/lib/dc.py
+++ b/src/datachain/lib/dc.py
@@ -54,6 +54,7 @@ if TYPE_CHECKING:
     P = ParamSpec("P")
 
 C = Column
+cast = sqlalchemy.cast
 
 _T = TypeVar("_T")
 D = TypeVar("D", bound="DataChain")
--- a/src/datachain/query/dataset.py
+++ b/src/datachain/query/dataset.py
@@ -33,6 +33,7 @@ from sqlalchemy.sql.elements import ColumnClause, ColumnElement
 from sqlalchemy.sql.expression import label
 from sqlalchemy.sql.schema import TableClause
 from sqlalchemy.sql.selectable import Select
+from sqlalchemy.sql.sqltypes import NullType
 
 from datachain.asyn import ASYNC_WORKERS, AsyncMapper, OrderedMapper
 from datachain.catalog import (
@@ -1390,6 +1391,13 @@ class DatasetQuery:
             >>> ds.mutate(size10x=C.size * 10).order_by(C.size10x).results()
         """
         query_args = [v.label(k) for k, v in dict(args, **kwargs).items()]
+        for arg in query_args:
+            if isinstance(arg.type, NullType):
+                raise TypeError(
+                    f"Cannot infere type for {arg.name}, use cast(..) function"
+                    " to explicitly set correct type"
+                )
+
         query = self.clone()
         query.steps.append(SQLMutate((*query_args,)))
         return query
--- a/tests/func/test_dataset_query.py
+++ b/tests/func/test_dataset_query.py
@@ -14,7 +14,7 @@ import numpy as np
 import pytest
 import sqlalchemy
 from dateutil.parser import isoparse
-from sqlalchemy import tuple_
+from sqlalchemy import cast, tuple_
 
 from datachain.catalog import QUERY_SCRIPT_CANCELED_EXIT_CODE
 from datachain.dataset import DatasetDependencyType, DatasetStatus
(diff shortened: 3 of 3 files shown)


dmpetrov:
```
    .mutate(size_diff = cast(C("file__size") - 2, Int))
```

cast in the user code does not look good - does not seems like a solution. Can we use this inside mutate() somehow?

ilongin:
> ```
>     .mutate(size_diff = cast(C("file__size") - 2, Int))
> ```
> 
> cast in the user code does not look good - does not seems like a solution. Can we use this inside mutate() somehow?

We can call `cast` inside `.mutate()` but I don't think we can infer type for every expression that might come in. I'm sure there is a reason why SQLAlchemy cannot do it even for a simple subtraction so I think, at least for now, we need type information in `.mutate()` for those expression that fail to infer.

I'm not sure what is the best way to send it to `.mutate()` though...maybe values of  kwargs could be expressions **or** tuple of expression and type where needed.
Example:
```python
.mutate(with_explicit_type = (C("file__size") - 2, Int))  # we explicitly send type
.mutate(without_explicit_type = C("file__size")  + 2)   # we don't send type as it's not needed in this case
```
But this also doesn't look ideal to me ...

dberenbaum:
If we can't infer the types, one option is to make `mutate()` syntax similar to `map()`-like methods that optionally define output types like `.mutate(abs_difference = func.abs(C.query_sim_old - C.query_sim_new)), output=float)`.

skshetry:
`mutate` is also a bad API name, it's not mutating anything, is it?

ilongin:
> `mutate` is also a bad API name, it's not mutating anything, is it?

I have the same feeling. We do mutate columns but still, I would call it with more explicit name. Even `add_signal()` is better IMO

tibor-mach:
> > `mutate` is also a bad API name, it's not mutating anything, is it?
> 
> I have the same feeling. We do mutate columns but still, I would call it with more explicit name. Even `add_signal()` is better IMO

Absolutely. basically it does the same thing as `.withColumn` in Spark (which is not a great name either, but somewhat better). I like `add_signal`. Or even `add_cols` or something like that.

ilongin:
> If we can't infer the types, one option is to make `mutate()` syntax similar to `map()`-like methods that optionally define output types like `.mutate(abs_difference = func.abs(C.query_sim_old - C.query_sim_new)), output=float)`.

`.mutate()` is little bit different in a sense that it accepts multiple columns and their expressions, one for each column (that's why it's signature is currently just `**kwargs`), while `map`-like functions accept one function and `output` defines output columns and their types.

I'm not sure how we can just introduce `output` to `.mutate()` except maybe change columns and their expressions to be a `dict` and set output types as separate argument but that looks strange to me (because of that separation and defining columns 2 times), e.g
```python
.mutate({"col1": C("file__size")  + 2,  "col2": C("file__size")  / 2}. output={"col1": Integer})
```

One other option is to limit number of colums that can be generated with `.mutate()` to only one, then your suggestion would work with slight modification, e.g:
 ```python
.mutate("abs_difference", func.abs(C.query_sim_old - C.query_sim_new), output=float)
```
I actually think this is maybe the best option as I don't see that much of a value of being able to add multiple of them in one method call....and chaining method calls, one for each new columns looks even cleaner to me. Also with this signature it's even more clearer that we need to rename `.mutate`

**Comment by dmpetrov** [comment]
We should consider how to implement this in a more user-friendly way.

One possible approach is: https://github.com/iterative/dvcx/issues/1637 However, I hope there is a way to solve it within sqlalchemy.

**Comment by dmpetrov** [comment]
> ```python
> from sqlalchemy import column
> print((column("x") + 5).type)   # Integer()
> print((column("x") * 5).type)   # Integer()
> print((column("x") - 5).type)   # NullType()
> print((column("x") / 5).type)   # NullType()
> ```

I was unable to confirm that this is an issue (see in #183):
```python
def test_column_math():
    fib = [1, 1, 2, 3, 5, 8]
    chain = DataChain.from_values(num=fib)

    ch = chain.mutate(add2=Column("num") - 2)
    assert list(ch.collect("add2")) == [x - 2 for x in fib]

    ch = chain.mutate(div2=Column("num") / 2.0)
    assert list(ch.collect("div2")) == [x / 2.0 for x in fib]

    ch2 = ch.mutate(x=1 - Column("div2"))
    assert list(ch2.collect("x")) == [1 - (x / 2.0) for x in fib]
```

dmpetrov:
I see two ways of solving this issue and we need both of the method (1 is a prerequisite to 2):

1. Getting typed-columns from chain.

```python
chain =  DataChain.from_storage("s3://ldb-public/remote/data-lakes/dogs-and-cats/", anon=True)

(  chain
    .mutate(size_diff = chain.column("file.size") - 2)  # DataCHain.column() resolves columns AND types
    .save("example")
)
```

2. 
Do this (1) conversion under the hood.

mutate() should get this no-typed expression and convert it to typed (as in (1)).
```python
chain =  DataChain.from_storage("s3://ldb-public/remote/data-lakes/dogs-and-cats/", anon=True)

(  chain
    .mutate(size_diff = Column("file.size") - 2)
    .save("example")
)
```

dmpetrov:
sorry, I accidentally closed it - fat finger

ilongin:
Closing in favor of https://github.com/iterative/datachain/pull/226

PR #154, opened 2024-07-23
Fix studio tests

dreadatour:


shcheklein:
Doesn't look like a fix, but let's do this if it works for now

PR #155, opened 2024-07-23
Update imports & instructions for 3-train-model.py

mnrozhkov:


examples/computer_vision/fashion_product_images/scripts/3-train-model.py
```diff
@@ -1,82 +1,13 @@
...
-        num_classes (int): Number of classes in the dataset.
-        num_epochs (int, optional): Number of epochs to train the model. Defaults to 20.
-        lr (float, optional): Learning rate for the optimizer. Defaults to 0.001.
-
-    Returns:
-        tuple: A tuple containing the trained model and the optimizer used for training.
-
-    """
+# For Local Development
+from src.train import train_model, transform
```
mattseddon, line 7: Have you run the script locally?

mnrozhkov: yes, at least to test it works.

mattseddon: I get this: 

```
❯ python examples/computer_vision/fashion_product_images/scripts/3-train-model.py
  File "/datachain/examples/computer_vision/fashion_product_images/scripts/3-train-model.py", line 7, in <module>
    from src.train import train_model, transform
ModuleNotFoundError: No module named 'src'
```

mnrozhkov: Ok, I got it!  To test, I run scripts from inside the `fashion_product_images/` directory. 
I don't expect users will run scripts outside the Studio.

mattseddon: We want to keep all the scripts working and usable for anyone who stumbles across them. 

I am working on adding all of the scripts to a test suite that will run nightly. If we use `nbconvert` to create an exact (script) copy of each notebook we can verify that all steps run without error.

My $0.02 is that the duplication isn't that bad and having all of the code exposed and visible (as opposed to hidden) would be a better idea for tutorial purposes. This also keeps the instructions for running things in Studio simple(r). For example, we have copies of the same `CNN` across different scripts already (`examples/get_started/torch-loader.py` / `examples/computer_vision/fashion_product_images/src/train.py`.) and it doesn't seem to be a problem.

mnrozhkov: @mattseddon, let's sync on this testing. May be a short call will be faster. 
I'm also have a solution for nightly tests, but with `papermill` :))) I'll share with you, let's discuss then

mnrozhkov: I have an additional requirement here - I need to keep Studio examples somewhere and make it as simple as possible to copy-paste them into Studio (along with additional imported modules). 
- that's why we have `scripts` in `fashion...` examples dir 
- they are in the Studio format 
- the `train.py` is designed intentionally to be attached to Studio 
- to test them I use `datachain query ...` CLI, not `python  ... `

mattseddon: no worries, you can maintain the examples. I'll close my PRs

mattseddon: Please tag me in the papermill PR. I'm interested to see how that works.

Issue #103, opened 2024-07-19
`subtract()` and maybe other methods relying on old file signals are broken


dberenbaum:
### Description

Example to reproduce:

```python
>>> from datachain import C, DataChain
>>> source = "gs://dvcx-datalakes/dogs-and-cats/"
>>> chain = DataChain.from_storage(source)
>>> chain.subtract(chain.limit(5)).collect()
Listing gs://dvcx-datalakes: 400 objects [00:00, 1214.35 objects/s]
Processed: 400 rows [00:00, 13201.67 rows/s]
Processed: 400 rows [00:00, 25120.48 rows/s]
Traceback (most recent call last):
  File "/Users/dave/micromamba/envs/dvcx-py311/lib/python3.11/site-packages/sqlalchemy/sql/base.py", line 1617, in __getattr__
    return self._index[key][1]
           ~~~~~~~~~~~^^^^^
KeyError: 'source'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/dave/Code/datachain/src/datachain/lib/dc.py", line 774, in collect
    return list(self.iterate(*cols))
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dave/Code/datachain/src/datachain/lib/dc.py", line 757, in iterate
    with super().select(*db_signals).as_iterable() as rows:
  File "/Users/dave/micromamba/envs/dvcx-py311/lib/python3.11/contextlib.py", line 137, in __enter__
    return next(self.gen)
           ^^^^^^^^^^^^^^
  File "/Users/dave/Code/datachain/src/datachain/query/dataset.py", line 1251, in as_iterable
    query = self.apply_steps().select()
            ^^^^^^^^^^^^^^^^^^
  File "/Users/dave/Code/datachain/src/datachain/query/dataset.py", line 1191, in apply_steps
    result = step.apply(
             ^^^^^^^^^^^
  File "/Users/dave/Code/datachain/src/datachain/query/dataset.py", line 278, in apply
    diff_q = self.query(source_query, target_query)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dave/Code/datachain/src/datachain/query/dataset.py", line 310, in query
    return self.catalog.warehouse.subtract_query(source_query, target_query)
[...]

PR #157, opened 2024-07-23
Update DataChain.subtract() to work without legacy file signals

rlamy:
Partial fix for #103. Also includes a unit test for `DataChain.union()` which was mentioned there but appears to work.

Note that `chain1.subtract(chain2)` doesn't actually work for real datasets, even though the unit tests pass. The workaround in to use `on` as in `new.subtract(old, on=["file.name"])`.

### DIFF
--- a/src/datachain/lib/dc.py
+++ b/src/datachain/lib/dc.py
@@ -951,6 +951,41 @@ class DataChain(DatasetQuery):
 
         return ds
 
+    def subtract(  # type: ignore[override]
+        self,
+        other: "DataChain",
+        on: Optional[Union[str, Sequence[str]]] = None,
+    ) -> "Self":
+        """Remove rows that appear in another chain.
+
+        Parameters:
+            other: chain whose rows will be removed from `self`
+            on: columns to consider for determining row equality. If unspecified,
+                defaults to all common columns between `self` and `other`.
+        """
+        if isinstance(on, str):
+            on = [on]
+        if on is None:
+            other_columns = set(other._effective_signals_schema.db_signals())
+            signals = [
+                c
+                for c in self._effective_signals_schema.db_signals()
+                if c in other_columns
+            ]
+            if not signals:
+                raise DataChainParamsError("subtract(): no common columns")
+        elif not isinstance(on, Sequence):
+            raise TypeError(
+                f"'on' must be 'str' or 'Sequence' object but got type '{type(on)}'",
+            )
+        elif not on:
+            raise DataChainParamsError(
+                "'on' cannot be empty",
+            )
+        else:
+            signals = self.signals_schema.resolve(*on).db_signals()
+        return super()._subtract(other, signals)
+
     @classmethod
     def from_values(
         cls,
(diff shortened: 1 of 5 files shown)


dmpetrov (approved):
Thank you @rlamy! I really like how's done - more on app level.

src/datachain/query/dataset.py
```diff
@@ -292,23 +292,21 @@ def q(*columns):
...
-        return self.catalog.warehouse.subtract_query(source_query, target_query)
+        sq = source_query.alias("source_query")
+        tq = target_query.alias("target_query")
+        on_clause = sqlalchemy.and_(
+            getattr(sq.c, col_name) == getattr(tq.c, col_name) for col_name in self.on
+        )  # type: ignore[arg-type]
+        outer_join = sqlalchemy.join(sq, tq, on_clause, isouter=True)
+
+        where_clause = sqlalchemy.and_(
+            getattr(tq.c, col_name) == None  # noqa: E711
```
ilongin, line 306: Because of this test fails in SaaS as in Clickhouse we don't have `null`-able columns so value of empty column is empty value of a specific type (for `str` is `""`).  We do have a static method to get default values for each sql type, e.g:
```python
from datachain.sql.types import String
String.default_value()  # this will be None in CLI and "" in SaaS
```
I'm not sure if it's easy to get these SQL types from signals schema though ... if we have python types only then we prob have a map somwehere between python types and SQL types that can be used.

ilongin (changes requested):

dberenbaum (approved):
Looks like it has the same issue as #126 on studio but otherwise LGTM

rlamy:
The new implementation (`return sq.select().except_(sq.select().where(where_clause))`) avoids issues with non-nullable columns and also with NULL values, thanks to `IS NOT DISTINCT FROM`. That fixes `chain1.subtract(chain2)` in all cases, so no workaround is needed.
It works both on sqlite and ClickHouse. Performance seems decent on sqlite, but I didn't check CH.

dberenbaum (approved):

src/datachain/lib/dc.py
```diff
@@ -951,6 +951,41 @@ def merge(
 
         return ds
 
+    def subtract(  # type: ignore[override]
+        self,
+        other: "DataChain",
+        on: Optional[Union[str, Sequence[str]]] = None,
```
dmpetrov, line 957: So, for the files it will be like:
```
diff = dc1.subtract(dc2, ["file.source" "file.name", "file.etag"])
```

QUestions:
1) Is it possible to handle use cases when file name do not match? Like: `"file.name" == "wds.laion.file.name" AND "file.etag" == "wds.laion.file.etag"`

2) Is it possible to reuse the existing merge or join to get the diff/subtract?

dmpetrov (changes requested):
@rlamy thank you for the change and make it it work everywhere!

Please answer questions - it would be great to know more.

ilongin (approved):

dmpetrov (approved):
Looks good! Thank you @rlamy 

Could you please create a followup issue for `right` to make sure we can handle different-files use case properly.

Also, we can consider implementing this using merge/join. We might need an issue as well.

Issue #36, opened 2024-07-01
Redesign UniqueId


dmpetrov:
- [ ] Fit UniqueId to File/pydentic class hierarchy
    - Let's make something like `class BaseFile(BaseModel)` instead of `UniqueId`. `File` should be just `class File(BaseFile)`.
    - Another alternative: Fit everything to `File` class and just stop using UniqueId.
- [x] stop using `vtype`
- [ ] stop making location as String. It has to be list[dict] (or dict - depending on implementation)
- [x] add `version` to the hash
- [x] Remove `dir_type`

Note 1, let's not use `owner` signals at this moment.
Note 2, in case `catalog` class is needed, more general `session` should be used (which contains catalog inside).

It's a part of iterative/dvcx#1635

PR #165, opened 2024-07-24
DataChain.from_storage: add last_modified and is_latest to the columns

### LINKED ISSUE #163: File modification time does not work
### Description

It looks like last modificaiont time does not work. I always see something like `1970-01-01 00:00:00+00:00`

### Version Info

_No response_

skshetry:
Fixes #163.

I have also changed the order of some fields in `File`.

### DIFF
--- a/src/datachain/lib/file.py
+++ b/src/datachain/lib/file.py
@@ -20,7 +20,7 @@ from datachain.cache import UniqueId
 from datachain.client.fileslice import FileSlice
 from datachain.lib.data_model import DataModel
 from datachain.lib.utils import DataChainError
-from datachain.sql.types import JSON, Int, String
+from datachain.sql.types import JSON, Boolean, DateTime, Int, String
 from datachain.utils import TIME_ZERO
 
 if TYPE_CHECKING:
@@ -126,11 +126,13 @@ class File(DataModel):
         "source": String,
         "parent": String,
         "name": String,
+        "size": Int,
         "version": String,
         "etag": String,
-        "size": Int,
-        "vtype": String,
+        "is_latest": Boolean,
+        "last_modified": DateTime,
         "location": JSON,
+        "vtype": String,
     }
 
     _unique_id_keys: ClassVar[list[str]] = [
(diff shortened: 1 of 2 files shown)


src/datachain/lib/file.py
```diff
@@ -115,22 +115,24 @@ class File(DataModel):
...
         "name": String,
-        "version": String,
-        "etag": String,
         "size": Int,
+        "etag": String,
+        "version": String,
+        "is_latest": Boolean,
         "vtype": String,
         "location": JSON,
+        "last_modified": DateTime,
```
skshetry, line 135: Should we add `owner_id` and `owner_name` as well?

src/datachain/lib/file.py
```diff
@@ -115,22 +115,24 @@ class File(DataModel):
...
     size: int = Field(default=0)
-    version: str = Field(default="")
     etag: str = Field(default="")
+    version: str = Field(default="")
     is_latest: bool = Field(default=True)
-    last_modified: datetime = Field(default=TIME_ZERO)
-    location: Optional[Union[dict, list[dict]]] = Field(default=None)
     vtype: str = Field(default="")
+    location: Optional[Union[dict, list[dict]]] = Field(default=None)
+    last_modified: datetime = Field(default=TIME_ZERO)
```
rlamy, line 123: Changing the order of fields is (to some extent) a compatibility break. I wouldn't change it unless we have a good reason.

skshetry: I would not worry about compatibility yet. Also, we are on zero-ver, so breakages are expected.

Anyway, I'll change the `_datachain_column_types` order to keep them in sync.

src/datachain/lib/file.py
```diff
@@ -375,6 +378,7 @@ def get_file_type(
             size=size,
             vtype=vtype,
             location=location,
+            last_modified=last_modified,
```
rlamy, line 381: `is_latest` needs to be passed in as well. (it's never False in tests because we're not testing cloud versioning enough)

src/datachain/lib/file.py
rlamy: TBH, I'm not really sure what this change does. `File._datachain_column_types` is used in a few unrelated places for things that seem mostly irrelevant.
BTW, `owner_id` and `owner_name` are deprecated and should be removed soon, see #36.

skshetry: By default, DataChain starts with the signal schema from `File._datachain_column_types`.

https://github.com/iterative/datachain/blob/f7fdf6b70f323c74f6bbf8bfb9ae08b66755d00c/src/datachain/lib/dc.py#L209

https://github.com/iterative/datachain/blob/f7fdf6b70f323c74f6bbf8bfb9ae08b66755d00c/src/datachain/query/dataset.py#L1102


https://github.com/iterative/datachain/blob/f7fdf6b70f323c74f6bbf8bfb9ae08b66755d00c/src/datachain/lib/dc.py#L218


So, this is just adding them to `SignalSchema` so that `map` can use it to pass to the function.

PR #171, opened 2024-07-24
Fix for using new column from `.mutate()` in `.order_by()`

ilongin:
This adds small fix for using new column added in `.mutate()` in `.order_by(). 

Example:
```python

(
    .mutate(size_10x = C("file__size") * 10)
    .order_by("size_10x")
    .save("example")
)
```


dtulga (approved):
LGTM, thanks for fixing this!

PR #172, opened 2024-07-24
Renaming `File.write()` to `File.save()`

ilongin:
Simple rename of `.write()` to `.save()` to not break built-in `io.FileIO` interface which our `File` should follow.


dtulga (approved):
LGTM, thanks!

PR #174, opened 2024-07-25
ci: disable azure and gs remote tests on macOS

skshetry:
The CI is flaky because of colima/docker usage, which does not seem to be stable on macOS. Also, we already run those tests on Linux, so we don't really need to run it on macOS.

Closes #51.

PR #175, opened 2024-07-25
ci: run s3 tests on Windows, be more careful while skipping

skshetry:
Follow up to #174.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -92,7 +92,7 @@ jobs:
 
       - name: Skip all remotes on Windows
         if: runner.os == 'Windows'
-        run: echo 'DISABLE_REMOTES_ARG=--disable-remotes=all' >> $env:GITHUB_ENV
+        run: echo 'DISABLE_REMOTES_ARG=--disable-remotes=azure,gs' >> $env:GITHUB_ENV
 
       - name: Run tests
         run: nox -s tests-${{ matrix.pyv }} -- $DISABLE_REMOTES_ARG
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -92,7 +92,7 @@ jobs:
 
       - name: Skip all remotes on Windows
         if: runner.os == 'Windows'
-        run: echo 'DISABLE_REMOTES_ARG=--disable-remotes=all' >> $env:GITHUB_ENV
+        run: echo 'DISABLE_REMOTES_ARG=--disable-remotes=azure,gs' >> $env:GITHUB_ENV
```
skshetry, line 95: I forgot that we can also run `s3` tests runs on Windows.

tests/conftest.py
```diff
@@ -384,7 +387,7 @@ def pytest_collection_modifyitems(config, items):
     for item in items:
         if "cloud_server" in item.fixturenames:
             cloud_type = item.callspec.params.get("cloud_type")
-            if cloud_type == "file":
+            if cloud_type not in cloud_types:
```
skshetry, line 390: Trying to be more defensive, so that we don't accidentally skip tests that we should not have.

.github/workflows/tests.yml
skshetry: Maybe that's what causes performance issues. `azure` and `gs` are skipped as there's no docker in Windows.

0x2b3bfa0 (approved):

.github/workflows/tests.yml
skshetry: Yup, s3 tests slows down Windows tests.

PR #176, opened 2024-07-25
avoid instantiating filesystem for path operations

skshetry:
`S3Client.create_fs()` tries to sign an s3 uri, which might hit the remote. So, all of the current
`get_destination_path` and `get_path` tests are
hitting AWS S3, which might fail if you don't have credentials set.

https://github.com/iterative/datachain/blob/0ed1e2cf2e66ba1913301c81b0a907941b0c6953/src/datachain/client/s3.py#L46

So this PR tries to avoid constructing them wherever possible (at least in these path operations).

An alternative fix might be to run tests inside `s3_server`.

Also removes one test that hits gcs.

### DIFF
--- a/src/datachain/lib/file.py
+++ b/src/datachain/lib/file.py
@@ -12,7 +12,6 @@ from urllib.parse import unquote, urlparse
 from urllib.request import url2pathname
 
 from fsspec.callbacks import DEFAULT_CALLBACK, Callback
-from fsspec.implementations.local import LocalFileSystem
 from PIL import Image
 from pydantic import Field, field_validator
 
--- a/src/datachain/lib/file.py
+++ b/src/datachain/lib/file.py
@@ -283,9 +283,8 @@ class File(DataModel):
     def get_path(self) -> str:
         """Returns file path."""
         path = unquote(self.get_uri())
-        fs = self.get_fs()
-        if isinstance(fs, LocalFileSystem):
-            # Drop file:// protocol
+        source = urlparse(self.source)
+        if source.scheme == "file":
             path = urlparse(path).path
             path = url2pathname(path)
         return path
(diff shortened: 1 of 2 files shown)


src/datachain/lib/file.py
```diff
@@ -283,9 +282,8 @@ def get_uri(self):
     def get_path(self) -> str:
         """Returns file path."""
         path = unquote(self.get_uri())
-        fs = self.get_fs()
-        if isinstance(fs, LocalFileSystem):
-            # Drop file:// protocol
+        source = urlparse(self.source)
+        if source.scheme == "file":
```
skshetry, line 286: @ilongin, is there any reason why we keep drive in `source`? I think all of this would be much simpler if we had everything we need in `parent + name` (well, soon that will be called `path`).

skshetry: Also, if we want to support s3-compatible storage and use of multiple storages at the same time, we need to also preserve `fsid` or endpointurl. Otherwise, how will we know if the path is from AWS s3 or some compatible storage?

skshetry:
Need to see if Windows test passes.

src/datachain/lib/file.py
skshetry, line 286: We always have `self.source` set, right?

rlamy: Normally yes - but if not, `get_fs()` would fail anyway.

rlamy (approved):

src/datachain/lib/file.py
ilongin: Having `file:///` as source was part of the solution to remove old  `add_storage` method that was needed to be called before being able to index local FS, and to have things like partial indexing for local FS indexes (avoid re-indexing already indexed folders). The idea was to look local FS similar as one bucket and every directory in it is similar to directories in cloud buckets. This way every time someone tries to index some dir, it will create another partial index and can check if that dir is already indexed in some other partial index. Also, with this user will have only one entry in `buckets` table for local FS and that is the root of the system itself (`file:///`).

ilongin (approved):
LGTM, thanks for the improvements

src/datachain/lib/file.py
skshetry: Sorry, I should have been clearer. We have `file:///<drive>/` (`file:///C:/`) as source for Windows.

PR #177, opened 2024-07-25
drop owner_id and owner_name signals

skshetry:


rlamy:
LGTM, however datachain_server needs to be updated as well. Also, I'm wondering about the consequences in terms of backward compatibility: how does this affect an existing datachain install?

dmpetrov:
Let's merge this together with the path change. So, it will breaks compatibility only once.

Ps: yeah, server fixes

skshetry:
> LGTM, however datachain_server needs to be updated as well. 

I have an open PR there.

skshetry:
> LGTM, however datachain_server needs to be updated as well. Also, I'm wondering about the consequences in terms of backward compatibility: how does this affect an existing datachain install?

That is difficult to say. In a few places, we do `select()` all columns and pass it to `Node` directly, so that's where things might fail.

rlamy:
This PR should probably be rebased on top of #263 so we can merge all the breaking changes together.

PR #183, opened 2024-07-26
from_values with array of arrays

dmpetrov:
Just a small fix with more unit tests

### DIFF
--- a/src/datachain/lib/convert/values_to_tuples.py
+++ b/src/datachain/lib/convert/values_to_tuples.py
@@ -71,7 +71,10 @@ def values_to_tuples(  # noqa: C901, PLR0912
                     f"signal '{k}' has unsupported type '{typ.__name__}'."
                     f" Please use DataModel types: {DataTypeNames}",
                 )
-            types_map[k] = typ
+            if typ is list:
+                types_map[k] = list[type(v[0][0])]
+            else:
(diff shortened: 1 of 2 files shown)


src/datachain/lib/convert/values_to_tuples.py
```diff
@@ -71,7 +71,10 @@ def values_to_tuples(  # noqa: C901, PLR0912
                     f"signal '{k}' has unsupported type '{typ.__name__}'."
                     f" Please use DataModel types: {DataTypeNames}",
                 )
-            types_map[k] = typ
+            if typ is list:
+                types_map[k] = list[type(v[0][0])]  # type: ignore[misc]
+            else:
+                types_map[k] = typ
```
dberenbaum, line 77: WDYT about moving this whole type mapping logic into a small helper function? `values_to_tuples()` is getting long, and it would allow for doing the list check recursively so this doesn't break for deeper nested lists.

dmpetrov: yeah, this function needs redesign. I'd do that as a separate PR.

dberenbaum (approved):
LGTM as long as we are aware that it will still break for deeper nested lists like `[[[1,2,3]]]`.

PR #184, opened 2024-07-26
fixing a missed import for codegen schemas

volkfox:


src/datachain/lib/dc.py
```diff
@@ -342,7 +342,7 @@ def from_json(
         spec: Optional[DataType] = None,
         schema_from: Optional[str] = "auto",
         jmespath: Optional[str] = None,
-        object_name: str = "",
+        object_name: Optional[str] = "",
```
dmpetrov, line 345: Please not make it optional. An empty value is enough to identify the missing value.

dmpetrov (approved):

src/datachain/lib/dc.py
skshetry: FYI `Optional` does not mean the argument is optional. It just means it takes either of. `None` value or some other value.

Issue #170, opened 2024-07-24
Inefficency - no "batched" Inference, and some major 


michaelfeil:
I noticed a number of various things are incorrectly implemented.

```python
classifier = pipeline("sentiment-analysis", device="cpu",
                model="distilbert/distilbert-base-uncased-finetuned-sst-2-english")

def is_positive_dialogue_ending(file) -> bool:
    dialogue_ending = file.read()[-512:] # 512 chars != 512 tokens
    return classifier(dialogue_ending)[0]["label"] == "POSITIVE" # NOT performing batched inference, performs threaded, blocking single inference -> really bad once you are on GPU.
```

```python
classifier = pipeline("sentiment-analysis", device="cpu",  
                # tokenizer from left 
                truncation="right/left", # sets truncation from the right side, I think this is not possible
                model="distilbert/distilbert-base-uncased-finetuned-sst-2-english")

def is_positive_dialogue_ending(files: list) -> list[bool]:
    dialogue_endings = [file.read()for file in files] # something ike this
    return (label  == "POSITIVE" for label in  classifier(dialogue_endings ))  # something like this
```

I launched https://github.com/michaelfeil/embed and https://github.com/michaelfeil/infinity for sentence classification. 

```python
from embed import BatchedInference
from concurrent.futures import Future

def is_positive_dialogue_ending(file) -> bool:
    dialogue_ending = file.read()[-512:] # 512 chars != 512 tokens
    return classifier(dialogue_ending)[0]["label"] == "POSITIVE" # NOT performing batched inference, performs threaded, blocking single inference -> really bad once you are on GPU.

# Run any model
register = BatchedInference(
    model_id=[
        # classification models
        "distilbert/distilbert-base-uncased-finetuned-sst-2-english")
    ],
    # engine to `torch` or `optimum`
    engine="torch",
    # device `cuda` (Nvidia/AMD) or `cpu`
    device="cpu",
)


def is_positive_dialogue_ending(file) -> bool:
    """not multiprocessing recommended, but neither is transformers. Threading is encouraged."
    dialogue_ending = file.read()[-512:] # 512 chars != 512 tokens
    future: "Future" = register.classify(model_id="philschmid/tiny-bert-sst2-distilled", sentences=[dialogue_ending])
    # best: defer to later stage
    return future.result()[0]["label"] == "POSITIVE"
```

volkfox:
need to fix batching ASAP for many similar HuggingFace models to work

dmpetrov:
[...]

PR #191, opened 2024-07-26
Support for `DataChain.batch_map()`

### LINKED ISSUE #84: `batch` and `batch_map` are broken
### Description

`batch` and `batch_map` do not work and it's not clear what the syntax should be.

Take this example:

```python
DataChain.from_storage(path="gs://dvcx-datalakes/dogs-and-cats/").settings(
        batch=10
).map(
        lambda file: len(file.parent + file.name),
        # lambda files: [len(file.parent + file.name) for file in files],
        params=["file"],
        output={"path_len": int}
).show()
```

It fails with error `TypeError: DatasetQuery.add_signals() got an unexpected keyword argument 'batch'`. Do we want to be passing the batch size to the non-batched `map()` method, or should we always assume batch size is 1 here? If `batch` > 1, should the udf expect batched inputs and outputs?

Do we need support for this and `batch_map()`? It's not clear what each one should do.

Note that `batch_map()` now looks like it's just a copy of `gen()` and fails with the same error.

dberenbaum:
Fixes #84.

I think we should keep open #170. That request seems to be specifically about using futures to batch individual results using the existing `.map` without needing a separate `.batch_map()`. I think `.batch_map()` may be both simpler to implement and explain for now, but I think we could come back to the ideas in #170 in the future.

### DIFF
--- a/docs/references/udf.md
+++ b/docs/references/udf.md
@@ -13,6 +13,8 @@ steps need to happen before or after the processing function runs.
 
 ::: datachain.lib.udf.Aggregator
 
+::: datachain.lib.udf.BatchMapper
+
 ::: datachain.lib.udf.Generator
 
 ::: datachain.lib.udf.Mapper
--- a/src/datachain/lib/dc.py
+++ b/src/datachain/lib/dc.py
@@ -238,7 +238,6 @@ class DataChain(DatasetQuery):
     def settings(
         self,
         cache=None,
-        batch=None,
         parallel=None,
         workers=None,
         min_task_size=None,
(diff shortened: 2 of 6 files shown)


dmpetrov (approved):
Great ✨  Thank you Dave for the change!

Shortening the param name - good idea!

rlamy (changes requested):
This conflates 2 settings that I consider logically unrelated:
1. The size of the input row batches passed to a batched UDF
2. The size of the buffer holding output rows to insert in the warehouse DB

Case 1 needs to be under user control, but case 2 is more of an implementation detail that mostly depends on the warehouse engine. Also, the default value seems OK for case 2 but way too high for case 1.

Exposing case 2 doesn't seem that useful, but it doesn't hurt - as long as it's not conflated with 1. I'd suggest calling it "insert_buffer_size" and keeping "batch" for case 1.

dberenbaum:
Thanks @rlamy.

I have completely dropped `batch` from `DataChain.settings` since I agree this is not nearly as important and could be confusing alongside `batch_map()`. I was trying to find some usefulness for this already existing setting but in the end I think it's not needed right now.

rlamy (approved):

PR #192, opened 2024-07-27
fix nested list of pydentics. close #182

dmpetrov:

Issue #182, opened 2024-07-25
Object serialization fails on List of List


volkfox:
### Description

Here is the desired schema:

```
from datachain.lib.dc import Column, DataChain
DataChain.from_storage("gs://datachain-demo/openimages-jsonl/open_images_validation_localized_narratives.jsonl").show_jsonl_schema(model_name="Narrative").exec()
```

```
Processed: 1 rows [00:00, 231.44 rows/s]
Download: 19.7kB [00:00, 202kB/s]
# generated by datamodel-codegen:
#   filename:  <stdin>
#   timestamp: 2024-07-25T22:55:08+00:00

from __future__ import annotations

from typing import List

from pydantic import BaseModel


class TimedCaptionItem(BaseModel):
    utterance: str
    start_time: float
    end_time: float


class Trace(BaseModel):
    x: float
    y: float
    t: float


class Narrative(BaseModel):
    dataset_id: str
    image_id: str
    annotator_id: int
    caption: str
    timed_caption: List[TimedCaptionItem]
    traces: List[List[Trace]]
    voice_recording: str


from datachain.lib.data_model import DataModel


DataModel.register(Narrative)


spec=Narrative


Download: 30.6kB [00:00, 51.1kB/s]]
Processed: 1 rows [00:00, 344.33 rows/s]
<datachain.lib.dc.DataChain at 0x13d3ba5b0>
```

If we execute the above, we will have a registered Pydantic class "Narrative".
It successfully validates the manually input data, but fails to put it on the chain:

```
[...]

PR #195, opened 2024-07-29
Fix for nested lists of models in schema

ilongin:
The problem was if we had `list[list[SomeModel]]` as a field in our custom model as flattening worked only for one level of lists with models.

This PR fixes `flatten(obj)` function to be able to flatten any number of nested lists and adds tests.

### DIFF
--- a/src/datachain/lib/convert/flatten.py
+++ b/src/datachain/lib/convert/flatten.py
@@ -41,21 +41,22 @@ def flatten_list(obj_list):
     )
 
 
-def _flatten_fields_values(fields, obj: BaseModel):
-    def _flatten_list(value):
-        assert isinstance(value, list)
-        if value and ModelStore.is_pydantic(type(value[0])):
-            return [val.model_dump() for val in value]
-        if value and isinstance(value[0], list):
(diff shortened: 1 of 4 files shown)


dtulga (commented):
This looks good to me, but I see that the `test_custom_model_with_nested_lists` test is failing in ClickHouse with Pydantic ValidationErrors. That would ideally be fixed before merging.

src/datachain/lib/convert/flatten.py
```diff
@@ -42,16 +42,20 @@ def flatten_list(obj_list):
 
 
 def _flatten_fields_values(fields, obj: BaseModel):
+    def _flatten_list(value):
```
dmpetrov, line 45: Please extract this func from func.

dmpetrov (approved):
Is this an alternative implementation to #192?

ilongin:
> Is this an alternative implementation to #192? Please make sure it solves #182 and close 192.

Yes, I somehow missed your PR and saw that there was no connected PR to the issue itself and issue was unassigned so I took it.

PR #196, opened 2024-07-29
Ignore Incorrect ClickHouse Warning

dtulga:
This ignores SQLAlchemy warnings about cartesian products that only occur in one test and only on ClickHouse, as after some investigation it appears that these warnings are not accurate, and are generated due to a bug in ClickHouse-SQLAlchemy. I have checked all generated queries, and they do have `ON` clauses as expected, and this warning is not generated in SQLite. (And as such this ignores only on ClickHouse, not on SQLite.)

This is a followup to #158 and has a companion Studio/ClickHouse PR here: https://github.com/iterative/studio/pull/10376

### DIFF
--- a/tests/examples/test_wds_e2e.py
+++ b/tests/examples/test_wds_e2e.py
@@ -1,10 +1,13 @@
 import io
 import json
+import os
 import tarfile
+import warnings
 from pathlib import Path
 
 import pandas as pd
 import pytest
+from sqlalchemy.exc import SAWarning
 
 from datachain.client.local import FileClient
 from datachain.lib.dc import DataChain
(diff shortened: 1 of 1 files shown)


tests/examples/test_wds_e2e.py
```diff
@@ -96,35 +99,45 @@ def test_wds(catalog, webdataset_tars):
 
 
 def test_wds_merge_with_parquet_meta(catalog, webdataset_tars, webdataset_metadata):
-    wds = DataChain.from_storage(Path(webdataset_tars).as_uri()).gen(
-        laion=process_webdataset(spec=WDSLaion), params="file"
-    )
+    with warnings.catch_warnings():
```
shcheklein, line 102: should we also put a limit of some sort to the ClickHouse-SQLAlchemy? so that we can remove this code later 

also, worth mentioning the version at list where it is still the case

(not a blocker to the PR)

tests/examples/test_wds_e2e.py
```diff
@@ -96,35 +99,45 @@ def test_wds(catalog, webdataset_tars):
...
+        # Ignore SQLAlchemy warnings about cartesian products as these warnings are
+        # not accurate, and are generated due to a bug in ClickHouse-SQLAlchemy.
+        # All queries do have ON clauses, and this warning is not generated in SQLite.
+        if os.environ.get("DATACHAIN_METASTORE") or os.environ.get(
```
shcheklein, line 106: Q: why this `or` ? is there a better way to detect that CH version is used ... e.g. run some kind of instance of ?

rlamy:
I think it would be better to fix the underlying issue if possible. Looking into it, the issue seems to be that the ON clause refers to different table objects than the ones in the JOIN.

dtulga:
Closing this, as there is some controversy around whether ignoring this warning is a good idea, or not.

PR #199, opened 2024-07-30
add example script smoke tests

mattseddon:
part of https://github.com/iterative/dvcx/issues/1480

relates to #190

currently tests -
```
examples/get_started/udfs/stateful.py
examples/get_started/udfs/simple.py
examples/get_started/udfs/parallel.py
examples/get_started/common_sql_functions.py
examples/get_started/json-csv-reader.py
examples/get_started/torch-loader.py # not on linux as model training times out
examples/llm_and_nlp/unstructured-text.py # only where unstructured[all-documents] is available
examples/multimodal/clip_inference.py
examples/multimodal/wds_filtered.py
examples/multimodal/wds.py # this is by far the longest-running example
examples/computer_vision/iptc_exif_xmp_lib.py
examples/computer_vision/openimage-detect.py
```

excludes - 
```
examples/multimodal/hf_pipeline.py # model download too slow
examples/multimodal/openai_image_desc_lib.py # no Open AI API key
examples/llm_and_nlp/llm-claude-aggregate-query.py # no Anthropic API key
examples/llm_and_nlp/llm-claude-simple-query.py # no Anthropic API key
examples/llm_and_nlp/llm-claude.py # no Anthropic API key
examples/computer_vision/blip2_image_desc_lib.py # model download too slow
examples/computer_vision/llava2_image_desc_lib.py # model download too slow
examples/computer_vision/fashion_product_images/** # scripts are for Studio
examples/**/*.ipynb # out of scope
```

Next, we need to decide on:

- if we want to remove some of the untested examples
- if we want to add the appropriate Open AI and Anthropic API keys 
- if we want to self host/cache some of the huggingface models
- what to do with notebooks and the "scripts to be copy pasted into Studio"

To give some more context on the HF issues - After downloading ~5GB of model I get stuck on the following message: 
```
Download: 0.00B [00:00, ?B/s]We detected that you are passing `past_key_values` as a tuple and this is deprecated and will be removed in v4.43. Please use an appropriate `Cache` class (https://huggingface.co/docs/transformers/v4.41.3/en/internal/generation_utils#transformers.Cache)
```

and after some amount of time I can cancel and see:

```
Download: 16.5kB [09:20, 30.1B/s]
```
at that rate I don't think the download will ever finish


examples/get_started/torch-loader.py
```diff
@@ -64,7 +64,7 @@ def forward(self, x):
     optimizer = optim.Adam(model.parameters(), lr=0.001)
 
     # Train the model
-    num_epochs = 10
+    num_epochs = 3
```
mattseddon, line 67: [C] 10 seems excessive for the example.

examples/get_started/json-csv-reader.py
```diff
@@ -88,16 +87,7 @@ def main():
...
     static_csv_ds = DataChain.from_csv(uri, output=ChatDialog, object_name="chat")
     static_csv_ds.print_schema()
-    print(static_csv_ds.to_pandas())
-
-    uri = "gs://datachain-demo/laion-aesthetics-csv"
-    print()
-    print("========================================================================")
-    print("dynamic CSV with header schema test parsing 3/3M objects")
-    print("========================================================================")
-    dynamic_csv_ds = DataChain.from_csv(uri, object_name="laion", nrows=3)
```
mattseddon, line 98: [F] `nrows` with a `csv` from `gs` does not work on my machine (never returns). We could leave the example in with `nrows=3` removed but we will get a blow out in run time. I am happy to go with whatever is decided.

examples/multimodal/wds.py
```diff
@@ -9,25 +7,36 @@
     .filter(C("file.name").glob("00000000.tar"))
     .settings(cache=True)
     .gen(laion=process_webdataset(spec=WDSLaion), params="file")
+    .save()  # materialize chain to avoid downloading data multiple times
```
mattseddon, line 10: For this example, the run time before this change was > 2hrs (for me). It now takes ~25 minutes to run.

examples/multimodal/wds.py
```diff
@@ -9,25 +7,36 @@
...
+    on=["stem", "emd.index"],
+    right_on=["stem", "source.index"],
 )
 
 res = wds.merge(meta, on="laion.json.uid", right_on="uid")
 
-df = res.limit(10).to_pandas()
-with pd.option_context("display.max_columns", None):
-    print(df)
+res.show(3)
```
mattseddon, line 42: [F] There are some characters in the first 20 records that cannot be printed on Windows

noxfile.py
```diff
@@ -74,3 +74,19 @@ def dev(session: nox.Session) -> None:
 
     python = os.path.join(venv_dir, "bin/python")
     session.run(python, "-m", "pip", "install", "-e", ".[dev]", external=True)
+
+
+@nox.session(python=["3.9", "3.10", "3.11", "3.12", "pypy3.9", "pypy3.10"])
+def examples(session: nox.Session) -> None:
+    session.install(".[examples]")
+    try:
+        session.install("unstructured[all-docs]")
```
mattseddon, line 83: [F] This package cannot be installed for all of the operating systems we are testing against. I thought it was better to include it where possible so that we can at least partially cover `examples/llm_and_nlp/unstructured-text.py`.

tests/examples/test_examples.py
```diff
@@ -0,0 +1,91 @@
...
+    if "torch" not in filename or os.environ.get("RUNNER_OS") != "Linux"
+]
+
+llm_and_nlp_examples = [
+    filename
+    for filename in glob.glob("examples/llm_and_nlp/**/*.py", recursive=True)
+    # no anthropic token
+    if "claude" not in filename
+    and ("unstructured" not in filename or can_import_unstructured())
+] or [NO_EXAMPLES]
```
mattseddon, line 32: [F] If we cannot `import unstructured` then we have no `llm_and_nlp_examples` and the group fails on Windows with this OR statement.

tests/examples/test_examples.py
```diff
@@ -0,0 +1,91 @@
...
+def can_import_unstructured():
+    try:
+        import unstructured  # noqa: F401
+
+        return True
+    except ImportError:
+        return False
+
+
+get_started_examples = [
```
mattseddon, line 20: [F] New examples will be added automatically to the test suite.

.github/workflows/tests.yml
```diff
@@ -180,3 +180,32 @@ jobs:
...
+
+  examples:
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 60
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest-16-cores, macos-latest, windows-latest-8-cores]
+        pyv: ['3.9', '3.12']
+        group: ['get_started', 'llm_and_nlp or computer_vision', 'multimodal']
```
mattseddon, line 193: [F] The tests are manually grouped using `@pytest.mark` functionality.

examples/computer_vision/iptc_exif_xmp_lib.py
```diff
@@ -63,7 +64,7 @@ def image_description(file):
         DataChain.from_storage(source, type="image")
         .settings(parallel=-1)
         .filter(C("file.name").glob("*.jpg"))
-        .limit(10000)
+        .limit(5000)
```
mattseddon, line 67: [F] We still get results but cut the run time in half

dtulga (approved):
I am somewhat concerned about the large amount of additional test runs (and test runtime) this adds for every PR, although I cannot think of a better approach at the moment to test these examples and ensure they are updated properly. So this looks good to me, and we will see if the extra test runs (or flakiness within them) cause any problems in the future (or not).

mattseddon:
We would've caught the breakage from https://github.com/iterative/datachain/pull/226 if we had merged this already

noxfile.py
shcheklein: Q: what is `all-docs`? I see that we deal only with pdf in that example, would it simplify / speedup if we use only `[pdf]`?

tests/examples/test_examples.py
```diff
@@ -0,0 +1,92 @@
...
+    filename
+    for filename in glob.glob("examples/get_started/**/*.py", recursive=True)
+    # torch-loader will not finish within an hour on Linux runner
```
shcheklein, line 23: models?

mattseddon: yeah, that example is training a torch model. From earlier tests, it did not finish within an hour on the Linux runners.

examples/get_started/common_sql_functions.py
```diff
@@ -25,19 +25,20 @@ def num_chars_udf(name):
...
 (
-    ds.mutate(
-        a=array.length(string.split(C("file.parent"), literal("/"))),
-        b=array.length(string.split(C("file.name"), literal("0"))),
-    )
-    .mutate(
-        greatest=greatest(C("a"), C("b")),
-        least=least(C("a"), C("b")),
+    chain.mutate(
+        greatest=greatest(chain.column("a"), chain.column("b")),
```
mattseddon, line 40: [F] This is a patch for https://github.com/iterative/datachain/pull/226

mattseddon: alternative patch is 

```python
(
    dc.mutate(
        a=array.length(string.split(C("file.parent"), literal("/"))),
        b=array.length(string.split(C("file.name"), literal("0"))),
    )
    .mutate(
        greatest=greatest(C("a", Int64), C("b", Int64)),
        least=least(C("a", Int64), C("b", Int64)),
    )
    .select("a", "b", "greatest", "least")
    .show(10)
)
```

mattseddon: and the last is to give `greatest`/`least` a `type=Float()` property and change it from inheriting `ReturnTypeFromArgs` to `GenericFunction`

```python
class greatest(GenericFunction):  # noqa: N801
    inherit_cache = True

    type = Float()


class least(GenericFunction):  # noqa: N801
    inherit_cache = True

    type = Float()

```

.github/workflows/tests.yml
```diff
@@ -180,3 +180,41 @@ jobs:
...
+
+      - name: Set up Python ${{ matrix.pyv }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.pyv }}
+          cache: 'pip'
+
+      - name: Install libmagic for unstructured package
+        if: ${{ runner.os == 'macOS' &&  matrix.group == 'llm_and_nlp or computer_vision' }}
+        run: brew install libmagic
```
shcheklein, line 206: just to double check - do we still need libmagic?

noxfile.py
```diff
@@ -74,3 +74,18 @@ def dev(session: nox.Session) -> None:
 
     python = os.path.join(venv_dir, "bin/python")
     session.run(python, "-m", "pip", "install", "-e", ".[dev]", external=True)
+
+
+@nox.session(python=["3.9", "3.10", "3.11", "3.12", "pypy3.9", "pypy3.10"])
+def examples(session: nox.Session) -> None:
+    session.install(".[examples]")
+    try:
+        session.install("unstructured[pdf]")
```
shcheklein, line 83: can it be simplified now (move into the regular requirements?)

.github/workflows/tests.yml
mattseddon: yes, this is a pre-requirement of `unstructured`

noxfile.py
mattseddon: No, it does not build for all combinations of OS + Python versions. [example](https://github.com/iterative/datachain/actions/runs/10276130816/job/28436039948?pr=199). This is why we need `can_import_unstructured`.

.github/workflows/tests.yml
shcheklein: we can simplify it with `from unstructured.partition.pdf import partition_pdf as partition` I think (since we deal only with PDFs)

noxfile.py
shcheklein: Hmm, I don't see any error (specific to this) in that link 🤔 - can you share a specific line in the logs / job?

tests/examples/test_examples.py
```diff
@@ -0,0 +1,92 @@
...
+    if "claude" not in filename
+    and ("unstructured" not in filename or can_import_unstructured())
+] or [NO_EXAMPLES]
+
+multimodal_examples = [
+    filename
+    for filename in glob.glob("examples/multimodal/**/*.py", recursive=True)
+    # no OpenAI token
+    # and hf download painfully slow
+    if "openai" not in filename and "hf" not in filename
```
shcheklein, line 40: minor: checks like `"hf" not in filename` a bit concerning (can easily lead to false positives w/o noticing)

optional: I would do a file `.skip.tests`

or we can even put a list of file names right here to be 100% explicit

- first get the list of files
- for each exception `assert filename in the list`
- remove each exception

noxfile.py
mattseddon: https://github.com/iterative/datachain/actions/runs/10276130816/job/28436039948?pr=199#step:6:19 - unstructured depends on pycrypto which doesn't build.

shcheklein:
So, the longest time we run it is 19m? I don't think we can allow such a drastic drop in run time for CI tbh. We should parallelize or cache? any other ideas?

noxfile.py
shcheklein: ah, I see this is `uv` bug! add `pdfplumber==0.11.1` as a dependency to force uv to use a proper version of it

mattseddon:
> So, the longest time we run it is 19m? I don't think we can allow such a drastic drop in run time for CI tbh. We should parallelize or cache? any other ideas?

The longest-running example is `wds.py` and the run time is about to increase dramatically. For this example, we can either 

1. add an environment variable to use a different bucket as a base
2. add an environment variable to limit the result
4. drop the example
5. move long-running examples out to daily.

We could try patching some of the internals to limit the output of each step but we'd be getting further away from testing the code/syntax.

shcheklein:
> add an environment variable to use a different bucket as a base
> add an environment variable to limit the result

both options are totally fine I think, if it gives us something like 4 minutes

mattseddon:
> > add an environment variable to use a different bucket as a base
> > add an environment variable to limit the result
> 
> both options are totally fine I think, if it gives us something like 4 minutes

The slowest existing step for the CI is 12-15 minutes for Windows

shcheklein:
> The slowest existing step for the CI is 12-15 minutes for Windows

yep, but we should be optimizing that as well. It's way way too long I think.

shcheklein (approved):
Good stuff!

mattseddon:
I'll iterate on this

PR #203, opened 2024-07-30
validation error handling improved

volkfox:
Minor improvements to JSON validation code:

1. Offending row numbers now reported to pinpoint data errors in JSON arrays and JSON lines
2. Methods show_json*_schema() were renamed to print_json*_schema() for consistency
3. examples modified to reflect these changes


dmpetrov (approved):
LG!

PS: I'm a bit skeptical about having these print/show options as a params. We should probably revisit this practice.

PR #207, opened 2024-07-31
JSON tutorial

volkfox:
first shot


dmpetrov (approved):
LG!

PS: we'll need to move all notebooks to outside repo.

PR #213, opened 2024-08-01
fix select then distinct chain

### LINKED ISSUE #209: Datachain `distinct` method doesn't work after `select`
### Description

I have a scenario where I need to `select` some columns first and get `distinct` rows. I found out that in this order it doesn't work. But it works when `distinct` goes first. 

Is it a bug or a feature? What is the reason? 

Code to reproduce: 
```python 
from datachain.sql.functions.array import cosine_distance


class Embedding(BaseModel):
    id: int
    name: str
    values: list[float]


( DataChain
 .from_values(
        embedding=[
            Embedding(id=1, name="name", values=[0.1, 0.3]),
            Embedding(id=2, name="name", values=[0.1, 0.5]),
            Embedding(id=3, name="name", values=[0.1, 0.4]),
            Embedding(id=4, name="name", values=[0.1, 0.7]),
            Embedding(id=4, name="name", values=[0.1, 0.6]),
        ],)
 .select("embedding.id", "embedding.name")
 .distinct("embedding.id")
 .show()
)
``` 

Error 
```
File [~/.pyenv/versions/3.11.6/lib/python3.11/site-packages/datachain/lib/signal_schema.py:237](http://localhost:8888/lab/tree/crewai/~/.pyenv/versions/3.11.6/lib/python3.11/site-packages/datachain/lib/signal_schema.py#line=236), in SignalSchema.resolve(self, *names)
    235     if not isinstance(field, str):
    236         raise SignalResolvingTypeError("select()", field)
--> 237     schema[field] = self._find_in_tree(field.split("."))
    239 return SignalSchema(schema)

File [~/.pyenv/versions/3.11.6/lib/python3.11/site-packages/datachain/lib/signal_schema.py:253](http://localhost:8888/lab/tree/crewai/~/.pyenv/versions/3.11.6/lib/python3.11/site-packages/datachain/lib/signal_schema.py#line=252), in SignalSchema._find_in_tree(self, path)
    250     i += 1
    252 if curr_type is None:
--> 253     raise SignalResolvingError(path, "is not found")
    255 return curr_type

SignalResolvingError: cannot resolve signal name 'embedding.id': is not found
```  

### Version Info

```Text
0.2.14
Python 3.11.6
```

mattseddon:
closes #209

The issue is that `distinct` strips the parent path from the `signals_schema`. E.g. if we do distinct by `file.name` then `file` is removed from the schema. `select` then fails because it cannot find `file` in the tree.

see the `test_select_distinct` test to see exactly what is being fixed.

### DIFF
--- a/src/datachain/lib/signal_schema.py
+++ b/src/datachain/lib/signal_schema.py
@@ -243,9 +243,9 @@ class SignalSchema:
         curr_type = None
         i = 0
         while curr_tree is not None and i < len(path):
-            if val := curr_tree.get(path[i], None):
+            if val := curr_tree.get(path[i]):
                 curr_type, curr_tree = val
-            elif i == 0 and (val := curr_tree.get(".".join(path), None)):
+            elif i == 0 and len(path) > 1 and (val := curr_tree.get(".".join(path))):
                 curr_type, curr_tree = val
                 break
             else:
--- a/src/datachain/lib/signal_schema.py
+++ b/src/datachain/lib/signal_schema.py
@@ -245,6 +245,9 @@ class SignalSchema:
         while curr_tree is not None and i < len(path):
             if val := curr_tree.get(path[i], None):
                 curr_type, curr_tree = val
+            elif i == 0 and (val := curr_tree.get(".".join(path), None)):
+                curr_type, curr_tree = val
+                break
             else:
                 curr_type = None
             i += 1
--- a/tests/unit/lib/test_datachain.py
+++ b/tests/unit/lib/test_datachain.py
@@ -612,6 +612,41 @@ def test_select_restore_from_saving(catalog):
     assert n == len(features_nested)
 
 
+def test_select_distinct(catalog):
+    class Embedding(BaseModel):
+        id: int
+        filename: str
+        values: list[float]
+
+    expected = [
+        [0.1, 0.3],
+        [0.1, 0.4],
+        [0.1, 0.5],
+        [0.1, 0.6],
+    ]
+
+    actual = (
+        DataChain.from_values(
+            embedding=[
+                Embedding(id=1, filename="a.jpg", values=expected[0]),
+                Embedding(id=2, filename="b.jpg", values=expected[2]),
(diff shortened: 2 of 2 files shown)


src/datachain/lib/signal_schema.py
```diff
@@ -245,6 +245,9 @@ def _find_in_tree(self, path: list[str]) -> DataType:
         while curr_tree is not None and i < len(path):
             if val := curr_tree.get(path[i], None):
                 curr_type, curr_tree = val
+            elif i == 0 and (val := curr_tree.get(".".join(path), None)):
+                curr_type, curr_tree = val
```
dtulga, line 249: Won't this crash trying to unpack two values from `None` if the requested "."-based path is not found in `curr_tree` above, since the get will return `None`?

Also, `get` usually has a default of returning `None` if the requested entry is not found, providing `None` as a default value is usually not needed. See the Python documentation here: https://docs.python.org/3.12/library/stdtypes.html#dict.get

dtulga (approved):
LGTM, with one optional comment about possibly having a better error message and/or cleaner `get` call - these are completely optional to fix for this quick fix PR.

src/datachain/lib/signal_schema.py
mattseddon: > Won't this crash trying to unpack two values from None if the requested "."-based path is not found in curr_tree above, since the get will return None?

If I understand the question correctly I don't think either of the `if` or `elif` statements will evaluate to `True` under those circumstances because of the `:=` walrus operator. I.e. to re-assign `curr_type, curr_tree` we need `val is True`. Please correct me if I am wrong.

> Also, get usually has a default of returning None if the requested entry is not found, providing None as a default value is usually not needed. See the Python documentation here: https://docs.python.org/3.12/library/stdtypes.html#dict.get

I did this on auto-pilot, starting with the above line. I'll update both.

dtulga: Yep, that's correct - sorry I misread the `:=` operator usage (as I don't use it much). Thanks for the changes!

PR #215, opened 2024-08-01
rename DataChain's create_empty to from_records

mattseddon:
closes https://github.com/iterative/dvcx/issues/1555

### DIFF
--- a/src/datachain/lib/dc.py
+++ b/src/datachain/lib/dc.py
@@ -1124,7 +1124,7 @@ class DataChain(DatasetQuery):
         def _func_fr() -> Iterator[tuple_type]:  # type: ignore[valid-type]
             yield from tuples
 
-        chain = DataChain.create_empty(DataChain.DEFAULT_FILE_RECORD, session=session)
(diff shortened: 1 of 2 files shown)


ilongin (approved):

shcheklein:
It seems this PR broke CI?

mattseddon:
> It seems this PR broke CI?

it was broken prior to this PR

src/datachain/lib/dc.py
```diff
@@ -1441,22 +1441,23 @@ def to_parquet(
         )
 
     @classmethod
-    def create_empty(
+    def from_records(
```
skshetry, line 1444: The problem with `create_empty()` is that it cannot support arbitrary record dictionaries. (See the hardcoded `_datachain_column_types` columns usage below).

We need to improve this API to detect schema from the records (and/or additionally take `schema` parameter for users to pass) before calling it `from_records` (although I don't have strong opinion towards renaming).


Sorry, I should have written something in the issue. We should at least re-open the issue for that to be worked on.

PR #216, opened 2024-08-01
test union of chains with different columns

mattseddon:
Closes https://github.com/iterative/dvcx/issues/1202

~Adds a regression test for the bug shown in the above issue which is no longer relevant.~

There is an issue in Clickhouse with `out.append(sqlalchemy.cast(sqlalchemy.null(), col.type).label(n))`

In order to fix the syntax / fields need to be updated to be Nullable, i.e. : 

```sql
63f81a59e8fe :) SELECT cast(Null,'Nullable(String)') as name FROM system.databases;

SELECT CAST(NULL, 'Nullable(String)') AS name
FROM system.databases

Query id: 27a1bb4c-8e70-4110-bca5-bd93f2c56adb

   ┌─name─┐
1. │ ᴺᵁᴸᴸ │
2. │ ᴺᵁᴸᴸ │
3. │ ᴺᵁᴸᴸ │
4. │ ᴺᵁᴸᴸ │
5. │ ᴺᵁᴸᴸ │
   └──────┘

5 rows in set. Elapsed: 0.010 sec.

63f81a59e8fe :) SELECT cast(Null,'String') as name FROM system.databases;

SELECT CAST(NULL, 'String') AS name
FROM system.databases

Query id: 4faa70fb-ac15-48c0-a604-140fe53aa36c


Elapsed: 0.006 sec.

Received exception from server (version 24.5.3):
Code: 70. DB::Exception: Received from localhost:9010. DB::Exception: Cannot convert NULL to a non-nullable type: In scope SELECT CAST(NULL, 'String') AS name FROM system.databases. (CANNOT_CONVERT_TYPE)
```

### DIFF
--- a/src/datachain/query/dataset.py
+++ b/src/datachain/query/dataset.py
@@ -859,16 +859,17 @@ class SQLUnion(Step):
         temp_tables.extend(self.query1.temp_table_names)
         q2 = self.query2.apply_steps().select().subquery()
         temp_tables.extend(self.query2.temp_table_names)
-        columns1, columns2 = fill_columns(q1.columns, q2.columns)
+        columns = coalesce_columns(q1.columns, q2.columns)
 
         def q(*columns):
-            names = {c.name for c in columns}
-            col1 = [c for c in columns1 if c.name in names]
-            col2 = [c for c in columns2 if c.name in names]
             res = (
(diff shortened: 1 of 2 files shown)


mattseddon:
Call this a failed experiment

Issue #148, opened 2024-07-23
Column type inference fails after using mutate to subtract two columns


tibor-mach:
### Description

I am creating two columns with `cosine_distance` and then I try to create a new column by subtracting one from the other. The code looks like this:

```
from sqlalchemy import func

embeddings_differences = (
    DataChain
    .from_dataset("embeddings")
    .mutate(
        query_sim_new = 1 - cosine_distance(C.document.embeddings_new, embedded_query_new),
        query_sim_old = 1 - cosine_distance(C.document.embeddings_old, embedded_query_old),
        )
    .mutate(abs_difference = func.abs(C.query_sim_old - C.query_sim_new))
    .save("embeddings-differences")
)
```

After running this I get this error:

```
CompileError: (in table 'ds_embeddings-differences_1', column 'abs_difference'): Can't generate DDL for NullType(); did you forget to specify a type on this Column?
```

This fails even if I remove `func.abs`, i.e. when this is just an ordinary subtraction of the two columns. If I try to save the dataset after the first `.mutate`, load it and then save it as a different dataset after calling the second `.mutate` it fails as well.

If I simply try to show the results with `.show` or convert it to pandas, it works as expected.

How to reproduce:

Run [this notebook](https://github.com/iterative/datachain/blob/llm-tutorial/examples/llm/llm_chatbot_evaluation.ipynb) and see the output of the cell with the code above.

### Version Info

```Text
0.2.12.dev36+g0b3cf72
Python 3.12.4
```

ilongin:
@tibor-mach  Added PR https://github.com/iterative/datachain/pull/152

This PR just adds better error handling for this as I don't think we can fix it in better way.
The workaround, and only solution I think is valid, is to manually cast to specific type when type cannot be infered, like in subtract example.

Until we have `cast` function in `DataChain` (PR is merged) you can use it from `SQLAlchemy` directly, e.g `from sqlalchemy import cast`. Examples of how to cast to specific types are in the PR

tibor-mach:
@ilongin hmm, this is kind of awkward, particularly with these simple operations it is kind of a hassle to have to manually cast the output. I wonder if we couldn't just include our custom automatic inference. 

But I will use the workaround for now ,thanks!

dmpetrov:
That's right! We need to handle it under the hood, not through a new API call.

dmpetrov:
@tibor-mach I was not able to reproduce it.

This seems return a correct result:
```
from sqlalchemy import func

[...]

PR #217, opened 2024-08-01
Fix renaming object or normal signal with `.mutate()`

ilongin:
Fixes renaming objects with `.mutate()`, e.g `.mutate(my_file=Column("file"))` . This renames all sub-signals of `file` so we end up with `my_file__name`, `my__file_size ` etc.

Also fixes renaming signals in general to work with `.save()` as this was not working before - it was throwing error as new columns were without type (they were `NullType())` which is the same issue as in https://github.com/iterative/datachain/issues/148

This also makes sure we have correct `SignalsSchema` values after mutating.


src/datachain/lib/dc.py
```diff
@@ -831,8 +831,27 @@ def mutate(self, **kwargs) -> "Self":
...
+        mutated = {}
+        schema = self.signals_schema
+        for name, value in kwargs.items():
+            if isinstance(value, Column) and schema.has_object(value.name):
+                # renaming existing object
+                old_signals = schema.db_signals(obj_name=value.name, as_columns=True)
+                for signal in old_signals:
+                    mutated[signal.name.replace(value.name, name, 1)] = signal
+            elif isinstance(value, Column) and value.name in schema.db_signals():
+                # renaming existing non object signal
```
dmpetrov, line 843: What is a use case for non-object signals? Let's get rid of theses.

src/datachain/lib/signal_schema.py
```diff
@@ -281,8 +298,22 @@ def clone_without_file_signals(self) -> "SignalSchema":
                 del schema[signal]
         return SignalSchema(schema)
 
+    def has_object(self, name: str) -> bool:
```
dmpetrov, line 301: Renaming is needed. Please avoid using abstract terms like `object`.
`contains()` is enough, `has_signal()` is ok, ...

tests/unit/lib/test_datachain.py
```diff
@@ -1228,3 +1228,60 @@ class Nested(BaseModel):
...
+def test_rename_non_object_column_name_with_mutate(catalog):
+    ds = DataChain.from_values(ids=[1, 2, 3])
+    ds = ds.mutate(my_ids=Column("ids"))
+
+    assert ds.signals_schema.values == {"my_ids": int}
+
+    ds.save("mutated")
+
+    ds = DataChain(name="mutated")
+    assert ds.signals_schema.values == {"my_ids": int, "sys": Sys}
```
dmpetrov, line 1242: please avoid touching parts that do not related to the test (sys). It can be just:
```python
assert ds.signals_schema.values.get("my_ids", None) == int
```

tests/unit/lib/test_datachain.py
```diff
@@ -1228,3 +1228,60 @@ class Nested(BaseModel):
...
+    files = [File(name=name, size=size) for name, size in zip(names, sizes)]
+
+    ds = DataChain.from_values(file=files, ids=[1, 2, 3])
+    ds = ds.mutate(fname=Column("file.name"))
+
+    assert list(ds.order_by("fname").collect("fname")) == ["a", "b", "c"]
+    assert ds.signals_schema.values == {"file": File, "ids": int, "fname": str}
+
+    # check that persist after saving
+    ds.save("mutated")
```
dmpetrov, line 1258: Can "mutated" ds name confclict with "mutated" ds from another test? We don't create separate sessions for each test, do we?

tests/unit/lib/test_datachain.py
```diff
@@ -1228,3 +1228,60 @@ class Nested(BaseModel):
...
+
+    ds = DataChain(name="mutated")
+    assert ds.signals_schema.values == {
+        "file": File,
+        "ids": int,
+        "fname": str,
+        "sys": Sys,
```
dmpetrov, line 1265: the same. please avoid touching parts that do not related to the test (sys).

tests/unit/lib/test_datachain.py
```diff
@@ -1228,3 +1228,60 @@ class Nested(BaseModel):
...
+    ds = ds.mutate(my_file=Column("file"))
+
+    assert list(ds.order_by("my_file.name").collect("my_file.name")) == ["a", "b", "c"]
+
+    assert ds.signals_schema.values == {"my_file": File, "ids": int}
+
+    ds.save("mutated")
+
+    ds = DataChain(name="mutated")
+    assert ds.signals_schema.values == {"my_file": File, "ids": int, "sys": Sys}
```
dmpetrov, line 1286: sys

src/datachain/lib/dc.py
```diff
@@ -831,8 +831,27 @@ def mutate(self, **kwargs) -> "Self":
         )
         ```
         """
-        chain = super().mutate(**kwargs)
-        chain.signals_schema = self.signals_schema.mutate(kwargs)
+        mutated = {}
+        schema = self.signals_schema
```
dmpetrov, line 835: It's not clear why the logic is separated between SignalSchema.mutete() and this code.
Ideally, we do all the logic in SignalSchema and DC just triggers this.
WDYT?

dmpetrov (changes requested):
Thank you for the PR! 

Some comments are inline.

tests/unit/lib/test_datachain.py
ilongin: No, there should not be any conflict

tests/unit/lib/test_datachain.py
ilongin: Done

src/datachain/lib/signal_schema.py
ilongin: Removed method as it was not needed after refactoring

tests/unit/lib/test_datachain.py
ilongin: Removed

src/datachain/lib/dc.py
ilongin: Refactored so there is only one `if` clause

src/datachain/lib/dc.py
ilongin: That was my initial idea but the thing is that those 2 methods are doing different things:
1. `SignalsSchema.mutate(...)` only needs to update it's `values` object which is pretty straightforward. 
2. `DataChain.mutate(...)` must prepare fields that need change for underlying `DatasetQuery.mutate(...)` which doesn't have knowledge of pydantic objects...so if pydantic object is renamed it needs to gather all of it's sub-fields and send to underlying mutate. 

I think now that I simplified `DataChain.mutate()` it's decent but maybe this can be refactored even more, idk.

src/datachain/lib/signal_schema.py
```diff
@@ -282,7 +299,18 @@ def clone_without_file_signals(self) -> "SignalSchema":
         return SignalSchema(schema)
 
     def mutate(self, args_map: dict) -> "SignalSchema":
-        return SignalSchema(self.values | sql_to_python(args_map))
+        new_values = self.values.copy()
+
+        for name, value in args_map.items():
+            if isinstance(value, Column) and value.name in self.values:
+                # renaming existing signal
+                del new_values[value.name]
```
dmpetrov, line 307: misc: del is not needed becasue of the next line

dmpetrov (approved):
Looks good!

src/datachain/lib/signal_schema.py
ilongin: It's needed. Old signals is removed and new one added. It's different keys

PR #221, opened 2024-08-01
Arrow nrows fix

### LINKED ISSUE #210: `parse_tabular(nrows=3)` is slow
### Description

Thanks to both @mattseddon and @volkfox for raising this. Any of `from_csv/from_parquet/parse_tabular` will be slow if used with `nrows` and a cloud path. 

To reproduce, run https://github.com/iterative/datachain/blob/main/examples/get_started/json-csv-reader.py. The last example takes a long time to complete even though it uses `nrow=3`. After diving into it a bit, it looks to be an issue with either pyarrow or fsspec. Opened https://github.com/apache/arrow/issues/43497 to track the issue upstream.

### Version Info

```
$ datachain --version
0.2.10.dev6+gcc0b63b.d20240716
```

dberenbaum:
Fixes #210 

Writes the first nrows to a local temp file to avoid the pyarrow issues with partial reads from cloud filesystems. This means nrows won't work with parquet or other binary formats.

I find this fix hacky but a quick look at pyarrow makes me think it would be much more effort to fix it there, so this might be the best we can do for now.

### DIFF
--- a/src/datachain/lib/arrow.py
+++ b/src/datachain/lib/arrow.py
@@ -54,7 +54,7 @@ class ArrowGenerator(Generator):
             )
         index = 0
         with tqdm(desc="Parsed by pyarrow", unit=" rows") as pbar:
-            for record_batch in ds.to_batches(use_threads=False):
+            for record_batch in ds.to_batches():
                 for record in record_batch.to_pylist():
                     vals = list(record.values())
                     if self.output_schema:
--- a/src/datachain/lib/arrow.py
+++ b/src/datachain/lib/arrow.py
@@ -135,8 +135,8 @@ def _nrows_file(file: File, nrows: int) -> str:
     with file.open(mode="r") as reader:
         with open(tf.name, "a") as writer:
             for row, line in enumerate(reader):
-                writer.write(line)
-                writer.write("\n")
                 if row >= nrows:
                     break
+                writer.write(line)
+                writer.write("\n")
     return tf.name
--- a/src/datachain/lib/arrow.py
+++ b/src/datachain/lib/arrow.py
@@ -7,7 +7,7 @@ import pyarrow as pa
 from pyarrow.dataset import dataset
 from tqdm import tqdm
 
-from datachain.lib.file import File, IndexedFile, TextFile
+from datachain.lib.file import File, IndexedFile
 from datachain.lib.udf import Generator
 
 if TYPE_CHECKING:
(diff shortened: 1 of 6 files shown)


src/datachain/lib/arrow.py
```diff
@@ -125,3 +128,15 @@ def _arrow_type_mapper(col_type: pa.DataType) -> type:  # noqa: PLR0911
...
+
+
+def _nrows_file(file: File, nrows: int) -> str:
+    tf = NamedTemporaryFile(delete=False)
+    with file.open(mode="r") as reader:
+        with open(tf.name, "a") as writer:
+            for row, line in enumerate(reader):
+                writer.write(line)
+                writer.write("\n")
+                if row >= nrows:
```
shcheklein, line 140: just to double check: will it support headers?

src/datachain/lib/arrow.py
```diff
@@ -43,13 +44,17 @@ def __init__(
...
+            path = _nrows_file(file, self.nrows)
+            ds = dataset(path, schema=self.input_schema, **self.kwargs)
+        else:
+            path = file.get_path()
+            ds = dataset(
+                path, filesystem=file.get_fs(), schema=self.input_schema, **self.kwargs
+            )
         index = 0
         with tqdm(desc="Parsed by pyarrow", unit=" rows") as pbar:
-            for record_batch in ds.to_batches(use_threads=False):
```
shcheklein, line 52: why removing use_threads?

src/datachain/lib/dc.py
```diff
@@ -1263,8 +1263,20 @@ def parse_tabular(
...
+        from pyarrow.dataset import CsvFileFormat, JsonFileFormat
+
         from datachain.lib.arrow import ArrowGenerator, infer_schema, schema_to_output
 
+        if nrows:
+            format = kwargs.get("format")
+            if format not in ["csv", "json"] and not isinstance(
+                format, (CsvFileFormat, JsonFileFormat)
+            ):
+                raise ValueError(
```
shcheklein, line 1275: minor: raise NotImplemented?

dberenbaum: Changed to `DatasetPrepareError` since that's used as the main type for datachain errors.

src/datachain/lib/arrow.py
dberenbaum: Good catch. I forgot to support no-header use cases. Fixed in the latest commit by adding a row if there is a header.

src/datachain/lib/arrow.py
dberenbaum: `use_threads=False` was a hack to make nrows not hang forever. Previously nrows broke the batch iteration and hung forever without `use_threads=False`. We no longer are messing with the batch iteration, so this isn't needed.

mattseddon (approved):

PR #206, opened 2024-07-30
make datachain show respect existing limit

mattseddon:
closes #204

PR #225, opened 2024-08-02
do not modify datachain max limit in show

### LINKED ISSUE #219: `limit()` ignored after `show()`
In some cases, `.limit()` seems to have no effect. Here's a reproducer:
```python

img_dc = DataChain.from_storage("gs://datachain-demo/newyorker_caption_contest/images")
img_dc.show(3)
assert img_dc.limit(10).count() == 10
```
Commenting out `img_dc.show(3)` makes the assert succeed as expected.

Note that I've only been able to reproduce the failure with a real bucket, but not in tests.

mattseddon:
Closes #219 by fixing the error introduced in #206. The approach now more closely aligns with the existing mechanics around `clone`.

### DIFF
--- a/src/datachain/lib/dc.py
+++ b/src/datachain/lib/dc.py
@@ -193,8 +193,6 @@ class DataChain(DatasetQuery):
         ```
     """
 
-    max_row_count: Optional[int] = None
-
     DEFAULT_FILE_RECORD: ClassVar[dict] = {
         "source": "",
         "name": "",
@@ -1603,18 +1601,7 @@ class DataChain(DatasetQuery):
     @detach
     def limit(self, n: int) -> "Self":
         """Return the first n rows of the chain."""
-        n = max(n, 0)
-
-        if self.max_row_count is None:
-            self.max_row_count = n
-            return super().limit(n)
-
-        limit = min(n, self.max_row_count)
-        if limit == self.max_row_count:
-            return self
-
-        self.max_row_count = limit
-        return super().limit(self.max_row_count)
+        return super().limit(n)
 
     @detach
     def offset(self, offset: int) -> "Self":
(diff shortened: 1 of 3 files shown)


src/datachain/lib/dc.py
```diff
@@ -1603,18 +1601,7 @@ def filter(self, *args) -> "Self":
     @detach
     def limit(self, n: int) -> "Self":
```
dreadatour, line 1602: 🤔 do we still need this function here as it only calls "super" method?

dreadatour (approved):
Looks good to me! Only one comment above.

src/datachain/lib/dc.py
mattseddon: This is how I found it before #206 (https://github.com/iterative/datachain/pull/206/files#diff-0b1cfd3d89c057cc57811e07eb20298eeaaee60c54cc66e5128f618d144f42eeL1601). There are a few methods in the class like this - e.g. `sample`, `shuffle`, `offset`, `filter`.

dreadatour: Got it! 👍 We can address this later than, if needed.

skshetry: This makes it clear what APIs are internal (from `DatasetQuery`) and what are exposed to the user. This helps us in writing a good API reference for `DataChain` and in the future, will help us from decoupling `DatasetQuery` and `DataChain`.

src/datachain/query/dataset.py
```diff
@@ -1383,6 +1383,9 @@ def order_by(self, *args) -> "Self":
     @detach
     def limit(self, n: int) -> "Self":
         query = self.clone(new_table=False)
+        for step in query.steps:
+            if isinstance(step, SQLLimit) and step.n < n:
+                return query
```
rlamy, line 1388: This will miscompile if the chain contains steps that increase the size of the dataset, like `.gen()`.

PR #229, opened 2024-08-02
Fixing too many files open, and adding reconnect

### LINKED ISSUE #160: If UDF execution fails, SQLite database closes
### Description

This is a copy of iterative/dvcx#1663 from dvcx. Raising priority because of the frequent occurence

### Description

Let us assume we have a wrong API key to simulate a UDF error.
Now execute the following code:

```python
import os
import pandas as pd
from datachain.lib.feature import Feature
from mistralai.client import MistralClient
from mistralai.models.chat_completion import ChatMessage
from datachain.lib.dc import Column, DataChain            

source = "gs://datachain-demo/chatbot-KiT/"           
PROMPT = "Was this dialog successful? Describe the 'result' as 'Yes' or 'No' in a short JSON"

model = "mistral-large-latest"
api_key = os.environ["MISTRAL_API_KEY"]

# TODO: parallelize w/same client? cannot pickle 'SSLContext'

chain = DataChain.from_storage(source)                       \
              .filter(Column("file.name").glob("*.txt"))     \
              .limit(5)                                      \
              .settings(cache=True, parallel = 5)            \
              .map(
		              mistral_response = lambda file:            \
				   MistralClient(api_key=api_key)        \
						.chat(
						    model=model,
                                                    response_format={"type": "json_object"},
				                    messages= [                          
						         ChatMessage(role="user", content=f"{PROMPT}: {file.get_value()}") 
						    ]                                                       
					         ).choices[0].message.content, 
				   )
	 
try:
   print(chain.select("mistral_response").results())
except Exception as e:
   print(f"do you have the right Mistral API key? {e}")

```

As expected, the UDFs will fail:

```
....
datachain.lib.utils.DataChainError: Error in user code in class 'Mapper': Status: 401. Message: {
  "message":"Unauthorized",
  "request_id":"8dd5091032fd6d4e706aee4e9fdc614b"
}
do you have the right Mistral API key? UDF Execution Failed!
```

However, the runtime failure renders entire chain unusable due to a closed database:

```
>>> chain.select("mistral_response").results()
  File "/Users/dkh/dvcx/src/datachain/query/dataset.py", line 1287, in results
    with self.as_iterable(**kwargs) as result:
  File "/usr/local/Cellar/python@3.9/3.9.4/Frameworks/Python.framework/Versions/3.9/lib/python3.9/contextlib.py", line 117, in __enter__
    return next(self.gen)
  File "/Users/dkh/dvcx/src/datachain/query/dataset.py", line 1296, in as_iterable
    query = self.apply_steps().select()
  File "/Users/dkh/dvcx/src/datachain/query/dataset.py", line 1239, in apply_steps
    result = step.apply(
  File "/Users/dkh/dvcx/src/datachain/query/dataset.py", line 687, in apply
    query, tables = self.process_input_query(query)
  File "/Users/dkh/dvcx/src/datachain/query/dataset.py", line 737, in process_input_query
    table = self.create_pre_udf_table(query)
  File "/Users/dkh/dvcx/src/datachain/query/dataset.py", line 717, in create_pre_udf_table
    table = self.catalog.warehouse.create_udf_table(self.udf_table_name(), columns)
  File "/Users/dkh/dvcx/src/datachain/data_storage/warehouse.py", line 890, in create_udf_table
    self.db.create_table(tbl, if_not_exists=True)
  File "/Users/dkh/dvcx/src/datachain/data_storage/sqlite.py", line 208, in create_table
    self.execute(CreateTable(table, if_not_exists=if_not_exists))
  File "/Users/dkh/dvcx/src/datachain/data_storage/sqlite.py", line 72, in wrapper
    return func(*args, **kwargs)
  File "/Users/dkh/dvcx/src/datachain/data_storage/sqlite.py", line 149, in execute
    result = self.db.execute(*self.compile_to_args(query))
sqlite3.ProgrammingError: Cannot operate on a closed database.

```

The workaround is to define the chain again.

### Version Info

```Text
(clean-datachain) DN0a1e26a5:dvcx dkh$ datachain -V; python -V
0.1.14.dev8+gd94054a
Python 3.9.4
```

### Version Info

```Text
0.2.6.dev4+gef2347f
```

dtulga:
This includes fixes for the "too many open files" errors on tests when running locally, and also provides a reconnect option, so that if a UDF fails, then the dataset query or datachain can be used again, even if the database was closed upon failure (or other conditions). This is primarily accomplished through adding context manager support to metastore, warehouse, etc. to be used with `clone()`, adding more auto-close behavior on Session and in tests, using a test session for datachain tests so that duplicate Catalogs are not created, and adding `client_config` support to Session.

This fixes https://github.com/iterative/dvcx/issues/1685 and fixes #160 This has been tested locally to not raise the "too many open files" errors, even when leaving `ulimit` at the default value. This has also been tested with ClickHouse, with the companion ClickHouse PR code here: https://github.com/iterative/studio/pull/10398


src/datachain/data_storage/id_generator.py
```diff
@@ -33,6 +33,15 @@ def init(self) -> None:
     def cleanup_for_tests(self):
         """Cleanup for tests."""
 
+    def close(self) -> None:
+        """Closes any active database connections."""
+
+    def close_on_exit(self) -> None:
+        """Closes any active database or HTTP connections, called on exit or for test
```
mattseddon, line 40: [C] could mention session here as it appears that is where close_on_exit is currently called. I.e. - on `Session` exit

mattseddon: same for the other `close_on_exit` methods

tests/conftest.py
```diff
@@ -168,6 +171,11 @@ def catalog(id_generator, metastore, warehouse):
     return Catalog(id_generator=id_generator, metastore=metastore, warehouse=warehouse)
 
 
+@pytest.fixture
+def test_session(catalog):
```
mattseddon, line 175: [Q] How does this fit with the `autouse=True` - `clean_session` fixture? Do you know if we still need that fixture now?

mattseddon: Would using the following code mean that we could simplify these fixtures a little?

```python
with Session("TestSession", catalog=catalog) as session: 
    yield session
```

mattseddon: [Q] and (this is just a suggestion as it might be a bit too "magic") but could we change `Session.get` to return the `test_session` (e.g. use environment variable or make this session the global one) and remove the need to always declare `test_session` and pass it?

tests/unit/test_catalog_loader.py
```diff
@@ -24,37 +23,34 @@ def __init__(self, **kwargs):
         self.kwargs = kwargs
 
 
-def test_get_id_generator():
-    db = SQLiteDatabaseEngine.from_db_file(":memory:")
-
-    id_generator = SQLiteIDGenerator(db, table_prefix="prefix")
-    assert id_generator.db == db
+def test_get_id_generator(sqlite_db):
```
mattseddon, line 26: [Unrelated] - are these tests run in the Studio test suite?

mattseddon (approved):

src/datachain/data_storage/id_generator.py
dtulga: Yep, changed the comment to reference Session exit.

tests/unit/test_catalog_loader.py
dtulga: As far as I can tell, yes they are run - see the run here for example: https://github.com/iterative/datachain/actions/runs/10258065848/job/28381752999?pr=229

We probably could add a skip for these SQLite-specific tests, but that should probably be in a followup.

PR #230, opened 2024-08-03
only combine final step for limit

mattseddon:
Follow up to [this comment](https://github.com/iterative/datachain/pull/225#discussion_r1701687436)


rlamy (approved):
Thanks! I like the test - simple yet exhausitve!

Issue #237, opened 2024-08-05
Make show() not crash on empty dataset


volkfox:
### Description

Currently show() command crashes when getting an empty dataset.
It should handle it gracefully.

```
from datachain.lib.dc import Column, DataChain
a = DataChain.from_values(a=['a', 'b'])
 a.count()

>> Processed: 1 rows [00:00, 786.19 rows/s]
>> Generated: 2 rows [00:00, 1695.01 rows/s]
>> 2

 a.filter(a == '3').count()
>> Processed: 1 rows [00:00, 906.48 rows/s]
>> Generated: 2 rows [00:00, 1996.34 rows/s]
>> 0

a.filter(a == '3').show()

>>>
Processed: 1 rows [00:00, 1548.28 rows/s]
Generated: 2 rows [00:00, 3034.95 rows/s]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/dkh/datachain/src/datachain/lib/dc.py", line 1198, in show
    df = dc.to_pandas(flatten)
  File "/Users/dkh/datachain/src/datachain/lib/dc.py", line 1175, in to_pandas
    df.columns = [".".join(filter(None, header)) for header in headers]
  File "/Users/dkh/venv/datachain-1/lib/python3.9/site-packages/pandas/core/generic.py", line 6313, in __setattr__
    return object.__setattr__(self, name, value)
  File "properties.pyx", line 69, in pandas._libs.properties.AxisProperty.__set__
  File "/Users/dkh/venv/datachain-1/lib/python3.9/site-packages/pandas/core/generic.py", line 814, in _set_axis
    self._mgr.set_axis(axis, labels)
  File "/Users/dkh/venv/datachain-1/lib/python3.9/site-packages/pandas/core/internals/managers.py", line 238, in set_axis
    self._validate_set_axis(axis, new_labels)
  File "/Users/dkh/venv/datachain-1/lib/python3.9/site-packages/pandas/core/internals/base.py", line 98, in _validate_set_axis
    raise ValueError(
ValueError: Length mismatch: Expected axis has 0 elements, new values have 1 elements
```

### Version Info

```Text
0.2.16.dev2+gf019b0f
```

PR #241, opened 2024-08-06
fix(to_pandas): handle empty datachain in to_pandas (and show)

shcheklein:
Closes https://github.com/iterative/datachain/issues/237 + simplifies the code a bit and makes sure that column names are preserved in `df` even if there are no values - consistently across flatten and non-flatten version.

```
files = [File(size=s, path=p) for p, s in zip(list("abcde"), range(5))]
DataChain.from_values(file=files).limit(0).show()
```

returns:

```
Empty result
Columns: [('file', 'source'), ('file', 'path'), ('file', 'size'), ('file', 'version'), ('file', 'etag'), ('file', 'is_latest'), ('file', 'last_modified'), ('file', 'location'), ('file', 'vtype')]
```

(similar to how pandas returns the list of columns to make it easier to debug)

Once thing that I was not able to imitate for now is a schema-less (no columns, no-values DataChain), but let's consider this as a minor edge case that we can test / fix later if we decide to allow that.

### DIFF
--- a/src/datachain/lib/dc.py
+++ b/src/datachain/lib/dc.py
@@ -1206,14 +1206,14 @@ class DataChain(DatasetQuery):
         """
         headers, max_length = self._effective_signals_schema.get_headers_with_length()
         if flatten or max_length < 2:
-            df = pd.DataFrame.from_records(self.to_records())
+            columns = []
             if headers:
-                df.columns = [".".join(filter(None, header)) for header in headers]
-            return df
+                columns = [".".join(filter(None, header)) for header in headers]
+            return pd.DataFrame.from_records(self.to_records(), columns=columns)
 
-        transposed_result = list(map(list, zip(*self.results())))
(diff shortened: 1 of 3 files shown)


tests/func/test_dataset_query.py
```diff
@@ -1185,8 +1185,8 @@ def name_len_interrupt(_name):
 
     q = (
         DatasetQuery(name="animals", version=1, catalog=catalog)
-        .filter(C.size < 13)
-        .filter(C.parent.glob("cats*") | (C.size < 4))
+        # .filter(C.size < 13)
```
shcheklein, line 1188: disregard this, unrelated, this PR is primarily for @mattseddon and can be closed

mattseddon: closed my PR, you can take it

shcheklein: sounds good

mattseddon (approved):

PR #245, opened 2024-08-06
Llm bruteforce blog

volkfox:
adding a notebook for the blog


shcheklein:
we should have already a repo for notebooks, I think @mnrozhkov and @tibor-mach created it. Let's move it there please.

dmpetrov (approved):
LG!

shcheklein (commented):
The trick looks quite cool :) But what is the value of datachain here? (I guess I can wait for the blog post itself as well)

shcheklein:
@volkfox I think we can close it here now?

PR #248, opened 2024-08-07
Adding schema param to `from_records`

ilongin:
Adding optional `schema` param to `DataChain.from_records()`
This allows us creating datasets with specific schema where important use case is creating empty datasets with no rows but with defined schema. If no rows are provided and there is no explicit schema, exception is thrown. 

Follow up is to infer schema from rows itself.


src/datachain/lib/dc.py
```diff
@@ -1490,6 +1490,7 @@ def from_records(
         cls,
         to_insert: Optional[Union[dict, list[dict]]],
         session: Optional[Session] = None,
+        schema: Optional[dict[str, DataType]] = None,
     ) -> "DataChain":
```
shcheklein, line 1494: can we update the docstring please?

src/datachain/lib/dc.py
```diff
@@ -1499,6 +1500,7 @@ def from_records(
         Parameters:
             to_insert : records (or a single record) to insert. Each record is
                         a dictionary of signals and theirs values.
+            schema : describes chain signals and their corresponding types
```
shcheklein, line 1503: does the `empty = DataChain.from_records()` still work?

shcheklein: (see below in the doc string)

src/datachain/lib/dc.py
```diff
@@ -1509,12 +1511,31 @@ def from_records(
         session = Session.get(session)
         catalog = session.catalog
 
+        if not to_insert and not schema:
+            raise ValueError("Schema is required for creating empty dataset")
```
shcheklein, line 1515: message seems to be wrong or misleading (?) considering the condition `Schema is required for creating empty dataset`. Condition should be `if not schema` then

ilongin: Changed message, please check if now it seems better

src/datachain/lib/dc.py
ilongin: Doesn't work, removing it

src/datachain/lib/dc.py
```diff
@@ -1499,22 +1500,41 @@ def from_records(
         Parameters:
             to_insert : records (or a single record) to insert. Each record is
                         a dictionary of signals and theirs values.
+            schema : describes chain signals and their corresponding types
 
         Example:
             ```py
-            empty = DataChain.from_records()
             single_record = DataChain.from_records(DataChain.DEFAULT_FILE_RECORD)
```
shcheklein, line 1507: okay, but is the meaning, purpose of this?

tests/unit/lib/test_datachain.py
```diff
@@ -128,6 +128,31 @@ def test_from_features(test_session):
...
+    ds_sys = ds.settings(sys=True)
+
+    ds_name = "my_ds"
+    ds.save(ds_name)
+    ds = DataChain(name=ds_name)
+
+    assert isinstance(ds.feature_schema, dict)
+    assert isinstance(ds.signals_schema, SignalSchema)
+    assert ds.schema.keys() == {"my_file", "my_col"}
+    assert set(ds.schema.values()) == {File, int}
```
dmpetrov, line 143: can these two lines be compared in a single line? like `assert ds.schema == {"my_file": FIle, "my_col": int}`

tests/unit/lib/test_datachain.py
```diff
@@ -128,6 +128,31 @@ def test_from_features(test_session):
...
+    schema = {"my_file": File, "my_col": int}
+    ds = DataChain.from_records([], schema=schema, session=test_session)
+    ds_sys = ds.settings(sys=True)
+
+    ds_name = "my_ds"
+    ds.save(ds_name)
+    ds = DataChain(name=ds_name)
+
+    assert isinstance(ds.feature_schema, dict)
+    assert isinstance(ds.signals_schema, SignalSchema)
```
dmpetrov, line 141: These two assertions seems unnecessary since `.schema` is the only API we should care about.

tests/unit/lib/test_datachain.py
```diff
@@ -128,6 +128,31 @@ def test_from_features(test_session):
...
+    assert ds.schema.keys() == {"my_file", "my_col"}
+    assert set(ds.schema.values()) == {File, int}
+    assert ds.count() == 0
+
+    # check that columns have actually been created from schema
+    dr = ds_sys.catalog.warehouse.dataset_rows(ds_sys.catalog.get_dataset(ds_name))
+    assert sorted([c.name for c in dr.c]) == sorted(ds.signals_schema.db_signals())
```
dmpetrov, line 148: This doesn't belong in this test. It should either be removed or extracted to a separate test if the need is clear.

tests/unit/lib/test_datachain.py
```diff
@@ -128,6 +128,31 @@ def test_from_features(test_session):
...
+
+
+def test_from_records_empty_chain_without_schema():
+    with pytest.raises(ValueError):
+        DataChain.from_records([], schema=None)
```
dmpetrov, line 153: misc: schema should be None by default

tests/unit/lib/test_signal_schema.py
```diff
@@ -2,8 +2,9 @@
 from typing import Optional, Union
 
 import pytest
+from sqlalchemy import Column
```
dmpetrov, line 5: Please use DataChain's Column, not sqlite. There is a chance we will be replacing the implementation.

dmpetrov (approved):
Looks good!

A couple of comments are inline. Mostly about test - some tests are testing so many parts. We should be careful with this.

Also, it feels like DC is becoming a fat class. We should do something about it (not in this PR of course).

tests/unit/lib/test_datachain.py
ilongin: It is but I wanted to make that explicit here in test

tests/unit/lib/test_signal_schema.py
ilongin: I had some strange issues when I was using `Column` from `DataChain` [here](https://github.com/iterative/datachain/pull/248/files#diff-69a1fc2d77ff106392ad7649ed8cf7ec344dad230af2e4d0ea9c64b816cee15bR236) so I change to use `sa.Column` directly. For some reason `index` field was defined when using `DataChain.Column` for each column and that was causing failure later on when creating table with those columns. 
I need to investigate why is that (maybe we accidentally broke something when defining our `Column` class which extends SQLAlchemy one) but regardless, this is all hidden from user, i.e is not in the API itself so it should be good.

tests/unit/lib/test_datachain.py
ilongin: Extracted to separate test as I think it's important to test it

src/datachain/lib/dc.py
ilongin: You are referring to that example of creating dataset from single record or?

shcheklein: yep, I would say this specific record - DEFAULT_FILE_RECORD - what is the point of this example?

src/datachain/lib/dc.py
```diff
@@ -1513,22 +1514,41 @@ def from_records(
...
             ```
         """
         session = Session.get(session)
         catalog = session.catalog
 
+        if not to_insert and not schema:
+            raise ValueError("Non empty records to insert or schema must be defined")
```
shcheklein, line 1528: can we define both? what takes precedence then?

shcheklein: btw, why can't we do a completely empty schema and no records?

shcheklein (approved):

shcheklein:
LGTM, please take a look one more time and address some unresolved discussions if they make sense. Thanks @ilongin !

src/datachain/lib/dc.py
ilongin: Yea, I removed this exception. Now if empty records and no schema is set, default columns are created (as in general when no schema is set) ... in followup issue we should try to infer schema from records, if explicit schema is not defined, and then if there is no records no columns will be created

ilongin: About precedence, schema now takes precedence if it's defined

shcheklein: > default columns are created 

what are the default columns though? It probably comes back to my question re the `DEFAULT_FILE_RECORD` - I don't understand its meaning tbh.

PR #253, opened 2024-08-07
Autodetect Studio branch

dreadatour:
If Studio repo have branch with the same name PR have use it instead of hardcoded `develop` branch.

It was tested in [this PR](https://github.com/iterative/datachain/pull/252), see [this action](https://github.com/iterative/datachain/actions/runs/10290404411/job/28480287352?pr=252).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,6 +8,7 @@ on:
 
 env:
   FORCE_COLOR: "1"
+  BRANCH: ${{ github.head_ref || github.ref_name }}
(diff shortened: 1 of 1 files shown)


0x2b3bfa0 (approved):
Hacky, but...

ilongin (approved):
Thanks for fixing this

dtulga (approved):
LGTM, thanks! This will be very helpful to avoid having to change this ref for tests, then remember to change it back before merging.

PR #260, opened 2024-08-08
Added generator function to create dataset out of bucket listing

ilongin:
Goal of this PR was to add generator function that is able to list specific bucket and yield `File` objects for each bucket entry. This way we can easily create dataset out of bucket listing.

Example :
```python
ds = (
    DataChain.from_records(DataChain.DEFAULT_FILE_RECORD)
    .gen(file=list_bucket("s3://ldb-public", client_config={"anon": True}))
    .save("ldb_public_listing")
)
```


src/datachain/lib/listing.py
```diff
@@ -0,0 +1,106 @@
...
+from datachain.error import ClientError as DataChainClientError
+from datachain.lib.file import File
+
+ResultQueue = asyncio.Queue[Optional[Sequence[File]]]
+
+DELIMITER = "/"  # Path delimiter
+FETCH_WORKERS = 100
+
+
+async def _fetch_dir(client, prefix, result_queue) -> set[str]:
```
ilongin, line 19: These private functions were copied from [our fsspec client](https://github.com/iterative/datachain/blob/main/src/datachain/client/fsspec.py) and adapted for usage in listing generator

tests/func/test_listing.py
```diff
@@ -0,0 +1,34 @@
...
+from datachain.lib.listing import list_bucket
+from tests.data import ENTRIES
+
+
+def test_listing_generator(cloud_test_catalog, cloud_type):
+    ctc = cloud_test_catalog
+
+    uri = f"{ctc.src_uri}/cats"
+
+    dc = DataChain.from_records(DataChain.DEFAULT_FILE_RECORD).gen(
```
dmpetrov, line 12: How can we avoid this DEFAULT_FILE_RECORD? 🤔

tests/func/test_listing.py
```diff
@@ -0,0 +1,34 @@
...
+    )
+    assert dc.count() == 2
+
+    entries = sorted(
+        [e for e in ENTRIES if e.path.startswith("cats/")], key=lambda e: e.path
+    )
+    files = sorted(dc.collect("file"), key=lambda f: f.path)
+
+    for cat_file, cat_entry in zip(files, entries):
+        if cloud_type == "file":
```
dmpetrov, line 23: It looks like we can avoid this. Otherwise, we will be always treating local files as special type which will lead to tons of extra logic.

Ideally, the records should be compared like `assert cat_file.model_dump() == cat_entry.model_dump()` or just `assert files == entries` for the whole list.

Is it possible to achive this?

tests/unit/test_asyn.py
```diff
@@ -161,3 +161,31 @@ async def process(input):
...
+        yield 2
+
+    res = iter_over_async(gen(), get_loop())
+    assert list(res) == [1, 2]
+
+
+def test_iter_over_async_exception():
+    async def gen():
+        yield 1
+        raise Exception("Some error")
```
dmpetrov, line 178: It might be better to use more specialized exception

dmpetrov (changes requested):
Amazing ✨ 

One question about local files.

tests/func/test_listing.py
ilongin: The problem is that generator generates new rows for each row in the chain so if we have empty chain no rows will be generated. If we want some function to run on empty chain, then we would need to re-design some parts of internal generator logic I think. This needs to be investigated but I would do it in separate issue / PR as it could be big change by itself.

tests/unit/test_asyn.py
ilongin: Why is it needed here? It's just test for util function to check what happens when any exception is thrown in the code

tests/func/test_listing.py
ilongin: Yea ... that is one downside of current file system listing implementation where source is always `file:///` ... upside is that partial indexing works and no need for that `add_storage` method. I will think how to re-implement it again now with new listing in API since we should get rid of partial indexing anyway. I think the biggest question is what is exactly `source` field for listing rows .... with s3 bucket it's clear - it's the name of the bucket itself, but with local FS it's not exactly clear.

src/datachain/asyn.py
```diff
@@ -224,3 +224,23 @@ async def _end_iteration(self) -> None:
     async def _break_iteration(self) -> None:
         self.heap = []
         self._push_result(self._next_yield, None)
+
+
+def iter_over_async(ait, loop):
```
mattseddon, line 229: [Q] Should this handle exceptions in the same way as `AsyncMapper.iterate`? Is it possible/worth the effort to refactor and have `AsyncMapper.iterate` call this function?

ilongin: I don't think it makes sense to refactor `AsyncMapper.iterate()` with this method as `AsyncMapper` is implemented for different scenario where we know in advance how many tasks we expect so it's `.iterate()` method doesn't really deal with `AsyncIterator` .   
When doing bucket listing we don't know in advance how many async tasks / futures will be there as we don't know the depth of the directory tree so, as a result, we get `AsyncIterator` and this function just does simple conversion to the sync one.

Regarding exception handling, I'm not sure about that  either. I've added a test with exception and it works fine. From what I see in `AsyncMapper` we explicitly run through futures / tasks to check for exceptions but I don't see the need to do so here. I mean, I could maybe do:
```python
future = asyncio.run_coroutine_threadsafe(get_next(), loop)
if err := future.exception():
    raise err
done, obj = future.result()
...
```
but it seems to work without that as well (which is expected IMO)

dmpetrov (approved):
Looks good!

Issue #220, opened 2024-08-01
Disable WAL in SQLite


dreadatour:
Some performance measurements.

With `WAL` on:
```
$ time python wal-on.py
Listing file:///: 2 objects [00:00, 750.46 objects/s]
Processed: 2 rows [00:00, 4040.76 rows/s]
Download: 5.78GB [07:36, 13.6MB/s]rows]
Processed: 2 rows [07:36, 228.39s/ rows]
Generated: 1009874 rows [07:36, 2214.38 rows/s]
Processed: 1009874 rows [05:19, 3162.63 rows/s]
python wal-on.py  2707.84s user 439.81s system 81% cpu 1:04:03.64 total
$
```

With `WAL` off:
```
$ time python wal-off.py
Listing file:///: 2 objects [00:00, 687.08 objects/s]
Processed: 2 rows [00:00, 4148.67 rows/s]
Download: 5.78GB [06:10, 16.7MB/s]rows]
Processed: 2 rows [06:10, 185.44s/ rows]
Generated: 1009874 rows [06:10, 2727.49 rows/s]
Processed: 1009874 rows [05:15, 3196.07 rows/s]
python wal-off.py  426.31s user 217.22s system 57% cpu 18:47.79 total
$
```

19 minutes VS 1 hour.

---

Need to check/discuss pros/cons on `WAL` disabling for SQLite.

PR #211, opened 2024-07-31
Parallel UDF optimizations

dreadatour:
"Parallel" mode is not optimal now. It took more time and much more memory (script can even be killed by OOMKiller). In this PR we are going to profile and optimize parallel mode.

---

### Before optimization

Test dataset with `109662` rows:
```
$ ls -l dvcx-datacomp-small/metadata-small
total 1315952
-rw-r--r--@ 1 vlad  staff  673764696 Jul 24 17:15 036d6b9ae87a00e738f8fc554130b65b.npz
$
```

#### Single process mode

Test query `laion_emb.py`:
```python
from datachain.lib.dc import DataChain
from datachain.lib.webdataset_laion import process_laion_meta

(
    DataChain.from_storage("dvcx-datacomp-small/metadata-small/")
    .gen(emd=process_laion_meta)
    .map(stem=lambda file: file.get_file_stem(), params=["emd.file"], output=str)
    .save("laion_emb")
)
```

##### Run time

```
$ time python laion_emb.py
Processed: 1 rows [00:00, 2532.79 rows/s]
Download: 643MB [00:53, 12.7MB/s]s]
Processed: 1 rows [00:53, 53.10s/ rows]
Generated: 109662 rows [00:52, 2071.59 rows/s]
Processed: 109662 rows [00:18, 6039.80 rows/s]
python laion_emb.py  58.26s user 37.74s system 62% cpu 2:32.64 total
$
```

##### Memory consumption

Pasted image 20240806224139

#### Parallel mode

Test query `laion_emb_parallel.py`:

```python
from datachain.lib.dc import DataChain
from datachain.lib.webdataset_laion import process_laion_meta

(
    DataChain.from_storage("dvcx-datacomp-small/metadata-small/")
    .settings(cache=True, parallel=4)
    .gen(emd=process_laion_meta)
    .map(stem=lambda file: file.get_file_stem(), params=["emd.file"], output=str)
    .save("laion_emb")
)
```

##### Run time

```
$ time python laion_emb_parallel.py
Processed: 1 rows [00:00, 2236.96 rows/s]
Processed: 1 rows [00:34, 34.10s/ rows]
Generated: 109662 rows [00:34, 3216.25 rows/s]
Processed: 109662 rows [03:54, 467.85 rows/s]
python laion_emb_parallel.py  372.29s user 61.47s system 84% cpu 8:30.86 total
$
```

##### Memory consumption

Pasted image 20240806225238

#### Comparing

```
[...]

PR #261, opened 2024-08-08
Switch SQLite journal mode from 'WAL' to 'MEMORY'

dreadatour:
Update SQLite PRAGMAs:
- [journal_mode](https://www.sqlite.org/pragma.html#pragma_journal_mode) from `WAL` to `MEMORY`
- [synchronous](https://www.sqlite.org/pragma.html#pragma_synchronous) no change

---

### Some measurements

Using small dataset with 109662 rows (same as in https://github.com/iterative/datachain/pull/211)

##### journal_mode = WAL, synchronous = NORMAL

```
$ time python laion_emb.py
Processed: 1 rows [00:00, 2666.44 rows/s]
Download: 643MB [00:42, 15.9MB/s]s]
Processed: 1 rows [00:42, 42.31s/ rows]
Generated: 109662 rows [00:42, 2600.67 rows/s]
Created pre-UDF table udf_ultqNZ in 16.39s
Processed: 109662 rows [00:11, 9141.12 rows/s]
Saving result in 17.60s
python laion_emb.py  56.47s user 32.31s system 96% cpu 1:31.95 total
$
```

##### journal_mode = OFF, synchronous = NORMAL

```
$ time python laion_emb.py
Processed: 1 rows [00:00, 2427.26 rows/s]
Download: 643MB [00:39, 17.2MB/s]s]
Processed: 1 rows [00:39, 39.29s/ rows]
Generated: 109662 rows [00:39, 2799.99 rows/s]
Created pre-UDF table udf_UAa7zL in 6.25s
Processed: 109662 rows [00:13, 7909.44 rows/s]
Saving result in 6.28s
python laion_emb.py  45.90s user 21.10s system 97% cpu 1:09.02 total
$
```

##### journal_mode = MEMORY, synchronous = NORMAL

```
$ time python laion_emb.py
Processed: 1 rows [00:00, 2373.69 rows/s]
Download: 643MB [00:39, 17.0MB/s]s]
Processed: 1 rows [00:39, 39.61s/ rows]
Generated: 109662 rows [00:39, 2778.42 rows/s]
Created pre-UDF table udf_jgnpOu in 6.20s
Processed: 109662 rows [00:14, 7815.49 rows/s]
Saving result in 7.35s
python laion_emb.py  46.04s user 21.20s system 94% cpu 1:10.85 total
$
```

##### journal_mode = MEMORY, synchronous = OFF

```
$ time python laion_emb.py
Processed: 1 rows [00:00, 2750.36 rows/s]
Download: 643MB [00:36, 18.3MB/s]s]
Processed: 1 rows [00:36, 36.74s/ rows]
Generated: 109662 rows [00:36, 2994.50 rows/s]
Created pre-UDF table udf_MG3nvv in 4.76s
Processed: 109662 rows [00:18, 5947.26 rows/s]
Saving result in 6.43s
python laion_emb.py  45.99s user 17.15s system 89% cpu 1:10.46 total
$
```

### Memory consumption

Just to be sure putting journal into memory did not affect memory consumption a lot.

##### Memory with journal_mode = WAL

image

##### Memory with journal_mode = MEMORY

image

### Summary
```
journal_mode | synchronous | generating           | copying table | processing           | saving | total
-------------+-------------+----------------------+---------------+----------------------+--------+-------
WAL          | NORMAL      | 0:42, 2600.67 rows/s | 0:16          | 0:11, 9141.12 rows/s | 0:18   | 1:32
OFF          | NORMAL      | 0:39, 2799.99 rows/s | 0:06          | 0:13, 7909.44 rows/s | 0:06   | 1:09
MEMORY       | NORMAL      | 0:39, 2778.42 rows/s | 0:06          | 0:14, 7815.49 rows/s | 0:07   | 1:11
MEMORY       | OFF         | 0:36, 2994.50 rows/s | 0:05          | 0:18, 5947.26 rows/s | 0:06   | 1:10
```

Moving journal into memory have almost the same effect as turning journal OFF, but at the same time is a bit more robust.

It gives 50% performance boost for small (100k) dataset and 300% boost for big (1M) dataset (see measurements in the original issue here: https://github.com/iterative/datachain/issues/220).

### DIFF
--- a/src/datachain/data_storage/sqlite.py
+++ b/src/datachain/data_storage/sqlite.py
@@ -127,7 +127,7 @@ class SQLiteDatabaseEngine(DatabaseEngine):
             db.execute("PRAGMA foreign_keys = ON")
             db.execute("PRAGMA cache_size = -102400")  # 100 MiB
             # Enable Write-Ahead Log Journaling
-            db.execute("PRAGMA journal_mode = WAL")
+            db.execute("PRAGMA journal_mode = MEMORY")
             db.execute("PRAGMA synchronous = NORMAL")
             db.execute("PRAGMA case_sensitive_like = ON")
             if os.environ.get("DEBUG_SHOW_SQL_QUERIES"):


shcheklein:
So, from what I'm reading:

https://www.sqlite.org/howtocorrupt.html#cfgerr

```
If protections are disabled, database corruption may occur.
```

```
Using PRAGMA journal_mode=OFF or PRAGMA journal_mode=MEMORY and taking an application crash in the middle of a write transaction.
```

so, I'm not sure if this is safe tbh.

--------------

Q: do we understand why exactly we are getting this boost?

E.g. we just run a lot of small transactions (instead of a single bulk insert)? Do we know these numbers - like transaction numbers?

We probably also create quite a lot of intermediate tables - that do not affect the end result (?). That we even drop as part of the query execution. Is it true?

Can we use the second DB file (and disable WAL, etc) for that one - so we are fine it goes corrupted? And write the final result or `save()` in the main file properly?

dreadatour:
> If protections are disabled, database corruption may occur.

Yes, this is the main downside of this change.

> so, I'm not sure if this is safe tbh.

It is not, but it may be worth it. That's the main reason of having this discussion.

Also in SQLite docs here: https://www.sqlite.org/wal.html we can see:

> WAL might be very slightly slower (perhaps 1% or 2% slower) than the traditional rollback-journal approach in applications that do mostly reads and seldom write.

I was also trying to run exactly the same queries from SQLite console and see exactly the same results and timings.

> Q: do we understand why exactly we are getting this boost?
> 
> E.g. we just run a lot of small transactions (instead of a single bulk insert)? Do we know these numbers - like transaction numbers?

From "Summary" section we can see the main boost is in `copying table` and `save` steps. This is how we prepare data for UDF, we basically create new table and then make `insert from select` into this table. This is the main place where disabling WAL gives a huge performance boost. Also we don't have any progress bar here so it looks like query was stuck doing nothing.

> We probably also create quite a lot of intermediate tables - that do not affect the end result (?). That we even drop as part of the query execution. Is it true?
> 
> Can we use the second DB file (and disable WAL, etc) for that one - so we are fine it goes corrupted? And write the final result or `save()` in the main file properly?

We can, yes. This is one of the future performance optimizations approach we can find in the "performance optimizations" epic.

At the same time disabling WAL looks like low-hanging fruit for now, 50% boost for 100k dataset and 300% boost for 1 million rows dataset may worth it for now, also it will be easy to revert this change after we do more complex optimizations and improvements.

dreadatour:
Here is an example of SQL query from "copying table" step:

```sql
CREATE TABLE IF NOT EXISTS "udf_eWKkak" (
	sys__id INTEGER NOT NULL,
	sys__rand INTEGER,
	emd__file__source VARCHAR,
	emd__file__path VARCHAR,
	emd__file__size INTEGER,
	emd__file__version VARCHAR,
	emd__file__etag VARCHAR,
	emd__file__is_latest BOOLEAN,
	emd__file__last_modified DATETIME,
	emd__file__location JSON,
	emd__file__vtype VARCHAR,
	emd__index INTEGER,
	emd__b32_img ARRAY,
	emd__b32_txt ARRAY,
	emd__l14_img ARRAY,
	emd__l14_txt ARRAY,
	emd__dedup ARRAY,
	PRIMARY KEY (sys__id)
)


INSERT INTO "udf_eWKkak" (sys__rand, emd__file__source, emd__file__path, emd__file__size, emd__file__version, emd__file__etag, emd__file__is_latest, emd__file__last_modified, emd__file__location, emd__file__vtype, emd__index, emd__b32_img, emd__b32_txt, emd__l14_img, emd__l14_txt, emd__dedup) SELECT anon_1.sys__rand AS sys__rand, anon_1.emd__file__source AS emd__file__source, anon_1.emd__file__path AS emd__file__path, anon_1.emd__file__size AS emd__file__size, anon_1.emd__file__version AS emd__file__version, anon_1.emd__file__etag AS emd__file__etag, anon_1.emd__file__is_latest AS emd__file__is_latest, anon_1.emd__file__last_modified AS emd__file__last_modified, anon_1.emd__file__location AS emd__file__location, anon_1.emd__file__vtype AS emd__file__vtype, anon_1.emd__index AS emd__index, anon_1.emd__b32_img AS emd__b32_img, anon_1.emd__b32_txt AS emd__b32_txt, anon_1.emd__l14_img AS emd__l14_img, anon_1.emd__l14_txt AS emd__l14_txt, anon_1.emd__dedup AS emd__dedup
FROM (SELECT "udf_N3EBeZ".sys__id AS sys__id, "udf_N3EBeZ".sys__rand AS sys__rand, "udf_N3EBeZ".emd__file__source AS emd__file__source, "udf_N3EBeZ".emd__file__path AS emd__file__path, "udf_N3EBeZ".emd__file__size AS emd__file__size, "udf_N3EBeZ".emd__file__version AS emd__file__version, "udf_N3EBeZ".emd__file__etag AS emd__file__etag, "udf_N3EBeZ".emd__file__is_latest AS emd__file__is_latest, "udf_N3EBeZ".emd__file__last_modified AS emd__file__last_modified, "udf_N3EBeZ".emd__file__location AS emd__file__location, "udf_N3EBeZ".emd__file__vtype AS emd__file__vtype, "udf_N3EBeZ".emd__index AS emd__index, "udf_N3EBeZ".emd__b32_img AS emd__b32_img, "udf_N3EBeZ".emd__b32_txt AS emd__b32_txt, "udf_N3EBeZ".emd__l14_img AS emd__l14_img, "udf_N3EBeZ".emd__l14_txt AS emd__l14_txt, "udf_N3EBeZ".emd__dedup AS emd__dedup
FROM "udf_N3EBeZ") AS anon_1
```

This particular query looks redundant (it is copying one table to the same another table) and this is also one of the future optimizations, but this step may be more complex and in general we can not easily avoid it.

shcheklein:
> It is not, but it may be worth it.

yep, but how? I am still no sure why it can be possible tbh.

> We can, yes. This is one of the future performance optimizations approach we can find in the "performance optimizations" epic.

so, sounds that it would give us the same result. Also doesn't sound like something very expensive to implement (clearly need a bit of research)

> At the same time disabling WAL looks like low-hanging fruit for now, 50% boost for 100k dataset and 300% boost for 1 million rows dataset may worth it for now, also it will be easy to revert this change after we do more complex optimizations and improvements.

Again, either I'm missing something, or I don't understand why it can be possible to do.

dmpetrov:
Folks, we might be overthinking this 🙂 

We are not dealing with OLTP use cases and corruption is not as a big concern for us. DB can be corrupted anyway - WAL just increases a chance of recovering.

I'd suggest to go ahead and turn this off (in memory - might be). Especially given the a decent chance that DB will switch to duckdb or another OLAP DB instead of SQLite.

shcheklein:
> We are not dealing with OLTP use cases and corruption is not as a big concern for us.

just to clarify:

I'm assuming that we might lost the saved datasets with this (people won't be able to even launch DataChain w/o deleting the corrupted file

shcheklein:
Hmm, I'm still quire concerned tbh or I'm missing something.

> Especially given the a decent chance that DB will switch to duckdb or another OLAP DB instead of SQLite.

Duckdb - I'm pretty sure has WAL and does pretty much the same (?).

At then end of the day _any_ DB has to have a way to put stuff back to disk in a reliable way so that data is persistent and do this is a way that it's consistent and internal data structures are consistent. We are pretty much disabling all of that AFAIU (?).

> DB can be corrupted anyway 

it's extremely rare and should be considered a bug in my experience (or wrong hardware setup -e..g some HDDs and controllers don't do sync, etc).

------

Can it be the case that in our specific workload it can be safe? - it can be, but requires way better understanding on how WAL and sync works in SQLite.

Losing saved datasets can be a really bad experience for users though.

dreadatour:
One more measurement. `journal_mode = WAL`, 1 million rows:

```
Generated: 1009874 rows [07:26, 2264.26 rows/s]
Created pre-UDF table udf_P0OyWS in 1389.55s
Processed: 1009874 rows [04:41, 3591.48 rows/s]
```

~23 minutes (1389 seconds) took "Created pre-UDF table" step which is execution of this query only: https://github.com/iterative/datachain/pull/261#issuecomment-2276327022

shcheklein:
I wonder if we can disable / tune checkpoints https://www.sqlite.org/wal.html#ckpt and do it once at the end (or on each save) - it can be also a safer alternative (can result in a memory growth though?)

dreadatour:
> I wonder if we can disable / tune checkpoints https://www.sqlite.org/wal.html#ckpt and do it once at the end (or on each save) - it can be also a safer alternative (can result in a memory growth though?)

It is single query processing by db. I am checking now if it is possible to split this query into smaller chunks.

dmpetrov (approved):
Looks good!
Please create a followup issue for turning this on. Like `.settings(journal=True)`

skshetry:
Dropping WAL should be the last thing that we should do.  There are still some things to investigate, like removing `auto_commit`before doing this.

Also, ROLLBACK journals have to lock the database for every read operation, unlike WAL that allows concurrent read/write access. 

So, could you please share a benchmark/profiling result that shows where the time is being spent with MEMORY/WAL, and how much difference this makes on read and write compared to WAL?

I'll also suggest doing native profiling (with `py-spy` or other tools) to see where the time is being spent on sqlite3. Native profiling does not work well on macOS AFAIR.

dreadatour:
Closing this PR as it looks like https://github.com/iterative/datachain/pull/278 is a better approach.

PR #263, opened 2024-08-08
Remove legacy columns

rlamy:
This removes the following attributes or columns from various internal classes and tables: `vtype`, `dir_type`, `owner_name`, `owner_id`. Note that `Node.dir_type` is still used though.

The publicly visible, backwards-incompatible changes are:
* `File.vtype` is removed
* Existing listings table become invalid.

Part of #36


shcheklein (approved):
Looks good to me.

Do we need to invalidate listing on the Studio side or remove them? Does it also invalidate datasets (do they have these file fields inherited from listings)?

rlamy:
@shcheklein This makes existing listing tables invalid, but I think that's irrelevant after #294. For datasets, the only change is that `File.vtype` is removed, but it wasn't used before so I don't think it'll cause any issue. Old dataset tables will just have an extra "file__vtype" column that should be ignored everywhere (UI, queries, ...).

shcheklein:
@rlamy are we good to go here? :)

rlamy:
@shcheklein The datachain part works, but we need the Studio tests to pass...

PR #274, opened 2024-08-11
test hf computer vision example

mattseddon:


examples/computer_vision/blip2_image_desc_lib.py
```diff
@@ -1,100 +0,0 @@
-# pip install torch
```
mattseddon, line 1: [F] This example was very close to the retained one. It even included an unused version of `LLaVADescribe`.

mattseddon:
Will never finish within the required amount of time. We should consider deleting these.

PR #277, opened 2024-08-12
fix unstructured-text example

mattseddon:
This PR fixes the unstructured-text nlp example. Tl;dr is that the new version of `nltk` does not play nicely with `unstructured`.

### DIFF
--- a/examples/llm_and_nlp/unstructured-text.py
+++ b/examples/llm_and_nlp/unstructured-text.py
@@ -1,11 +1,6 @@
 #
-# pip install unstructured[all-docs] huggingface_hub[hf_transfer]
+# pip install unstructured[pdf] nltk==3.8.1 huggingface_hub[hf_transfer]
 #
-# partition_object supports via unstructured library:
-#
-# "csv", "doc", "docx", "epub", "image", "md", "msg", "odt", "org",
-# "pdf", "ppt", "pptx", "rtf", "rst", "tsv", "xlsx"
-
 import os
 
 os.environ["HF_HUB_ENABLE_HF_TRANSFER"] = "1"
--- a/examples/llm_and_nlp/unstructured-text.py
+++ b/examples/llm_and_nlp/unstructured-text.py
@@ -18,6 +18,7 @@ nltk.download("averaged_perceptron_tagger_eng")
 
 
 from transformers import pipeline
+from unstructured.partition.pdf import PartitionStrategy
 from unstructured.partition.pdf import partition_pdf as partition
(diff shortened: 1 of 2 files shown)


examples/llm_and_nlp/unstructured-text.py
```diff
@@ -39,9 +45,9 @@ def summarize(clean):
     return (summary,)
 
 
-ds = (
+dc = (
     DataChain.from_storage(source)
-    .filter(C("file.path").glob("*.pdf"))
+    .settings(cache=True)
```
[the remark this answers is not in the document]
mattseddon: and so it throws `ValueError: Invalid file-like object type`

examples/llm_and_nlp/unstructured-text.py
```diff
@@ -1,26 +1,39 @@
...
 # "pdf", "ppt", "pptx", "rtf", "rst", "tsv", "xlsx"
 
+import os
+
+os.environ["HF_HUB_ENABLE_HF_TRANSFER"] = "1"
+
+import nltk
+
+nltk.download("punkt_tab")
+nltk.download("averaged_perceptron_tagger_eng")
```
mattseddon, line 17: [F] These are needed for the "fast" strategy which is the only strategy supported without installing `poppler` or other non-easy-to-install across all platform tools.

mattseddon: edit: a new version of `nltk` came out and that meant that `unstructured` could no longer download these files by itself. I have pinned `nltk` back to the working version.

examples/llm_and_nlp/unstructured-text.py
```diff
@@ -1,26 +1,27 @@
...
-# libmagic
+# pip install unstructured[pdf] nltk==3.8.1 huggingface_hub[hf_transfer]
 #
-# partition_object supports via unstructured library:
-#
-# "csv", "doc", "docx", "epub", "image", "md", "msg", "odt", "org",
-# "pdf", "ppt", "pptx", "rtf", "rst", "tsv", "xlsx"
+import os
+
+os.environ["HF_HUB_ENABLE_HF_TRANSFER"] = "1"
```
mattseddon, line 6: [F] `pipeline` downloads `pszemraj/led-large-book-summary` so we might as well get it as quickly as possible

mattseddon:
Alternatively, we can wait for https://github.com/Unstructured-IO/unstructured/pull/3512 + the next release

dtulga (approved):
LGTM, thanks for fixing this!

PR #283, opened 2024-08-13
Adding In-Memory DataChain Option

dtulga:
This adds an option for in-memory-only databases on DataChain (and on DatasetQuery as well), as described in iterative/dvcx#1719 - although with the option on creation, instead of settings, as the database can't be changed after creation. This is accomplished by adding this option to Session, and setting this automatically on DataChain creation, so users don't need to specify this on Session (unless they want to). This means that any other DataChain objects that use the same session will also automatically be in-memory. This also changes more tests to use in-memory databases, by utilizing the `test_session` fixture.

This has been tested locally, and fixes iterative/dvcx#1719


src/datachain/catalog/loader.py
```diff
@@ -43,25 +43,31 @@ def get_id_generator() -> "AbstractIDGenerator":
...
-        id_generator_class = SQLiteIDGenerator
+    id_generator_args: dict[str, Any] = {
+        k.lower(): v for k, v in id_generator_arg_envs.items()
+    }
+
+    if not id_generator_import_path:
+        id_generator_args["in_memory"] = in_memory
+        return SQLiteIDGenerator(**id_generator_args)
+    if in_memory:
+        raise RuntimeError("In-memory databased are only supported on SQLite")
```
mattseddon, line 54: [Q] databased?

mattseddon: looks like the same string has been copied elsewhere, not sure what the intended sentence is. Maybe something like `in_memory is only supported by SQLite` :shrug:

src/datachain/data_storage/sqlite.py
```diff
@@ -391,6 +396,9 @@ def __init__(
         # foreign keys
         self.default_table_names: list[str] = []
 
+        if in_memory:
```
mattseddon, line 399: [Q] is it possible that a `db_file` has been provided and has been overwritten? Should an error be thrown under those circumstances?

tests/unit/lib/test_datachain.py
```diff
@@ -786,6 +859,20 @@ def test_parse_tabular(tmp_dir, test_session):
     assert df1.equals(df)
 
 
+@skip_if_not_sqlite
+def test_parse_tabular_in_memory(tmp_dir):
```
mattseddon, line 863: [Q] Should all the different `from_` methods be tested individually for `in_memory`?

mattseddon (approved):

src/datachain/catalog/loader.py
dtulga: Whoops, that's a typo - fixed it, and made this error message a constant to avoid duplication.

src/datachain/data_storage/sqlite.py
dtulga: Yep, added a function to check for this error case, and reduce code duplication, and added a test for this as well.

tests/unit/lib/test_datachain.py
dtulga: I tested all the methods where a Session can be specified directly, which includes `from_storage`, `datasets`, `from_values`, `from_pandas`, and `from_records`. I also added `in_memory` tests for `from_csv` and `from_parquet`. The other methods all appear to either not allow a session to be specified (which is where `in_memory` is implemented), or pass the `session` and `in_memory` args through a generic `**kwargs` argument into another method that is already tested, such as `from_json` and `from_jsonl` pass to `from_storage` for example.

dreadatour (approved):
Looks good to me! Thank you for this update ❤️

PR #293, opened 2024-08-14
add skip_if_not_sqlite to failing Studio tests

mattseddon:
alternative to https://github.com/iterative/datachain/pull/292

~TBH I would probably pick the above PR given the choice~

I'll revert this after #296

### DIFF
--- a/tests/unit/lib/test_datachain_merge.py
+++ b/tests/unit/lib/test_datachain_merge.py
@@ -46,8 +46,8 @@ team = [
 ]
 
 
+@skip_if_not_sqlite
 def test_merge_objects(test_session):
-    skip_if_not_sqlite()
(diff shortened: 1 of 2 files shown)


tests/unit/lib/test_datachain_merge.py
```diff
@@ -46,8 +46,8 @@ class TeamMember(BaseModel):
 ]
 
 
+@skip_if_not_sqlite
 def test_merge_objects(test_session):
-    skip_if_not_sqlite()
```
mattseddon, line 50: [F] The need to update to a decorator must've got lost in a merge

dberenbaum (approved):

mattseddon:
was going to merge to unblock myself but that could create more work for others

PR #294, opened 2024-08-14
Refactor `DataChain.from_storage()` to use new listing generator

ilongin:
Refactoring `DataChain.from_storage()` to use new listing generator instead of calling underlying `DatasetQuery` indexing step which uses deprecated listing codebase.

For each listing requested in `.from_storage()` method, new dataset will be created with `lst_` prefix and uri, e.g
`DataChain.from_storage("s3://ldb-public/dogs-cats")` -> this will create dataset with name `lst_s3://ldb-public/dogs-cats` which holds listing data.

Schema of listing dataset will always be `{"file": "File@v1"}`, but user can specify specific type (e,g `ImageFile` or `TextFile`) by adding argument `type` e.g `.from_storage("s3://ldb-public", type="text")` which would return chain with schema consisting of that specific type instead of generic `File`. This way we can, for example, filter out images from chain and set correct `image` type, e.g 
```python
# listing is done in first from_storage() call where schema is File but reading it as ImageFile
img_dc = DataChain.from_storage("gs://datachain-demo", type="image").filter(C("file.path").glob("*.jpg"))
# using cached listing from above but reading it as TextFile this time
txt_dc = DataChain.from_storage("gs://datachain-demo", type="text").filter(C("file.path").glob("*.txt"))  
```

Partials listing is also done without partials table which makes them obsolete.
Storage table is also obsolete as now all information is in that special listing dataset from above.

Next steps:

- Make listing caching lazy
- Refactor `CLI` to use `DataChain.from_storage()` to enlist sources
- Remove obsolete codebase


mattseddon:
`test_from_storage_reindex_expired[tree0-file-False]` seems to be a legit failure on Windows

src/datachain/lib/dc.py
```diff
@@ -305,21 +318,23 @@ def add_schema(self, signals_schema: SignalSchema) -> "Self":  # noqa: D102
     @classmethod
     def from_storage(
         cls,
-        path,
+        uri,
```
dreadatour, line 321: 👍

src/datachain/lib/dc.py
```diff
@@ -305,21 +318,23 @@ def add_schema(self, signals_schema: SignalSchema) -> "Self":  # noqa: D102
...
         *,
         type: Literal["binary", "text", "image"] = "binary",
         session: Optional[Session] = None,
         in_memory: bool = False,
         recursive: Optional[bool] = True,
         object_name: str = "file",
         update: bool = False,
+        client_config=None,
+        anon: bool = False,
```
dreadatour, line 330: Add info to docstrings about these two params?

src/datachain/lib/dc.py
```diff
@@ -411,7 +484,7 @@ def jmespath_to_name(s: str):
             object_name = jmespath_to_name(jmespath)
         if not object_name:
             object_name = meta_type
-        chain = DataChain.from_storage(path=path, type=type, **kwargs)
+        chain = DataChain.from_storage(uri=path, type=type, **kwargs)
```
dreadatour, line 487: Rename `path` param to `uri` as well?

src/datachain/lib/listing.py
```diff
@@ -1,19 +1,31 @@
...
+if TYPE_CHECKING:
+    from datachain.lib.dc import DataChain
+
 
 ResultQueue = asyncio.Queue[Optional[Sequence[File]]]
 
 DELIMITER = "/"  # Path delimiter
 FETCH_WORKERS = 100
+LISTING_TTL = 4 * 60 * 60  # cached listing lasts 4 days
+LISTING_PREFIX = "lst__"  # listing datasets start with this name
```
dreadatour, line 28: We have `LISTING_PREFIX` in three places for now:
- `src/datachain/dataset.py`
- `src/datachain/lib/dc.py`
- and here
Just wanted to be sure we don't want to have all these prefix constants in one place 👀

dreadatour (approved):
Looks good to me 👍 Awesome job! ❤️

src/datachain/dataset.py
```diff
@@ -443,7 +444,10 @@ def is_bucket_listing(self) -> bool:
         For bucket listing we implicitly create underlying dataset to hold data. This
         method is checking if this is one of those datasets.
         """
-        return Client.is_data_source_uri(self.name)
+        # TODO remove first part of condition when we remove legacy indexing / listing
```
shcheklein, line 447: put it an item in the cleanup ticket?

tests/unit/lib/test_signal_schema.py
```diff
@@ -356,6 +357,24 @@ def test_slice_nested():
     assert list(sliced.values.items()) == [("feature.aa", int)]
 
 
+def test_mutate_rename():
```
shcheklein, line 360: are these tests related?

tests/scripts/feature_class_parallel.py
```diff
@@ -18,7 +18,7 @@ class Embedding(BaseModel):
 ds_name = "feature_class"
 ds = (
     DataChain.from_storage("gs://dvcx-datalakes/dogs-and-cats/")
-    .filter(C.path.glob("*cat*.jpg"))  # type: ignore [attr-defined]
+    .filter(C("file.path").glob("*cat*.jpg"))  # type: ignore [attr-defined]
```
shcheklein, line 21: Q: why do we have these changes?

src/datachain/lib/dc.py
```diff
@@ -331,22 +346,80 @@ def from_storage(
...
+
+        client, path = Client.parse_url(
+            uri, session.catalog.cache, **session.catalog.client_config
+        )
+
+        # clean path without globs
+        lst_path = (
+            posixpath.dirname(path)
+            if glob.has_magic(os.path.basename(os.path.normpath(path)))
+            or client.fs.isfile(uri)
```
shcheklein, line 372: can we now treat file as a single entry dataset? (and avoid indexing the whole parent bucket) - can be a followup, but seems an important one

src/datachain/lib/dc.py
```diff
@@ -331,22 +346,80 @@ def from_storage(
...
+            else path
+        )
+        lst_uri = f"{client.uri}/{lst_path.lstrip('/')}"
+
+        ds_name = listing_dataset_name(client.uri, posixpath.join(lst_path, ""))
+
+        for ds in cls.datasets(
```
shcheklein, line 379: does it return all datasets + listsings? can we return _only_ listings?

otherwise, code like `listing_expired(ds.created_at)` below look a bit strange if it's applied to a regular datasets

src/datachain/lib/dc.py
```diff
@@ -331,22 +346,80 @@ def from_storage(
...
+        ).collect("dataset"):
+            if listing_expired(ds.created_at):  # type: ignore[union-attr]
+                continue
+            if (
+                listing_subset(ds.name, ds_name)  # type: ignore[union-attr]
+                and not update
+            ):
+                # we can use found listing as it contains the one from input
+                dc = cls.from_dataset(ds.name, **kwargs)  # type: ignore[union-attr]
+                dc.signals_schema = dc.signals_schema.mutate(
```
shcheklein, line 390: Q: why is this needed?

src/datachain/lib/dc.py
```diff
@@ -331,22 +346,80 @@ def from_storage(
...
+            session=session, in_memory=in_memory, include_listing=True
+        ).collect("dataset"):
+            if listing_expired(ds.created_at):  # type: ignore[union-attr]
+                continue
+            if (
+                listing_subset(ds.name, ds_name)  # type: ignore[union-attr]
+                and not update
+            ):
+                # we can use found listing as it contains the one from input
+                dc = cls.from_dataset(ds.name, **kwargs)  # type: ignore[union-attr]
```
shcheklein, line 389: code looks very repetitive with the https://github.com/iterative/datachain/pull/294/files#diff-0b1cfd3d89c057cc57811e07eb20298eeaaee60c54cc66e5128f618d144f42eeR410-R413 - can we please try to use the same?

otherwise it raises questions right away - e.g. why in one case we pass sessions, and don't do this in the second case, etc 

+ plus it'll be harder to main

src/datachain/lib/dc.py
```diff
@@ -331,22 +346,80 @@ def from_storage(
...
+
+        return ls(dc, path, recursive=recursive, object_name=object_name)
+
     @classmethod
-    def from_dataset(cls, name: str, version: Optional[int] = None) -> "DataChain":
+    def from_dataset(
+        cls,
+        name: str,
+        version: Optional[int] = None,
+        session: Optional[Session] = None,
```
shcheklein, line 420: why do we need session now here?

src/datachain/lib/meta_formats.py
```diff
@@ -54,10 +54,10 @@ def read_schema(source_file, data_type="csv", expr=None, model_name=None):
     try:
         with source_file.open() as fd:  # CSV can be larger than memory
             if data_type == "csv":
-                data_string += fd.readline().decode("utf-8", "ignore").replace("\r", "")
```
shcheklein, line 57: seems a bit unrelated? just to make sure?

tests/conftest.py
```diff
@@ -173,8 +173,8 @@ def catalog(id_generator, metastore, warehouse):
 
 
 @pytest.fixture
-def test_session(catalog):
-    with Session("TestSession", catalog=catalog) as session:
+def test_session(cloud_test_catalog):
```
shcheklein, line 176: why is this change?

src/datachain/lib/listing.py
```diff
@@ -97,15 +109,68 @@ async def _scandir(client, prefix, fetch_workers) -> AsyncIterator:
...
+
+        # path has glob syntax so we are returning glob filter
+        return dc.filter(_file_c("path").glob(path))
+    # returning only first level children by path
+    return dc.filter(pathfunc.parent(_file_c("path")) == path.lstrip("/").rstrip("/*"))
+
+
+def listing_dataset_name(uri: str, path: str) -> str:
+    """Returns special listing dataset name from uri and path"""
+    return f"{LISTING_PREFIX}{uri}/{path.lstrip('/')}"
```
shcheklein, line 164: important: does dataset name always end with `/` or not? we need put some logic to always check this rule and / or assert on it (it affects the listing_subset - if we always have `/` at the end `startswith` might be enough, otherwise it should be a different implementation

src/datachain/lib/listing.py
```diff
@@ -97,15 +109,68 @@ async def _scandir(client, prefix, fetch_workers) -> AsyncIterator:
...
+def listing_expired(created_at: datetime) -> bool:
+    """Checks if listing has expired based on it's creation date"""
+    return datetime.now(timezone.utc) > created_at + timedelta(seconds=LISTING_TTL)
+
+
+def listing_subset(ds1_name: str, ds2_name: str) -> bool:
+    """
+    Checks if one listing contains another one by comparing corresponding dataset names
+    """
+    return ds2_name.startswith(ds1_name)
```
shcheklein, line 176: let's put asserts that we have both names end with `/` - otherwise this implementation is broken I think

src/datachain/lib/listing.py
```diff
@@ -97,15 +109,68 @@ async def _scandir(client, prefix, fetch_workers) -> AsyncIterator:
...
+    Return files by some path from DataChain instance which contains bucket listing.
+    Path can have globs.
+    If recursive is set to False, only first level children will be returned by
+    specified path
+    """
+
+    def _file_c(name: str) -> Column:
+        return Column(f"{object_name}.{name}")
+
+    dc = dc.filter(_file_c("is_latest") == true())
```
shcheklein, line 141: Q: what is `is_latest`?

src/datachain/lib/listing.py
```diff
@@ -97,15 +109,68 @@ async def _scandir(client, prefix, fetch_workers) -> AsyncIterator:
...
+            # file and we are adding proper filter for it
+            return dc.filter(
+                (_file_c("path") == path)
+                | (_file_c("path").glob(path.rstrip("/") + "/*"))
+            )
+
+        # path has glob syntax so we are returning glob filter
+        return dc.filter(_file_c("path").glob(path))
+    # returning only first level children by path
+    return dc.filter(pathfunc.parent(_file_c("path")) == path.lstrip("/").rstrip("/*"))
```
shcheklein, line 159: why do we strip only `/*` - can it be a different glob there?

src/datachain/lib/listing.py
```diff
@@ -97,15 +109,68 @@ async def _scandir(client, prefix, fetch_workers) -> AsyncIterator:
...
+        return Column(f"{object_name}.{name}")
+
+    dc = dc.filter(_file_c("is_latest") == true())
+
+    if recursive:
+        if not path or path == "/":
+            # root of a bucket, returning all latest files from it
+            return dc
+
+        if not glob.has_magic(os.path.basename(os.path.normpath(path))):
```
shcheklein, line 148: seems we can move into some util this stuff `glob.has_magic(os.path.basename(os.path.normpath(path)))` and test it

src/datachain/lib/listing.py
shcheklein: should we also return glob filter here even if it's nor recursive?

src/datachain/lib/dc.py
dmpetrov: yes, we need session and we don't need kwargs. See my other comments on Session, Catalog and client_config. In short - we can afford only one config object in user API, not 3.

src/datachain/lib/dc.py
```diff
@@ -331,22 +346,80 @@ def from_storage(
...
+        **kwargs,
```
dmpetrov, line 421: PLease get rid of kwargs (I assume these are used for config and catalog which we cannot expose to users).

src/datachain/lib/dc.py
```diff
@@ -491,6 +564,7 @@ def datasets(
         session: Optional[Session] = None,
         in_memory: bool = False,
         object_name: str = "dataset",
+        **kwargs,
```
dmpetrov, line 567: remove kwargs

tests/func/test_datachain.py
```diff
@@ -25,34 +33,157 @@ def test_catalog_anon(tmp_dir, catalog, anon):
...
+    assert DataChain.from_storage(path, session=test_session).count() == 1
 
     pd.DataFrame({"name": ["Charlie", "David"]}).to_parquet(tmp_dir / "test2.parquet")
-    assert DataChain.from_storage(path, catalog=catalog).count() == 1
-    assert DataChain.from_storage(path, catalog=catalog, update=True).count() == 2
+    assert DataChain.from_storage(path, session=test_session).count() == 1
+    assert DataChain.from_storage(path, session=test_session, update=True).count() == 2
+
+
+@pytest.mark.parametrize(
```
dmpetrov, line 78: Could you please explain why these settings are needed? :)

tests/func/test_datachain.py
```diff
@@ -25,34 +33,157 @@ def test_catalog_anon(tmp_dir, catalog, anon):
...
+
+    DataChain.from_storage(f"{src_uri}/cats", session=test_session)
+    assert _datasets(test_session) == sorted(
+        [
+            f"{listing_dataset_name(src_uri, 'dogs/')}@v1",
+            f"{listing_dataset_name(src_uri, '')}@v1",
+        ]
+    )
+
+    DataChain.from_storage(f"{src_uri}/cats", session=test_session, update=True)
```
dmpetrov, line 153: I'd suggest to extract the updates to a separate test since these logic can change independently

tests/unit/lib/test_signal_schema.py
```diff
@@ -3,17 +3,18 @@
 from typing import Optional, Union
 
 import pytest
-from sqlalchemy import Column
+import sqlalchemy as sa
```
dmpetrov, line 6: Why is this needed? 

I'd try to use our column `from datachain import Column` because we can eventuallly switch to a custom Column.

tests/unit/lib/test_signal_schema.py
```diff
@@ -3,17 +3,18 @@
...
 from datachain.lib.convert.flatten import flatten
-from datachain.lib.file import File
+from datachain.lib.file import File, TextFile
 from datachain.lib.signal_schema import (
     SetupError,
     SignalResolvingError,
     SignalSchema,
     SignalSchemaError,
 )
+from datachain.query.schema import Column
```
dmpetrov, line 17: This should be ` from datachain import Column`, right?

tests/unit/lib/test_signal_schema.py
```diff
@@ -296,7 +297,7 @@ def test_db_signals_as_columns():
     spec = {"name": str, "age": float, "fr": MyType2}
     lst = list(SignalSchema(spec).db_signals(as_columns=True))
 
-    assert all(isinstance(s, Column) for s in lst)
+    assert all(isinstance(s, sa.Column) for s in lst)
```
dmpetrov, line 300: Any reason why regular `datachain.Column` does not work?

tests/unit/lib/test_signal_schema.py
```diff
@@ -356,6 +357,24 @@ def test_slice_nested():
     assert list(sliced.values.items()) == [("feature.aa", int)]
 
 
+def test_mutate_rename():
+    schema = SignalSchema({"name": str})
+    schema = schema.mutate({"new_name": Column("name")})
```
dmpetrov, line 362: why not a standard way `mutate(new_name=Column("name"))` ?

tests/unit/lib/test_signal_schema.py
```diff
@@ -356,6 +357,24 @@ def test_slice_nested():
...
+    assert schema.values == {"new_name": str}
+
+
+def test_mutate_new_signal():
+    schema = SignalSchema({"name": str})
+    schema = schema.mutate({"age": Column("age", Float)})
```
dmpetrov, line 368: It should throw an exception (missing signal), shouldn't it?

src/datachain/lib/signal_schema.py
```diff
@@ -311,6 +311,10 @@ def mutate(self, args_map: dict) -> "SignalSchema":
                 # renaming existing signal
                 del new_values[value.name]
                 new_values[name] = self.values[value.name]
+            elif name in self.values:
```
dmpetrov, line 314: Is it a part of this PR? Could you please explain why this is needed?

tests/unit/lib/test_signal_schema.py
```diff
@@ -356,6 +357,24 @@ def test_slice_nested():
...
+    assert schema.values == {"name": str, "age": float}
+
+
+def test_mutate_change_type():
+    schema = SignalSchema({"name": str, "age": float, "f": File})
+    schema = schema.mutate({"age": int, "f": TextFile})
```
dmpetrov, line 374: It should be like `mutate(age=Column("age", int), f=Column("f", TextFile))`

It will require introducing our constructor for Column that does the type conversion to sqlalchemy types. Like:
```
class Column(sa.Column):
   def __init__(self, name, type):
      super().__init__(self, name, python_to_sql(type))
```

src/datachain/lib/signal_schema.py
dmpetrov: I'd suggest avoiding tricks like `mutate(aa=Float)` - it exposes internals to users.

src/datachain/lib/listing.py
```diff
@@ -97,15 +109,68 @@ async def _scandir(client, prefix, fetch_workers) -> AsyncIterator:
...
+    # returning only first level children by path
+    return dc.filter(pathfunc.parent(_file_c("path")) == path.lstrip("/").rstrip("/*"))
+
+
+def listing_dataset_name(uri: str, path: str) -> str:
+    """Returns special listing dataset name from uri and path"""
+    return f"{LISTING_PREFIX}{uri}/{path.lstrip('/')}"
+
+
+def listing_expired(created_at: datetime) -> bool:
```
dmpetrov, line 167: please add `is_` prefix to all the bool functions.

src/datachain/lib/dc.py
dmpetrov, line 330: `anon` is ok to keep but we should think on how to use only Session

src/datachain/lib/dc.py
```diff
@@ -331,22 +346,80 @@ def from_storage(
...
+            client_config=client_config,
+            in_memory=in_memory,
+        )
+
+        client, path = Client.parse_url(
+            uri, session.catalog.cache, **session.catalog.client_config
+        )
+
+        # clean path without globs
+        lst_path = (
```
dmpetrov, line 369: is it possible to extract all the logic to lib/listing.py? dc.py is becoming an unmanageble fat file

tests/unit/lib/test_signal_schema.py
dmpetrov: If that's not possible, we can consider introducing factory methods like Column.create("name", int)

dmpetrov (changes requested):
@ilongin great change! Listing looks really good in app level and I like the simplified logic on partial resolving!

A couple changes are needed:
1. Avoid using `catalog`, `config` and `kwargs`. We are intoroducing 3 redundant ways of settings for users - not good. We have to stick with one. I think it's `session`.
2. Type machinery in mutate() exposes internal types to users. We should to avoid this. And notation mutate(my_col=int) does not look obvious.

An extra discussion point (likely not in this change): how to remove these methods from DataChain class - it's becoming too fat.

src/datachain/lib/dc.py
ilongin: This is another method so I would like to leave this to another PR / refactoring as this one is already too big ... but I agree with you, it should be `uri` there as well

src/datachain/lib/listing.py
ilongin: I've removed them from `dc.py` and about `src/datachain/dataset.py` - they are there because of legacy and backward compatibility and will be removed in one of the followup PRs (I've added a TODO where it's used)

src/datachain/dataset.py
ilongin: Added and added ticket number here

tests/unit/lib/test_signal_schema.py
ilongin: No, I just realized we are missing tests for that one so added them. They are pretty simple ones so I didn't want to open a new PR just for it.

tests/scripts/feature_class_parallel.py
ilongin: New implementation of `from_storage()` is not creating deprecated root level file signals, instead they are nested inside feature object, by default it's called `file`.

src/datachain/lib/dc.py
ilongin: Currently, If user enters specific file  it's parent folder is listed .. e.g
```python
DataChain.from_storage("s3://my-bucket/animals/dogs/dog5.jpg")
```
Bucket listing will have the whole `dogs` folder and dataset holding data will be called `lst_s3://my-bucket/animals/dogs/`

We can do a followup to optimize this if needed.

src/datachain/lib/dc.py
ilongin: Yes, this will return datasets + listings (special kind of datasets). 
In followup [PR](https://github.com/iterative/datachain/pull/331) after this one I've added `def listings()`  [method](https://github.com/iterative/datachain/pull/331/files#diff-0b1cfd3d89c057cc57811e07eb20298eeaaee60c54cc66e5128f618d144f42eeR597-R633) that returns only listing datasets which I will add to be used here instead of `cls.datasets(...)`

src/datachain/lib/dc.py
ilongin: When we list bucket, listing dataset that holds data will always have generic `File` in schema, e.g `{"file": "File@v1"}`, but we need to be able to return chain from `.from_storage()` that is using different type, e.g `TextFile` which is done with `type` argument.
Example:

```python
txt_dc = DataChain.from_storage("gs://datachain-demo", type="text").filter(C("file.path").glob("*.txt")) 
```


So that line of code just changes type of object holding listing, e.g `File` -> `TextFile` in the schema.

src/datachain/lib/dc.py
shcheklein: yes, I understand the logic (it is the same as it was before). Mu question was - if it's easier to drop it now and do it properly right away. By properly I mean - index a single file (may be even don't cache it).

shcheklein: It think this is very needed.

shcheklein: can be a followup though

tests/scripts/feature_class_parallel.py
shcheklein: sound good, I think it might be used in many places still

src/datachain/lib/dc.py
ilongin: You are right, `session` arg is missing in that line. Adding it.

src/datachain/lib/meta_formats.py
ilongin: I had some issue with examples that are using this one after changes, and solution was to use `type="text"` when doing `.from_storage()` in `read_meta` function, with which we avoided need to decode to string as it was reading automatically as `TextFile` instead of `File` which is binary by default. I think it's more elegant this way anyway.

shcheklein: > I had some issue with examples that are using this one after changes

do you remember the context by chance? any explanation to that?

src/datachain/lib/dc.py
shcheklein: can we please deduplicate and simplify this?

tests/conftest.py
ilongin: We have `cloud_test_catalog` thing which has `catalog` inside that has set up for all mocked clouds to return data from `tests/data.py`.  We are using it extensively across the tests to be able to test with all the clouds.

In one of the tests I needed `test_session`  but also this special `catalog` from `cloud_test_catalog` to test for all the clouds, so instead of creating duplicate `test_session` I just used this catalog instance inside existing one. I think it's convenient to have this one here regardless.

src/datachain/lib/listing.py
ilongin: Yes, it should always end up with `/`. This might change if we allow enlisting single file though but that's separate issue anyway and for now we always enlist directory.
I will add some additional checks for this

tests/conftest.py
shcheklein: `cloud_test_catalog` does it mean that we run some extra code, initialization and cleanup for more tests where we don't need to?

src/datachain/lib/listing.py
shcheklein: yep, let's put asserts to document it everywhere to enforce this and catch if it's broken

src/datachain/lib/listing.py
ilongin: This will be broken if we introduce enlisting single file where listing dataset name won't end with `/`.
I will add asserts for now and this, alongside `listing_dataset_name`, will need to be changed when we introduce that.

src/datachain/lib/listing.py
ilongin: I think it's related to file versioning in clouds. Latest means it's last version of specific file if cloud versioning is turned on. Honestly I don't think we ever use this, so it's just passed through from the beginning of datachain.

src/datachain/lib/listing.py
ilongin: Good question. I was kind of just c/p this ones from the old code so not sure if there was any specific reason to do so but I agree it does look strange. 

I'm not even sure if globs + non recursive even make sense TBO ... and `recursive` naming is also a little bit strange. From what I remember idea here was to return only first level children for some specific path.

src/datachain/catalog/catalog.py
```diff
@@ -1422,17 +1422,19 @@ def get_dataset_dependencies(
...
 
-    def ls_datasets(self) -> Iterator[DatasetRecord]:
+    def ls_datasets(self, include_listing: bool = False) -> Iterator[DatasetRecord]:
         datasets = self.metastore.list_datasets()
         for d in datasets:
-            if not d.is_bucket_listing:
-                yield d
+            if not include_listing and d.is_bucket_listing:
+                continue
+            yield d
```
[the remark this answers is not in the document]
ilongin: Yes, applied!

examples/get_started/udfs/parallel.py
```diff
@@ -31,7 +31,7 @@ def path_len_benchmark(path):
 
 # Run in chain
 DataChain.from_storage(
-    path="gs://datachain-demo/dogs-and-cats/",
+    uri="gs://datachain-demo/dogs-and-cats/",
```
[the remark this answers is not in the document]
ilongin: Done

src/datachain/lib/dc.py
ilongin: Added for `anon` ... `client_config` is being discussed now if we are going to leave it or not

src/datachain/lib/dc.py
ilongin: I would rather do it in followup

src/datachain/lib/dc.py
ilongin: Refactored

src/datachain/lib/meta_formats.py
ilongin: Looks like example that was failing to me (`/examples/get_started/json-csv-reader.py`) works even without these changes so they are not really required in this PR. Maybe issue occurred to me during development that was not related, idk ... anyway, I still think these changes are good as we don't need to explicitly decode strings if we read storage as `text` and I would leave them.

tests/func/test_datachain.py
```diff
@@ -25,34 +33,157 @@ def test_catalog_anon(tmp_dir, catalog, anon):
 
 def test_from_storage(cloud_test_catalog):
     ctc = cloud_test_catalog
-    dc = DataChain.from_storage(ctc.src_uri, catalog=ctc.catalog)
+    dc = DataChain.from_storage(ctc.src_uri, client_config=ctc.catalog.client_config)
```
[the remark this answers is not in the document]
ilongin: I agree about `Catalog`, but I'm not sure about session. If we go only with sessions, user will even for a simplest example need to construct `Session` object and will need to learn about another class / context manager etc. 
I think sessions are more for advanced usage like those ephemeral datasets where they are absolutely required, and we should allow users to do simple chains in the most easier way possible which in this case is providing simple config.
Also, I think we should remove `Catalog` from `Session` constructor as it could be confusing to the users.

BTW `client_config` also doesn't really sounds "right" to me. It would be more correct `cloud_config` IMO, but this is another discussion. 

WDYT?

tests/conftest.py
ilongin: It just does simple setup, but the bigger problem is that enforces running tests for every cloud type and cloud versioning which means it will slow tests significantly where it doesn't need to (e.g unit tests where we manually create local dirs / files and test on them) so I'm reverting this change and adding `session` property on `CloudTestCatalog` class to be used where we need that mocked clouds and running tests for each of those (usually func tests)

src/datachain/lib/listing.py
ilongin: Extracted to util and added tests

tests/func/test_datachain.py
ilongin: Removed them, please ignore. Related to https://github.com/iterative/datachain/pull/294#discussion_r1730478181

tests/func/test_datachain.py
ilongin: Extracted to separated test case

tests/unit/lib/test_signal_schema.py
ilongin: It was throwing some strange SQLAlchemy errors when creating table with `datachain.Column`s so this was a workaround (I wanted to fix it in followup as I didn't have time for deeper investigation). Now I did some research and it turns out the problem was that our `datachain.Column` is extending [sa.ColumnClause](https://pydocbrowser.github.io/sqlalchemy/latest/sqlalchemy.sql.elements.ColumnClause.html) which is a parent class of [sa.Column](https://pydocbrowser.github.io/sqlalchemy/latest/sqlalchemy.sql.schema.Column.html#index) which is needed if you want to create a table from it (what I was doing in `DataChain.from_records()` which was calling it). `sa.ColumnClause` cannot be used when defining DB table.

TL;DR I managed to fix it to return `datachain.Column` instead and doing convertion to `sa.Column` in `DataChain.from_records()` which moves these column hacking from user facing method to internal implementation so it's good now I think.

tests/unit/lib/test_signal_schema.py
ilongin: `SignalSchema.mutate()` accepts normal dict, not kwargs

tests/unit/lib/test_signal_schema.py
ilongin: `SignalSchema.mutate()` was from beginning implemented to accept new signals so this should not throw exception, instead it will just add new signal.

tests/unit/lib/test_signal_schema.py
ilongin: Make sense!

ilongin: The problem here is that we cannot just put `TextFile` as type of `datachain.Column` (SQLAlchemy complains about it), and also we cannot convert it to something SQL-like. 
Also, it seems little bit strange to put it into `Column` which is clearly 1-1 with DB columns (I think users will have this intuition as well)

src/datachain/lib/signal_schema.py
ilongin: About avoiding tricks, I'm not sure I understand what is the real issue there ... by setting `Float` as a type? ... Note that we accept python type instead when changing signal type which is expected behavior for user I would say.

src/datachain/lib/dc.py
ilongin: Removed

src/datachain/lib/dc.py
ilongin: Removed

tests/unit/lib/test_signal_schema.py
ilongin: Explained in comment below

src/datachain/lib/dc.py
ilongin: I've tried but code looks worse than here. I've refactored it and reduced line numbers by ~half and there is not much logic left anyway except check if listing exist and if not call listing generator to create it, so I think it's good now.

src/datachain/lib/dc.py
```diff
@@ -305,21 +318,23 @@ def add_schema(self, signals_schema: SignalSchema) -> "Self":  # noqa: D102
...
-        path,
+        uri,
         *,
         type: Literal["binary", "text", "image"] = "binary",
         session: Optional[Session] = None,
         in_memory: bool = False,
         recursive: Optional[bool] = True,
         object_name: str = "file",
         update: bool = False,
+        client_config=None,
```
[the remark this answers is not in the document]
shcheklein: is there a discussion / resolution for this?

src/datachain/lib/dc.py
shcheklein: @ilongin same question here - can we iterate / discuss the API?

src/datachain/lib/listing.py
shcheklein: was is the similar logic before / in the current implementation?

src/datachain/lib/listing.py
shcheklein: is there any resolution to this? I'm not sure I completely understand this listing logic in case of globs +/- recursive

src/datachain/lib/dc.py
ilongin: I put my thoughts in the comment below https://github.com/iterative/datachain/pull/294#discussion_r1730464596

src/datachain/lib/listing.py
ilongin: I tried to remember and I'm pretty sure this was decision way back in the early days of dql where we decided to support just this simple glob when listing/indexing and to add support others in future but we just never added it

src/datachain/lib/listing.py
```diff
@@ -1,19 +1,31 @@
...
+
+if TYPE_CHECKING:
+    from datachain.lib.dc import DataChain
+
 
 ResultQueue = asyncio.Queue[Optional[Sequence[File]]]
 
 DELIMITER = "/"  # Path delimiter
 FETCH_WORKERS = 100
+LISTING_TTL = 4 * 60 * 60  # cached listing lasts 4 days
```
rlamy, line 27: That's 4 hours, not 4 days

src/datachain/lib/file.py
```diff
@@ -390,3 +357,13 @@ class IndexedFile(DataModel):
 
     file: File
     index: int
+
+
+def get_file_type(type_: Literal["binary", "text", "image"] = "binary"):
```
rlamy, line 362:
```suggestion
def get_file_type(type_: Literal["binary", "text", "image"] = "binary") -> type[File]:
```

src/datachain/lib/dc.py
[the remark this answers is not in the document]
shcheklein: @ilongin let's for now keep it part of kwargs (like it was before I assume)? we don't have to do this a public interface atm.

src/datachain/lib/dc.py
```diff
@@ -1583,15 +1635,18 @@ def from_records(
             ```
         """
         session = Session.get(session, in_memory=in_memory)
-        catalog = session.catalog
+        catalog = kwargs.get("catalog") or session.catalog
```
shcheklein, line 1638: do we pass catalog anywhere directly? why do we need this change?

shcheklein (commented):
It seems just a few things left. I'm fine if we hide client_config / catalog under kwargs for now (as it was before). `session` in some APIs is fine as well.

Re the globs - if it was the same logic as before - only `/*` is supported - I'm fine with that for this PR

There are a few questions left by Ronan and I had a new minor one I think.

I resolved almost everything else.

Let's get this merged soon :) Thanks for your patience @ilongin !

tests/func/test_datachain.py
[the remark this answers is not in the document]
dmpetrov: It would be great to know what configuration option we really need to expose to users (not unit tests). Off the top of my head, I can only think of anon.

In general, exposing multiple similar classes to uses seems like a big issue for users.

src/datachain/lib/dc.py
ilongin: This is a leftover, I removed it

tests/func/test_datachain.py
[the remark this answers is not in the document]
ilongin: Ok, I removed `client_config` and left only `anon` and `session`. User will need to create `Session` and add `client_config` to it if he want's to use different cloud credentials than set default in his environment (when using from CLI) which is ok I guess

src/datachain/lib/dc.py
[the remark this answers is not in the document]
ilongin: @shcheklein I removed both `kwargs` and `client_config`.
`kwargs` are bad in public API in general as it's not really explicit, and regarding `client_config` please look at message below: https://github.com/iterative/datachain/pull/294#discussion_r1736679962

src/datachain/lib/listing.py
ilongin: This was basically moved from legacy indexing codebase so nothing has been changed.

shcheklein (approved):

dmpetrov (approved):
Looks great! PLease remove the kwargs and config and it;s good to merge.

PR #249, opened 2024-08-07
handle nan and inf float values

dberenbaum:
Partial fix for https://github.com/iterative/dvcx/issues/1697#issuecomment-2269457257:

- [x] make sure NaN and Inf are supported in SQLite
- [ ] NaN and Inf supported in CH - see https://github.com/iterative/studio/pull/10429
- [x] from_parquet() support NaN and Inf
- [ ] make sure `Array(Float)` support that - needs follow-up since orjson does not support [nan/inf](https://github.com/ijl/orjson/issues/488)

This PR treats all missing values in float columns as NaN for a couple reasons:
1. Consistency with the pydata ecosystem. Numpy and [pandas](https://pandas.pydata.org/docs/user_guide/missing_data.html) both treat all missing values in float columns as NaN.
2. Sqlite does not differentiate between null and NaN (see [sqlite](https://stackoverflow.com/questions/67360775/sql-command-to-insert-inf-nan-and-inf-into-sqlite3-database) [discussions](https://stackoverflow.com/questions/20619957/nan-in-sqlite3-have-invalid-numerical-comparisons) and [source code](https://github.com/sqlite/sqlite/blob/210f6f89befbe4e4c561fc34aabc175f552450f4/src/vdbemem.c#L980)), although [CH does](https://clickhouse.com/docs/en/sql-reference/data-types/float#nan-and-inf).

PR #296, opened 2024-08-14
Implement database default values

dreadatour:
Fix studio tests by implementing database default values for tables. See details below.
Follow-up for the https://github.com/iterative/datachain/pull/249

Tests to be added if this approach looks ok for everyone.


src/datachain/data_storage/schema.py
```diff
@@ -67,7 +67,11 @@ def convert_rows_custom_column_types(
     for row in rows:
         row_list = list(row)
         for idx, t in custom_columns_types:
-            row_list[idx] = t.on_read_convert(row_list[idx], dialect)
+            row_list[idx] = (
+                t.default_value(dialect)
```
dreadatour, line 71: Return default value if no value is set.

src/datachain/data_storage/schema.py
```diff
@@ -136,7 +140,15 @@ def __init__(
...
-    def copy_column(column: sa.Column):
+    def copy_column(
+        column: sa.Column,
+        primary_key: Optional[bool] = None,
+        index: Optional[bool] = None,
+        nullable: Optional[bool] = None,
+        default: Optional[Any] = None,
+        server_default: Optional[Any] = None,
+        unique: Optional[bool] = None,
+    ) -> sa.Column:
```
dreadatour, line 151: These params are useful if we want to copy column with updated params.

src/datachain/node.py
```diff
@@ -47,7 +47,7 @@ class DirTypeGroup:
 @attrs.define
 class Node:
     sys__id: int = 0
-    sys__rand: int = -1
+    sys__rand: int = 0
```
dreadatour, line 50: `sys__rand` is uint in Studio at least, negative default value cause an exception.

src/datachain/sql/types.py
```diff
@@ -439,6 +506,50 @@ def binary(self):
         return None
 
 
+class DBDefaults:
```
dreadatour, line 509: Use default values in `CREATE TABLE` query based on DB implementations.
It is used in Studio, see related PR.

dreadatour (commented):
Provide default values on a DB level. This PR is needed to support `NaN` default value for `float` types.

dberenbaum:
So this moves everything up a level so it applies to all warehouse types instead of being specific to sqlite, right? If so, LGTM, thanks!

dreadatour:
> So this moves everything up a level so it applies to all warehouse types instead of being specific to sqlite, right? If so, LGTM, thanks!

It is 👀 Thank you! 🥰

dtulga (approved):
LGTM, thanks for fixing this! That being said, at least a few more tests would be nice to add.

src/datachain/sql/types.py
```diff
@@ -328,7 +396,11 @@ def uint64(self, value):
         return value
 
     def float(self, value):
-        return float("nan") if value is None else value
+        if value is None:
+            return float("nan")
+        if isinstance(value, str) and value.lower() == "nan":
+            return float("nan")
```
dberenbaum, line 402: Were you encountering this with either sqlite or clickhouse?

dreadatour: It is SQLite.

dreadatour: SQLite CLI:

```
$ sqlite3 .datachain/db
SQLite version 3.43.2 2023-10-10 13:08:14
Enter ".help" for usage hints.
sqlite> create table foo (id int, val float not null default nan);
sqlite> insert into foo (id) values (1);
sqlite> select * from foo;
1|nan
sqlite> 
$
```

Python:

```
$ ipython
Python 3.12.5 (main, Aug  6 2024, 19:08:49) [Clang 15.0.0 (clang-1500.3.9.4)]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.25.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import sqlite3

In [2]: conn = sqlite3.connect('.datachain/db')

In [3]: cur = conn.cursor()

In [4]: cur.execute('select * from foo')
Out[4]: <sqlite3.Cursor at 0x11064f4c0>

In [5]: cur.fetchall()
Out[5]: [(1, 'nan')]

In [6]:
Do you really want to exit ([y]/n)? ^D
$
```

src/datachain/sql/types.py
```diff
@@ -432,11 +503,49 @@ def json(self):
...
+
+class DBDefaults:
+    def string(self):
+        return sa.text("''")
+
+    def boolean(self):
+        return sa.text("False")
+
+    def int(self):
+        return sa.text("0")
```
dberenbaum, line 518: Are these all the default values in sqlite now also? Does it mean there are no longer null values in sqlite?

dreadatour: It depends on how we are creating table. If we will pass `NOT NULL` (`nullable=False`) and default value when table is created — there will be no nulls.

But for SQLite we don't do this — nothing is changed.

dreadatour: This class is used only if we set not nullable columns on table creation. So I have added tests for this case and updated these fields to provide correct values for SQLite implementation.

src/datachain/sql/types.py
dberenbaum: Not sure there is a better solution, but want to note some edge cases when setting the default to `nan` in sqlite:
- unexpected sorting and equality operations as discussed in https://stackoverflow.com/questions/20619957/nan-in-sqlite3-have-invalid-numerical-comparisons
- sqlite stores nan values like 10/0 as nulls in sqlite rather than `nan`, which again may cause unexpected results like null and nan being sorted differently

dreadatour: Absolutely agree 👍 Hopefully we will not need default values in SQLite.

PR #304, opened 2024-08-15
standardise union behaviour between db implementations

mattseddon:
follow up to https://github.com/iterative/datachain/pull/216

closes https://github.com/iterative/dvcx/issues/1202

This PR changes the behaviour of our union operator to throw an obvious error (in all db implementations) when there is a mismatch in columns between the two chains being unioned. Previously we "cast"ed all columns with missing values to nulls. This only worked for SQLite as we do not/do not want to store nulls in Clickhouse. For more details see the above issue.

### DIFF
--- a/src/datachain/query/dataset.py
+++ b/src/datachain/query/dataset.py
@@ -878,17 +878,14 @@ class SQLUnion(Step):
         temp_tables.extend(self.query1.temp_table_names)
         q2 = self.query2.apply_steps().select().subquery()
         temp_tables.extend(self.query2.temp_table_names)
-        columns1, columns2 = fill_columns(q1.columns, q2.columns)
+
+        columns1, columns2 = _order_columns(q1.columns, q2.columns)
 
         def q(*columns):
             names = {c.name for c in columns}
             col1 = [c for c in columns1 if c.name in names]
             col2 = [c for c in columns2 if c.name in names]
-            res = (
-                sqlalchemy.select(*col1)
-                .select_from(q1)
-                .union_all(sqlalchemy.select(*col2).select_from(q2))
-            )
+            res = sqlalchemy.select(*col1).union_all(sqlalchemy.select(*col2))
 
             subquery = res.subquery()
             return sqlalchemy.select(*subquery.c).select_from(subquery)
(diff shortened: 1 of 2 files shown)


src/datachain/query/dataset.py
```diff
@@ -1007,23 +1004,46 @@ def q(*columns):
...
-    result: list[list[ColumnElement]] = [[] for _ in column_dicts]
-    for n in combined_columns:
-        col = next(col_dict[n] for col_dict in column_dicts if n in col_dict)
-        for col_dict, out in zip(column_dicts, result):
-            if n in col_dict:
-                out.append(col_dict[n])
-            else:
-                # Cast the NULL to ensure all columns are aware of their type
-                # Label it to ensure it's aware of its name
-                out.append(sqlalchemy.cast(sqlalchemy.null(), col.type).label(n))
```
mattseddon, line 1025: [F] We could probably use `col.type.db_default_value(self.db.dialect)` but that would lead to more user confusion IMO.

src/datachain/query/dataset.py
```diff
@@ -1021,23 +1018,46 @@ def q(*columns):
...
+        for missing_columns, found_side in zip(
+            [
+                missing_right,
+                missing_left,
+            ],
+            ["left", "right"],
+        )
+        if missing_columns
+    ]
+    msg = f"Cannot perform union. {'. '.join(msg_parts)}"
```
dtulga, line 1047: Would it make sense to have a period (full stop) at the end of this message always? It looks a bit odd to have a period at the end of "Cannot perform union." but not at the end of the message, unless there are missing columns on both sides, where there is a period in between these two statements as well, but not at the end.

dtulga (approved):
LGTM, thanks!

src/datachain/query/dataset.py
mattseddon: I can update but I was trying to be consistent with the bulk of the other error messages in the repo. Generally the "standard" seems to be a capital first letter and no period at the end of the error message. WDYT?

dtulga: I'm not really sure if it's an official standard, or just something we've been doing. It's totally fine keeping the current way, it just looked a bit odd with the other periods in the message, but not at the end, although it doesn't really matter much to me. (Possibly I only noticed this due to the other periods, which we usually don't have.)

PR #308, opened 2024-08-15
Split Studio tests in CI

dreadatour:
Split Studio tests into separate workflow. This way we able to see when tests are fails, but "tests" badge in readme will only represent DataChain CLI tests status.


shcheklein (approved):
LTGM is that works and gives us what we need in terms of badges!

dreadatour:
> LTGM is that works and gives us what we need in terms of badges!

This should works since badge URL looks like this: `https://github.com/iterative/datachain/actions/workflows/tests.yml/badge.svg`

We still will be able to see if Studio tests fails in PRs, but this will not affects badge.

dtulga (approved):
LGTM, thanks!

mattseddon:
[N] Can we keep the action name as `Tests / studio`?

image

amritghimire (approved):

dreadatour:
> [N] Can we keep the action name as `Tests / studio`?

Hm, interesting. Let me check 👍

dreadatour:
> > [N] Can we keep the action name as `Tests / studio`?
> 
> Hm, interesting. Let me check 👍

It looks like workflows with duplicate names are not running by GitHub CI. Leave it as "Studio Tests" for now, please, feel free to update it if you'll find a better name or find a way to use same "Tests" name ❤️

PR #309, opened 2024-08-16
fix example - use last working version of unstructured

mattseddon:
similar to #277

It looks like the team over at unstructured is having nearly as much fun with `nltk` as me. 

It looks like `0.15.2` was yanked from pypi. `0.15.3` was a failed release that `0.15.4` quickly fixed but that version hasn't been released yet.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -98,7 +98,7 @@ examples = [
   "numpy>=1,<2",
   "defusedxml",
   "accelerate",
-  "unstructured[pdf]",
+  "unstructured[pdf]==0.15.1",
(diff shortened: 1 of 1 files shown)


dberenbaum (approved):

dtulga (approved):
LGTM, and it's unfortunate that this package is causing so many problems. 😕

mattseddon:
Looks like this isn't needed for now but I will raise another one to unpin `nltk` once `unstructured` jumps the shark

mattseddon:
replaced by https://github.com/iterative/datachain/pull/332

PR #314, opened 2024-08-18
add_regexp_replace

### LINKED ISSUE #303: Need regexp_replace() function
### Description

It is common to transform text columns in datasets.

For example, if  class is encoded as part of the filename, one currently needs to write a mapper to extract label "cat" from the name "cat.1001.jpg" :

```
import re

def extract_label(filepath: str) -> str:
    match = re.search(r'\.(\pL+)\.', filepath)
    
    if match:
        return match.group(1)
    else:
        return None

map(label = lambda file:  extract_label(file.path))
```

It should be straightforward in many SQL engines to use regexp_replace() for this part:

```
mutate(label = regexp_replace(Column("file.path"), '\.(\pL+)\.', '\1') ))

```

EdwardLi-coder:
Adds `regexp_replace()` function. 

Closes #303

### DIFF
--- a/src/datachain/sql/functions/string.py
+++ b/src/datachain/sql/functions/string.py
@@ -37,7 +37,6 @@ class regexp_replace(GenericFunction):  # noqa: N801
     inherit_cache = True
 
 
-compiler_not_implemented(regexp_replace)
-
 compiler_not_implemented(length)
 compiler_not_implemented(split)
+compiler_not_implemented(regexp_replace)
--- a/src/datachain/sql/functions/string.py
+++ b/src/datachain/sql/functions/string.py
@@ -26,5 +26,18 @@ class split(GenericFunction):  # noqa: N801
     inherit_cache = True
 
 
+class regexp_replace(GenericFunction):  # noqa: N801
+    """
+    Replaces substring that match a regular expression.
+    """
+
+    type = String()
+    package = "string"
+    name = "regexp_replace"
+    inherit_cache = True
+
+
+compiler_not_implemented(regexp_replace)
+
 compiler_not_implemented(length)
 compiler_not_implemented(split)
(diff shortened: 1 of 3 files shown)


shcheklein (commented):
Thanks @EdwardLi-coder ! It would be great to add a test for this.

EdwardLi-coder:
> Thanks @EdwardLi-coder ! It would be great to add a test for this.

Thank you @shcheklein  for pointing that out. I apologize for forgetting to add a test. I will add one now.

shcheklein (approved):

shcheklein:
Looks good, but we need also implement the ClickHouse part (that is used in the private Studio product). I'm getting atm:

```python
NotImplementedError: Compiler not implemented for the SQLAlchemy function, string.regexp_replace, with dialect, clickhouse. For information on adding dialect-specific compilers, see https://docs.sqlalchemy.org/en/14/core/compiler.html
```

let me try to help you here and create a companion PR is possible.

shcheklein (changes requested):
Needs a companion PR on the Studio side. I'll try to look into this.

shcheklein:
I've added a companion PR to the Studio (private) repo. Made small changes here (@EdwardLi-coder please review). It should be good to go @iterative/datachain

mattseddon (approved):

shcheklein (approved):

EdwardLi-coder:
> I've added a companion PR to the Studio (private) repo. Made small changes here (@EdwardLi-coder please review). It should be good to go @iterative/datachain

Sorry. I only considered SQLite and wasn't aware that ClickHouse support also needed to be considered. In the future, when I try to solve similar problems, I will take it into account. I appreciate that you modified the code to make it more concise. Thank you.

Issue #12, opened 2024-07-11
Add support for `Float64` for Clickhouse (SaaS)


ilongin:
The problem is that if number that can only be represented with double precision, e.g `0.3813476562` is inserted using CH database (in SaaS), it's actually inserted as single precision (`Float32`) as python `float` is saved as `Float32` column type in database by default. This way precision of that value is **lost** for user. 

One solution is to support `datachain.sql.types` when defining output schema (currently only python basic values are supported together with `Feature` type).

Example that works for SQLite but fails for Clickhouse:
```python
import pandas as pd
from datachain.lib.dc import DataChain

num = 0.3813476562   # number that requires 64 bits to not loose precision
chain = DataChain.from_pandas(pd.DataFrame([[num]], columns=['number']))
assert chain.collect_one("number")[0] == num
```

dmpetrov:
@ilongin is it still relevant or we can close?
Based on my understanding we are working with float64 only rn. Other types are not supported yet.

ilongin:
@dmpetrov yes, looks like example in description works now so I'm closing this issue.

skshetry:
I don't think we have made any changes to this. @ilongin, can you point me the commit that fixed this please?

ilongin:
@skshetry I made a mistake testing this on SQLite DB instead of Clickhouse. Reopening it agan.

dreadatour:
Let's discuss this approach, please? https://github.com/iterative/studio/pull/10528 🙏👀

dreadatour:
Should be fixed now by related Studio issue.

PR #319, opened 2024-08-19
Support all column types in SignalSchema.from_column_types

dreadatour:
Having registered dataset:

```python

DataChain.from_storage("s3://dql-50k-laion-files/").save("laion")
```

Making basic query on it:

```python

DataChain(name="laion").filter(C.file__path.glob("*.jpg"))
```

Fails with error:

```
  File "<string>", line 4, in <module>
  File "/Users/vlad/work/iterative/datachain/src/datachain/lib/dc.py", line 227, in __init__
    self.signals_schema |= SignalSchema.from_column_types(self.column_types)
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/vlad/work/iterative/datachain/src/datachain/lib/signal_schema.py", line 108, in from_column_types
    raise SignalSchemaError(
datachain.lib.signal_schema.SignalSchemaError: signal schema cannot be obtained for column 'sys__rand': unsupported type 'None'
  File "<string>", line 4, in <module>
  File "/Users/vlad/work/iterative/datachain/src/datachain/lib/dc.py", line 227, in __init__
    self.signals_schema |= SignalSchema.from_column_types(self.column_types)
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/vlad/work/iterative/datachain/src/datachain/lib/signal_schema.py", line 108, in from_column_types
    raise SignalSchemaError(
datachain.lib.signal_schema.SignalSchemaError: signal schema cannot be obtained for column 'sys__rand': unsupported type 'None'
```

This is because `sys__rand` column type is `UInt64` (at least in SaaS), and this type is not exists in `DATACHAIN_TO_TYPE`, which leads to an error here: https://github.com/iterative/datachain/blob/main/src/datachain/lib/signal_schema.py#L107-L111


mattseddon (approved):
is there a regression test somewhere?

src/datachain/lib/convert/flatten.py
```diff
@@ -10,19 +10,25 @@
     Boolean,
     DateTime,
     Float,
+    Float32,
+    Float64,
```
mattseddon, line 14: Does this close https://github.com/iterative/datachain/issues/12 ?

skshetry: If it does, please don't forget to rollback this change: https://github.com/iterative/datachain/pull/13.

ilongin (approved):
LGTM

src/datachain/lib/convert/flatten.py
dreadatour: This doesn't close #12, but it will be easy to fix, I'll make separate PR for that.

dreadatour:
> is there a regression test somewhere?

Wasn't sure about this approach. Updated the PR with better code, added test, please, take a look 🙏

dtulga (approved):
LGTM, thanks!

mattseddon (approved):

PR #276, opened 2024-08-12
Use free GitHub Actions workers whenever possible

0x2b3bfa0:
This will slow down tests considerably; let's see if it's acceptable.

PR #325, opened 2024-08-20
update examples tests timeout to be 30 minutes

mattseddon:
should be more than enough time

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -112,7 +112,7 @@ jobs:
 
   examples:
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 60
+    timeout-minutes: 30
     strategy:
       fail-fast: false
       matrix:


ilongin (approved):

dtulga (commented):
It looks like one of the example test runs timed out at 25 minutes. Should the maximum runtime be increased (perhaps to a compromise value) or should those examples be sped up? I think this looks fine to me, as long as it doesn't introduce more flakiness.

skshetry:
Do we need to run examples on all platforms and 2 python versions?

mattseddon:
> Do we need to run examples on all platforms and 2 python versions?

I would say yes. If you're looking to save money see https://github.com/iterative/datachain/pull/276

dreadatour (approved):
Agree. Looks good to me 👀

mattseddon:
less important now https://github.com/iterative/datachain/pull/276 has been merged

PR #326, opened 2024-08-20
Handle carriage return to support progress bar in logs

amritghimire:
This change breaks the read by reading each character and
yielding the breaks in <CR> or new lines instead of just waiting on new
line.

Studio PR: https://github.com/iterative/studio/pull/10509


shcheklein (approved):

dreadatour (approved):
Looks good to me! Thank you! ❤️

PR #330, opened 2024-08-20
use default delimiter to flatten columns

### LINKED ISSUE #328: Method `.to_pandas(flatten=True)` returns DataFrame of NaN values
### Description

Code to reproduce: 
```python
DataChain.from_storage('gs://datachain-demo/dogs-and-cats/cat.1009.jpg', type='image').to_pandas(flatten=True)
```

image

### Version Info

```Text
0.3.4
Python 3.12.4
```

shcheklein:
Fixes #328 

~`to_pandas(flatten=True)` now returns columns in the same way as `to_records` and in the same way we store them in the DB (to fix the bug and for consistency)~ 

changing this use `.`. It is simpler for end users and we don't want them to see DB details (at least it's not how our API operates atm).

### DIFF
--- a/src/datachain/lib/dc.py
+++ b/src/datachain/lib/dc.py
@@ -48,7 +48,7 @@ from datachain.query.dataset import (
     PartitionByType,
     detach,
 )
-from datachain.query.schema import Column, DatasetRow
+from datachain.query.schema import DEFAULT_DELIMITER, Column, DatasetRow
 from datachain.sql.functions import path as pathfunc
 from datachain.utils import inside_notebook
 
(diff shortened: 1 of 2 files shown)


shcheklein:
Probably not the right way to do this, tbh. I'll change it back (and replace `__` with `.`). @iterative/datachain what was the reason to use `__` instead of a `.` on the DB level? We now have two ways of interacting with names (even within public API - e.g. to_records).

src/datachain/lib/dc.py
```diff
@@ -1224,14 +1224,11 @@ def to_pandas(self, flatten=False) -> "pd.DataFrame":
         """
         headers, max_length = self._effective_signals_schema.get_headers_with_length()
         if flatten or max_length < 2:
-            columns = []
-            if headers:
-                columns = [".".join(filter(None, header)) for header in headers]
-            return pd.DataFrame.from_records(self.to_records(), columns=columns)
```
dberenbaum, line 1230: So the problem was this line (passing data as a dict and then assigning different column names)? Is the rest refactoring? Just want to make sure I understand the bug and the fix.

mnrozhkov (approved):

PR #331, opened 2024-08-20
Add `DataChain.listings()` method and use it in getting storages

ilongin:
This PR adds class method similar to `DataChain.datasets()` called `DataChain.listings()` which returns list of `ListingInfo` objects from chain.
`ListingInfo` describes one specific cached listing (version) and is simply sub-classing `DatasetInfo` which is returned from mentioned `DataChain.datasets()` as listing is just special form of dataset.
The idea is to use this instead of deprecated `Storage` class which will be, along side with related codebase, removed in upcoming followups. 

Additional changes:
1. Using `Dataset.listings()` in CLI method `ls_local` which lists storages / listings
2. Removed couple of unused `Catalog` methods: `unlist_source`, `storage_stats`, `ls_storage_uris`, `get_storage`, `ls_storages` and underlying metastore methods
3. Fixing dataset dependencies as now there is no special "storage" table which can be references and `storage` type dependency is also pointing to dataset (just listing type of dataset)

Blocked by https://github.com/iterative/datachain/pull/294

### DIFF
--- a/src/datachain/data_storage/metastore.py
+++ b/src/datachain/data_storage/metastore.py
@@ -168,13 +168,6 @@ class AbstractMetastore(ABC, Serializable):
         """
 
     @abstractmethod
-    def mark_storage_not_indexed(self, uri: StorageURI) -> None:
-        """
-        Mark storage as not indexed.
-        This method should be called when storage index is deleted.
-        """
-
-    @abstractmethod
     def update_last_inserted_at(self, uri: Optional[StorageURI] = None) -> None:
         """Updates last inserted datetime in bucket with current time."""
 
--- a/src/datachain/data_storage/metastore.py
+++ b/src/datachain/data_storage/metastore.py
@@ -179,19 +179,11 @@ class AbstractMetastore(ABC, Serializable):
         """Updates last inserted datetime in bucket with current time."""
 
     @abstractmethod
-    def get_all_storage_uris(self) -> Iterator[StorageURI]:
-        """Returns all storage uris."""
-
-    @abstractmethod
     def get_storage(self, uri: StorageURI) -> Storage:
         """
         Gets storage representation from database.
         E.g. if s3 is used as storage this would be s3 bucket data.
         """
-
-    @abstractmethod
-    def list_storages(self) -> list[Storage]:
-        """Returns all storages."""
 
     @abstractmethod
     def mark_storage_pending(self, storage: Storage) -> Storage:
@@ -906,11 +898,6 @@ class AbstractDBMetastore(AbstractMetastore):
             self._storages_update().where(s.c.uri == uri).values(**updates)  # type: ignore [attr-defined]
         )
 
-    def get_all_storage_uris(self) -> Iterator[StorageURI]:
-        """Returns all storage uris."""
-        s = self._storages
-        yield from (r[0] for r in self.db.execute(self._storages_select(s.c.uri)))
-
     def get_storage(self, uri: StorageURI, conn=None) -> Storage:
         """
         Gets storage representation from database.
@@ -925,13 +912,6 @@ class AbstractDBMetastore(AbstractMetastore):
             raise StorageNotFoundError(f"Storage {uri} not found.")
 
         return self.storage_class._make(result)
-
-    def list_storages(self) -> list[Storage]:
-        result = self.db.execute(self._storages_select())
-        if not result:
-            return []
-
-        return [self.storage_class._make(r) for r in result]
 
     def mark_storage_pending(self, storage: Storage, conn=None) -> Storage:
         # Update status to pending and dates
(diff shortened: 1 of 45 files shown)


src/datachain/data_storage/metastore.py
```diff
@@ -167,32 +167,17 @@ def mark_storage_indexed(
...
 
-    @abstractmethod
-    def mark_storage_not_indexed(self, uri: StorageURI) -> None:
-        """
-        Mark storage as not indexed.
-        This method should be called when storage index is deleted.
-        """
-
     @abstractmethod
     def update_last_inserted_at(self, uri: Optional[StorageURI] = None) -> None:
```
shcheklein, line 171: is this one still relevant?

src/datachain/data_storage/metastore.py
```diff
@@ -167,32 +167,17 @@ def mark_storage_indexed(
...
         """Updates last inserted datetime in bucket with current time."""
 
-    @abstractmethod
-    def get_all_storage_uris(self) -> Iterator[StorageURI]:
-        """Returns all storage uris."""
-
     @abstractmethod
     def get_storage(self, uri: StorageURI) -> Storage:
```
shcheklein, line 175: is this one still relevant?

src/datachain/dataset.py
```diff
@@ -92,33 +101,31 @@ def parse(
...
+
+        if not dataset_id:
+            return None
+
+        assert dataset_name is not None
+        dependency_type = DatasetDependencyType.DATASET
+        dependency_name = dataset_name
+
+        if is_listing_dataset(dataset_name):
+            dependency_type = DatasetDependencyType.STORAGE  # type: ignore[arg-type]
```
shcheklein, line 114: should we use LISTING as a type (since we use `listing` as term in some other places)

src/datachain/lib/dc.py
```diff
@@ -577,6 +572,43 @@ def datasets(
...
 
+    @classmethod
+    def listings(
+        cls,
+        session: Optional[Session] = None,
+        in_memory: bool = False,
+        object_name: str = "listing",
+        **kwargs,
+    ) -> "DataChain":
+        """Generate chain with list of cached listing.
```
shcheklein, line 583: listing -> listings 

Q: do we really need to make it a public dc method? (I don't have any strong opinion on this - just thinking if we can for now expose less and see if there are scenarios for this)

shcheklein: does it mean we will need `from_listing` btw? (just trying to think if we can keep at all about datasets / datachain and avoid exposing "listing" concept to end users (?)

shcheklein, line 583: it would be great to explain a bit what the listing is - just a few words

src/datachain/lib/dc.py
```diff
@@ -577,6 +572,43 @@ def datasets(
...
+
+        Example:
+            ```py
+            from datachain import DataChain
+
+            chain = DataChain.listings()
```
shcheklein, line 589: can we use `show()` instead? makes it simpler

shcheklein (approved):
LGTM, my only question if we really want to expose this as method that returns datachain (what is the use case for this?)

also the name `listing` vs `storage` that we use in some places

src/datachain/data_storage/metastore.py
ilongin: Yes, this is still being used in old deprecated listing from CLI. It will be removed when we refactor that last left part (CLI)

src/datachain/data_storage/metastore.py
ilongin: Yes, it's the same story as above

src/datachain/dataset.py
ilongin: I was thinking about this as well. Not sure, but both is correct IMO

src/datachain/lib/dc.py
ilongin: Yea, we do have usage of both `listings` and `storages` words around interface which is ok IMO. Listing is just listed storage so I don't think those two words cannot live together. 
I don't think we need `from_listing` as we have `from_storage` which in background creates new listing if doesn't exist and uses it then.

I think it's ok for us to expose this as we will be using it in Studio for example which can also be seen as one client.

src/datachain/lib/dc.py
ilongin: Yes, that's good idea

src/datachain/lib/dc.py
ilongin: Added one sentence

ilongin:
> LGTM, my only question if we really want to expose this as method that returns datachain (what is the use case for this?)
> 
> also the name `listing` vs `storage` that we use in some places

About exposing it as a method that returns `DataChain` instance, I wanted to be consistent with similar `.datasets()` method. If we want to change this, both methods needs to be refactored together

src/datachain/dataset.py
shcheklein: > Not sure, but both is correct IMO

yes, but it just make it simpler to maintain, search, understand

not critical, but I would try to go everywhere with the same term

PR #332, opened 2024-08-21
remove nltk pin

mattseddon:
The most recent version of `unstructured` has moved to `nltk==3.9.1`. We want to use that version of `unstructured` now so we need to remove the pin.


mattseddon:
if this works this PR will replace #309

shcheklein (approved):

PR #338, opened 2024-08-22
remove blip2 image desc example

mattseddon:
This example is close to a duplicate of examples/computer_vision/llava2_image_desc_lib.py. We can't get either of them under test due to the run time in the CI environment. I don't see the value in carrying both.


examples/computer_vision/blip2_image_desc_lib.py
```diff
@@ -1,100 +0,0 @@
...
-
-        generated_ids = self.model.generate(**inputs, max_new_tokens=self.max_tokens)
-        generated_text = self.processor.batch_decode(
-            generated_ids, skip_special_tokens=True
-        )
-        desc = generated_text[0]
-        return desc.strip(), ""
-
-
-class LLaVADescribe(Mapper):
```
mattseddon, line 58: [C] This is a duplicate of the class from examples/computer_vision/llava2_image_desc_lib.py

PR #339, opened 2024-08-22
Convert custom columns types in dataset_select_paginated

dreadatour:
We use [convert_rows_custom_column_types](https://github.com/iterative/datachain/blob/dfa206812227d5ded173a33d1998c98161368b06/src/datachain/data_storage/schema.py#L49-L76) in [dataset_rows_select](https://github.com/iterative/datachain/blob/dfa206812227d5ded173a33d1998c98161368b06/src/datachain/data_storage/sqlite.py#L654C9-L658), but did not use it in `dataset_select_paginated`. This PR fixes this.

### DIFF
--- a/src/datachain/data_storage/warehouse.py
+++ b/src/datachain/data_storage/warehouse.py
@@ -227,10 +227,7 @@ class AbstractWarehouse(ABC, Serializable):
                     if limit < page_size:
                         paginated_query = paginated_query.limit(None).limit(limit)
 
-                results = self.db.execute(paginated_query.offset(offset))
-                results = convert_rows_custom_column_types(
-                    query.selected_columns, results, self.db.dialect
-                )
+                results = self.dataset_rows_select(paginated_query.offset(offset))
 
                 processed = False
                 for row in results:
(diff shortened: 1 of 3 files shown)


src/datachain/data_storage/warehouse.py
```diff
@@ -227,6 +228,9 @@ def dataset_select_paginated(
                         paginated_query = paginated_query.limit(None).limit(limit)
 
                 results = self.db.execute(paginated_query.offset(offset))
+                results = convert_rows_custom_column_types(
+                    query.selected_columns, results, self.db.dialect
+                )
```
skshetry, line 233: We already have `dataset_rows_select()` in SQLite and in CH, that do similar things. 
Maybe we can extract this out and unify them? That way, we will only have one place to `execute` and convert rows.

dreadatour: 🤔 I like the idea. Let me think.

dreadatour: Please, check the update [here](https://github.com/iterative/datachain/pull/339/commits/ccd0572d7549bb2ff22036b1dbc5cdaa5f875061) and in related SaaS PR.

skshetry: Can we also change this above line to use `dataset_rows_select`?
```suggestion
                results = self.dataset_rows_select(paginated_query.offset(offset))
```

skshetry (approved):

PR #311, opened 2024-08-16
`DataChain.from_hf`

dberenbaum:
Partially addresses #236.

Example (adapted from https://huggingface.co/docs/datasets/en/image_load):

```python
>>> from datachain import DataChain
>>> chain = DataChain.from_hf("beans")
>>> chain.show()
Processed: 1 rows [00:00,  3.64 rows/s]
Generated: 1295 rows [00:00, 4707.65 rows/s]
Cleanup: 1 tables [00:00,  4.82 tables/s]
    split                                    image_file_path                                              image             labels  labels
                                                                                                            img             string integer
0   train  /home/albert/.cache/huggingface/datasets/downl...  b'\xff\xd8\xff\xe0\x00\x10JFIF\x00\x01\x01\x00...  angular_leaf_spot       0
1   train  /home/albert/.cache/huggingface/datasets/downl...  b'\xff\xd8\xff\xe0\x00\x10JFIF\x00\x01\x01\x00...  angular_leaf_spot       0
2   train  /home/albert/.cache/huggingface/datasets/downl...  b'\xff\xd8\xff\xe0\x00\x10JFIF\x00\x01\x01\x00...  angular_leaf_spot       0
3   train  /home/albert/.cache/huggingface/datasets/downl...  b'\xff\xd8\xff\xe0\x00\x10JFIF\x00\x01\x01\x00...  angular_leaf_spot       0
4   train  /home/albert/.cache/huggingface/datasets/downl...  b'\xff\xd8\xff\xe0\x00\x10JFIF\x00\x01\x01\x00...  angular_leaf_spot       0
5   train  /home/albert/.cache/huggingface/datasets/downl...  b'\xff\xd8\xff\xe0\x00\x10JFIF\x00\x01\x01\x00...  angular_leaf_spot       0
6   train  /home/albert/.cache/huggingface/datasets/downl...  b'\xff\xd8\xff\xe0\x00\x10JFIF\x00\x01\x01\x00...  angular_leaf_spot       0
7   train  /home/albert/.cache/huggingface/datasets/downl...  b'\xff\xd8\xff\xe0\x00\x10JFIF\x00\x01\x01\x00...  angular_leaf_spot       0
8   train  /home/albert/.cache/huggingface/datasets/downl...  b'\xff\xd8\xff\xe0\x00\x10JFIF\x00\x01\x01\x00...  angular_leaf_spot       0
[...]

PR #342, opened 2024-08-22
Add ability to set row size for flushing udf results to database

dberenbaum:
Needed as part of #311 but wanted to break it out in a separate PR because it's more generally applicable. For all `DataChain.from_...` methods that generate many rows from a single file or object (like `.from_csv()`), datachain could run into memory issues for a few reasons:
- for any complex output, a udf bug resulted in materializing all udf results in memory
- datachain had a hard-coded batch size of 10,000 rows before it would flush to the db and there was no setting to override it
- even if we had a setting, settings could only be applied after the `from_...` methods

With this PR, you can avoid such memory issues by setting `DataChain.from_csv(.., settings={"flush": 1000})` to flush to the database every 1000 rows.

### DIFF
--- a/src/datachain/query/dataset.py
+++ b/src/datachain/query/dataset.py
@@ -383,9 +383,10 @@ def process_udf_outputs(
     udf_table: "Table",
     udf_results: Iterator[Iterable["UDFResult"]],
     udf: UDFBase,
-    batch_size=INSERT_BATCH_SIZE,
+    batch_size: Optional[int] = None,
     cb: Callback = DEFAULT_CALLBACK,
 ) -> None:
+    batch_size = batch_size or INSERT_BATCH_SIZE
     rows: list[UDFResult] = []
     # Optimization: Compute row types once, rather than for every row.
     udf_col_types = get_udf_col_types(warehouse, udf)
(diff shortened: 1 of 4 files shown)


dreadatour:
Looks great, thank you! 👍
I only afraid of backward compatibility. If we will implement auto flush size based on row size, for example, these settings will be deprecated and broke existing scripts.

shcheklein:
QQ - is there a possibility to make it about size (in MBs) vs number of rows?

dberenbaum:
Reference to how dask manages memory: https://distributed.dask.org/en/stable/worker-memory.html

src/datachain/query/dataset.py
```diff
@@ -396,7 +397,9 @@ def process_udf_outputs(
         for row in udf_output:
             cb.relative_update()
             rows.append(adjust_outputs(warehouse, row, udf_col_types))
-            if len(rows) >= batch_size:
+            if len(rows) >= batch_size or (
+                len(rows) % 10 == 0 and psutil.virtual_memory().percent > 80
+            ):
```
dberenbaum, line 402: @shcheklein @dreadatour Thoughts on this heuristic as a simple way to automatically check for memory limits instead of having to set them manually? For comparison, dask checks every 200 ms and pauses workers when they reach 80% of the memory limit.

shcheklein: sounds good to me. We should try this.

dreadatour: Wow, sounds good! Love the idea and simplicity.

Issue #327, opened 2024-08-20
Empty save() does not works as query


dreadatour:
Empty `save()` in datachain query fails with `Internal error on creating dataset` error.

Query example:

```python
from datachain import DataChain

DataChain.from_storage("gs://dvcx-datacomp-small/metadata", anon=True).save()
```

This happens because if `name` is not set, [temp dataset name will be used](https://github.com/iterative/datachain/blob/main/src/datachain/query/dataset.py#L1659-L1660). At the same time, after `save()` is finishes, [temp datasets are cleaned up](https://github.com/iterative/datachain/blob/main/src/datachain/query/dataset.py#L1700).

Some explanation:

> Empty save is needed because operations are lazy. Chains are not executed immediately and intermediate results are not stored.
For example, if we want to create `dc_filtered_1` and `dc_embeddings` from `dc`, without `save()` intermediate `dc` chain will be executed twice, for each children.

Couple options here:
1. Skip saved temp dataset on cleanup
2. Use `exec()` instead of empty `save()`. Need to check if this is proper replace + need to add `exec()` to docs.

skshetry:
Note that this happens as a part of session cleanup, not during `DatasetQuery.cleanup()`.

https://github.com/iterative/datachain/blob/0eb79599a0d774d89b5ff5a34db2b3dcba9a5016/src/datachain/query/session.py#L70

skshetry:
If I remember correctly, this was intended to work as it is.

Btw, can you share what the "Internal Error" is? It's working fine for me on CLI.

dreadatour:
> If I remember correctly, this was intended to work as it is.

Interesting, @mnrozhkov says it works before, also he have workflows based in this behavior. May be he can provide more details on this? 🙏

> Btw, can you share what the "Internal Error" is? It's working fine for me on CLI.

This is literally the error 👀 From this line: https://github.com/iterative/datachain/blob/0eb79599a0d774d89b5ff5a34db2b3dcba9a5016/src/datachain/catalog/catalog.py#L100

skshetry:
Looks like this is raised on Studio side.

https://github.com/iterative/studio/blob/6e7ed3fe9b30ef2631b7cee97e5eb621263e155b/backend/datachain_worker/tasks/run_job.py#L367

skshetry:
I think we should disable Session cleanup on Studio inside the scripts.

We can decide if we want to delete the datasets or hide it, but that should happen after the script runs.

[...]

PR #343, opened 2024-08-22
session: support disabling temp datasets cleanup on exit via an envvar

skshetry:
- Adds `DATACHAIN_SESSION_CLEANUP_ON_EXIT` envvar to enable/disable temp datasets cleanup on exit. (by default is enabled).
- Minor refactoring to move environ-related utils to `datachain.environ`. Also added `get_bool_env` which is based on datachain-server's implementation (`default` is renamed to `undefined`).
- Removed unused `get_env_list`.

Related #327.

### DIFF
--- a/src/datachain/catalog/loader.py
+++ b/src/datachain/catalog/loader.py
@@ -14,7 +14,7 @@ from datachain.data_storage.sqlite import (
     SQLiteMetastore,
     SQLiteWarehouse,
 )
-from datachain.utils import get_envs_by_prefix
+from datachain.environ import get_envs_by_prefix
 
 ID_GENERATOR_SERIALIZED = "DATACHAIN__ID_GENERATOR"
 ID_GENERATOR_IMPORT_PATH = "DATACHAIN_ID_GENERATOR"
--- a/src/datachain/environ.py
+++ b/src/datachain/environ.py
@@ -1,0 +1,36 @@
+import os
+from typing import TypeVar, Union, overload
+
+T = TypeVar("T")
+
+
+@overload
+def get_bool_env(name: str) -> bool: ...
+
+
+@overload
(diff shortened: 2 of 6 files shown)


dreadatour (approved):
Looks good to me! Thank you for these improvements ❤️

amritghimire (approved):

shcheklein (changes requested):
Can we briefly review the expected product behavior on this? It feels we are "hiding" the product issue a bit under this env var. By solving it for a specific `save` we are disable potentially a cleanup of a lot of tables, right?

- How do we save in Studio a dataset returned by query and prevent it from GC?
- Can we do the same for this `save` situation? Can we make `save()` transparent for the Studio logic (`save()` does extra table that is cleaned up, but we also return and somehow in Studio-way save the result as well)?

Sorry if my question a bit of touch - I haven't reviewed the whole code logic for this. I'm primarily coming to this from the product perspective. So, apologies, if I'm missing something obvious here.

skshetry:
Responded on https://github.com/iterative/datachain/issues/327#issuecomment-2306100171. This `save()` is working as expected.

> How do we save in Studio a dataset returned by query and prevent it from GC?

We create a dataset on a separate prefix on detached queries (i.e. one without save).

https://github.com/iterative/datachain/blob/49141a5b069c7d13d90a1798659faacf494e1e97/src/datachain/dataset.py#L27

Garbage collect only cleans up temporary tables.

https://github.com/iterative/datachain/blob/49141a5b069c7d13d90a1798659faacf494e1e97/src/datachain/data_storage/warehouse.py#L933-L933

https://github.com/iterative/datachain/blob/49141a5b069c7d13d90a1798659faacf494e1e97/src/datachain/data_storage/warehouse.py#L67

https://github.com/iterative/datachain/blob/49141a5b069c7d13d90a1798659faacf494e1e97/src/datachain/data_storage/warehouse.py#L66

PR #345, opened 2024-08-22
Convert 'Union[str, Literal[...]]' type to string

dreadatour:
Fix for https://github.com/iterative/datachain-examples/issues/7

Special case for `Union[str, Literal[...]]` type.

(tests to be added if this fix is OK)


volkfox (approved):

volkfox:
verified works with Claude

dmpetrov (approved):
Looks good!  @dreadatour thank you for quick turnaround!

It seems like a workaround. But let's merge it!

mnrozhkov (approved):

Issue #91, opened 2024-07-18
Serialize dataset to/from parquet


dmpetrov:
Users should be able to save datasets in parquet files and read these back as a dataset with all the types. For example, save Mistral responses and then convert them back with the correct structure type.

- [x] Research: Find a way to preserve schema in parquet files (or outside - user should specify it)
- [x] Implement basic parquet saving using pandas
- [x] Out-of memory parquet serialization
- [x] Out-of memory csv serialization

PR #348, opened 2024-08-23
Adding Custom Type (De)Serialization to Signal Schema

dtulga:
This adds the ability to save and restore model classes (feature classes) to SignalSchema, by saving these in a `_custom_types` key, then restoring them by creating a dynamic pydantic class with the stored definition. This is useful for when a custom model class is declared and used to create a dataset, but then this dataset is viewed or used in a DataChain somewhere else where this original definition is not available. This will also be useful for any other import / export functions where the original types may need to be changed, such as in parquet import/export, although this will be included in a followup/separate PR(s) to avoid this getting too complex and keep these PRs focused.

This has been tested locally and with Studio. This fixes https://github.com/iterative/studio/issues/10417 and is part of #91

### DIFF
--- a/src/datachain/lib/signal_schema.py
+++ b/src/datachain/lib/signal_schema.py
@@ -75,10 +75,12 @@ class SignalResolvingTypeError(SignalResolvingError):
         )
 
 
-dynamic_feature_models: dict[str, type[BaseModel]] = {}
+_dynamic_feature_models: dict[str, type[BaseModel]] = {}
 
 
-def get_or_create_feature_model(name: str, fields: dict[str, type]) -> type[BaseModel]:
+def get_or_create_feature_model(
+    name: str, fields: dict[str, Union[type, tuple[type, Any]]]
+) -> type[BaseModel]:
     """
     This gets or returns a dynamic feature model for use in restoring a model
     from the custom_types stored within a serialized SignalSchema. This is useful
@@ -87,15 +89,18 @@ def get_or_create_feature_model(name: str, fields: dict[str, type]) -> type[Base
     is used in a DataChain in a separate script where that model is not declared.
     """
     name = name.replace("@", "_")
-    if name not in dynamic_feature_models:
+    if name not in _dynamic_feature_models:
         cls = create_model(
             name,
             __base__=DataModel,  # type: ignore[call-overload]
             # These are tuples for each field of: annotation, default (if any)
-            **{field_name: (anno, None) for field_name, anno in fields.items()},
+            **{
+                field_name: anno if isinstance(anno, tuple) else (anno, None)
+                for field_name, anno in fields.items()
+            },
         )
-        dynamic_feature_models[name] = cls
-    return dynamic_feature_models[name]
+        _dynamic_feature_models[name] = cls
+    return _dynamic_feature_models[name]
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


src/datachain/lib/signal_schema.py
```diff
@@ -117,40 +146,119 @@ def from_column_types(col_types: dict[str, Any]) -> "SignalSchema":
...
+        orig = get_origin(fr)
+        if orig in (Literal, LiteralEx):
+            # Literal has no __name__ in Python 3.9
+            name = "Literal"
+        else:
+            name = str(fr.__name__)
+        if (
+            not ModelStore.is_pydantic(fr)
+            or not issubclass(fr, BaseModel)
+            or name in ("File", "TextFile", "ImageFile")
```
dreadatour, line 162: Will it be useful in future to have the list of these types somewhere as a constant?

dreadatour (approved):
Code looks good to me as long as it has been tested with Studio 👍

src/datachain/lib/signal_schema.py
```diff
@@ -69,6 +75,29 @@ def __init__(self, method: str, field):
         )
 
 
+dynamic_feature_models: dict[str, type[BaseModel]] = {}
```
shcheklein, line 78: should it be `_dynamic_feature_models` (protected) or even `__dynamic_feature_models`?

do we use it as a cache?

src/datachain/lib/signal_schema.py
```diff
@@ -117,40 +146,119 @@ def from_column_types(col_types: dict[str, Any]) -> "SignalSchema":
...
+        dict, and returns the name of the type provided."""
+        orig = get_origin(fr)
+        if orig in (Literal, LiteralEx):
+            # Literal has no __name__ in Python 3.9
+            name = "Literal"
+        else:
+            name = str(fr.__name__)
+        if (
+            not ModelStore.is_pydantic(fr)
+            or not issubclass(fr, BaseModel)
```
dberenbaum, line 161: Isn't this already covered by `not ModelStore.is_pydantic(fr)`?

dmpetrov:
@dtulga great change - neat and clean! I see there are already good comment inline please address these and merge.

src/datachain/lib/signal_schema.py
dberenbaum: How do this and `custom_types` differ from the `ModelStore`? Is it possible to reuse `ModelStore` for either of these?

src/datachain/lib/signal_schema.py
dtulga: I agree, although I'm not sure where to put them. Where do you think would be best for a "Built-In Feature Types" list? (As in, easy to maintain and not miss that these are listed somewhere.)

src/datachain/lib/signal_schema.py
dtulga: Sure, changed it to `_dynamic_feature_models` (protected) and yes, this is used as a cache.

This is intended more as a cache of the defined dynamic models, instead of the registered list of `ModelStore` - as this only includes the created dynamic models, not all models. I'm not 100% clear on how `ModelStore` is used and what caveats there would be with using it for this cache, especially since it does not have any mechanism for creating new models, only registering existing ones. I could look into integrating these two features together if you think it would be useful, but I'm not sure how complex that would be (and what side-effects there might be) and so it should probably be done in a followup.

src/datachain/lib/signal_schema.py
dtulga: Yes, but the type checker kept throwing errors, so I added this in addition. I changed this to just include the checks from `is_pydantic` instead so there are no duplicate calls.

src/datachain/lib/signal_schema.py
```diff
@@ -117,40 +151,116 @@ def from_column_types(col_types: dict[str, Any]) -> "SignalSchema":
...
-                orig = get_origin(fr_type)
-                args = get_args(fr_type)
-                # Check if fr_type is Optional
-                if orig == Union and len(args) == 2 and (type(None) in args):
-                    fr_type = args[0]
-                signals[name] = str(fr_type.__name__)  # type: ignore[union-attr]
+                type_name, fr_type = SignalSchema._get_name_original_type(fr_type)
+                signals[name] = type_name
+            self.serialize_custom_model_fields(type_name, fr_type, custom_types)
+        if custom_types:
```
dberenbaum, line 215: Am I missing something, or is `custom_types` always empty here?

dberenbaum: Should it be an attribute of `SignalSchema`? It looks like a local variable but doesn't seem to be used like one AFAICT.

src/datachain/lib/signal_schema.py
dberenbaum: Is there much harm in dropping this condition and serializing custom fields even for those built-in types? That way, even if a future version were to drop or rename any of these, old datasets could still be read as expected.

src/datachain/lib/signal_schema.py
dberenbaum: Have you tried just dropping `_dynamic_feature_models` and seeing if it makes any difference? AFAIU because you use `DataModel` as the base when creating the dynamic models, they should already be registered to the `ModelStore` and should already be cached and should be caught [here](https://github.com/iterative/datachain/blob/2a86ad17b910d39e197ea24e3fd4a507b62249d0/src/datachain/lib/signal_schema.py#L227-L229) before getting duplicated.

src/datachain/lib/signal_schema.py
dtulga: The `custom_types` dict is set within `serialize_custom_model_fields`, see above for example: https://github.com/iterative/datachain/pull/348/files/2a86ad17b910d39e197ea24e3fd4a507b62249d0#diff-69a1fc2d77ff106392ad7649ed8cf7ec344dad230af2e4d0ea9c64b816cee15bR200

I didn't want to just make this a return-based set as it would lead to extra code complexity (due to the recursive nature of `serialize_custom_model_fields`) and I didn't want to make this an attribute of `SignalSchema` as this information is only valid during the call to `serialize` and may become out of date or incorrect at any other times (and is not used at any other times).

src/datachain/lib/signal_schema.py
dtulga: Upon further consideration, I decided to just serialize those built-in models as well, since the default is to use the existing and defined model in the code first, this should be harmless unless they are renamed or deleted in the future, in which case having them serialized would be useful for reading old datasets.

dberenbaum (approved):

PR #350, opened 2024-08-24
Remove Catalog.merge_datasets()

### LINKED ISSUE #349: Remove `Catalog.merge_datasets()`
Current implementation doesn't really work with our new feature schemas from `lib` and is also not really used from anywhere except from CLI. 

We already have`DataChain.merge()` which maybe can be used instead, but we need to check if dataset dependencies are handled in a right way with that one. 
Maybe we can also remove CLI command for this as well as it seems kind of random to have that there(this is more of a product decision though)

EdwardLi-coder:
Fix: Remove Catalog.merge_datasets() #349

Closes #349


shcheklein (approved):
thanks @EdwardLi-coder

rlamy (approved):

PR #351, opened 2024-08-25
Add `device` parameter to convert functions and update usage model de…

ayasyrev:
Add `device` parameter to convert functions and update usage model device in clip_similarity_scores.
related https://github.com/iterative/datachain-examples/issues/6

### DIFF
--- a/src/datachain/lib/clip.py
+++ b/src/datachain/lib/clip.py
@@ -1,5 +1,5 @@
 import inspect
-from typing import TYPE_CHECKING, Any, Callable, Literal, Union
+from typing import TYPE_CHECKING, Any, Callable, Literal, Optional, Union
 
 import torch
 from transformers.modeling_utils import PreTrainedModel
@@ -39,6 +39,7 @@ def clip_similarity_scores(
     tokenizer: Callable,
     prob: bool = False,
     image_to_text: bool = True,
+    device: Optional[Union[str, torch.device]] = None,
 ) -> list[list[float]]:
     """
     Calculate CLIP similarity scores between one or more images and/or text.
@@ -52,6 +53,7 @@ def clip_similarity_scores(
         prob : Compute softmax probabilities.
         image_to_text : Whether to compute for image-to-text or text-to-image. Ignored
             if only one of images or text provided.
+        device : Device to use. Defaults is None - use model's device.
 
 
     Example:
@@ -130,7 +132,13 @@ def clip_similarity_scores(
         ```
     """
 
-    device = next(model.parameters()).device
+    if device is None:
+        if hasattr(model, "device"):
+            device = model.device
+        else:
+            device = next(model.parameters()).device
+    else:
+        model = model.to(device)
     with torch.no_grad():
         if images is not None:
             encoder = _get_encoder(model, "image")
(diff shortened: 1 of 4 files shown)


src/datachain/lib/clip.py
```diff
@@ -130,17 +130,20 @@ def clip_similarity_scores(
         ```
     """
 
+    device = next(model.parameters()).device
```
dberenbaum, line 133: Probably worth noting for future reference that this is not guaranteed to work but is likely good enough for almost all use cases. See https://stackoverflow.com/questions/58926054/how-to-get-the-device-type-of-a-pytorch-module-conveniently.

dberenbaum: Maybe the function should also include a `device` parameter but default to this if no device is passed?

dberenbaum (commented):
Looks great! See one comment about adding a `device` parameter to `clip_similarity_scores()`. Thanks for the contribution!

src/datachain/lib/clip.py
ayasyrev: I suggest this first. Let's add `device`.

dberenbaum:
One more thought: I think we should also add a `device` parameter to `DataChain.to_pytorch()` for model training. Looking at the clip fine tuning example, the image, text, and label need to be sent to the correct device during training.

@ayasyrev Let me know if you are up for this or would prefer to leave it out of this PR. If you want to take a look, most of the code is in https://github.com/iterative/datachain/blob/main/src/datachain/lib/pytorch.py.

ayasyrev:
> One more thought: I think we should also add a `device` parameter to `DataChain.to_pytorch()` for model training. Looking at the clip fine tuning example, the image, text, and label need to be sent to the correct device during training.
> 
> @ayasyrev Let me know if you are up for this or would prefer to leave it out of this PR. If you want to take a look, most of the code is in https://github.com/iterative/datachain/blob/main/src/datachain/lib/pytorch.py.
Will check it.
Usually (OK, IMHO) move "batch" to `device`  used after dataloader. Will look for implementation.

dberenbaum:
> Usually (OK, IMHO) move "batch" to `device` used after dataloader. Will look for implementation.

Yes, I guess this is needed in complex use cases where there are multiple devices. However, I think it makes sense to have a simple way to do this similar to https://huggingface.co/docs/datasets/en/use_with_pytorch.

ayasyrev:
> Yes, I guess this is needed in complex use cases where there are multiple devices. However, I think it makes sense to have a simple way to do this similar to https://huggingface.co/docs/datasets/en/use_with_pytorch.
OK, check this example. 
Can do it.
Do you think to add it to this pr or better do another one?

dberenbaum:
> Do you think to add it to this pr or better do another one?

Whatever you prefer @ayasyrev. If you want to do a separate PR, let me know.

dberenbaum (approved):

dberenbaum:
Could you please fix the tests that got broken @ayasyrev? The tests use a mock for the model to speed up the tests, so you likely need to add the `parameters` attribute to the mocked model [here](https://github.com/iterative/datachain/blob/main/tests/unit/lib/conftest.py).

PR #352, opened 2024-08-25
improve datachain subtract

### LINKED ISSUE #181: DataChain.subtract() on differently named columns
We need to handle use cases where signal names do not match. Like: "file.name" == "wds.laion.file.name" AND "file.etag" == "wds.laion.file.etag". That probably means adding a `right_on` argument to `subtract()`, similarly to `merge()`.
Follow-up to #103 and #157.

EdwardLi-coder:
Fix: DataChain.subtract() on differently named columns 

Fixes #181

### DIFF
--- a/src/datachain/query/dataset.py
+++ b/src/datachain/query/dataset.py
@@ -296,15 +296,23 @@ class DatasetDiffOperation(Step):
 
 @frozen
 class Subtract(DatasetDiffOperation):
-    on: Sequence[str]
+    on: Sequence[Union[str, tuple[str, str]]]
 
     def query(self, source_query: Select, target_query: Select) -> sa.Selectable:
         sq = source_query.alias("source_query")
         tq = target_query.alias("target_query")
         where_clause = sa.and_(
-            getattr(sq.c, col_name).is_not_distinct_from(getattr(tq.c, col_name))
-            for col_name in self.on
-        )  # type: ignore[arg-type]
+            *[
+                getattr(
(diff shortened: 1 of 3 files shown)


src/datachain/query/dataset.py
```diff
@@ -296,15 +296,23 @@ def q(*columns):
 
 @frozen
 class Subtract(DatasetDiffOperation):
-    on: Sequence[str]
+    on: Sequence[Union[str, tuple[str, str]]]
```
rlamy, line 299: Shouldn't this be `Sequence[tuple[str, str]]`?

EdwardLi-coder: Yeah.I will update it.Thank you.

EdwardLi-coder: > Shouldn't this be `Sequence[tuple[str, str]]`?

Yeah.I will update it.Thank you.

tests/unit/lib/test_datachain.py
```diff
@@ -1811,3 +1846,32 @@ def test_from_csv_nan_inf(tmp_dir, test_session):
     assert np.isnan(res[0])
     assert np.isposinf(res[1])
     assert np.isneginf(res[2])
+
+
+def test_subtract_with_different_column_names(test_session):
```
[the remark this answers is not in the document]
EdwardLi-coder: > Is this test useful? It seems to duplicate the addition to `test_subtract()`.

Sorry.I forgot to delete it and will do so now

PR #357, opened 2024-08-27
Fix empty 'save()' as query last statement

dreadatour:
Fix for https://github.com/iterative/datachain/issues/327
See [this comment](https://github.com/iterative/datachain/issues/327#issuecomment-2311622854).

Closes #343.

### DIFF
--- a/src/datachain/query/dataset.py
+++ b/src/datachain/query/dataset.py
@@ -1778,7 +1778,7 @@ def query_wrapper(dataset_query: DatasetQuery) -> DatasetQuery:
     save = bool(os.getenv("DATACHAIN_QUERY_SAVE"))
     save_as = os.getenv("DATACHAIN_QUERY_SAVE_AS")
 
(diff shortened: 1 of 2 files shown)


src/datachain/query/dataset.py
```diff
@@ -1778,6 +1778,10 @@ def query_wrapper(dataset_query: DatasetQuery) -> DatasetQuery:
     save = bool(os.getenv("DATACHAIN_QUERY_SAVE"))
     save_as = os.getenv("DATACHAIN_QUERY_SAVE_AS")
 
+    is_session_dataset = dataset_query.name.startswith(
+        Session.DATASET_PREFIX + dataset_query.session.name + "_"
+    )
```
skshetry, line 1783:
```suggestion
    is_session_dataset = dataset_query.name is not None and dataset_query.name.startswith(
        Session.DATASET_PREFIX + dataset_query.session.name + "_"
    )
```

`dataset_query` could be detached, so this needs a guard.

skshetry (approved):

PR #358, opened 2024-08-27
session: add get_temp_prefix

skshetry:
Small cleanup to avoid creating hardcoded prefixes.


dreadatour (approved):
Thank you! 🙏 Looks awesome!

PR #362, opened 2024-08-27
added embeddings/gen example

tibor-mach:


examples/llm_and_nlp/unstructured-embeddings-gen.py
```diff
@@ -0,0 +1,69 @@
...
+            key=file.path,
+            text=chunk.text,
+            embeddings=chunk.embeddings,
+        )
+
+
+dc = (
+    DataChain.from_storage(source)
+    .settings(parallel=-1)
+    .filter(C.file.path.glob("*.pdf"))
```
shcheklein, line 63: should we put some limit here, at least for tests?

mattseddon:
Looks like the example won't run on Windows due to `ImportError: DLL load failed while importing onnx_cpp2py_export: A dynamic link library (DLL) initialization routine failed.`. You'll need to poke around in `unstructured` to see what version of `onnx` they are using and whether or not it can be downgraded/pinned to fix the issue. [This issue](https://github.com/onnx/onnx/issues/6267) is likely the cause.

examples/llm_and_nlp/unstructured-embeddings-gen.py
```diff
@@ -0,0 +1,69 @@
+from collections.abc import Iterator
```
mattseddon, line 1: [C] In other examples there is usually a list of packages to install that don't come with the base datachain install.

examples/llm_and_nlp/unstructured-embeddings-gen.py
mattseddon: could parameterise the source variable (as per `examples/multimodal/wds.py`) to only ingest a subset of files in the test

shcheklein:
@tibor-mach is this one done? any progress on it? what are the blockers?

mattseddon:
> @tibor-mach is this one done? any progress on it? what are the blockers?

The new example needs parameterized as it takes > 1hr to run on Windows.

examples/llm_and_nlp/unstructured-embeddings-gen.py
mattseddon: The idea is to have the same kind of block as shown in https://github.com/iterative/datachain/pull/426 unless a user installs `datachain[examples]` this example won't work.

examples/llm_and_nlp/unstructured-embeddings-gen.py
mattseddon: I think there is a bug in the Windows CI where `parallel` mode breaks. I've put up https://github.com/iterative/datachain/pull/424 as an attempt to fix the issue.

mattseddon (approved):

PR #366, opened 2024-08-28
unstructured example - remove misleading install instructions

mattseddon:
Missed in https://github.com/iterative/datachain/pull/332


dtulga (approved):
LGTM

PR #376, opened 2024-08-30
Simplify datachain.lib.listing by reusing Cilent.scandir()

### LINKED ISSUE #373: Code duplication: `datachain.lib.listing` should use `Client.scandir()`
`datachain.lib.listing` contains a partial copy of the code used by `Client.scandir()`, which only includes a generic version of the "nested" algorithm. This has some disadvantages:
* Code duplication is always bad
* It misses customisations added for some of the clients
* Using the "nested" algorithm will cause a regression on https://github.com/iterative/dvcx/issues/1403 as soon as we start using it in #294.

A quick fix is to convert `list_bucket()` into a wrapper around `Client.scandir()`, converting the `Entry` objects the latter returns into `File`, but it shouldn't be hard to make `Client.scandir()` return `File` directly, as already planned in #33.

rlamy:
Fixes #373

### DIFF
--- a/src/datachain/lib/listing.py
+++ b/src/datachain/lib/listing.py
@@ -1,103 +1,14 @@
-import asyncio
-from collections.abc import AsyncIterator, Iterator, Sequence
-from typing import Callable, Optional
+from collections.abc import Iterator
+from typing import Callable
 
-from botocore.exceptions import ClientError
 from fsspec.asyn import get_loop
 
 from datachain.asyn import iter_over_async
 from datachain.client import Client
-from datachain.error import ClientError as DataChainClientError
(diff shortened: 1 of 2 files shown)


shcheklein (approved):
Thanks @rlamy !

ilongin:
Even though I wanted to to this exact thing at first, my understanding was  that we need to move as much "core" logic to "user / lib" level as possible for listing/indexing (that was part of the requirement). For `scandir` it was kind of questionable whether it really needs to be in the `Client` or we can move it to "upper" level and since it basically has most of the codebase used in listing I decided to lift it up. 
I'm not against this refactoring, just explaining the reasoning.

PR #378, opened 2024-09-01
Use stderr for sql debug prints

shcheklein:
Quick fix for `DEBUG_SHOW_SQL_QUERIES` when queries with `src/datachain/lib/meta_formats.py` `read_meta` are used. It's broken since we have:

`# ugly hack: datachain is run redirecting printed outputs to a variable`

there, where we execute command output. It breaks if there are some debug statements.

We need to get back and review `src/datachain/lib/meta_formats.py`.

### DIFF
--- a/src/datachain/data_storage/sqlite.py
+++ b/src/datachain/data_storage/sqlite.py
@@ -143,7 +143,9 @@ class SQLiteDatabaseEngine(DatabaseEngine):
             db.execute("PRAGMA synchronous = NORMAL")
             db.execute("PRAGMA case_sensitive_like = ON")
             if os.environ.get("DEBUG_SHOW_SQL_QUERIES"):
-                db.set_trace_callback(print)
+                import sys
+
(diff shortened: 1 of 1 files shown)


mattseddon (approved):
This won't break debugging in Studio... right?

shcheklein:
I think this SQLite only for now. So, no - it should be fine.

I'll check if we need something similar on the CH side.

On the Studio UI we should be capturing both - stdout and stderr I think.

shcheklein:
```
        engine = sqlalchemy.create_engine(
            dsn,
            echo=debug,
            echo_pool=debug,
            logging_name="warehouse.clickhouse",
        )
```

`echo=debug` on the CH side, and no easy way to redirect this - https://github.com/sqlalchemy/sqlalchemy/discussions/5878

so, this PR won't break it, but also doesn't fix this issue if someone decides to use `DATACHAIN_CH_DATABASE_DEBUG` on the Studio side.

PR #379, opened 2024-09-02
remove unused finally block

mattseddon:


src/datachain/data_storage/warehouse.py
```diff
@@ -218,35 +218,26 @@ def dataset_select_paginated(
...
+
+        while True:
+            if limit is not None:
+                limit -= num_yielded
+                if limit == 0:
+                    break
+                if limit < page_size:
+                    paginated_query = paginated_query.limit(None).limit(limit)
+
+            results = self.dataset_rows_select(paginated_query.offset(offset))
```
mattseddon, line 230: https://github.com/iterative/datachain/pull/339 changed the code to 

```python
results =  yield from convert_rows_custom_column_types(
                query.selected_columns, rows, self.db.dialect
            )
```

which I don't think will ever have a `close` attribute.

shcheklein (approved):
thanks @mattseddon , I wonder though why we didn't catch this with tests?

src/datachain/data_storage/warehouse.py
```diff
@@ -318,10 +309,21 @@ def dataset_rows_select(
         """
         Fetch dataset rows from database.
         """
-        rows = self.db.execute(query, **kwargs)
-        yield from convert_rows_custom_column_types(
-            query.selected_columns, rows, self.db.dialect
-        )
+
```
skshetry, line 312:
```suggestion
        rows = None
```

src/datachain/data_storage/warehouse.py
skshetry: Since the `results` is an iterator, I think we still need to close this.

mattseddon: I moved it underneath into `dataset_rows_select` but it doesn't look like it is required there either. The original intention from https://github.com/iterative/dvcx/pull/1369 was to get `test_row_generator_with_new_columns_wrong_type` passing. We can now run the entire suite without having the `finally` statement in either method.

Issue #360, opened 2024-08-27
Remove support for executing last expression in the script


skshetry:

skshetry:
A couple of questions:

1) Do we need to show previews for multiple datasets, given that there can be multiple `save()`s within the script? Or, is it okay to show the last one that wins? 

2) Since we don't have any custom wrapper, we won't be able to show previews for temp datasets or `persist()`-ed dataset in `datachain query` command. I guess that would be okay? Same in Studio.

Since we won't be wrapping the datasets, the only way for us to know what datasets were created in the script is by using `job_id`.

skshetry:
Looks like we will still need some kind of wrapper for "metrics" to work.

shcheklein:
> Do we need to show previews for multiple datasets, given that there can be multiple save()s within the script? Or, is it okay to show the last one that wins?

Yes, I think that's fine to show the last one.

> Since we don't have any custom wrapper, we won't be able to show previews for temp datasets or persist()-ed dataset in datachain query command. I guess that would be okay? Same in Studio.

could you point to the code that wraps it / add a bit more details?

PR #385, opened 2024-09-03
avoid splitting script into feature files/scripts

skshetry:
Reverts https://github.com/iterative/dvcx/pull/1500.

I don't think this is needed after we started using cloudpickle for parallel udfs since https://github.com/iterative/datachain/pull/65.

Part of #360 .


skshetry:
ping @dtulga.

amritghimire (approved):
Lets wait for Studio CI to be green before merging this

skshetry:
All of the tests passed in Studio. See above referenced link.

PR #388, opened 2024-09-04
allow merge on expressions

### LINKED ISSUE #299: Introduce @ operator for File path manipulations
### Description

There is a number of common operations to File path, e.g.   {stem, dir, basename, ext} which are poorly supported.

For example, if we need to merge two datasets "images" and "meta" by the file stem, we currently require many manipulations to create and remove merge keys – which is not ideal:

```
import datachain.sql.functions.path

images = images.mutate(images_stem = path.stem("file.path")
meta = meta.mutate(meta_stem = path.stem("file.path")

annotated_images = images.merge(meta, on="image_stem", right_on="meta_stem")

annotated_images.select_except(["image_stem", "meta_stem"])
```

A preferred way is to create path-related functions on the fly:

```
annotated = images.merge(meta, on="file.path@stem", right_on="meta_file.path@stem")
```

mattseddon:
Closes #299

This PR allows users to merge DataChains using `Column`s and/or expressions (e.g. `path.file_stem(dc.c("source.path"))`). Examples of the new syntax are shown in `examples/multimodal/clip_inference.py` & `examples/multimodal/wds.py`. Once this change has been merged I will make the appropriate changes in the `datachain-examples` repo.

It should be noted that the change is backwards compatible and `merge` will continue to accept strings for the `on` and `right_on` parameters.

### DIFF
--- a/src/datachain/lib/dc.py
+++ b/src/datachain/lib/dc.py
@@ -56,7 +56,7 @@ from datachain.query.dataset import (
     PartitionByType,
     detach,
 )
-from datachain.query.schema import Column, DatasetRow
+from datachain.query.schema import DEFAULT_DELIMITER, Column, DatasetRow
 from datachain.sql.functions import path as pathfunc
 from datachain.utils import inside_notebook
 
@@ -112,6 +112,16 @@ class DatasetFromValuesError(DataChainParamsError):  # noqa: D101
         super().__init__(f"Dataset{name} from values error: {msg}")
 
 
+def _get_merge_error_str(col: Union[str, sqlalchemy.ColumnElement]) -> str:
+    if isinstance(col, str):
+        return col
+    if isinstance(col, sqlalchemy.Column):
+        return col.name
+    if isinstance(col, GenericFunction):
+        return f"{col.name} expression"
+    return str(col)
+
+
 class DatasetMergeError(DataChainParamsError):  # noqa: D101
     def __init__(  # noqa: D107
         self,
@@ -120,18 +130,9 @@ class DatasetMergeError(DataChainParamsError):  # noqa: D101
         msg: str,
     ):
         def get_str(on: Sequence[Union[str, sqlalchemy.ColumnElement]]) -> str:
-            str_list = []
             if not isinstance(on, Sequence):
                 return str(on)  # type: ignore[unreachable]
-            for col in on:
-                if isinstance(col, sqlalchemy.ColumnElement):
-                    str_list.append(col.name)
-                if isinstance(col, str):
-                    str_list.append(col)
-                if isinstance(col, GenericFunction):
-                    str_list.append(f"{col.name} expression")
-
-            return ", ".join(str_list)
+            return ", ".join([_get_merge_error_str(col) for col in on])
 
         on_str = get_str(on)
         right_on_str = (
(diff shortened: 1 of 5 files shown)


src/datachain/query/dataset.py
```diff
@@ -1175,8 +1175,12 @@ def attached(self) -> bool:
         """
         return self.name is not None and self.version is not None
 
-    def c(self, name: Union[C, str]) -> "ColumnClause[Any]":
-        col = sqlalchemy.column(name) if isinstance(name, str) else name
+    def c(self, column: Union[C, str]) -> "ColumnClause[Any]":
+        col: sqlalchemy.ColumnClause = (
+            sqlalchemy.column(column)
+            if isinstance(column, str)
+            else sqlalchemy.column(column.name, column.type)
```
mattseddon, line 1182: [F] It does not seem possible to overwrite the table of a column which already has the table set. I.e `test_merge_with_itself_column` fails without this change.

src/datachain/lib/dc.py
```diff
@@ -112,10 +112,29 @@ def __init__(self, name, msg):  # noqa: D107
...
+        on: Sequence[Union[str, sqlalchemy.ColumnElement]],
+        right_on: Optional[Sequence[Union[str, sqlalchemy.ColumnElement]]],
+        msg: str,
+    ):
+        def get_str(on: Sequence[Union[str, sqlalchemy.ColumnElement]]) -> str:
+            str_list = []
+            if not isinstance(on, Sequence):
+                return str(on)  # type: ignore[unreachable]
+            for col in on:
+                if isinstance(col, sqlalchemy.ColumnElement):
```
mattseddon, line 126: Should be `Column`

src/datachain/query/dataset.py
mattseddon: This change is no longer required for this PR to work. I'd be happy to split it out if required.

src/datachain/lib/dc.py
```diff
@@ -252,13 +272,24 @@ def column(self, name: str) -> Column:
         """Returns Column instance with a type if name is found in current schema,
         otherwise raises an exception.
         """
-        name_path = name.split(".")
+        if "." in name:
+            name_path = name.split(".")
+        elif DEFAULT_DELIMITER in name:
+            name_path = name.split(DEFAULT_DELIMITER)
```
dreadatour, line 278: Just wonder, if it is possible to have both dot (`.`) and `DEFAULT_DELIMITER` (`__`) in column name? 🤔

src/datachain/lib/dc.py
```diff
@@ -1175,54 +1215,55 @@ def merge(
...
             if len(right_on) != len(on):
                 raise DatasetMergeError(
                     on, right_on, "'on' and 'right_on' must have the same length'"
                 )
 
-            right_on_columns: list[str] = right_signals_schema.resolve(
-                *right_on
-            ).db_signals()  # type: ignore[assignment]
-
-            if len(right_on_columns) != len(on_columns):
```
dreadatour, line 1202: Are we loosing this check after these changes or I miss something? 🤔

src/datachain/query/dataset.py
dreadatour: As for me personally I am OK to leave it here. It feels natural to have this change here.

dreadatour (approved):
Looks good to me! 👍
Couple comments below, also Studio tests fails 😥. But it is not related to this PR, as far as I can see.

src/datachain/lib/dc.py
mattseddon: I don't think so. Seems like one is always swapped for the other.

src/datachain/lib/dc.py
mattseddon: The check is moved down. If any of the columns fail to resolve then we will have an entry in `errors` and we'll raise a `DatasetMergeError`.

PR #389, opened 2024-09-04
query: remove support for saving dataset query with a given name

skshetry:
Reverts https://github.com/iterative/dvcx/pull/1443.

Since we want to make `save` explicits via users, and are dropping last expression wrapping thing in #360, we won't have any mechanism to support this feature.


src/datachain/catalog/catalog.py
```diff
@@ -1297,19 +1297,6 @@ def create_dataset_from_sources(
 
         return self.get_dataset(name)
 
-    def register_new_dataset(
```
skshetry, line 1300: This API was only used in a single place inside the `query_wrapper()`.

tests/func/test_query.py
```diff
@@ -194,11 +189,12 @@ def test_query(
...
 
-    result = catalog.query(query_script, save=save, save_as=save_as)
-    if save_as:
-        assert result.dataset.name == save_as
-        assert catalog.get_dataset(save_as)
-    elif save_dataset:
+    result = catalog.query(query_script, save=save)
+    if not save:
+        assert result.dataset is None
+        return
```
skshetry, line 195: Previous test did not test with `save=False` and when `save_as=None`. I am testing this just to make parametrization easier. #360 will require adjusting these anyway.

dreadatour (approved):
My favorite kind of PRs with removing outdating code 😅👍

PR #392, opened 2024-09-04
cli: remove preview from `datachain query` command

skshetry:
Part of #360.

After removing last expression wrapping, we will not be able to get access to the dataset query, nor we'll be able to show previews.

Also, CLI is hidden at the moment. So, it is better if we disable the preview support altogether for now, as this simplifes a lot for #360.

### DIFF
--- a/src/datachain/catalog/catalog.py
+++ b/src/datachain/catalog/catalog.py
@@ -156,7 +156,6 @@ class QueryResult(NamedTuple):
     dataset: Optional[DatasetRecord]
     version: Optional[int]
     output: str
-    preview: Optional[list[dict]]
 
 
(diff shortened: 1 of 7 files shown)


dreadatour (approved):
I love this series of PRs with cleanup. Can't wait for the next episode :)

Question: do we even need `datachain query` command? 🤔

skshetry:
> I love this series of PRs with cleanup. Can't wait for the next episode :)
> 
> Question: do we even need `datachain query` command? 🤔

Good question. I find it useful for debugging, as it runs as a job, so it imitates what happens on SaaS to some degree. But I guess it's on @dmpetrov to decide if we need a user-facing command or not. It's hidden to the users anyway. :)

PR #393, opened 2024-09-04
query: remove use of pipe for communication

skshetry:
Now, we fetch all dataset versions created with a given `job_id`, and then choose the latest one by its `datetime`. (If there are multiple dataset versions with the same datetime, the first one encountered is used but that may be rare).

This removes the need for using a pipe to read and write dataset version from a script, which is too complicated.

### DIFF
--- a/src/datachain/catalog/catalog.py
+++ b/src/datachain/catalog/catalog.py
@@ -1416,7 +1416,8 @@ class Catalog:
 
         for d in datasets:
             yield from (
-                (d, v, jobs.get(v.job_id) if v.job_id else None) for v in d.versions
+                (d, v, jobs.get(str(v.job_id)) if v.job_id else None)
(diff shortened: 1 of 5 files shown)


src/datachain/catalog/catalog.py
```diff
@@ -1931,19 +1931,37 @@ def query(
                 output=output,
             )
 
+        def _get_latest_dataset():
+            for dataset, dv, job in self.list_datasets_versions():
+                if job and str(job.id) == job_id:
+                    yield dv.created_at, (dataset, dv)
```
skshetry, line 1937: We might need an API for listing dataset versions by a given job id in the future for performance reasons.

ilongin (approved):

dreadatour (approved):
🙏

PR #394, opened 2024-09-05
fix issues with new version of huggingface datasets package

mattseddon:
Fixing the breakage in the pipeline demonstrated [here](https://github.com/iterative/datachain/actions/runs/10713276424)

### DIFF
--- a/src/datachain/lib/hf.py
+++ b/src/datachain/lib/hf.py
@@ -99,7 +99,9 @@ class HFGenerator(Generator):
 
 def stream_splits(ds: Union[str, HFDatasetType], *args, **kwargs):
     if isinstance(ds, str):
-        ds = load_dataset(ds, *args, streaming=True, **kwargs)
+        if "streaming" not in kwargs:
+            kwargs["streaming"] = True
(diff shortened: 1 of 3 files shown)


tests/unit/lib/test_datachain.py
```diff
@@ -9,7 +9,6 @@
 import pyarrow.parquet as pq
 import pytest
 from datasets import Dataset
-from datasets.exceptions import DatasetNotFoundError
```
mattseddon, line 12: No longer exists. The package now throws `FileNotFoundError` as demonstrated below

src/datachain/lib/hf.py
```diff
@@ -99,7 +99,8 @@ def process(self, split: str = ""):
 
 def stream_splits(ds: Union[str, HFDatasetType], *args, **kwargs):
     if isinstance(ds, str):
-        ds = load_dataset(ds, *args, streaming=True, **kwargs)
+        kwargs["streaming"] = True
+        ds = load_dataset(ds, *args, **kwargs)
```
mattseddon, line 103: [F] This was a mypy error

skshetry (approved):

PR #395, opened 2024-09-05
do not require last statement to be an expression or an instance of DatasetQuery

skshetry:
This PR removes the requirement for the last statement to be either an expression or an instance of DatasetQuery.

After this PR, scripts without a final expression will function properly in both Datachain and Studio.

The last statement wrappers and related infrastructure will be removed in subsequent PRs.
However, this PR demonstrates that everything already works and is usable.

```python
from datachain.query import DatasetQuery
DatasetQuery("s3://dql-test-e2e", catalog=catalog).save("temp")
print("hello world!")
```

```console
$ datachain query script.py
Processed: 1 rows [00:00, 512.00 rows/s]
Generated: 9 rows [00:00, 4950.00 rows/s]
Cleanup: 1 tables [00:00, 3498.17 tables/s]
Saving: 279 rows [00:00, 68945.43 rows/s]
hello world!
```

### DIFF
--- a/tests/func/test_catalog.py
+++ b/tests/func/test_catalog.py
@@ -971,42 +971,6 @@ DatasetQuery('{src_uri}')
         assert str(exc_info.value).startswith("Query script exited with error code 1")
 
 
-def test_query_last_statement_not_expression(mock_popen, cloud_test_catalog):
-    mock_popen.configure_mock(returncode=10)
-    catalog = cloud_test_catalog.catalog
-    src_uri = cloud_test_catalog.src_uri
-
-    query_script = f"""
-from datachain.query import DatasetQuery, C
-ds = DatasetQuery('{src_uri}')
-    """
-
-    with pytest.raises(QueryScriptCompileError) as exc_info:
-        catalog.query(query_script)
-        assert str(exc_info.value).startswith(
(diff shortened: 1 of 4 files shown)


tests/func/test_catalog.py
```diff
@@ -971,42 +971,6 @@ def test_query_subprocess_wrong_return_code(mock_popen, cloud_test_catalog):
         assert str(exc_info.value).startswith("Query script exited with error code 1")
 
 
-def test_query_last_statement_not_expression(mock_popen, cloud_test_catalog):
```
skshetry, line 974: These tests are mocked tests. And since these tests are no longer a failure, they are not useful.

dreadatour (approved):
Looks good to me, thank you! 🙏

We also need to make sure it all works on Studio side (including frontend part).

skshetry:
> Looks good to me, thank you! 🙏
> 
> We also need to make sure it all works on Studio side (including frontend part).

It does work well on Studio. We might need to adjust some stuff like "Register dataset" button which is no longer possible as we need explicit saves.

But I'd not consider this as completely functional, that will happen when we completely remove this wrapping thing.

dreadatour:
Thank you for handling this ❤️

PR #397, opened 2024-09-05
Fixes settings

dberenbaum:
#342 in part added `settings` as a kwarg to be passed to the initial DataChain, but it wasn't working in some cases due to not being passed everywhere.


dberenbaum:
Not sure why the example test fails for `examples/get_started/json-csv-reader.py`. That example runs locally for me and the test passes if I force it to run locally. @mattseddon @iterative/datachain Any ideas?

mattseddon:
> Not sure why the example test fails for `examples/get_started/json-csv-reader.py`. That example runs locally for me and the test passes if I force it to run locally. @mattseddon @iterative/datachain Any ideas?
> 
> Edit: There is also a Studio test failure but it appears to be unrelated since I get the same error on `main`.

It looks like there is an error with `datamodel-code-generator` (new version released 3 days ago).

By running the code, we end up with `TypeError: Cannot recognize type ForwardRef('List[str]')`. Extracting `List[str]` from the `ForwardRef` doesn't help as we don't support `List` in `PYTHON_TO_SQL`. 

Do you know the history of why we are using `datamodel-codegen`? Looks suspect.

mattseddon:
> > Not sure why the example test fails for `examples/get_started/json-csv-reader.py`. That example runs locally for me and the test passes if I force it to run locally. @mattseddon @iterative/datachain Any ideas?
> > Edit: There is also a Studio test failure but it appears to be unrelated since I get the same error on `main`.
> 
> It looks like there is an error with `datamodel-code-generator` (new version released 3 days ago).
> 
> By running the code, we end up with `TypeError: Cannot recognize type ForwardRef('List[str]')`. Extracting `List[str]` from the `ForwardRef` doesn't help as we don't support `List` in `PYTHON_TO_SQL`.
> 
> Do you know the history of why we are using `datamodel-codegen`? Looks suspect.

More likely: 

image

rlamy (approved):
LGTM

PR #375, opened 2024-08-29
Support for reading from huggingface hub with `hf://` filesystem

dberenbaum:
Closes #368.

Example:

```python
>>> DataChain.from_parquet("hf://datasets/ylecun/mnist/**.parquet").show()
Preparing: 2 rows [00:00, 2100.83 rows/s]
Processed: 2 rows [00:00, 3178.71 rows/s]
Cleanup: 2 tables [00:00, 4447.83 tables/s]
Preparing: 2 rows [00:00, 1408.43 rows/s]
Processed: 2 rows [00:00, 2016.01 rows/s]
Parsed by pyarrow: 10000 rows [00:00, 10172.99 rows/s]
Parsed by pyarrow: 60000 rows [00:03, 18752.09 rows/s]
Processed: 2 rows [00:04,  2.21s/ rows]0000.68 rows/s]
Generated: 70000 rows [00:03, 19081.33 rows/s]
Cleanup: 5 tables [00:00, 63.35 tables/s]
           source                                          source   source                                    source                                    source    source  \
             file                                            file     file                                      file                                      file      file
           source                                            path     size                                   version                                      etag is_latest
0   hf://datasets  ylecun/mnist/mnist/test-00000-of-00001.parquet  2595890  77f3279092a1c1579b2250db8eafed0ad422088c  447ec884dcdd99cfb34e5fc7792062787508bfe5         1
1   hf://datasets  ylecun/mnist/mnist/test-00000-of-00001.parquet  2595890  77f3279092a1c1579b2250db8eafed0ad422088c  447ec884dcdd99cfb34e5fc7792062787508bfe5         1
2   hf://datasets  ylecun/mnist/mnist/test-00000-of-00001.parquet  2595890  77f3279092a1c1579b2250db8eafed0ad422088c  447ec884dcdd99cfb34e5fc7792062787508bfe5         1
3   hf://datasets  ylecun/mnist/mnist/test-00000-of-00001.parquet  2595890  77f3279092a1c1579b2250db8eafed0ad422088c  447ec884dcdd99cfb34e5fc7792062787508bfe5         1
4   hf://datasets  ylecun/mnist/mnist/test-00000-of-00001.parquet  2595890  77f3279092a1c1579b2250db8eafed0ad422088c  447ec884dcdd99cfb34e5fc7792062787508bfe5         1
[...]

PR #396, opened 2024-09-05
Explore downloading hf datasets and read images from cache

dberenbaum:
Some exploration related to #369. I don't plan to merge this but wanted to share the results. 

This PR splits `from_hf` into two methods:
- `from_hf_streaming`: This is the method that was implemented in #311. It streams the datasets and store their contents in the warehouse. Images and other binary objects are stored as bytes in the warehouse.
- `from_hf_download`: This method downloads the dataset to the hf cache, which saves them as arrow files (the `datasets` library first downloads the files and then converts them to a uniform arrow format). Images are saved as bytes in those cached arrow files and datachain reads them from there rather than storing them in the warehouse.

I used a sample of coco data to show performance considerations of using each of these approaches:

Here is `from_hf_streaming`:

```python
>>> chain = DataChain.from_hf_streaming("lmms-lab/COCO-Caption2017", split="val")
>>> images = list(chain.select("image").to_pytorch())
Processed: 1 rows [00:00, 1062.66 rows/s]
Generated: 1 rows [00:00, 1109.90 rows/s]
Parsed Hugging Face dataset: 5000 rows [02:21, 35.44 rows/s]
Processed: 1 rows [02:22, 142.01s/ rows]02:21, 77.73 rows/s]
Generated: 5000 rows [02:19, 35.72 rows/s]
Saving: 5000 rows [00:01, 3476.78 rows/s]
Cleanup: 2 tables [00:00,  3.17 tables/s]
Parsed PyTorch dataset for rank=0 worker: 5000 rows [00:11, 423.39 rows/s]
```

Here is `from_hf_download`, including the initial download to the cache:

```python
>>> chain = DataChain.from_hf_download("lmms-lab/COCO-Caption2017", split="val")
Downloading readme: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1.78k/1.78k [00:00<00:00, 9.73MB/s]
Downloading data: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 424M/424M [00:27<00:00, 15.3MB/s]
Downloading data: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 391M/391M [00:24<00:00, 16.3MB/s]
Downloading data: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 476M/476M [00:25<00:00, 18.4MB/s]
[...]

PR #398, opened 2024-09-05
Auto-detect huggingface datasets when reading tabular data

dberenbaum:
This PR makes the `from_parquet("hf://...")` syntax from #375 more consistent with the `from_hf()` syntax from #311. `from_parquet("hf://...")` will auto-detect that the file is from a huggingface dataset and automatically generate special hf model classes for images, class labels, etc.

For example:

```python
>>> chain = DataChain.from_parquet("hf://datasets/lmms-lab/COCO-Caption2017/data/val*")
Processed: 1 rows [00:00, 5349.88 rows/s]
Cleanup: 1 tables [00:00, 8065.97 tables/s]
Listing hf://datasets: 16 objects [00:00, 27.36 objects/s]
Processed: 1 rows [00:00,  1.70 rows/s]0, 27.41 objects/s]
Generated: 16 rows [00:00, 4873.56 rows/s]
Saving: 16 rows [00:00, 16806.63 rows/s]
Cleanup: 1 tables [00:00, 5405.03 tables/s]
>>> images = list(chain.select("image").to_pytorch())
Parsed by pyarrow: 2500 rows [00:42, 58.64 rows/s]
Parsed by pyarrow: 2500 rows [00:36, 69.22 rows/s]
Processed: 2 rows [01:21, 40.63s/ rows].31 rows/s]
Generated: 5000 rows [01:17, 64.42 rows/s]
Saving: 5000 rows [00:02, 2270.16 rows/s]
Cleanup: 1 tables [00:01,  1.70s/ tables]
Parsed PyTorch dataset for rank=0 worker: 5000 rows [00:16, 305.86 rows/s]
```

The images in the above example are automatically converted from bytes objects into to torch tensors like they would be using `from_hf()`. This is also faster to process than `from_hf()` and allows granular reading of individual files, although it does not support all `from_hf()` features like splits or non-tabular datasets.


shcheklein:
> from_parquet("hf://...") will auto-detect that the file is from a huggingface dataset and automatically generate special hf model classes for images, class labels, etc.

@dberenbaum what is the difference with `from_hf` then? should it be doing the same?

should we keep from_parquet the same for all parquet files (no special treatment depending on the source) and keep the higher level / order logic to the methods like `from_hf`?

dberenbaum:
> @dberenbaum what is the difference with `from_hf` then? should it be doing the same?

They are the same in this respect. The difference with `from_hf` is:

> This is also faster to process than `from_hf()` and allows granular reading of individual files, although it does not support all `from_hf()` features like splits or non-tabular datasets.

shcheklein:
@dberenbaum 

> than from_hf() and allows granular reading of individual files,

could you clarify a bit please?

and one more questions bothers me still tbh:

> should we keep from_parquet the same for all parquet files (no special treatment depending on the source) and keep the higher level / order logic to the methods like from_hf?

dberenbaum:
> could you clarify a bit please?

`from_parquet` is faster to process than `from_hf` because `from_hf` applies all the high-level logic from hf `datasets.load_dataset` to read the dataset and convert it to a uniform format. `from_parquet` simply reads the raw parquet files. If you look at the top example in #396, you will see that it takes about 2:22 to process this dataset with `from_hf` vs 1:21 above using `from_parquet`.

Also, you can process a single parquet file from the dataset like `from_parquet("hf://datasets/lmms-lab/COCO-Caption2017/data/val-00000-of-00002.parquet")`. `from_hf` has no concept of individual files, so it must process a whole dataset or defined split from the dataset.

However, `from_parquet` can only read datasets which are already saved as parquet files (hf supports [many dataset formats](https://huggingface.co/docs/datasets/en/create_dataset)) and which require no other processing steps (hf datasets can have [custom scripts](https://huggingface.co/docs/datasets/en/dataset_script) applied as processing steps). Also, `from_hf` supports any additional args that can be passed to [`datasets.load_dataset`](https://huggingface.co/docs/datasets/en/package_reference/loading_methods#datasets.load_dataset), such as passing `from_hf(..., split="train")`. Since `from_parquet` does not call `datasets.load_dataset`, it cannot support such arguments.

Overall, `from_hf` is a more robust way to handle hf datasets but has significant drawbacks (performance at larger scale is particularly painful).

> should we keep from_parquet the same for all parquet files (no special treatment depending on the source) and keep the higher level / order logic to the methods like from_hf?

Of course we can just close this PR, but we lose the advantages mentioned above for `from_parquet`, and I don't see a way we can apply them to `from_hf`. It also means that you will get different results for `from_hf("llms-lab/COCO-Caption2017", split="val")` and `from_parquet("hf://datasets/lmms-lab/COCO-Caption2017/data/val*")`, whereas this PR makes the output of each much closer to one another.

The hf logic applied in this PR does not actually depend on the source -- it depends on the parquet schema having metadata describing a hf dataset. This matters because even a locally saved hf dataset would get parsed the same way as a dataset on hf hub (this is shown in `test_arrow_generator_hf` added here). In this case, we know it's a hf dataset, so I thought it makes sense to treat it like one rather than a generic parquet file, but we can close this if you feel it's too complex/magical. We can also add some flag for the behavior in this PR, but since we already know whether it is a hf dataset, I felt it wasn't necessary.

shcheklein:
I see, thanks @dberenbaum for a good explanation!

> but we can close this if you feel it's too complex/magical. We can also add some flag for the behavior in this PR, but since we already know whether it is a hf dataset, I felt it wasn't necessary.

no strong opinion, but it feels a bit like too much magic for `from_parquet`. If we decide to add a parameter, can be added to `from_hf` itself? (like `loader=hf | parquet ...`) ? wdyt?

or can we make `from_hf` itslf more magical - e.g. if a single parquet file is passed - don't apply high level logic, or if it's parquet format - do what you do in this case?

dberenbaum:
> no strong opinion, but it feels a bit like too much magic for `from_parquet`. If we decide to add a parameter, can be added to `from_hf` itself? (like `loader=hf | parquet ...`) ? wdyt?
> 
> or can we make `from_hf` itslf more magical - e.g. if a single parquet file is passed - don't apply high level logic, or if it's parquet format - do what you do in this case?

This doesn't make sense to me since the interface and args to each are completely different. `from_parquet` takes a path and optional args to pass to the arrow dataset loader (like file format and partitioning schema). `from_hf` takes a hf dataset name and optional args to pass to the hf dataset loader (like split name and revision). Also, we need to support not only parquet but also csv and other tabular file formats. This all would be much more complicated and magical if we do it inside `from_hf` IMHO.

I mentioned above that another option is to add a flag in `from_parquet` and related methods that would enable the hf parsing, like `from_parquet("hf://...", hf=True)`. I opted against it since it feels a bit redundant and unnecessary when we can already tell whether this is a hf dataset, but not that strong an opinion if you prefer it not being automatic. Would that help?

shcheklein:
> I opted against it since it feels a bit redundant and unnecessary when we can already tell whether this is a hf dataset, but not that strong an opinion if you prefer it not being automatic. Would that help?

I'm fine with that. Just wanted to explore a bit more options since it indeed feels like a magic to me. Anyways, I think it's a good options considering other alternatives.

shcheklein (approved):

PR #399, opened 2024-09-05
pin pydantic < 2.9

mattseddon:
2.9.0 has breaking changes that are beyond me solving in the next 20 minutes

Failures: 

- https://github.com/iterative/datachain/actions/runs/10726012473
- https://github.com/iterative/datachain/actions/runs/10724648636

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,7 +40,7 @@ dependencies = [
   "dill==0.3.8",
   "cloudpickle",
   "orjson>=3.10.5",
-  "pydantic>=2,<3",
+  "pydantic>2,<2.9",
   "jmespath>=1.0",
   "datamodel-code-generator>=0.25",
   "Pillow>=10.0.0,<11",


pyproject.toml
```diff
@@ -40,7 +40,7 @@ dependencies = [
   "dill==0.3.8",
   "cloudpickle",
   "orjson>=3.10.5",
-  "pydantic>=2,<3",
+  "pydantic==2.8.2",
```
dmpetrov, line 43: Shouldn't it be '>2,<2.9'?

dmpetrov (approved):
LG
See a comment

pyproject.toml
mattseddon: I don't understand why we are so laissez-faire with dependency versions

rlamy: Exact pins are good for apps, but terrible for libraries.

PR #400, opened 2024-09-06
unpin pydantic, use python API for datamodel_codegen

skshetry:
Latest update of pydantic broke `read_schema()`, which fails to correctly resolve annotations.

Please read more here: https://github.com/pydantic/pydantic/pull/10113#issuecomment-2334100659.

### DIFF
--- a/src/datachain/lib/meta_formats.py
+++ b/src/datachain/lib/meta_formats.py
@@ -2,14 +2,14 @@
 # pip install jmespath
 #
 import csv
-import io
 import json
-import subprocess
-import sys
+import tempfile
 import uuid
 from collections.abc import Iterator
+from pathlib import Path
 from typing import Any, Callable
 
+import datamodel_code_generator
 import jmespath as jsp
 from pydantic import BaseModel, ConfigDict, Field, ValidationError  # noqa: F401
 
@@ -47,9 +47,8 @@ def read_schema(source_file, data_type="csv", expr=None, model_name=None):
     data_string = ""
     # using uiid to get around issue #1617
     if not model_name:
-        uid_str = str(generate_uuid()).replace(
-            "-", ""
-        )  # comply with Python class names
+        # comply with Python class names
+        uid_str = str(generate_uuid()).replace("-", "")
         model_name = f"Model{data_type}{uid_str}"
     try:
         with source_file.open() as fd:  # CSV can be larger than memory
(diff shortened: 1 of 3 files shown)


src/datachain/lib/meta_formats.py
```diff
@@ -115,26 +108,17 @@ def read_meta(  # noqa: C901
 
     # ugly hack: datachain is run redirecting printed outputs to a variable
```
skshetry, line 109:
```suggestion
```

src/datachain/lib/meta_formats.py
```diff
@@ -70,33 +69,27 @@ def read_schema(source_file, data_type="csv", expr=None, model_name=None):
...
-            input=data_string,
-            text=True,
-            capture_output=True,
-            check=True,
+
+    input_file_types = {i.value: i for i in datamodel_code_generator.InputFileType}
+    input_file_type = input_file_types[data_type]
+    with tempfile.NamedTemporaryFile() as tmpfile:
+        output = Path(tmpfile.name)
+        datamodel_code_generator.generate(
```
skshetry, line 77: Ideally, we need a `jsonschema_to_pydantic` like APIs instead of using datamodel_codegen.

src/datachain/lib/meta_formats.py
```diff
@@ -70,33 +69,27 @@ def read_schema(source_file, data_type="csv", expr=None, model_name=None):
...
+
+    input_file_types = {i.value: i for i in datamodel_code_generator.InputFileType}
+    input_file_type = input_file_types[data_type]
+    with tempfile.TemporaryDirectory() as tmpdir:
+        output = Path(tmpdir) / "model.py"
+        datamodel_code_generator.generate(
+            data_string,
+            input_file_type=input_file_type,
+            output=output,
+            target_python_version=datamodel_code_generator.PythonVersion.PY_39,
```
skshetry, line 81: This should probably be dependent on `sys.version_info[1]`, but previous implementation was left to default which is `PY_37`.

src/datachain/lib/meta_formats.py
```diff
@@ -70,33 +69,27 @@ def read_schema(source_file, data_type="csv", expr=None, model_name=None):
...
+            base_class="datachain.lib.meta_formats.UserModel",
+            class_name=model_name,
+            additional_imports=["datachain.lib.data_model.DataModel"],
+            use_standard_collections=True,
```
skshetry, line 85: This makes datamodel-codegen to use standard collection types, instead of importing from `typing` module which are deprecated.

rlamy (approved):
LGTM. My only concern is that we're apparently not testing that code.

skshetry:
> LGTM. My only concern is that we're apparently not testing that code.

I have added tests for `json` formats.

PR #403, opened 2024-09-07
remove pytest_asyncio lib

EdwardLi-coder:
When I use Python 3.12, I receive many warnings about the following issue:
```
/Users/lifei/PycharmProjects/datachain/.nox/tests-3-12/lib/python3.12/site-packages/pytest_asyncio/plugin.py:208: PytestDeprecationWarning: The configuration option "asyncio_default_fixture_loop_scope" is unset.
The event loop scope for asynchronous fixtures will default to the fixture caching scope. Future versions of pytest-asyncio will default the loop scope for asynchronous fixtures to function scope. Set the default fixture loop scope explicitly in order to avoid unexpected behavior in the future. Valid fixture loop scopes are: "function", "class", "module", "package", "session"

  warnings.warn(PytestDeprecationWarning(_DEFAULT_FIXTURE_LOOP_SCOPE_UNSET))
```
The reason is that the asyncio_default_fixture_loop_scope needs to be explicitly defined. I have added it in this PR.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -127,6 +127,7 @@ namespaces = false
 
 [tool.pytest.ini_options]
 addopts = "-rfEs -m 'not benchmark and not examples'"
+asyncio_default_fixture_loop_scope = "function"
 markers = [
   "benchmark: benchmarks.",
   "e2e: End-to-end tests",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -82,7 +82,6 @@ tests = [
   "pytest-mock>=3.12.0",
   "pytest-servers[all]>=0.5.5",
   "pytest-benchmark[histogram]",
-  "pytest-asyncio>=0.23.2",
   "pytest-xdist>=3.3.1",
   "virtualenv",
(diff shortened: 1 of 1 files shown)


skshetry:
I don't think we have any async tests. So, most likely, we can get rid of it from `tests` dependency and remove this config.

https://github.com/iterative/datachain/blob/ae493e776e29d2b735f768ecd695416f73e726f9/pyproject.toml#L139

(cc @rlamy).

EdwardLi-coder:
> I don't think we have any async tests. So, most likely, we can get rid of it from `tests` dependency and remove this config.
> 
> https://github.com/iterative/datachain/blob/ae493e776e29d2b735f768ecd695416f73e726f9/pyproject.toml#L139
> 
> (cc @rlamy).

I agree with this.And I will update it.Thank you.

shcheklein (approved):
thanks !

skshetry (approved):
Please adjust the title. Otherwise lgtm. Thank you.

PR #404, opened 2024-09-07
remove some warnings

EdwardLi-coder:
When I run the tests, I often encounter warnings related to library dependencies. In my PR, I've ignored warnings coming from third-party libs. And if they come from our own code, I'm rewriting it.
These warnings are need to rewrite code.
```
tests/unit/lib/test_text.py: 1 warning
  /Users/lifei/PycharmProjects/datachain/.nox/tests-3-12/lib/python3.12/site-packages/datachain/lib/text.py:36: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).
    tokens = torch.tensor(tokens)

tests/unit/lib/test_clip.py::test_similarity_scores_hf
tests/unit/lib/test_image.py::test_convert_image_hf
  /Users/lifei/PycharmProjects/datachain/.nox/tests-3-12/lib/python3.12/site-packages/datachain/lib/image.py:37: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).
    img = torch.tensor(img.pixel_values[0])  # type: ignore[assignment,attr-defined]

tests/unit/lib/test_clip.py::test_similarity_scores_hf
  /Users/lifei/PycharmProjects/datachain/.nox/tests-3-12/lib/python3.12/site-packages/datachain/lib/clip.py:21: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).
    return lambda x: method(torch.tensor(x))
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -141,6 +141,10 @@ filterwarnings = [
   "error::pytest_mock.PytestMockWarning",
   "error::pytest.PytestCollectionWarning",
   "error::sqlalchemy.exc.SADeprecationWarning",
+  "ignore::DeprecationWarning:timm.*",
+  "ignore::DeprecationWarning:botocore.auth",
+  "ignore::DeprecationWarning:datasets.utils._dill",
+  "ignore::DeprecationWarning:librosa.core.intervals",
   "ignore:Field name .* shadows an attribute in parent:UserWarning"  # datachain.lib.feature
 ]
 
--- a/src/datachain/lib/clip.py
+++ b/src/datachain/lib/clip.py
@@ -18,7 +18,7 @@ def _get_encoder(model: Any, type: Literal["image", "text"]) -> Callable:
         hasattr(model, method_name) and inspect.ismethod(getattr(model, method_name))
     ):
         method = getattr(model, method_name)
-        return lambda x: method(torch.tensor(x))
+        return lambda x: method(torch.as_tensor(x).clone().detach())
 
     # Check for model from clip or open_clip library
     method_name = f"encode_{type}"
--- a/src/datachain/lib/image.py
+++ b/src/datachain/lib/image.py
@@ -34,7 +34,7 @@ def convert_image(
             from transformers.image_processing_utils import BaseImageProcessor
 
             if isinstance(transform, BaseImageProcessor):
-                img = torch.tensor(img.pixel_values[0])  # type: ignore[assignment,attr-defined]
+                img = torch.as_tensor(img.pixel_values[0]).clone().detach()  # type: ignore[assignment,attr-defined]
         except ImportError:
             pass
         if device:
--- a/src/datachain/lib/text.py
+++ b/src/datachain/lib/text.py
@@ -33,7 +33,7 @@ def convert_text(
         res = tokenizer(text)
 
     tokens = res.input_ids if isinstance(tokenizer, PreTrainedTokenizerBase) else res
-    tokens = torch.tensor(tokens)
+    tokens = torch.as_tensor(tokens).clone().detach()
     if device:
         tokens = tokens.to(device)
 
--- a/tests/func/test_pytorch.py
+++ b/tests/func/test_pytorch.py
@@ -1,5 +1,6 @@
 import open_clip
 import pytest
+import torch
 from datasets import load_dataset
 from torch import Size, Tensor
 from torchvision.datasets import FakeData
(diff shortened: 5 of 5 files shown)


shcheklein:
can we ignore warning that come from third-party libs (e.g. some lib that we depend on is using torch in a deprecated way) and fix warnings on our end by rewriting the code, wdyt?

EdwardLi-coder:
> can we ignore warning that come from third-party libs (e.g. some lib that we depend on is using torch in a deprecated way) and fix warnings on our end by rewriting the code, wdyt?

You can review my code. I have already done this.

shcheklein:
> In my PR, I've ignored warnings coming from third-party libs. And if they come from our own code, I'm rewriting it.

will it ignore new warnings (let's say someone adds it by mistake tomorrow) in our code? just to double check ...

EdwardLi-coder:
> > In my PR, I've ignored warnings coming from third-party libs. And if they come from our own code, I'm rewriting it.
> 
> will it ignore new warnings (let's say someone adds it by mistake tomorrow) in our code? just to double check ...

These ignore settings are specific to warnings from certain third-party libraries. They will not affect new warnings in our own code. Each ignore has a specific module path, so even if someone accidentally adds new warnings in our code in the future, these settings will not ignore them. New warnings in our code will still be captured and displayed.
However, to ensure we don't miss important warnings, I suggest we periodically review these ignore settings and re-evaluate their necessity after updates to the third-party libraries.

pyproject.toml
```diff
@@ -141,6 +141,10 @@ filterwarnings = [
   "error::pytest_mock.PytestMockWarning",
   "error::pytest.PytestCollectionWarning",
   "error::sqlalchemy.exc.SADeprecationWarning",
+  "ignore::DeprecationWarning:timm.*",
+  "ignore::DeprecationWarning:botocore.auth",
+  "ignore::DeprecationWarning:datasets.utils._dill",
```
EdwardLi-coder, line 146: I think maybe only ignore DeprecationWarning is good.

src/datachain/lib/clip.py
```diff
@@ -18,7 +18,7 @@ def _get_encoder(model: Any, type: Literal["image", "text"]) -> Callable:
         hasattr(model, method_name) and inspect.ismethod(getattr(model, method_name))
     ):
         method = getattr(model, method_name)
-        return lambda x: method(torch.tensor(x))
+        return lambda x: method(torch.as_tensor(x).clone().detach())
```
EdwardLi-coder, line 21: This is the warning:
```
tests/unit/lib/test_clip.py::test_similarity_scores_hf
  /Users/lifei/PycharmProjects/datachain/.nox/tests-3-12/lib/python3.12/site-packages/datachain/lib/clip.py:21: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).
    return lambda x: method(torch.tensor(x))
```
Update Reason:
We use torch.as_tensor() to create tensors, which avoids copying data whenever possible. We then call clone().detach() to create a new tensor that is detached from the original data and will not affect the computation graph of the original data.

src/datachain/lib/image.py
```diff
@@ -34,7 +34,7 @@ def convert_image(
             from transformers.image_processing_utils import BaseImageProcessor
 
             if isinstance(transform, BaseImageProcessor):
-                img = torch.tensor(img.pixel_values[0])  # type: ignore[assignment,attr-defined]
+                img = torch.as_tensor(img.pixel_values[0]).clone().detach()  # type: ignore[assignment,attr-defined]
```
EdwardLi-coder, line 37: This is the warning:
```
tests/unit/lib/test_clip.py::test_similarity_scores_hf
tests/unit/lib/test_image.py::test_convert_image_hf
  /Users/lifei/PycharmProjects/datachain/.nox/tests-3-12/lib/python3.12/site-packages/datachain/lib/image.py:37: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).
    img = torch.tensor(img.pixel_values[0])  # type: ignore[assignment,attr-defined]
```
Update Reason:
We use torch.as_tensor() to create tensors, which avoids copying data whenever possible. We then call clone().detach() to create a new tensor that is detached from the original data and will not affect the computation graph of the original data.

EdwardLi-coder:
@shcheklein I only remain ignore the DeprecationWarning. I have provided some explanations for my modifications. If you have any other suggestions, I would be happy to make further changes.

dreadatour (commented):
Looks good to me! 👍

shcheklein (approved):

shcheklein:
thanks @EdwardLi-coder !

PR #405, opened 2024-09-09
increase timeout of test for windows

mattseddon:
Another failure here: https://github.com/iterative/datachain/actions/runs/10750425347/job/29816550855?pr=403

There is some more investigation and discussion in https://github.com/iterative/datachain/pull/406

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -50,7 +50,7 @@ jobs:
         run: nox -s lint
 
   datachain:
-    timeout-minutes: 40
+    timeout-minutes: 60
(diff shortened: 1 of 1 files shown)


skshetry:
Given that the 3.9 job finishes in ~19mins, the other job is most likely frozen, and we should investigate instead of increasing timeout.

mattseddon:
I think it is just luck of the draw with the windows runner that we pull. Some perform like dog shit and take 4x longer than a normal runner. The job was 92% of the way through.

skshetry:
It could be E2E tests too, that are frozen. We can add `pytest-timeout` and see if they fail.

PR #406, opened 2024-09-09
add timeout to e2e tests

mattseddon:
From https://github.com/iterative/datachain/pull/405#issuecomment-2336971826


skshetry:
I'm investigating here: https://github.com/iterative/datachain/actions/runs/10770282498.

I will merge this if E2E is the problem.

skshetry:
It looks like the issue is with parallel udf tests. Although I am not sure, but adding `--timeout` to the Windows tests started crashing `xdist` workers and it says it was running `test_parallel` (or, similar other tests).

Need to check if the crashes happen on `main` or not.

mattseddon:
> It looks like the issue is with parallel udf tests. Although I am not sure, but adding `--timeout` to the Windows tests started crashing `xdist` workers and it says it was running `test_parallel` (or, similar other tests).

From the docs: "For each test item the pytest-timeout plugin starts a timer thread which will terminate the whole process after the specified timeout. When a test item finishes this timer thread is cancelled and the test run continues." so these tests are timing out. Adding `timeout_func_only = true` to the config seems to improve things but there are still random timeouts in the suite. 

I am still seeing runs > 35 minutes long ([here](https://github.com/iterative/datachain/actions/runs/10784225443)) even with tests getting killed after the timeout. Given this and the fact that the original run in question from https://github.com/iterative/datachain/pull/405 was at 92% completed with 4 workers, I do not think it is unfeasible that the entire suite is just running slowly and needs > 40 minutes to complete.

skshetry:
> I do not think it is unfeasible that the entire suite is just running slowly and needs > 40 minutes to complete.

Check the timestamp on https://github.com/iterative/datachain/actions/runs/10750425347/job/29816550855?pr=403.

The second-to-last line is at 10-minute mark and the last line appears 30 minutes later when the job times out.

PR #407, opened 2024-09-09
disable execution of last query expression by default

### LINKED ISSUE #360: Remove support for executing last expression in the script

skshetry:
This PR hides the support for execution of last query expression behind a flag, which I plan to remove when Studio is updated.

Also, few other changes have been made to the `query` API:

1. `envs=` keyword argument has been renamed to `env`, similar to `subprocess.Popen(env=)`.
2. `python_executable` default has been changed from `None` to `sys.executable`. And it no longer accepts `None`.
3. The `query` API no longer returns `QueryResult` API. The responsibility is now with the caller, to find out latest dataset version. They are in much better place to do that, since they are the one responsible for creating a `job`.
4. On `capture_output=True`, `query` API no longer prints to the `stdout`. The `output_hook` is responsible to do so now.
5. Exceptions raised on `query` no longer have `output` set. 
6. `QueryScriptDatasetNotFound` has been removed.
7. We no longer set `script_output` and `query_script` to the `DatasetVersion`. This was not used anyway. 

Closes #360.

### DIFF
--- a/src/datachain/catalog/catalog.py
+++ b/src/datachain/catalog/catalog.py
@@ -12,7 +12,6 @@ import sys
 import time
 import traceback
 from collections.abc import Iterable, Iterator, Mapping, Sequence
-from contextlib import contextmanager, nullcontext
 from copy import copy
 from dataclasses import dataclass
 from functools import cached_property, reduce
@@ -58,7 +57,6 @@ from datachain.error import (
     PendingIndexingError,
     QueryScriptCancelError,
     QueryScriptCompileError,
-    QueryScriptDatasetNotFound,
     QueryScriptRunError,
 )
 from datachain.listing import Listing
@@ -115,38 +113,19 @@ def noop(_: str):
     pass
 
 
-@contextmanager
-def print_and_capture(
-    stream: "IO[bytes]|IO[str]", callback: Callable[[str], None] = noop
-) -> "Iterator[list[str]]":
-    lines: list[str] = []
-    append = lines.append
+def _process_stream(stream: "IO[bytes]", callback: Callable[[str], None]) -> None:
+    buffer = b""
+    while byt := stream.read(1):  # Read one byte at a time
+        buffer += byt
 
-    def loop() -> None:
-        buffer = b""
-        while byt := stream.read(1):  # Read one byte at a time
-            buffer += byt.encode("utf-8") if isinstance(byt, str) else byt
-
-            if byt in (b"\n", b"\r"):  # Check for newline or carriage return
-                line = buffer.decode("utf-8")
-                print(line, end="")
-                callback(line)
-                append(line)
-                buffer = b""  # Clear buffer for next line
-
-        if buffer:  # Handle any remaining data in the buffer
+        if byt in (b"\n", b"\r"):  # Check for newline or carriage return
             line = buffer.decode("utf-8")
-            print(line, end="")
             callback(line)
-            append(line)
+            buffer = b""  # Clear buffer for next line
 
-    thread = Thread(target=loop, daemon=True)
-    thread.start()
-
-    try:
-        yield lines
-    finally:
-        thread.join()
+    if buffer:  # Handle any remaining data in the buffer
+        line = buffer.decode("utf-8")
+        callback(line)
 
 
 class QueryResult(NamedTuple):
@@ -650,11 +629,6 @@ class Catalog:
                 ]
                 code_ast.body[-1:] = new_expressions
         return code_ast
-
-    def compile_query_script(self, script: str) -> str:
-        code_ast = ast.parse(script)
-        code_ast = self.attach_query_wrapper(code_ast)
-        return ast.unparse(code_ast)
 
     def parse_url(self, uri: str, **config: Any) -> tuple[Client, str]:
         config = config or self.client_config
@@ -1806,13 +1780,13 @@ class Catalog:
         self,
         query_script: str,
         envs: Optional[Mapping[str, str]] = None,
-        python_executable: Optional[str] = None,
+        python_executable: str = sys.executable,
         save: bool = False,
         capture_output: bool = True,
         output_hook: Callable[[str], None] = noop,
         params: Optional[dict[str, str]] = None,
         job_id: Optional[str] = None,
-    ) -> QueryResult:
+    ) -> None:
         """
         Method to run custom user Python script to run a query and, as result,
         creates new dataset from the results of a query.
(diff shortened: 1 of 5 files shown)


src/datachain/catalog/catalog.py
```diff
@@ -115,44 +112,19 @@ def noop(_: str):
...
-
+            buffer = b""  # Clear buffer for next line
 
-class QueryResult(NamedTuple):
-    dataset: Optional[DatasetRecord]
-    version: Optional[int]
-    output: str
+    if buffer:  # Handle any remaining data in the buffer
+        line = buffer.decode("utf-8")
+        callback(line)
```
skshetry, line 127: We'll no longer print to the stdout when `capture_output=True`.

src/datachain/catalog/catalog.py
```diff
@@ -1805,14 +1772,15 @@ def apply_udf(
     def query(
         self,
         query_script: str,
-        envs: Optional[Mapping[str, str]] = None,
-        python_executable: Optional[str] = None,
+        env: Optional[Mapping[str, str]] = None,
```
skshetry, line 1775: Renamed to `env`. I don't think it should be plural.

src/datachain/catalog/catalog.py
```diff
@@ -1805,14 +1772,15 @@ def apply_udf(
...
+        python_executable: str = sys.executable,
```
skshetry, line 1776: Changed `python_executable` to default to `sys.executable` and not take a `None` value.

src/datachain/catalog/catalog.py
```diff
@@ -1805,14 +1772,15 @@ def apply_udf(
...
         save: bool = False,
         capture_output: bool = True,
         output_hook: Callable[[str], None] = noop,
         params: Optional[dict[str, str]] = None,
         job_id: Optional[str] = None,
-    ) -> QueryResult:
+        _execute_last_expression: bool = False,
```
skshetry, line 1782: I plan to get rid of this when we update Studio.

src/datachain/error.py
```diff
@@ -42,10 +42,6 @@ def __init__(self, message: str, return_code: int = 0, output: str = ""):
         super().__init__(self.message)
 
 
-class QueryScriptDatasetNotFound(QueryScriptRunError):  # noqa: N818
```
skshetry, line 45: No longer used in `query`. I will create a similar exception in `Studio` side.

src/datachain/catalog/catalog.py
```diff
@@ -1835,92 +1803,21 @@ def query(
...
-            query_script=query_script,
-        )
-        self.update_dataset_version_with_warehouse_info(
-            dr,
-            dv.version,
-            script_output=output,
-            query_script=query_script,
-            job_id=job_id,
-            is_job_result=True,
-        )
```
skshetry, line 1902: This was not being used anywhere. `Job` replaces this.

src/datachain/catalog/catalog.py
```diff
@@ -1835,92 +1803,21 @@ def query(
...
-                _get_dataset_versions_by_job_id(), key=lambda x: x[1].created_at
-            )
-        except ValueError as e:
-            if not save:
-                return QueryResult(dataset=None, version=None, output=output)
-
-            raise QueryScriptDatasetNotFound(
-                "No dataset found after running Query script",
-                output=output,
-            ) from e
```
skshetry, line 1888: This will have to be done on the caller side. And eventually removed when we drop `_execute_last_expression` support.

src/datachain/catalog/catalog.py
```diff
@@ -1835,92 +1803,21 @@ def query(
...
+                code_ast = ast.parse(query_script)
+                code_ast = self.attach_query_wrapper(code_ast)
+                query_script_compiled = ast.unparse(code_ast)
+            except Exception as exc:
+                raise QueryScriptCompileError(
+                    f"Query script failed to compile, reason: {exc}"
+                ) from exc
+        else:
+            query_script_compiled = query_script
+            assert not save
```
amritghimire, line 1817: Reminder to remove this flag as well.

amritghimire (approved):
Looks good to me.

src/datachain/catalog/catalog.py
```diff
@@ -1835,92 +1803,21 @@ def query(
...
-    ) -> tuple[list[str], subprocess.Popen]:
-        try:
-            query_script_compiled = self.compile_query_script(query_script)
-        except Exception as exc:
-            raise QueryScriptCompileError(
-                f"Query script failed to compile, reason: {exc}"
-            ) from exc
-        envs = dict(envs or os.environ)
-        envs.update(
+        env = dict(env or os.environ)
```
dreadatour, line 1819: Hm, not sure why `or`? May be something like?

```suggestion
        env = {**os.environ, **(env or {})}
```

dreadatour (approved):
Code looks good to me, thank you for removing all this outdated code 🙏

src/datachain/catalog/catalog.py
skshetry: How would you provide a way to override envvars of the current process? 

This is how `subprocess.Popen` works, and given this is a thin wrapper around it, I think it's better to mimic it's API. 

Also, Studio already provides copy of all envvars.

src/datachain/catalog/catalog.py
```diff
@@ -115,44 +112,19 @@ def noop(_: str):
...
-@contextmanager
-def print_and_capture(
-    stream: "IO[bytes]|IO[str]", callback: Callable[[str], None] = noop
-) -> "Iterator[list[str]]":
-    lines: list[str] = []
-    append = lines.append
+def _process_stream(stream: "IO[bytes]", callback: Callable[[str], None]) -> None:
+    buffer = b""
+    while byt := stream.read(1):  # Read one byte at a time
+        buffer += byt
```
mattseddon, line 118: Seeing some errors coming through the test suite of the form: `TypeError: can't concat str to bytes`. 

Examples are here: https://github.com/iterative/datachain/actions/runs/10821775005/job/30024466321?pr=427#step:7:177

skshetry: Fixed tests in #431.

src/datachain/catalog/catalog.py
```diff
@@ -115,44 +112,19 @@ def noop(_: str):
...
 
-    def loop() -> None:
-        buffer = b""
-        while byt := stream.read(1):  # Read one byte at a time
-            buffer += byt.encode("utf-8") if isinstance(byt, str) else byt
```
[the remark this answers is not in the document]
skshetry: Fixed in #431.

PR #408, opened 2024-09-09
chore: support light mode, dark mode logo

yathomasi:
- Minor fix for dark and light mode in readme
- It looks like there is no proper method to fix this with RST
- Added css to svg to support the mode

Screenshot 2024-09-09 at 19 54 55

Screenshot 2024-09-09 at 19 54 58


skshetry (approved):

yathomasi:
I see there is issue with Safari and `prefers-color-scheme` for `SVG` files.

One option is to use the normal text for `Datachain` but it will be on new line below logo. Let me know if it's okay.

Screenshot 2024-09-09 at 22 28 05

skshetry:
> One option is to use the normal text for `Datachain` but it will be on new line below logo. Let me know if it's okay.

@yathomasi, this might be a better solution. https://github.blog/changelog/2021-11-24-specify-theme-context-for-images-in-markdown/

skshetry:
There is also support for showing images based on theme (see https://github.blog/changelog/2022-05-19-specify-theme-context-for-images-in-markdown-beta/), but for that we'll have to change README to markdown from rst.

yathomasi:
Thanks @djsauble @skshetry for your inputs.

I am sorry to mention that I had already went through all of those before suggesting the solution. Let's go with the suggestion mentioned if we want to support Safari.

yathomasi:
Thanks @djsauble. That's great. Merged your change.

PR #409, opened 2024-09-09
Move 'join' SQL implementation to warehouse

dreadatour:
Fix for the https://github.com/iterative/studio/issues/10602

In this PR:

- do not generate multiple `join` SQL queries as it could be quite heavy for some warehouse implementations (e.g. ClickHouse, see related PR), if there is `join` (or `union`) found in chain, save results in temporary table and run chain from there. See implementation [here](https://github.com/iterative/datachain/compare/sql-merge-refactoring?expand=1#diff-5cd0b1bd1147a2704c8c333fdb16b0fa21c66e6f5101e4889f7a624d5da4c7cfR916-R935).
- move `join` implementation to warehouse to be able to tweak `join` (e.g. in ClickHouse, see [related PR](https://github.com/iterative/studio/pull/10634))
- [remove](https://github.com/iterative/datachain/compare/sql-merge-refactoring?expand=1#diff-5cd0b1bd1147a2704c8c333fdb16b0fa21c66e6f5101e4889f7a624d5da4c7cfL1694) `tqdm` from `save` method as it was added in previous PR as an experimental and not really needed here. Questionable.

### DIFF
--- a/src/datachain/data_storage/sqlite.py
+++ b/src/datachain/data_storage/sqlite.py
@@ -40,7 +40,9 @@ if TYPE_CHECKING:
     from sqlalchemy.dialects.sqlite import Insert
     from sqlalchemy.engine.base import Engine
     from sqlalchemy.schema import SchemaItem
+    from sqlalchemy.sql._typing import _FromClauseArgument, _OnClauseArgument
     from sqlalchemy.sql.elements import ColumnElement
+    from sqlalchemy.sql.selectable import Join
     from sqlalchemy.types import TypeEngine
 
     from datachain.lib.file import File
@@ -787,6 +789,23 @@ class SQLiteWarehouse(AbstractWarehouse):
 
             if progress_cb:
                 progress_cb(len(batch_ids))
+
+    def join(
+        self,
+        left: "_FromClauseArgument",
+        right: "_FromClauseArgument",
+        onclause: "_OnClauseArgument",
+        inner: bool = True,
+    ) -> "Join":
+        """
+        Join two tables together.
+        """
+        return sqlalchemy.join(
+            left,
+            right,
+            onclause,
+            isouter=not inner,
+        )
 
     def create_pre_udf_table(self, query: "Select") -> "Table":
         """
--- a/src/datachain/data_storage/warehouse.py
+++ b/src/datachain/data_storage/warehouse.py
@@ -32,7 +32,6 @@ if TYPE_CHECKING:
         _FromClauseArgument,
         _OnClauseArgument,
     )
-    from sqlalchemy.sql.elements import ColumnElement
     from sqlalchemy.sql.selectable import Join, Select
     from sqlalchemy.types import TypeEngine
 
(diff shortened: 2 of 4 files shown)


shcheklein:
> do not generate multiple join SQL queries as it could be quite heavy for some warehouse implementations (e.g. ClickHouse, see related PR), if there is join (or union) found in chain, save results in temporary table and run chain from there. See implementation [here](https://github.com/iterative/datachain/compare/sql-merge-refactoring?expand=1#diff-5cd0b1bd1147a2704c8c333fdb16b0fa21c66e6f5101e4889f7a624d5da4c7cfR916-R935).

can we avoid this for now if we use partial join and such? (I'm worried that let's say next time we add BigQuery or something - we are replacing optimizer with our own engine - kinda replicating what DB should be doing) ? it's also not needed for SQLite, right?

dreadatour:
> can we avoid this for now if we use partial join and such? (I'm worried that let's say next time we add BigQuery or something - we are replacing optimizer with our own engine - kinda replicating what DB should be doing) ?

Good point! Let me update the PR.

> it's also not needed for SQLite, right?

Yes... I was thinking how to make it warehouse-dependent but can not find a good way to do this easily. Let me think a bit more.

shcheklein (approved):

dreadatour:
> can we avoid this for now if we use partial join and such?

> it's also not needed for SQLite, right?

Did some tests in SQLite.

Python script:

```python
from datachain import DataChain, Column, ModelStore
from datachain.lib.webdataset_laion import WDSLaion, LaionMeta

ModelStore.register(LaionMeta)
ModelStore.register(WDSLaion)

images = DataChain.from_dataset("laion_wds")
meta_pq = DataChain.from_dataset("laion_pq")
meta_emd = DataChain.from_dataset("laion_emb")

images_with_meta = images.merge(
    meta_pq,
    on="laion.json.uid",
    right_on="uid",
    inner=True
)

result = images_with_meta.merge(
    meta_emd,
    on=["stem", "source.index"],
    right_on=["stem", "emd.index"],
    inner=True
).save("laion")
```

Existing datasets:

```
$ datachain dataset-stats laion_wds
Number of objects: 447567
Total objects size:   13.4G
$ datachain dataset-stats laion_pq
Number of objects: 2
Total objects size:  230.5M
$ datachain dataset-stats laion_emb
Number of objects: 1009874
Total objects size:    2.8P
$
```

Execution on `main` branch (without changes in this PR), one merge on three tables:

```
$ time python wds_merge.py
python wds_merge.py  53.56s user 79.61s system 39% cpu 5:34.64 total
$
```

Execution in current branch, two merges:

```
$ time python wds_merge.py
python wds_merge.py  7.78s user 41.17s system 36% cpu 2:12.94 total
$
```

It is 2,5 times faster than one merge on three tables in SQLite.

> I'm worried that let's say next time we add BigQuery or something - we are replacing optimizer with our own engine - kinda replicating what DB should be doing

This is my main concern right now. Still looking for a proper solution to implement multiple joins separation based on warehouse type.

Also wanted to mention changes in [Studio PR](https://github.com/iterative/studio/pull/10634) with join tables reorder will not works without this fix with separating joins 👀

PR #410, opened 2024-09-09
read_schema: avoid passing locals to exec while executing pydantic models

skshetry:
See https://github.com/pydantic/pydantic/pull/10113#issuecomment-2337337368.

### DIFF
--- a/src/datachain/lib/meta_formats.py
+++ b/src/datachain/lib/meta_formats.py
@@ -1,13 +1,10 @@
-# pip install datamodel-code-generator
-# pip install jmespath
-#
 import csv
 import json
 import tempfile
 import uuid
 from collections.abc import Iterator
 from pathlib import Path
-from typing import Any, Callable
+from typing import Callable
 
 import datamodel_code_generator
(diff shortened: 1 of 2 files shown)


src/datachain/lib/meta_formats.py
```diff
@@ -122,9 +118,9 @@ def read_meta(  # noqa: C901
             print(f"{model_output}")
         # Below 'spec' should be a dynamically converted DataModel from Pydantic
         if not spec:
-            local_vars: dict[str, Any] = {}
-            exec(model_output, globals(), local_vars)  # type: ignore[arg-type] # noqa: S102
-            spec = local_vars["spec"]
+            gl = globals()
+            exec(model_output, gl)  # noqa: S102
```
skshetry, line 122: We could avoid using `exec` and use `importlib` or `runpy.run_path` instead.
But `importlib` needs to update `sys.modules`, and `run_path` is not thread-safe, and makes a lot of modifications to `sys`.

rlamy (approved):

PR #411, opened 2024-09-09
Introduce telemetry in datachain

### LINKED ISSUE #344: DataChain Telemetry needed
### Description

We need to add DataChain telemetry to the roadmap to begin tracking usage since it is open now! 😄 🚀 🎉

amritghimire:
Telemetry is added to datachain based on [iterative-telemetry](https://pypi.org/project/iterative-telemetry/) .

The example values sent as payloads are:
```py
2024-09-09 19:44:33 - iterative_telemetry - DEBUG - Sending payload {'interface': 'api', 'action': 'from_storage', 'error': None, 'extra': {'client': 'gs://'}, 'python_version': {'major': 3, 'minor': 12, 'patch': 4}, 'tool_name': 'datachain', 'tool_version': '0.3.10.dev12+g944029b.d20240905', 'user_id': 'cb4bc63c-f9f7-46ea-8ff3-e3201fbabeb7', 'group_id': None, 'os_name': 'mac', 'os_version': '15.1'}
2024-09-09 19:44:33 - datachain.telemetry - DEBUG - Telemetry is enabled.
2024-09-09 19:44:33 - iterative_telemetry - DEBUG - Sending payload {'interface': 'api', 'action': 'map', 'error': None, 'extra': {}, 'python_version': {'major': 3, 'minor': 12, 'patch': 4}, 'tool_name': 'datachain', 'tool_version': '0.3.10.dev12+g944029b.d20240905', 'user_id': 'cb4bc63c-f9f7-46ea-8ff3-e3201fbabeb7', 'group_id': None, 'os_name': 'mac', 'os_version': '15.1'}
2024-09-09 19:44:33 - datachain.telemetry - DEBUG - Telemetry is enabled.
2024-09-09 19:44:33 - iterative_telemetry - DEBUG - Sending payload {'interface': 'api', 'action': 'show', 'error': None, 'extra': {}, 'python_version': {'major': 3, 'minor': 12, 'patch': 4}, 'tool_name': 'datachain', 'tool_version': '0.3.10.dev12+g944029b.d20240905', 'user_id': 'cb4bc63c-f9f7-46ea-8ff3-e3201fbabeb7', 'group_id': None, 'os_name': 'mac', 'os_version': '15.1'}

```

```py
2024-09-09 20:06:20 - iterative_telemetry - DEBUG - Sending payload {'interface': 'api', 'action': 'from_storage', 'error': None, 'extra': {'client': 'gs://'}, 'python_version': {'major': 3, 'minor': 12, 'patch': 4}, 'tool_name': 'datachain', 'tool_version': '0.3.10.dev12+g944029b.d20240905', 'user_id': 'cb4bc63c-f9f7-46ea-8ff3-e3201fbabeb7', 'group_id': None, 'os_name': 'mac', 'os_version': '15.1'}
2024-09-09 20:06:20 - datachain.telemetry - DEBUG - Telemetry is enabled.
2024-09-09 20:06:20 - iterative_telemetry - DEBUG - Sending payload {'interface': 'api', 'action': 'map', 'error': None, 'extra': {}, 'python_version': {'major': 3, 'minor': 12, 'patch': 4}, 'tool_name': 'datachain', 'tool_version': '0.3.10.dev12+g944029b.d20240905', 'user_id': 'cb4bc63c-f9f7-46ea-8ff3-e3201fbabeb7', 'group_id': None, 'os_name': 'mac', 'os_version': '15.1'}
2024-09-09 20:06:20 - datachain.telemetry - DEBUG - Telemetry is enabled.
2024-09-09 20:06:20 - iterative_telemetry - DEBUG - Sending payload {'interface': 'api', 'action': 'show', 'error': None, 'extra': {}, 'python_version': {'major': 3, 'minor': 12, 'patch': 4}, 'tool_name': 'datachain', 'tool_version': '0.3.10.dev12+g944029b.d20240905', 'user_id': 'cb4bc63c-f9f7-46ea-8ff3-e3201fbabeb7', 'group_id': None, 'os_name': 'mac', 'os_version': '15.1'}

```

Closes #344

### DIFF
--- a/src/datachain/cli.py
+++ b/src/datachain/cli.py
@@ -15,7 +15,7 @@ import shtab
 from datachain import utils
 from datachain.cli_utils import BooleanOptionalAction, CommaSeparatedArgs, KeyValueArgs
 from datachain.lib.dc import DataChain
-from datachain.telemetry import api_telemetry, telemetry, track_cli_telemetry
+from datachain.telemetry import telemetry
 from datachain.utils import DataChainDir
 
 if TYPE_CHECKING:
@@ -692,7 +692,6 @@ def ls_remote(
         first = False
 
 
-@api_telemetry
 def ls(
     sources,
     long: bool = False,
@@ -718,14 +717,12 @@ def ls(
         )
 
 
-@api_telemetry
 def ls_datasets(catalog: "Catalog"):
     for d in catalog.ls_datasets():
         for v in d.versions:
             print(f"{d.name} (v{v.version})")
 
 
-@api_telemetry
 def rm_dataset(
     catalog: "Catalog",
     name: str,
@@ -735,7 +732,6 @@ def rm_dataset(
     catalog.remove_dataset(name, version=version, force=force)
 
 
-@api_telemetry
 def dataset_stats(
     catalog: "Catalog",
     name: str,
@@ -753,7 +749,6 @@ def dataset_stats(
             print(f"Total objects size: {utils.sizeof_fmt(stats.size, si=si): >7}")
 
 
-@api_telemetry
 def du(catalog: "Catalog", sources, show_bytes=False, si=False, **kwargs):
     for path, size in catalog.du(sources, **kwargs):
         if show_bytes:
@@ -762,7 +757,6 @@ def du(catalog: "Catalog", sources, show_bytes=False, si=False, **kwargs):
             print(f"{utils.sizeof_fmt(size, si=si): >7} {path}")
 
 
-@api_telemetry
 def index(
     catalog: "Catalog",
     sources,
@@ -771,7 +765,6 @@ def index(
     catalog.index(sources, **kwargs)
 
 
-@api_telemetry
 def show(
     catalog: "Catalog",
     name: str,
@@ -803,7 +796,6 @@ def show(
         dc.print_schema()
 
 
-@api_telemetry
 def query(
     catalog: "Catalog",
     script: str,
@@ -851,12 +843,10 @@ def query(
     catalog.metastore.set_job_status(job_id, JobStatus.COMPLETE)
 
 
-@api_telemetry
 def clear_cache(catalog: "Catalog"):
     catalog.cache.clear()
 
 
-@api_telemetry
 def garbage_collect(catalog: "Catalog"):
     temp_tables = catalog.get_temp_table_names()
     if not temp_tables:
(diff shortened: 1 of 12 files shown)


tests/test_telemetry.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+def test_is_enabled():
+    assert not is_enabled()
+
+
+def test_telemetry_api_call(mocker, tmp_dir):
+    patch_send = mocker.patch("iterative_telemetry.IterativeTelemetryLogger.send")
+
+    DataChain.from_storage(tmp_dir.as_uri())
```
mattseddon, line 12: [Q] What will be sent for a chain for the form?

`DataChain.from_storage(tmp_dir.as_uri()).map(...).show()` ? Will we get three events and will there be a way to link these events together into a "user session"?

skshetry (changes requested):
This change will require us to remember to add decorators to every API. 

I think you should change your approach to only send analytics when `DataChain` is first constructed, and ignore all APIs. I don't think those are useful and are quite intrusive, both for users, and for us since every function has to go through an indirection.


```python
class DataChain:
  def __init__(self, *args, **kwargs):
    global sent_telemetry
    if not sent_telemetry and not telemetry_disabled:
      send(...)
      sent_telemetry = True
```

tests/test_telemetry.py
amritghimire: Updated the logic. PTAL.

mattseddon: tl;dr - I appreciate the effort required to implement proper telemetry is high and this is more than likely good enough.

FWIW - This seems okay for a first pass, but it will not give us any information about how the product is being used. This should however be enough for the active loyal users metric which is probably all we care about as we currently do not have a person dedicated to looking at telemetry/making decisions from it. To give an idea of usage I think we would need to send an event as/when each chain is evaluated (e.g. `save`, `exec`, `to_records`, `collect`, etc).

src/datachain/lib/dc.py
```diff
@@ -247,6 +248,8 @@ def __init__(self, *args, settings: Optional[dict] = None, **kwargs):
             **kwargs,
             indexing_column_types=File._datachain_column_types,
         )
+        send_telemetry_once("datachain_init", **kwargs)
```
mattseddon, line 251: [Q] Would it simplify the code to run on import of `DataChain` instead of `init`?

tests/test_telemetry.py
jendefig: I will be looking at these metrics monthly as I have done every month for the last 3 years and 7 months.  That nobody pays attention in the company has nothing to do with the fact that someone is dedicated to this and is extremely frustrated that we do not do this better and make better decisions from this data. 

Questions/comments:
1. Is the loyal monthly users as being set up here now equal to how it is applied for DVC (5 times usage or more per month)?
2. I agree that as much information in the chain  as possible would be highly desired from the beginning - we will be able to see where people get stuck.  Otherwise we could just be counting trials. with no idea of next steps.  This will be of minimal value like the data from the VS Code extension.
3. It makes more sense to do it right from the beginning so that we can have good data. 
4. Adding for a wholistic view: Our new use of Reo.dev will help to some extent in determining user journey (what people use, where they get stuck, etc) because it will be able to track who and from what company hits our doc pages and when (😃 🎉). But it would be immensely better to know this from the product telemetry.

shcheklein: > This will be of minimal value like the data from the VS Code extension. 

qq - did you get / have access to Azure ? that's where all real product analytics for VS Code is (and it's very comprehensive).

we also have the breakdown on the DVC usage (anonymized, etc) - which commands are being used more, etc. 

And this is not about someone not paying attention (we did with @mattseddon and @dberenbaum ). I think it boils down to the fact that you need to get to some different numbers to start benefit from more granular events in a meaningful way.

On the other hand it can potentially slow down the product a bit, a risk of breaking the trust with the community, etc. My 2cs - I would start with basics and expand if needed later.

amritghimire: > On the other hand it can potentially slow down the product a bit, a risk of breaking the trust with the community, etc. My 2cs - I would start with basics and expand if needed later.

I agree. IMO, let's merge this and look at the metrics to see if the information we get are enough or not. If we need more information, we can then revive `api_telemetry` decorator and attach it to few function we want to look at like save, exec, to_records and so on.

dreadatour (approved):
Code looks good to me! It would me great if someone else will take a look and approved this PR as I am not very familiar with telemetry part 👀

src/datachain/lib/dc.py
```diff
@@ -246,6 +247,8 @@ def __init__(self, *args, settings: Optional[dict] = None, **kwargs):
...
```
shcheklein, line 250: can we make it part of the telemetry package please? to keep code here simpler 

I mean the "once" part implementation

shcheklein (commented):
Just to confirm - telemetry send events in a separate thread, right?

amritghimire:
> Just to confirm - telemetry send events in a separate thread, right?

Yes

src/datachain/lib/dc.py
amritghimire: https://github.com/iterative/telemetry-python/pull/59 PTAL cc. @shcheklein

src/datachain/cli.py
```diff
@@ -1053,3 +1057,5 @@ def main(argv: Optional[list[str]] = None) -> int:  # noqa: C901, PLR0912, PLR09
 
             pdb.post_mortem()
         return 1
+    finally:
+        telemetry.send_cli_call(args.command, error=error)
```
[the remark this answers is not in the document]
amritghimire: Even if it is not exposed, I believe it would be a good idea to get information if someone is using it IMO.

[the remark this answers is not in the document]
skshetry: Who is going to use it?

src/datachain/telemetry.py
```diff
@@ -0,0 +1,36 @@
...
+def is_enabled():
+    """
+    Determine if telemetry is enabled based on environment variables and configuration.
+    """
+    # Disable telemetry if running in test mode
+    if env2bool("DATACHAIN_TEST"):
+        return False
+
+    # Check if telemetry is disabled by environment variable
+    disabled = env2bool("DATACHAIN_NO_ANALYTICS")
```
skshetry, line 20: In dvc, this envvar disables analytics when it is set, regardless of it's value. 

```suggestion
    disabled = bool(os.getenv("DATACHAIN_NO_ANALYTICS"))
```

The envvar being set is an enough signal from the user, and the boolean values are confusing (is it `1`, `true`, `True`, `0`, 'y', 'yes', etc to disable analytics?).

skshetry (approved):

src/datachain/lib/listing.py
```diff
@@ -79,6 +80,7 @@ def parse_listing_uri(uri: str, cache, client_config) -> tuple[str, str, str]:
     """
     client = Client.get_client(uri, cache, **client_config)
     storage_uri, path = Client.parse_url(uri)
+    telemetry.log_param("client", client.PREFIX)
```
skshetry, line 83: Why do we have telemetry call inside a utility function?

PR #412, opened 2024-09-09
fix open file method for tar files

dberenbaum:
Opening files from tar archives like wds was broken. This also has a small fix to the wds example to read the file as images since that is what is needed for downstream tasks like model training.

### DIFF
--- a/src/datachain/lib/file.py
+++ b/src/datachain/lib/file.py
@@ -195,14 +195,15 @@ class File(DataModel):
             with VFileRegistry.resolve(self, self.location) as f:  # type: ignore[arg-type]
                 yield f
 
-        uid = self.get_uid()
-        client = self._catalog.get_client(self.source)
-        if self._caching_enabled:
-            client.download(uid, callback=self._download_cb)
-        with client.open_object(
-            uid, use_cache=self._caching_enabled, cb=self._download_cb
-        ) as f:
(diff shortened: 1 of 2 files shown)


src/datachain/lib/file.py
```diff
@@ -195,14 +195,15 @@ def open(self, mode: Literal["rb", "r"] = "rb"):
...
 
-        uid = self.get_uid()
-        client = self._catalog.get_client(self.source)
-        if self._caching_enabled:
-            client.download(uid, callback=self._download_cb)
-        with client.open_object(
-            uid, use_cache=self._caching_enabled, cb=self._download_cb
-        ) as f:
-            yield io.TextIOWrapper(f) if mode == "r" else f
+        else:
```
dberenbaum, line 198: Tar archives were still hitting this logic because `else:` was missing

src/datachain/lib/file.py
```diff
@@ -195,14 +195,15 @@ def open(self, mode: Literal["rb", "r"] = "rb"):
             with VFileRegistry.resolve(self, self.location) as f:  # type: ignore[arg-type]
                 yield f
```
skshetry, line 196: Nitpicky suggestion: Instead of `else`, you can early return from `if` branch, and keep the rest of the code same.


But no strong opinion.

```suggestion
                yield f
                return
```

skshetry (approved):

PR #418, opened 2024-09-10
Fix calculating datasets stats size

dreadatour:
It looks like calculating datasets stats is broken in some cases.

The way it works now is we are calculating the sum of either `file__size` or `size` fields in all dataset rows. I think `size` is not used anymore and `file__size` will works only for listings.

In this PR I am calculating `size` dataset stat as sum of all fields with name ending with `file__size`. This way we will be able to calculate listing size (`file__size` field) and other signals, like `laion.file.size`, `source.file.size`, `emd.file.size`, etc. Also we are now calculating sum of all `file.size` fields instead of calculating only one of them.

### DIFF
--- a/src/datachain/catalog/catalog.py
+++ b/src/datachain/catalog/catalog.py
@@ -1052,7 +1052,6 @@ class Catalog:
         if create_rows_table:
             table_name = self.warehouse.dataset_table_name(dataset.name, version)
             self.warehouse.create_dataset_rows_table(table_name, columns=columns)
-            self.update_dataset_version_with_warehouse_info(dataset, version)
 
         return dataset
 
@@ -1390,12 +1389,12 @@ class Catalog:
         dataset = self.get_dataset(name)
         return self.warehouse.dataset_table_export_file_names(dataset, version)
 
-    def dataset_stats(self, name: str, version: int) -> DatasetStats:
+    def dataset_stats(self, name: str, version: Optional[int]) -> DatasetStats:
         """
         Returns tuple with dataset stats: total number of rows and total dataset size.
         """
         dataset = self.get_dataset(name)
-        dataset_version = dataset.get_version(version)
+        dataset_version = dataset.get_version(version or dataset.latest_version)
         return DatasetStats(
             num_objects=dataset_version.num_objects,
             size=dataset_version.size,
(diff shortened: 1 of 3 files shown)


src/datachain/catalog/catalog.py
```diff
@@ -1091,7 +1091,6 @@ def create_new_dataset_version(
         if create_rows_table:
             table_name = self.warehouse.dataset_table_name(dataset.name, version)
             self.warehouse.create_dataset_rows_table(table_name, columns=columns)
-            self.update_dataset_version_with_warehouse_info(dataset, version)
```
dreadatour, line 1094: No need to calculate dataset stats for newly created empty datasets (?)

src/datachain/catalog/catalog.py
```diff
@@ -1429,12 +1428,12 @@ def dataset_table_export_file_names(self, name: str, version: int) -> list[str]:
...
         return self.warehouse.dataset_table_export_file_names(dataset, version)
 
-    def dataset_stats(self, name: str, version: int) -> DatasetStats:
+    def dataset_stats(self, name: str, version: Optional[int]) -> DatasetStats:
         """
         Returns tuple with dataset stats: total number of rows and total dataset size.
         """
         dataset = self.get_dataset(name)
-        dataset_version = dataset.get_version(version)
+        dataset_version = dataset.get_version(version or dataset.latest_version)
```
dreadatour, line 1436: `datachain dataset-stats ds-name` command now takes latest dataset version if version is not set instead of raising an error.

src/datachain/data_storage/warehouse.py
```diff
@@ -404,13 +404,12 @@ def dataset_stats(
...
-            expressions = (*expressions, sa.func.sum(table.c.file__size))
-        elif "size" in table.columns:
-            expressions = (*expressions, sa.func.sum(table.c.size))
+        for c in table.columns:
+            if c.name.endswith("file__size"):
+                expressions = (*expressions, sa.func.sum(c))
         query = select(*expressions)
         ((nrows, *rest),) = self.db.execute(query)
-        return nrows, rest[0] if rest else None
+        return nrows, sum(rest) if rest else 0
```
dreadatour, line 412: We are waiting for `int` in many places, errors happens if numbers of rows is set, but size is `None` (for example, [here](https://github.com/iterative/datachain/pull/418/files#diff-3b39a77247b3274508f28e3a5263a04189e57183b5bcfade572ccef288d1f293R1437-R1440)). It is better to return `0` in such cases.

dreadatour: Error looks like this:

```
$ datachain dataset-stats laion_pq
Number of objects: 2
Error: bad operand type for abs(): 'NoneType'
$
```

shcheklein (approved):

amritghimire (approved):

PR #421, opened 2024-09-10
Fix storage dependencies

ilongin:
Making sure we have only name of the bucket in storage dependencies, instead of bucket name + path.
Issue was starting to appear when we moved listing to app level with `DataChain.from_storage()`

Example:
```python

ds = DataChain.from_storage("s3://ldb-public/remote/data-lakes/", anon=True).save("dep")
```

This script generates storage dependency with name `s3://ldb-public` and before these changes it was generating `s3://ldb-public/remote/data-lakes/`


rlamy (approved):

dreadatour (approved):
👍

src/datachain/dataset.py
```diff
@@ -112,7 +112,11 @@ def parse(
 
         if is_listing_dataset(dataset_name):
             dependency_type = DatasetDependencyType.STORAGE  # type: ignore[arg-type]
-            dependency_name = listing_uri_from_name(dataset_name)
+            client, _ = Client.parse_url(
+                listing_uri_from_name(dataset_name),
+                None,  # type: ignore[arg-type]
```
amritghimire, line 117: Can we fix this instead of ignoring?

ilongin: Yes, this is on my plate for long time. I will see if it makes sense to do it in separate PR or this one

ilongin: Fix in separate PR https://github.com/iterative/datachain/pull/435

PR #422, opened 2024-09-10
Adding Complex Type Support to Signal Schema

### LINKED ISSUE #98: SignalSchemaError when saving a dataset with a list of floats.
### Description

When I am running [this notebook](https://github.com/iterative/datachain/blob/llm-tutorial/examples/llm/llm_chatbot_evaluation.ipynb) (WIP for the LLM guide for datachain), I run into the following error at the first cell in the #Evaluation section when I try to save the `dc_chunks_embeddings` datachain.
```
SignalSchemaError: cannot deserialize 'document.embeddings': unknown type 'List'. Try to add it with `ModelStore.add(List)`.
```

Also, it seems that running `dc_chunks_embeddings.save()` in the previous cell is ignored as the embeddings are recalculated again (if I save it with a specific name and then load it with `.from_dataset` it works fine).

I am not sure if I am misunderstanding something or if this is indeed a bug.

### Version Info

```Text
0.2.11
Python 3.12.4
```

dtulga:
This adds complex type support to Signal Schema, including in model serialization / deserialization. This includes old-style `List` and `Dict` support, along with many other types, nested types, and custom type support as well. This also includes fallbacks for when custom types are not resolvable or available, and has been tested to serialize and deserialize mistral types as well (which use many strange and custom types in their classes).

This has been tested to work locally, and with Studio for both data display and image previews, even with custom or third-party types, including mistral types. Note that this was tested before many of the current breaking changes for Studio were introduced, and I will test this again in Studio once those are resolved. (The Studio test failures appear unrelated to this PR, and are also failing on `main`.) This is part of https://github.com/iterative/studio/issues/10601 and fixes #98


src/datachain/lib/signal_schema.py
```diff
@@ -183,37 +167,102 @@ def serialize_custom_model_fields(
...
+            close_bracket_idx = type_name.rfind("]")
+            if close_bracket_idx == -1:
+                raise TypeError("Unclosed square bracket when parsing type")
+            if close_bracket_idx < bracket_idx:
+                raise TypeError("Square brackets are out of order when parsing type")
+            if close_bracket_idx == bracket_idx + 1:
+                raise TypeError("Empty square brackets when parsing type")
+            subtype_names = SignalSchema._split_subtypes(
+                type_name[bracket_idx + 1 : close_bracket_idx]
+            )
```
skshetry, line 251: I think we should investigate how others do it (eg: `pydantic`).
This looks very complicated.

skshetry: I wonder if we can use `eval` or `typing._eval_type()`/`typing.get_type_hints()`.

dtulga: I'm not aware of any pydantic-based string to type conversion, but please do mention if that is available somewhere. I did think of using `eval` but I'm not sure of the security implications / possibility for other problems from using the `eval` function. Also, this section of the code mostly just checks for syntax errors, as seen in the test `test_resolve_types_errors`.

skshetry: Aren't these trusted data (coming from `_custom_types`)?

dreadatour:
Thank you, @dtulga, looks great to me! 👍

I can confirm this PR fixes the issue in https://github.com/iterative/studio/issues/10601

`main` branch:
image

This PR:
image

I am worrying about E2E tests failing, but I don't really think it is related to this PR 👀

src/datachain/lib/signal_schema.py
dreadatour: My two cents: https://peps.python.org/pep-0563/#resolving-type-hints-at-runtime
I’m sure you’re already aware of this PEP 👀

dreadatour (approved):
Looks good to me overall. Great job! 👍

Tested this PR on my localhost, and it solved the bug (see [this comment](https://github.com/iterative/studio/issues/10601#issuecomment-2342600541)).

Also studio tests are failing 😥 But it looks like it is not related to this PR, because it fails in also `main` branch (e.g. [here](https://github.com/iterative/datachain/actions/runs/10804111765/job/29969044923)).

src/datachain/lib/signal_schema.py
dreadatour: > Aren't these trusted data (coming from _custom_types)?

It is also depends on where do we run this. If we run this only in DataChain (CLI or Studio workers) — it is absolutely OK to use `eval`, but if we are going to run this inside Studio backend, we should be very careful, because `_custom_types` can be easily updated by user with malformed data.

I am OK with this code, it is not covering all possible scenarios and errors might happens in case of bad data, but I think it works just fine with data generated by DataChain and I think we should not overcomplicate it.

Other option might be using regexps btw 🤔

skshetry: It's okay to `eval` even on Studio (they run inside a cluster). They are limited to types. (`typing.get_type_hints()` does exactly that).

> `_custom_types` can be easily updated by user with malformed data.

We should not allow anyone to do it through our API. But a malicious user with access to cluster can do anything they like. So that's not an issue that we need to solve (hence why we have clusters).

> but I think it works just fine with data generated by DataChain

DataChain generates arbitrary models, so it should work with anything. DataChain is no longer limited to `File` models.

We are implementing our own parser (see `_split_subtypes`) for stringized annotations. And we are adding a lot of complexity in a already complicated module.

> Other option might be using regexps btw

If we can avoid parsing this ourselves, that is my whole point.

dreadatour: > It's okay to `eval` even on Studio (they run inside a cluster).

No, I mean in Studio backend, not DataChain compute cluster. And that's my only concern agains `eval`.

> We should not allow anyone to do it through our API. But a malicious user with access to cluster can do anything they like. So that's not an issue that we need to solve (hence why we have clusters).

It is OK to do anything inside the cluster, but one can send malformed data into our Studio backend and it is not OK to run `eval` there. We can not even run checks with `eval` in API endpoint because it will be run inside Studio backend, and at the same time we can not trust to checks ran in DataChain cluster.

> DataChain generates arbitrary models, so it should work with anything. DataChain is no longer limited to `File` models.
> 
> We are implementing our own parser (see `_split_subtypes`) for stringized annotations. And we are adding a lot of complexity in a already complicated module.

I agree 😥

> > Other option might be using regexps btw
> 
> If we can avoid parsing this ourselves, that is my whole point.

Yes, this is not a replacement for this parser, this is just one more way to do this, worth mentioning.

dreadatour: Thank you @dtulga ❤️ For me it looks good! 👍

skshetry:
I don't think there's any need for being paranoid (they are just types).

At least, I'd have loved to see other alternatives and research into other tools before getting this merged. Again, parsing type annotations is already complicated, and it's a moving target.
It's also hard to guarantee that this approach _supports_ everything.

PR #424, opened 2024-09-11
patch psutil virtual memory for test suite in CI

mattseddon:
Related to #405 / #406

I have a theory that in the Windows CI `psutil.virtual_memory().percent > 80` is being hit often in `process_udf_outputs` (which is executed under parallel mode).  Code is below: 
https://github.com/iterative/datachain/blob/424b05b840a047a7801016cf91cb19a545ea5402/src/datachain/query/dataset.py#L383-L410

This would mean that we are concurrently writing to the warehouse more often, giving the UDF far more chance to fail due to SQLite locks.

This PR patches `psutil.virtual_memory` to be automatically pinned at 50% utilised for the entirety of the test suite.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -3,6 +3,7 @@ import os.path
 import uuid
 from collections.abc import Generator
 from pathlib import PosixPath
+from typing import NamedTuple
 
 import attrs
 import pytest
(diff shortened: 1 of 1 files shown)


skshetry:
@mattseddon, thank you for investigating. Can you demonstrate that this fixes the timing-out issue?

Although it looks like a good thing that we should do anyway.

mattseddon:
> @mattseddon, thank you for investigating. Can you demonstrate that this fixes the timing-out issue?
> 
> Although it looks like a good thing that we should do anyway.

@skshetry PTAL at #427 it looks like those changes should fix the test suite running against Windows. I think the main thing is moving all of the tests that use `_tmpfile` fixtures onto the same worker.

mattseddon:
closing in favour of #427

PR #425, opened 2024-09-11
simplify SQLiteIDGenerator tests (can fail due to concurrency)

mattseddon:
~deleted assertion failed here: https://github.com/iterative/datachain/actions/runs/10804888436/job/29971096918?pr=424~

~and here: https://github.com/iterative/datachain/actions/runs/10806124770/job/29974368026~

After the discussion (below), I deleted some extra assertions/behaviour from the `SQLiteIDGenerator` tests. I do not believe that the juice was worth the squeeze.

### DIFF
--- a/tests/unit/test_id_generator.py
+++ b/tests/unit/test_id_generator.py
@@ -43,7 +43,6 @@ def test_init_with_prefix(sqlite_db):
     assert id_generator.db == sqlite_db
     assert id_generator._table_prefix == "foo"
     assert sqlite_db.has_table("foo_id_generator")
-    assert not sqlite_db.has_table("id_generator")
     assert get_rows(id_generator) == set()
 
     id_generator.cleanup_for_tests()
--- a/tests/unit/test_id_generator.py
+++ b/tests/unit/test_id_generator.py
@@ -22,20 +22,14 @@ def test_init(sqlite_db):
     assert sqlite_db.has_table("id_generator")
     assert get_rows(id_generator) == set()
 
-    id_generator.cleanup_for_tests()
-    assert not sqlite_db.has_table("id_generator")
-
 
 @skip_if_not_sqlite
 def test_init_empty(tmp_dir):
-    id_generator = SQLiteIDGenerator()
+    id_generator = SQLiteIDGenerator(db_file=tmp_dir / "test.db")
     assert id_generator._table_prefix is None
     assert id_generator.db
     assert id_generator.db.has_table("id_generator")
     assert get_rows(id_generator) == set()
-
-    id_generator.cleanup_for_tests()
-    assert not id_generator.db.has_table("id_generator")
 
 
 def test_init_with_prefix(sqlite_db):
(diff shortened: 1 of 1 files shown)


dreadatour (commented):
🤔 interesting. This does not looks like fix to me, may be we should rather try to fix concurrent tests, because if it fails here and we are ignoring it this way, it still may fails somewhere else, we are just suppressing the issue.

At the same time I am kind of OK to merge this PR 👀

mattseddon:
> 🤔 interesting. This does not looks like fix to me, may be we should rather try to fix concurrent tests, because if it fails here and we are ignoring it this way, it still may fails somewhere else, we are just suppressing the issue.
> 
> At the same time I am kind of OK to merge this PR 👀

Do you think the assertion is important? It didn't seem worth the effort to me for `test_init_with_prefix` for the `SQLiteIDGenerator`.

dreadatour:
> Do you think the assertion is important? It didn't seem worth the effort to me for `test_init_with_prefix` for the `SQLiteIDGenerator`.

This assertion is not really important to be honest, but this PR hides the issue with concurrent tests and this looks important to me.

mattseddon:
> > Do you think the assertion is important? It didn't seem worth the effort to me for `test_init_with_prefix` for the `SQLiteIDGenerator`.
> 
> This assertion is not really important to be honest, but this PR hides the issue with concurrent tests and this looks important to me.

When we connect to the in-memory version of SQLite we `# Enable multithreaded usage of the same in-memory db`:

https://github.com/iterative/datachain/blob/43f5b21f70581f76fa26732b8234af3b6f76c0ea/src/datachain/data_storage/sqlite.py#L119-L123

The only way I can see to "fix" the concurrency issue would be to use `pytest-split` and mirror the approach we've taken for the Studio suite by running all tests sequentially in each split.

tests/unit/test_id_generator.py
```diff
@@ -22,33 +22,23 @@ def test_init(sqlite_db):
...
     assert get_rows(id_generator) == set()
 
-    id_generator.cleanup_for_tests()
-    assert not sqlite_db.has_table("id_generator")
-
 
 @skip_if_not_sqlite
 def test_init_empty(tmp_dir):
-    id_generator = SQLiteIDGenerator()
+    id_generator = SQLiteIDGenerator(db_file=tmp_dir / "test.db")
```
mattseddon, line 28: [I] If #427 gets merged this needs to go in the `tmpfile` xdist group.

mattseddon:
> > > Do you think the assertion is important? It didn't seem worth the effort to me for `test_init_with_prefix` for the `SQLiteIDGenerator`.
> > 
> > 
> > This assertion is not really important to be honest, but this PR hides the issue with concurrent tests and this looks important to me.
> 
> When we connect to the in-memory version of SQLite we `# Enable multithreaded usage of the same in-memory db`:
> 
> https://github.com/iterative/datachain/blob/43f5b21f70581f76fa26732b8234af3b6f76c0ea/src/datachain/data_storage/sqlite.py#L119-L123
> 
> The only way I can see to "fix" the concurrency issue would be to use `pytest-split` and mirror the approach we've taken for the Studio suite by running all tests sequentially in each split.

There might be a different fix for this: https://www.sqlite.org/draft/inmemorydb.html 

```
If two or more distinct but shareable in-memory databases are needed in a single process, then the [mode=memory](https://www.sqlite.org/draft/uri.html#coreqp) query parameter can be used with a [URI filename](https://www.sqlite.org/draft/uri.html) to create a named in-memory database:

rc = sqlite3_open("file:memdb1?mode=memory&cache=shared", &db);
```

mattseddon:
also rolled into #427

PR #426, opened 2024-09-11
fix dependency install instructions for examples

mattseddon:
I have simplified and standardised the install instructions at the top of examples that require dependencies that are not available through the base `datachain` installation.


dreadatour (approved):
❤️

ilongin (approved):

shcheklein (approved):

PR #427, opened 2024-09-11
prevent tests from hanging on CI (windows)

mattseddon:
This PR is designed to prevent the Windows portion of the CI's tests from hanging but the changes should benefit the entire pipeline. I have left explanatory inline comments which detail the reasons for the changes.

Links to two successive 50-iteration runs against Windows with Python 3.12 are shown below:

- https://github.com/iterative/datachain/actions/runs/10826685823
- https://github.com/iterative/datachain/actions/runs/10825863127


tests/conftest.py
```diff
@@ -32,6 +33,39 @@
 collect_ignore = ["setup.py"]
 
 
+@pytest.fixture(autouse=True)
+def virtual_memory(mocker):
```
mattseddon, line 37: might not need this anymore

tests/conftest.py
```diff
@@ -32,6 +33,39 @@
...
+    )
+
+
+@pytest.fixture(autouse=True)
+def per_thread_im_mem_db(mocker, worker_id):
+    if worker_id == "master":
+        return
+    mocker.patch(
+        "datachain.data_storage.sqlite._get_in_memory_uri",
+        return_value=f"file:in-mem-db-{worker_id}?mode=memory&cache=shared",
```
mattseddon, line 65: from https://www.sqlite.org/draft/inmemorydb.html: 

If two or more distinct but shareable in-memory databases are needed in a single process, then the [mode=memory](https://www.sqlite.org/draft/uri.html#coreqp) query parameter can be used with a [URI filename](https://www.sqlite.org/draft/uri.html) to create a named in-memory database:

rc = sqlite3_open("file:memdb1?mode=memory&cache=shared", &db);

We need one per worker to stop them from running into one another.

noxfile.py
```diff
@@ -40,6 +40,7 @@ def tests(session: nox.Session) -> None:
         "--cov-report=xml",
         "--durations=10",
         "--numprocesses=logical",
+        "--dist=loadgroup",
```
mattseddon, line 43: [F] All tests that use the `tmpfile` flavour of the database (i.e. the tests that use parallel/distributed processing) have been grouped onto the same worker.

tests/func/test_datachain.py
```diff
@@ -463,3 +465,162 @@ def test_mutate_existing_column(test_session):
         == "Error for column ids: Cannot modify existing column with mutate()."
         " Use a different name for the new column."
     )
+
+
+@pytest.mark.parametrize(
```
mattseddon, line 470: [F] These tests were timing out under the old `DatasetQuery` code/methods so I migrated them to code that is currently being used. More context/discussion [here](https://iterativeai.slack.com/archives/C04A9RWEZBN/p1726114474925839).

tests/test_cli_e2e.py
```diff
@@ -180,14 +180,20 @@ def verify_files(files, base=""):
...
-def run_step(step):
+def run_step(step, catalog):
     """Run an end-to-end test step with a command and expected output."""
     result = subprocess.run(  # noqa: S603
         step["command"],
         shell=False,
         capture_output=True,
         check=True,
         encoding="utf-8",
+        env={
```
mattseddon, line 191: [F] This is my attempt to ensure the database is isolated to this set of e2e tests

dreadatour (approved):
Thank you for fixing this! ❤️

PR #428, opened 2024-09-11
use the correct fixtures in tests

mattseddon:
Here I've switched over from catalog to test_session in a few tests and then from catalog to catalog_tmpfile in some tests that use parallel processing.

### DIFF
--- a/tests/test_cli_e2e.py
+++ b/tests/test_cli_e2e.py
@@ -213,7 +213,7 @@ def run_step(step):
 
 
 @pytest.mark.e2e
-def test_cli_e2e(tmp_dir, catalog):
+def test_cli_e2e(tmp_dir, catalog_tmpfile):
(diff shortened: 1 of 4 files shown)


tests/test_cli_e2e.py
```diff
@@ -213,7 +213,7 @@ def run_step(step):
 
 
 @pytest.mark.e2e
-def test_cli_e2e(tmp_dir, catalog):
+def test_cli_e2e(tmp_dir, catalog_tmpfile):
```
mattseddon, line 216: both this and test_query_e2e seem to use the parallel setting which I think means they should use the `tmp_file` variation of the catalog fixture

dreadatour (approved):
Looks reasonable 👍
Thank you for handling all these issues ❤️

ilongin (approved):

PR #431, opened 2024-09-12
tests: fix mock for subprocess stdout/stderr to return BytesIO

skshetry:
See https://github.com/iterative/datachain/pull/407#discussion_r1755915341.

`stdout`/`stderr` are now always a binary stream since #407, so adjusting the mocks in the tests to account for that.

Thanks @mattseddon and @amritghimire for identifying. :)


dreadatour (approved):
Thank you! ❤️

PR #436, opened 2024-09-12
fix last modified comparison in resolve file test

mattseddon:


shcheklein (approved):

tests/unit/lib/test_file.py
```diff
@@ -327,11 +329,21 @@ def test_read_text(tmp_path, catalog):
 def test_resolve_file(cloud_test_catalog):
     ctc = cloud_test_catalog
 
+    is_sqlite = isinstance(cloud_test_catalog.catalog.warehouse, SQLiteWarehouse)
```
mattseddon, line 332: there is a similar strategy in `test_from_storage_check_rows`

PR #440, opened 2024-09-13
Add `datachain.lib.tar.process_tar()` generator

### LINKED ISSUE #439: Convert  `index_tar` to a new-style generator
Though we have `process_webdataset`, we also need a simple generator that just returns `File` objects for each member of an archive similarly to the old-style UDF `index_tar`.

Usage should look like:
```python
chain = DataChain.from_storage("s3://my-bucket").filter(C("file.path").glob("*.tar").gen(file=process_tar)
```

This would allow us to test tar file handling in `datachain.lib` (which we're not currently doing, as mentioned in #412), and I consider it a blocker for #438 and #367.

rlamy:
This is a replacement for the deprecated `index_tar()` UDF. Also removes unused `TarStream` class.

Fixes #439

### DIFF
--- a/src/datachain/lib/tar.py
+++ b/src/datachain/lib/tar.py
@@ -1,0 +1,33 @@
+import hashlib
+import tarfile
+from collections.abc import Iterator
+
+from datachain.lib.file import File, TarVFile
+
+
+def build_tar_member(parent: File, info: tarfile.TarInfo) -> File:
+    new_parent = parent.get_full_name()
+    etag_string = "-".join([parent.etag, info.name, str(info.mtime)])
+    etag = hashlib.md5(etag_string.encode(), usedforsecurity=False).hexdigest()
+    return File(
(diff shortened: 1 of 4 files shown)


dberenbaum (approved):
LGTM - good generalization of the tar generator at a higher level than WDS.

One note that is outside the scope of this PR: I don't think compressed archives are being read correctly in `File.open/read()` with or without this PR because we don't decompress the archive before reading. I found this out when trying to add `File.read()` to the tests in `tests/examples/test_wds_e2e.py` for #441. That's one reason I chose to drop `FileSlice` there and use the `tarfile` utilities instead.

mattseddon (approved):

rlamy:
Compressed archives have never been supported, precisely because they don't support random access which was part of the original design for tar v-objects. We can revisit that, but note that it's a product decision - the current situation is WAD.

PR #438, opened 2024-09-13
remove legacy udf decorator

mattseddon:
This PR removes the legacy `@udf` decorator from the codebase. To do this I have migrated some tests from `DatasetQuery` to `DataChain`. 

I have deleted tests that relied on the udf decorator where: 

1. the code under test no longer seems relevant 
2. the test was explicitly testing the udf decorator
3. the code under test is covered elsewhere 
4. the test was marked xfail

PR #441, opened 2024-09-13
Vfile refactor

dberenbaum:
I started on this to try to clean up some of the arrow functionality and it expanded way beyond that into a full refactor of how to handle virtual files. I would prefer to separate this into multiple PRs but they all are interrelated. I'm happy to follow up on this beyond this week and/or leave it as a draft so others can take what's useful from it.

The PR tries to do a few things:
1. ~Reads arrow datasets from the cache.~ edit: extracted to #442 
2. Replaces `IndexedFile` with `ArrowVFile` - unlike `IndexedFile`, `ArrowVFile` has a `.read()` method that will read the row of data from the source file, somewhat similar to `TarVFile` (note that `ArrowVFile.read()` is very suboptimal for reading every row since it will create a new pyarrow dataset each time).
3. ~To implement `ArrowVFile`, we save the dataset kwargs as a dict, but they are read back as json. We previously expected this behavior but at this point I think json should be loaded automatically.~ edit: extracted to #444
4. Refactored `TarVFile` and webdataset functionalities to work with high-level `File` api and dropped old functionality. This has a lot of overlap with #440 and #438.


src/datachain/cache.py
```diff
@@ -48,8 +48,6 @@ def get_parsed_location(self) -> Optional[dict]:
 
     def get_hash(self) -> str:
         fingerprint = f"{self.storage}/{self.path}/{self.version}/{self.etag}"
-        if self.location:
-            fingerprint += f"/{self.location}"
```
dberenbaum, line 52: Stopped using location in vfiles, but even if we keep it, I don't think it should be part of the hash since the file itself is the same even if there are rows that point to different locations in the file.

src/datachain/lib/arrow.py
```diff
@@ -46,7 +46,10 @@ def __init__(
         self.kwargs = kwargs
 
     def process(self, file: File):
-        if self.nrows:
+        if file._caching_enabled:
+            path = file.get_local_path(download=True)
+            ds = dataset(path, schema=self.input_schema, **self.kwargs)
```
dberenbaum, line 51: Reads arrow dataset files from cache

src/datachain/lib/arrow.py
```diff
@@ -69,7 +72,12 @@ def process(self, file: File):
                                 vals_dict[field] = val
                         vals = [self.output_schema(**vals_dict)]
                     if self.source:
-                        yield [IndexedFile(file=file, index=index), *vals]
+                        kwargs: dict = self.kwargs
+                        # Can't serialize CsvFileFormat; may lose formatting options.
+                        if isinstance(kwargs.get("format"), CsvFileFormat):
+                            kwargs["format"] = "csv"
+                        arrow_file = ArrowVFile(file=file, index=index, kwargs=kwargs)
+                        yield [arrow_file, *vals]
```
dberenbaum, line 80: Save info to later open dataset and read rows from it

src/datachain/lib/file.py
```diff
@@ -257,14 +179,18 @@ def get_uid(self) -> UniqueId:
         dump = self.model_dump()
         return UniqueId(*(dump[k] for k in self._unique_id_keys))
 
-    def get_local_path(self) -> Optional[str]:
+    def get_local_path(self, download: bool = False) -> Optional[str]:
```
dberenbaum, line 182: Downloads file to cache if it doesn't exist

src/datachain/lib/file.py
```diff
@@ -409,22 +335,51 @@ class ImageFile(File):
 
     def read(self):
         """Returns `PIL.Image.Image` object."""
-        fobj = super().read()
-        return Image.open(BytesIO(fobj))
+        with self.open() as stream:
+            return Image.open(BytesIO(stream.read()))
```
dberenbaum, line 339: Needed to allow for mixins like `class LaionImageFile(ImageFile, TarVFile):`

src/datachain/lib/signal_schema.py
```diff
@@ -386,11 +386,20 @@ def row_to_features(
...
+    @staticmethod
+    def _set_file_stream(
+        obj: BaseModel, catalog: "Catalog", cache: bool = False
+    ) -> None:
+        if isinstance(obj, File):
+            obj._set_stream(catalog, caching_enabled=cache)
+        for field, finfo in obj.model_fields.items():
+            if ModelStore.is_pydantic(finfo.annotation):
+                SignalSchema._set_file_stream(getattr(obj, field), catalog, cache)
+
```
dberenbaum, line 402: This is a big fix that's needed to read files from wds regardless of the other changes in this PR. Right now, reading the wds files fails like this:

```python
>>> from datachain import C, DataChain
>>> from datachain.lib.webdataset import process_webdataset
>>> from datachain.lib.webdataset_laion import WDSLaion
>>> chain = DataChain.from_storage("gs://datachain-demo/datacomp-small/shards/00000000.tar", settings={"cache": True}).gen(laion=process_webdataset(spec=WDSLaion))
Processed: 1 rows [00:00, 1260.69 rows/s]
Generated: 2 rows [00:00, 2844.56 rows/s]
Cleanup: 1 tables [00:00, 5745.62 tables/s]
>>> row = next(chain.collect())
Processed: 1 rows [00:02,  2.66s/ rows]
Generated: 8953 rows [00:01, 8534.75 rows/s]
Cleanup: 1 tables [00:00, 39.20 tables/s]
>>> row[0].file.read()
  File "/Users/dave/Code/datachain/src/datachain/lib/file.py", line 213, in read
    with self.open() as stream:
  File "/Users/dave/micromamba/envs/datachain-py311/lib/python3.11/contextlib.py", line 137, in __enter__
    return next(self.gen)
           ^^^^^^^^^^^^^^
  File "/Users/dave/Code/datachain/src/datachain/lib/file.py", line 198, in open
    with VFileRegistry.resolve(self, self.location) as f:  # type: ignore[arg-type]
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dave/Code/datachain/src/datachain/lib/file.py", line 113, in resolve
    return reader.open(file, location)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dave/Code/datachain/src/datachain/lib/file.py", line 89, in open
    client = file._catalog.get_client(tar_file_uid.storage)
             ^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'get_client'
```

src/datachain/lib/webdataset_laion.py
```diff
@@ -41,7 +42,12 @@ def _reader(builder, item):
         return Laion.model_validate_json(builder.read_text(item))
 
 
-class WDSLaion(WDSBasic):
+class LaionImageFile(ImageFile, TarVFile):
```
dberenbaum, line 45: Read core files as images for downstream tasks like model training

src/datachain/lib/arrow.py
dberenbaum: Extracted to #442

src/datachain/lib/file.py
dberenbaum: Extracted to #442

src/datachain/lib/signal_schema.py
dberenbaum: Extracted to #443

dberenbaum:
Extracted all the parts into separate PRs so closing this one as too big

PR #442, opened 2024-09-13
Read arrow files from cache

dberenbaum:
Extracted from #441.

PR #445, opened 2024-09-13
`IndexedFile` -> `ArrowRow`

dberenbaum:
Extracted from #441.

Depends on #442 and #444.

Replaces `IndexedFile` with `ArrowRow` - unlike `IndexedFile`, `ArrowRow` has a `.read()` method that will read the row of data from the source file, somewhat similar to `TarVFile` (note that `ArrowRow.read()` is very suboptimal for reading every row since it will create a new pyarrow dataset each time).


src/datachain/lib/file.py
```diff
@@ -417,14 +418,30 @@ def save(self, destination: str):
...
+        else:
+            path = self.file.get_path()
+            ds = dataset(path, filesystem=self.file.get_fs(), **self.kwargs)
+
+        return ds.take([self.index]).to_reader()
+
+    def read(self):
+        """Returns row contents."""
+        with self.open() as record_batch:
+            return record_batch.to_pylist()[0]
```
shcheklein, line 444: should it be index here instead of `[0]`?

dberenbaum: No, we already take the index above in the open method. This is always a list of one and this ensures we return the dictionary instead of a list containing the dictionary.

dberenbaum:
@mattseddon One of the example tests if failing because there's nothing printed to stderr. What's the reason for this line?

https://github.com/iterative/datachain/blob/44c6a6ff9e5b1abb0d31239602ff8ac70f46d688/tests/examples/test_examples.py#L61

rlamy:
> @mattseddon One of the example tests if failing because there's nothing printed to stderr. What's the reason for this line?
> 
> https://github.com/iterative/datachain/blob/44c6a6ff9e5b1abb0d31239602ff8ac70f46d688/tests/examples/test_examples.py#L61

I opened #462 to discuss that failure. I think we can ignore it here.

rlamy (approved):
LGTM, though I find the name `ArrowVFile` confusing, since it's actually not a `VFile`.

dberenbaum:
Thanks @rlamy! Any alternate naming suggestions?

rlamy:
> Thanks @rlamy! Any alternate naming suggestions?

`ArrowData` or `ArrowRow`?

PR #446, opened 2024-09-13
`TarVFile` refactor

dberenbaum:
Extracted from #441.

Refactored `TarVFile` and webdataset functionalities to work with high-level `File` api and dropped old functionality. This has a lot of overlap with [Add `datachain.lib.tar.process_tar()` generator #440](https://github.com/iterative/datachain/pull/440) and [remove legacy udf decorator #438](https://github.com/iterative/datachain/pull/438).


dberenbaum:
Closing in favor of #448 since I can't update this PR any longer

mattseddon (approved):
looks fine to me, but I am an idiot

rlamy (changes requested):
I see several issues with this PR:
* Removing `FileSlice` means that we lose the ability to perform random reads inside a tar archive. To read a single image inside a webdataset file, we now need to get the whole file.
* We lose the ability to merge files and v-files in the same dataset - in fact, we're effectively doing away with the whole idea of v-files.
* (less important) `test_dir_expansion()` has been quite useful in the past to debug issues with the "Unix commands".

I'm not necessarily against these changes, but they're quite significant and we need to explore the trade-offs. @dberenbaum, what issues do you see with the current `TarVFile` and how do you think we should implement new kinds of v-objects (e.g. image patches/bounding boxes)?

dberenbaum:
@rlamy Yes, it is up to you if you think it's worth merging. You could also consider taking parts of it now or in the future.

In response to your specific concerns:

> * Removing `FileSlice` means that we lose the ability to perform random reads inside a tar archive. To read a single image inside a webdataset file, we now need to get the whole file.
I wonder how much the performance actually suffers here - it doesn't appear much worse in the examples. I did try to optimize this a bit so it streams the file and skips over each entry until hitting the image requested. The advantage would be that it is simpler and relies on `tarfile`, which also means it will support things like compression. Also, supporting other archive types could probably be achieved in a similar way easily.

However, it is possible to drop this part of the PR and bring back `FileSlice` along with a `TarVFile.offset` attribute. The rest of the PR could remain the same.

> * We lose the ability to merge files and v-files in the same dataset - in fact, we're effectively doing away with the whole idea of v-files.
Yes, I guess that's the primary benefit and drawback of the PR besides dropping a lot of code. Now that there is mature support for different `File` and `DataModel` classes, is `VFile` needed? IMHO the approach in this PR is more straightforward, and it will be easier to implement new v-objects and document how to do this since it only requires a new `File`-like class.

As you note, the drawback is the inability to merge files and v-files since they are two different `DataModel` classes. Is it useful to keep the raw tarfile records alongside the extracted objects? If so, is supporting two related classes in a dataset columns something that needs to be supported generally?

dberenbaum:
Just remembered another benefit to this PR: reading image files. The current wds examples won't work for model training since they don't read the file contents as images.

src/datachain/lib/file.py
```diff
@@ -206,18 +133,13 @@ def parent(self):
     @contextmanager
     def open(self, mode: Literal["rb", "r"] = "rb") -> Iterator[Any]:
         """Open the file and return a file object."""
-        if self.location:
```
shcheklein, line 209: does it mean we can drop location field after this PR?

src/datachain/lib/file.py
```diff
@@ -432,14 +354,32 @@ class ImageFile(File):
 
     def read(self):
         """Returns `PIL.Image.Image` object."""
-        fobj = super().read()
-        return Image.open(BytesIO(fobj))
+        with self.open() as stream:
```
shcheklein, line 357: hmm, am I missing something - or should it be the same (`super().read() still calls `self.open()`?)

src/datachain/lib/tar.py
```diff
@@ -5,28 +5,23 @@
 from datachain.lib.file import File, TarVFile
 
 
-def build_tar_member(parent: File, info: tarfile.TarInfo) -> File:
+def build_tar_member(
+    parent: File, info: tarfile.TarInfo, file_cls: type[TarVFile] = TarVFile
+) -> TarVFile:
```
shcheklein, line 10: should it be file_cls or substype of TarVFile?

src/datachain/lib/file.py
```diff
@@ -432,14 +354,32 @@ class ImageFile(File):
...
 
+class TarVFile(File):
+    """`DataModel` for files extracted from tar archives."""
+
+    tar: File
+
+    @contextmanager
+    def open(self):
+        """Stream file from tar archive."""
+        vpath = str(PurePosixPath(self.path).relative_to(self.tar.path))
```
shcheklein, line 373: just to make sure - I assume this should be fine with cache and things like `get_local_path` (?)

shcheklein:
I like the simplification and that we are relying on the existing abstractions (and optimization can be done on top even in this PR - model itself can stay simpler). 

> We lose the ability to merge files and v-files in the same dataset

@rlamy could you clarify please? what is the scenario / use case for this?

src/datachain/lib/file.py
dberenbaum: `super().read()` calls `super().open()`. So `LaionImageFile.read()` was calling `File.open()` instead of `TarVFile.open()` without this change.

src/datachain/lib/file.py
dberenbaum: Since `self.path` is just a pydantic str field, it should be fine here. Not sure `get_local_path` could be impacted in any way unless I'm missing something.

src/datachain/lib/tar.py
dberenbaum: Can you clarify? It should be a an instance of a subtype of `TarVFile`. Is there a better way you see to specify the type here?

src/datachain/lib/file.py
dberenbaum: Yup, it looks like we could. I opened to make #487 for that change, although since it's more "breaking," I'm not sure if/when it makes sense to merge.

src/datachain/lib/file.py
shcheklein: hmm, is it the same as:

```python
class Parent:
    def method1(self):
        print("Parent method1")

    def method2(self):
        print("Parent method2")
        self.method1()


class Child1(Parent):
    def method2(self):
        print("Child1 method2")
        super().method2()

class Child2(Parent):
    def method1(self):
        print("Child2 method1")


class Impl(Child1, Child2):
    pass

c = Impl()
c.method2()
```

it returns:

Child1 method2
Parent method2
Child2 method1

(so, even though we use super, it's important what is `self` in this case)?

dberenbaum: Sorry, guess I got confused here. I've reverted this change now.

src/datachain/lib/tar.py
shcheklein: yep, sorry, I think it goes into TypeVars, generics, or something. We want probably to bind type of the `file_cls` to the return type (to be the same). I'm not an expert here (I can take a look though), but probably it's achieved via TypeVars.

something like:

```python
T = TypeVar('T', bound=TarVFile)
def build_tar_member(parent: File, info: tarfile.TarInfo, file_cls: type[T] = TarVFile
) -> T:
   pass
```

It seems not possible though in mypy? https://github.com/python/mypy/issues/9003 ... feel free to disregard this.

dberenbaum: Got it. That makes sense, but it does appear it's not supported well enough in mypy. Your syntax above LGTM but fails to pass the mypy check, which doesn't seem to acknowledge that `TarVFile` is of type `T`.

shcheklein:
> We lose the ability to merge files and v-files in the same dataset

@rlamy could you clarify please? what is the scenario / use case for this?

rlamy:
> > We lose the ability to merge files and v-files in the same dataset
> 
> @rlamy could you clarify please? what is the scenario / use case for this?

An example would look like:
```python
base = DataChain.from_storage("s3://webdataset-stuff").gen(...)
extra = DataChain.from_storage("gs://raw-images").filter(C.path.glob("*.png").map(...)
combined = (base | extra).filter(...)
```

I agree that using `.location` is clumsy, but I think that the ability to combine data from different sources without copies is essential functionality. 

Now, `.location` is actually a work-around for the fact that we only support unions when the data models are exactly the same. So if we want to move on with this PR, I think we need to find a way to make `DataChain.union()` compatible with model inheritance (i.e. the schema of the union is the common base of the schemas, calling `collect()` on the union returns the original objects, etc.)

dberenbaum:
Closing for now since I don't see it progressing.

Issue #41, opened 2024-06-28
Decouple DataChain and DatasetQuery


dmpetrov:
It's needed for:
- a clean API - now these two concepts are too different to fit a single API
- a parts or DatasetQuery needs some redesign and should not be exposed to users
- depreciate old, flat signal based API: add_signals() and such

ToDo:
- [ ] DataChain should have params (and possible output values connected to metrics)
- [ ] DataChain should be reusable from another chains

rlamy:
Since we're now getting rid of most of the code that expected `DatasetQuery` to be user-facing, it might be time to tackle this. A simple first step should be to just remove the inheritance relationship and add a `_query: DatasetQuery` attribute to `DataChain` (and replace all super calls, etc.) That should make it easier to see which parts should be moved from `DatasetQuery` and what its exact role (and name!) should be.

dmpetrov:
Closing for now - we can return later when it's a priority

PR #459, opened 2024-09-19
Split DatasetQuery from DataChain

rlamy:
This removes the inheritance of DataChain from DatasetQuery and instead makes the latter an attribute of the former, as suggested in #41.

The de facto public properties and methods `name`, `version`, `session`, `union()`, `count()`, `exec()` and `chunk()` have been explicitly added to `DataChain`.

### DIFF
--- a/src/datachain/cli.py
+++ b/src/datachain/cli.py
@@ -770,7 +770,7 @@ def show(
     show_records(records, collapse_columns=not no_collapse)
     if schema and dataset_version.feature_schema:
         print("\nSchema:")
-        dc = DataChain(name=name, version=version, catalog=catalog)
+        dc = DataChain._create(name=name, version=version, catalog=catalog)
         dc.print_schema()
 
 
--- a/src/datachain/cli.py
+++ b/src/datachain/cli.py
@@ -12,7 +12,7 @@ from typing import TYPE_CHECKING, Optional, Union
 
 import shtab
 
-from datachain import utils
+from datachain import Session, utils
 from datachain.cli_utils import BooleanOptionalAction, CommaSeparatedArgs, KeyValueArgs
 from datachain.lib.dc import DataChain
 from datachain.telemetry import telemetry
@@ -770,7 +770,8 @@ def show(
     show_records(records, collapse_columns=not no_collapse)
     if schema and dataset_version.feature_schema:
         print("\nSchema:")
-        dc = DataChain._create(name=name, version=version, catalog=catalog)
+        session = Session.get(catalog=catalog)
+        dc = DataChain.from_dataset(name=name, version=version, session=session)
         dc.print_schema()
 
 
--- a/src/datachain/lib/dc.py
+++ b/src/datachain/lib/dc.py
@@ -286,12 +286,12 @@ class DataChain:
 
     @property
     def name(self) -> Optional[str]:
-        """Name of the chain."""
+        """Name of the underlying dataset, if there is one."""
         return self._query.name
 
     @property
     def version(self) -> Optional[int]:
-        """Version of the chain."""
+        """Version of the underlying dataset, if there is one."""
         return self._query.version
 
     def __or__(self, other: "Self") -> "Self":
--- a/src/datachain/lib/dc.py
+++ b/src/datachain/lib/dc.py
@@ -54,7 +54,6 @@ from datachain.query import Session
 from datachain.query.dataset import (
     DatasetQuery,
     PartitionByType,
-    detach,
 )
 from datachain.query.schema import DEFAULT_DELIMITER, Column, DatasetRow
 from datachain.sql.functions import path as pathfunc
@@ -159,7 +158,7 @@ class Sys(DataModel):
     rand: int
 
 
-class DataChain(DatasetQuery):
+class DataChain:
     """DataChain - a data structure for batch data processing and evaluation.
 
     It represents a sequence of data manipulation steps such as reading data from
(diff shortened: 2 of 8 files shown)


src/datachain/lib/pytorch.py
```diff
@@ -88,7 +88,9 @@ def __iter__(self) -> Iterator[Any]:
         if self.catalog is None:
             self.catalog = self._get_catalog()
         total_rank, total_workers = self.get_rank_and_workers()
-        ds = DataChain(name=self.name, version=self.version, catalog=self.catalog)
+        ds = DataChain._create(
```
shcheklein, line 91: why is it not complaining about accessing the protected method? can we avoid it here?

src/datachain/cli.py
```diff
@@ -770,7 +770,7 @@ def show(
     show_records(records, collapse_columns=not no_collapse)
     if schema and dataset_version.feature_schema:
         print("\nSchema:")
-        dc = DataChain(name=name, version=version, catalog=catalog)
+        dc = DataChain._create(name=name, version=version, catalog=catalog)
```
shcheklein, line 773: can we use `from_dataset` instead?

src/datachain/lib/dc.py
```diff
@@ -1278,6 +1317,10 @@ def _resolve(
 
         return ds
 
+    def union(self, other: "Self") -> "Self":
+        """Return the set union of the two datasets."""
```
shcheklein, line 1321: need proper doc string ? since it will become part of the docs?

src/datachain/lib/dc.py
```diff
@@ -1955,14 +1997,34 @@ def filter(self, *args) -> "Self":
...
-        return super().limit(n)
+        return self.clone(query=self._query.limit(n))
 
-    @detach
     def offset(self, offset: int) -> "Self":
         """Return the results starting with the offset row."""
-        return super().offset(offset)
+        return self.clone(query=self._query.offset(offset))
+
+    def count(self) -> int:
```
shcheklein, line 2010: are we removing them from the DatasetQuery?

shcheklein (commented):
Changes make sense to me. The biggest question - can we avoid `_create` method - it looks a bit weird and a bit like a hack.

src/datachain/cli.py
rlamy: `from_dataset` needs a session, not a catalog

src/datachain/lib/dc.py
rlamy: No, these should stay in DatasetQuery for now, I think DataChain shouldn't know anything about SQL.

src/datachain/lib/dc.py
```diff
@@ -290,18 +276,43 @@ def column(self, name: str) -> Column:
...
+    def __or__(self, other: "Self") -> "Self":
+        """Return `self.union(other)`."""
+        return self.union(other)
+
     def print_schema(self) -> None:
         """Print schema of the chain."""
         self._effective_signals_schema.print_tree()
 
-    def clone(self, new_table: bool = True) -> "Self":
+    def clone(self, *, query=None, new_table: bool = True) -> "Self":
```
rlamy, line 305: Since `clone()` is part of the public API, I'm actually not sure it's a good idea to coopt it like this.

shcheklein: I think that's fine for now

src/datachain/lib/dc.py
```diff
@@ -290,18 +276,43 @@ def column(self, name: str) -> Column:
...
         return c
 
+    @property
+    def session(self) -> Session:
+        """Session of the chain."""
+        return self._query.session
+
+    @property
+    def name(self) -> Optional[str]:
+        """Name of the chain."""
```
shcheklein, line 289: what is `name` of the chain? and version? are those dataset name / version or something else?

src/datachain/lib/dc.py
```diff
@@ -446,7 +457,24 @@ def from_dataset(
...
             ```
         """
-        return DataChain(name=name, version=version, session=session, settings=settings)
+        query = DatasetQuery(
+            name=name,
+            version=version,
+            session=session,
+            indexing_column_types=File._datachain_column_types,
+        )
+        telemetry.send_event_once("class", "datachain_init", name=name, version=version)
```
shcheklein, line 466: do we need to put it now into all other methods (e.g. from_json?)

src/datachain/lib/dc.py
```diff
@@ -765,9 +797,11 @@ def map(
         """
         udf_obj = self._udf_to_obj(Mapper, func, params, output, signal_map)
 
-        chain = self.add_signals(
-            udf_obj.to_udf_wrapper(),
-            **self._settings.to_dict(),
+        chain = self.clone(
```
shcheklein, line 800: just to make sure - does it mean that we do an extra table copy now?

dreadatour (commented):
Looks good to me 👍
Some tests (examples) are failing 👀

src/datachain/lib/dc.py
rlamy: No, for 2 reasons:
* `new_table` is ignored if you pass `query`.
* `clone()` never makes an actual copy anyway, it only sets an alias for the current query.

src/datachain/lib/dc.py
rlamy: Yes, it's the name/version of the dataset.

src/datachain/lib/dc.py
rlamy: Every `from_*` method ends up calling this, so I guess not.

src/datachain/lib/dc.py
shcheklein: let's please rephrase it then? wdyt?

dreadatour (approved):
Looks good to me, thank you! We definitely need this change 🥰
Not sure about telemetry part though, just have not enough expertise in it 👀

src/datachain/lib/dc.py
rlamy: Note that I'm not actually changing the logic, I'm just making the existing one more obvious. We should probably rethink it eventually but I guess it's good enough for now.

Issue #462, opened 2024-09-20
CI failure in examples/multimodal


rlamy:
For some reason, test_multimodal[examples/multimodal/wds_filtered.py] started failing today on Ubuntu ([example](https://github.com/iterative/datachain/actions/runs/10956268375/job/30421946162?pr=460)). It looks like that example stopped printing anything to stderr (`tqdm` output should go there if I'm not mistaken). I guess something got updated, but what?

PR #464, opened 2024-09-21
fix(tests): use unique dataset name to isolate from other examples

shcheklein:
Fixes https://github.com/iterative/datachain/issues/462

I think this is happening because of caching (e.g. we run one of the `wds` tests now before the other - order changed - and both have `save("wds")`, or some other caching is happening somewhere.

We probably should try to completely isolate tests by removing `.datachain` between them. It might slow down the whole thing though. @mattseddon was it intentional - that we are reusing the same `.datachain`? (is it true at all?)

### DIFF
--- a/examples/multimodal/wds_filtered.py
+++ b/examples/multimodal/wds_filtered.py
@@ -5,7 +5,7 @@ from datachain.lib.webdataset_laion import WDSLaion
 from datachain.sql import literal
 from datachain.sql.functions import array, greatest, least, string
 
-name = "wds"
+name = "wds_filtered"
 try:
     wds = DataChain.from_dataset(name=name)
 except datachain.error.DatasetNotFoundError:


dberenbaum:
I don't know whether it makes sense to keep the same name and try to read from an existing dataset in this example. However, it seems like the script still runs fine in #462 using the cached dataset, which seems to be the intent as currently written, but since it no longer has to create the dataset from scratch, tqdm is not called and nothing gets written to stderr. I think it would be better to drop the stderr assertion in the test regardless of this PR unless there's a reason it's helpful.

shcheklein:
>  which seems to be the intent as currently written

could you clarify please? I see that it's wrapped into `try`, so it's not exactly obvious that it expects the dataset to be cached (e.g. it can have to intents - run as a stand alone script + start from a cache version if it exists to save time). Then it's a matter of us to agree what is the primary intent we had in mind for that example and if expect users to run it stand alone - should we tests w/o cache?

dberenbaum:
Yup, I'm fine with whatever you want to do with this example. 

More generally, I think it will save more trouble for other examples to clarify whether stderr must contain some output. It's not clear to me if there is a reason for this assertion.

mattseddon:
> We probably should try to completely isolate tests by removing .datachain between them. It might slow down the whole thing though. @mattseddon was it intentional - that we are reusing the same .datachain? (is it true at all?)

Yes, it was intentional to have examples running against the same DB as this is closer to what a user will do/experience. I don't see a problem with the `wds_filtered`  example reading from an already-created dataset. The main reason that I kept it was because this could be a useful pattern for users to implement.

> More generally, I think it will save more trouble for other examples to clarify whether stderr must contain some output. It's not clear to me if there is a reason for this assertion.

I added this assertion because it is unclear whether datachain (partly due to using tqdm) will write to stdout or stderr but it was a mistake to implement both assertions. A more useful assertion would be `assert completed_process.stdout or completed_process.stderr` as we expect that SOMETHING is written to the user's screen during each example.

mattseddon:
I opened https://github.com/iterative/datachain/pull/468

PR #467, opened 2024-09-22
Removed `Catalog.open_object()` and refactor method to return file object from row

ilongin:
`Cattalog.open_object()` is not needed and not used any more.

Also refactoring method to return file signals from row to return actual `File` objects and instead returning always first found file subobject,  now returns file object by name

### DIFF
--- a/src/datachain/catalog/catalog.py
+++ b/src/datachain/catalog/catalog.py
@@ -80,6 +80,7 @@ if TYPE_CHECKING:
     )
     from datachain.dataset import DatasetVersion
     from datachain.job import Job
+    from datachain.lib.file import File
 
 logger = logging.getLogger("datachain")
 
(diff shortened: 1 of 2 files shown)


src/datachain/catalog/catalog.py
```diff
@@ -1438,26 +1439,34 @@ def get_file_signals(
...
+        if signal_name not in schema.get_signals(File):
+            raise RuntimeError(
+                f"File signal with path {signal_name} not found in ",
+                f"dataset {dataset_name}@v{dataset_version} signals schema",
+            )
 
-        file_signals = self.get_file_signals(dataset_name, dataset_version, row)
-        if not file_signals:
-            raise RuntimeError("Cannot open object without file signals")
+        prefix = signal_name.replace(".", DEFAULT_DELIMITER) + DEFAULT_DELIMITER
```
dreadatour, line 1460: Should we move this to some helper method? I've seen same line in another parts of the code.

dreadatour (approved):
Looks good to me, as long as it will be tested in both the CLI and SaaS environments and merged along with the related Studio PR 👍

src/datachain/catalog/catalog.py
ilongin: I've just remove other place where it was used as that whole method is not used any more so it's removed. I will see if it makes sense to put it to utils anyway though

PR #468, opened 2024-09-22
assert each example has some output instead of stdout and stderr

mattseddon:
Closes https://github.com/iterative/datachain/issues/462

This PR removes a dud assertion that each example should have some `stderr` output and replaces it with the assertion that each example will have some user-facing output (either stdout or stderr, doesn't matter which). It also removes the reliance on `glob` to order the examples for us.

More discussion can be found in #464

### DIFF
--- a/tests/examples/test_examples.py
+++ b/tests/examples/test_examples.py
@@ -6,34 +6,42 @@ from typing import Optional
 
 import pytest
 
-get_started_examples = [
-    filename
-    for filename in glob.glob("examples/get_started/**/*.py", recursive=True)
-    # torch-loader will not finish within an hour on Linux runner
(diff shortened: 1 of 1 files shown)


dberenbaum (approved):

tests/examples/test_examples.py
```diff
@@ -57,8 +65,8 @@ def smoke_test(example: str, env: Optional[dict] = None):
         print(e.stderr.decode("utf-8"))
         pytest.fail("subprocess returned a non-zero exit code")
 
-    assert completed_process.stdout
-    assert completed_process.stderr
+    example_has_some_output = bool(completed_process.stdout or completed_process.stderr)
```
shcheklein, line 68: should we always expect something in stdout (a stronger condition) ... at least for now. AFAIU we don't have (correctly working) examples where it's only stdout

PR #469, opened 2024-09-23
only use free runners for examples

mattseddon:
Checking whether or not we can drop `ubuntu-latest-4-cores` usage after #409


mattseddon:
still runs out of space

PR #472, opened 2024-09-23
remove unused is_job_result field

skshetry:


0x2b3bfa0 (approved):

dreadatour (approved):
Thank you! ❤️

`is_job_result` field have default value in DB, so it should be OK, but it is worth mentioning since we don't have migrations implemented yet it may be needed to remove `.datachain` directory after switching datachain to version with this PR merged.

skshetry:
Removing field is okay IIUC. We won't select for those columns. But this is a breaking change.

dreadatour:
> Removing field is okay IIUC. We won't select for those columns. But this is a breaking change.

Agree 🤗 Just wanted to highlight this, because it may be important next time and we don't have tests for these kind of cases.

PR #480, opened 2024-09-26
Parquet Import+Export with SignalSchema

dtulga:
This enables parquet import and export including the DataChain SignalSchema, which allows DataChain models to be exported to parquet files, then re-imported later on with the same data structure intact. This also chunks parquet export to allow for exporting parquet files larger than memory. (Parquet import should already support this.) The DataChain SignalSchema is stored in the parquet file's metadata (not separately) so this data schema will be accessible even when sharing only the parquet files.

This has been tested locally and is part of #91


shcheklein:
Thanks @dtulga ! Nice work. Just a few questions to clarify the implementation / expectations:

- What happens if some DataChain has two (not nested) models, e.g. schema is `File, Chunks`. Will it be handled (sequence of top level models). I hope my question makes sense :)
- When we read back parquet file and models are not registered in the `ModelRegistry` - are we returning dicts or we construct a Pydantic class dynamically (I think similar to what we do in the regular `from_parquet` flow?). Asking this primarily, because, trying to understand how will people use it. I can restore the file, okay, but how now I can do some operations if don't have a model class defined. Do we expect such use case?

dtulga:
Yes, multiple top-level models work as well, I added a test case for this too, just to be sure.

And yes, this creates a dynamic Pydantic class if the model class that is referenced in the dataset's SignalSchema cannot be found when restoring. This is done within the SignalSchema deserialization here: https://github.com/iterative/datachain/blob/main/src/datachain/lib/signal_schema.py#L279 and the model is created here: https://github.com/iterative/datachain/blob/main/src/datachain/lib/signal_schema.py#L86 This can happen if a dataset is exported then imported somewhere else where a custom model is not defined, so yes, this is an expected use case and should work. There are tests for this dynamic model creation in the SignalSchema tests, for example here: https://github.com/iterative/datachain/blob/main/tests/unit/lib/test_signal_schema.py#L350

shcheklein (approved):

PR #482, opened 2024-09-27
Implement chain group_by

dreadatour:
Implement chain `group_by`:

```
chain.group_by(
    total_size=func.sum("file.size"),
    cnt=func.count(),
    partition_by="path_ext",
)
```


src/datachain/lib/dc.py
```diff
@@ -1005,6 +1006,72 @@ def select_except(self, *args: str) -> "Self":
             query=self._query.select(*columns), signal_schema=new_schema
         )
 
+    def group_by(
+        self,
+        *,
+        partition_by: Union[str, Sequence[str]],
+        **kwargs: Func,
+    ) -> "Self":
```
dreadatour, line 1014: Need to simplify this method and move code to some helper functions.

src/datachain/query/dataset.py
```diff
@@ -956,6 +956,24 @@ def q(*columns):
...
+            for c in [*self.group_by, *self.cols]
+        ]
+        if not cols:
+            cols = subquery.c
+
+        return sqlalchemy.select(*cols).select_from(subquery).group_by(*self.group_by)
+
+
 @frozen
 class GroupBy(Step):
```
dreadatour, line 978: Need to remove this outdated code.

src/datachain/data_storage/sqlite.py
```diff
@@ -771,6 +771,14 @@ def copy_table(
...
         progress_cb: Optional[Callable[[int], None]] = None,
     ) -> None:
+        if len(query._group_by_clause) > 0:
+            select_q = query.with_only_columns(
+                *[c for c in query.selected_columns if c.name != "sys__id"]
+            )
+            q = table.insert().from_select(list(select_q.selected_columns), select_q)
+            self.db.execute(q)
+            return
+
```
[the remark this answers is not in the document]
dreadatour: Will do it in follow-up PRs.

src/datachain/lib/convert/sql_to_python.py
```diff
@@ -4,15 +4,11 @@
...
-        try:
-            type_ = sql_exp.type.python_type
-            if type_ == Decimal:
-                type_ = float
-        except NotImplementedError:
-            type_ = str
-        res[name] = type_
-
-    return res
+def sql_to_python(sql_exp: ColumnElement) -> Any:
```
dreadatour, line 7: This function was updated to be consistent with [python_to_sql](https://github.com/iterative/datachain/blob/main/src/datachain/lib/convert/python_to_sql.py#L37) function (both take single element). Also it was used [in one place only](https://github.com/iterative/datachain/blob/main/src/datachain/lib/signal_schema.py#L493)  and for single element, no need to have `dict` as input param and output result. Also it is now used to detect `partition_by` type in `group_by` in case function is used.

src/datachain/lib/dc.py
```diff
@@ -981,10 +983,9 @@ def distinct(self, arg: str, *args: str) -> "Self":  # type: ignore[override]
         row is left in the result set.
 
         Example:
-        ```py
-         dc.distinct("file.parent", "file.name")
-        )
```
dreadatour, line 986: Mistype — this bracket is not needed 👀

src/datachain/lib/func.py
```diff
@@ -0,0 +1,50 @@
...
+from sqlalchemy import func as sa_func
+
+from datachain.query.schema import DEFAULT_DELIMITER
+from datachain.sql import functions as dc_func
+
+if TYPE_CHECKING:
+    from datachain import DataType
+
+
+class Func:
```
dreadatour, line 12: Using custom `Func` class because I can not make it works in all possible cases with default SQLAlchemy functions. Sometimes type detection is not working (for example, for `sum` function) and it fails to run or to store result in DB with correct type.

I will appreciate any comments/recommendations on this 🙏

src/datachain/lib/signal_schema.py
```diff
@@ -490,7 +497,7 @@ def mutate(self, args_map: dict) -> "SignalSchema":
                 new_values[name] = args_map[name]
             else:
                 # adding new signal
-                new_values.update(sql_to_python({name: value}))
+                new_values[name] = sql_to_python(value)
```
dreadatour, line 500: See comment about `sql_to_python` above.

src/datachain/lib/signal_schema.py
```diff
@@ -534,12 +541,12 @@ def _build_tree(
             for name, val in values.items()
         }
 
-    def get_flat_tree(self) -> Iterator[tuple[list[str], type, bool, int]]:
+    def get_flat_tree(self) -> Iterator[tuple[list[str], DataType, bool, int]]:
```
dreadatour, line 544: Fix type here: `DataType` should be used because it is returned from `self.tree`. Otherwise we will have mypy warnings.

src/datachain/sql/functions/array.py
```diff
@@ -51,8 +51,16 @@ class avg(GenericFunction):  # noqa: N801
     inherit_cache = True
 
 
+class group_concat(GenericFunction):  # noqa: N801
```
dreadatour, line 54: This function is not implemented in ClickHouse, so we need custom function, see related Studio PR.

src/datachain/data_storage/sqlite.py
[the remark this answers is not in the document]
mattseddon: [Q] No update required on the Studio side?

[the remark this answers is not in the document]
dreadatour: No update required on the Studio side, SQLite implementation differs because of performance optimizations.

src/datachain/lib/func.py
dreadatour: Don't like this part. Decided to revisit this one more last time.

src/datachain/query/dataset.py
```diff
@@ -957,24 +957,24 @@ def q(*columns):
...
-        self, query_generator: QueryGenerator, temp_tables: list[str]
-    ) -> StepResult:
-        query = query_generator.select()
-        grouped_query = query.group_by(*self.cols)
+        subquery = query.subquery()
 
-        def q(*columns):
-            return grouped_query.with_only_columns(*columns)
+        cols = [
+            subquery.c[str(c)] if isinstance(c, (str, C)) else c
```
dtulga, line 973: Any reason why `query.selected_columns` won't work here instead of a subquery?

dtulga (approved):
LGTM, with just a couple of questions.

src/datachain/lib/dc.py
```diff
@@ -1010,6 +1011,96 @@ def select_except(self, *args: str) -> "Self":
...
+
+        schema_columns = self.signals_schema.db_columns_types()
+        schema_fields: dict[str, DataType] = {}
+
+        # validate partition_by columns and add them to the schema
+        partition_by_columns: list[Union[Column, GenericFunction]] = []
+        for col in partition_by:
+            if isinstance(col, GenericFunction):
+                partition_by_columns.append(col)
+                schema_fields[col.name] = sql_to_python(col)
```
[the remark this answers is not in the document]
dreadatour: Nice catch, thank you! 👍

src/datachain/query/dataset.py
dreadatour: The reason is if we are using `LIMIT` in the query, `group_by` will only work within the limited amount of columns. That's why I've added subquery here, to be sure limit will not affect aggregation.

src/datachain/lib/func.py
dreadatour: Wasn't able to win in this fight with sqlalchemy. I can not find a way to get list of columns with their types from sqlalchemy functions (tried to use [traverse](https://docs.sqlalchemy.org/en/20/core/visitors.html#sqlalchemy.sql.visitors.ExternalTraversal.traverse) but with no luck).

src/datachain/query/dataset.py
mattseddon: Are you sure that limit is applied before group by in SQL? I thought it was applied after group by. I.e 

```SQL

-- create
CREATE TABLE EMPLOYEE_SALES (
  empId INTEGER NOT NULL,
  name TEXT NOT NULL,
  dept TEXT NOT NULL,
  sale_value FLOAT NOT NULL
);

-- insert
INSERT INTO EMPLOYEE_SALES VALUES (0001, 'Clark', 'Sales', 10);
INSERT INTO EMPLOYEE_SALES VALUES (0001, 'Clark', 'Sales', 10);
INSERT INTO EMPLOYEE_SALES VALUES (0001, 'Clark', 'Sales', 300);
INSERT INTO EMPLOYEE_SALES VALUES (0001, 'Clark', 'Sales',400);
INSERT INTO EMPLOYEE_SALES VALUES (0001, 'Clark', 'Sales',1);
INSERT INTO EMPLOYEE_SALES VALUES (0001, 'Clark', 'Sales',5);
INSERT INTO EMPLOYEE_SALES VALUES (0002, 'Dave', 'Accounting', 20);
INSERT INTO EMPLOYEE_SALES VALUES (0003, 'Ava', 'Sales',400);

-- fetch 
SELECT 
  count(*) as count, 
  sum(sale_value) as total_sales, 
  name 
FROM EMPLOYEE_SALES group by name limit 3;
```
yields:

```console
Output:

1|400.0|Ava
6|726.0|Clark
1|20.0|Dave
```

dreadatour: Let me check again, but I've added subquery because of the limit issue (after using `.show()` chain method). I'll be back with the answer.

dreadatour: I can not find the way to reproduce issue with limit in current implementation, but removing subquery causes another issues with SQL functions and labels.
My suggestion is to leave it as is for now, because it is another issue — to get rid of subqueries. Right now they are actively used in all steps, even [basic select](https://github.com/iterative/datachain/blob/main/src/datachain/query/dataset.py#L709-L719).

I have some work done on this, removed almost all subqueries and in the end there is non performance boost without subqueries in both CLI and SaaS, but there was some issues with tests. Also @rlamy was looking into this. I can create a follow-up issue to solve subqueries issue.

rlamy (approved):

Issue #40, opened 2024-04-04
pre_fetch option in additional to cache for lib.File


dmpetrov:
We need to download items in async mode before processing them:
```python
chain.settings(pre_fetch=2, cache=True, parallel=True).gen(laion=process_webdataset(spec=WDSLaion))
```

- `pre_fetch` this should enable async file download (per thread) for a given limit of files (like, pre_fetch=10). Like pre_fetch in pytorch datasets. Default should be pre_fetch=2

OUTDATED:
- consider introducing pre_fetch=0 that returns Stream() descriptor with direct access to storage and no caching.

```python
ds.generate(WebDataset(spec=WDSLaion), parallel=4, cache=True, pre_fetch=10)
```

shcheklein:
Depends on the file API refactoring. Moving indexing to the app level. For now moving back to backlog.

shcheklein:
Since we are done with indexing more or less, moving it back to the ready stage cc @rlamy . Might still depend on some work that Ronan is doing now with decoupling datasetquery and datachain.

One of the use cases I have atm is:

- A lot of video files in the bucket
- Cache can help, but won't scale (we would need to pretty much download all at once). But it's a fine starting point.
- Pre-fetch would help if we have also an option to cleanup files (if cache is disabled)

------------------

One thing that is a bit annoying is that some tools (OpenCV) seems to require a local path. Yes, cache helps in that case and pre-fetch can help - but both require downloading the whole file, while for some operations I just need some header. If someone has ideas how that can be improved - let me know. Is there a way to create file-like-looking object but that is a stream from the cloud underneath?

rlamy:
Some notes:
* In order to implement this, we need to insert logic similar to `DatasetQuery.extract()` before (or maybe in) `udf.run()`.
* Fetching should be implemented, or at least controlled in some way, by the model. For instance, if we have an `ArrowRow` (which contains a `File` but doesn't inherit from it), we should fetch the row, not the whole file.

This means that `udf.run()` should receive model instances, not raw DB rows, which requires some refactoring...

shcheklein:
> This means that udf.run() should receive model instances, not raw DB rows, which requires some refactoring...

where do we receive raw DB rows there? (I wonder if this related or should be taken into account - https://github.com/iterative/studio/issues/10531#issuecomment-2379390308 )

rlamy:
After probably too much refactoring, I can confirm that this can be implemented inside `udf.run()` which means that:
[...]

PR #483, opened 2024-09-27
Merge datachain.query.udf into datachain.lib.udf and clean-up

rlamy:
This is the first part of the refactoring for #40.


shcheklein (approved):
Thanks!

PR #487, opened 2024-09-30
drop file.location

dberenbaum:
Drops `File.location` field after https://github.com/iterative/datachain/pull/446. See https://github.com/iterative/datachain/pull/446/files#r1780264304.


dberenbaum:
Looks like this would require someone to take over and do the studio part

dberenbaum:
Closing this per #446.

PR #488, opened 2024-10-01
Adding Streaming CSV Export

dtulga:
This adds a direct `to_csv` function for DataChain to export to CSV (or TSV, etc.) files. This also uses a results iterator and streaming CSV writer, so this should not be limited by memory size, unlike converting the results to pandas then converting to CSV. This has been tested locally and is part of #91


shcheklein (approved):

ilongin (approved):
LGTM

PR #491, opened 2024-10-02
Don't allow mappers to skip rows

rlamy:
AFAICT that obscure feature was introduced in https://github.com/iterative/dvcx/pull/863. It's undocumented (and even contradicts the docs) and untested, so I assume it's unused and broken.


shcheklein (approved):

dreadatour (approved):
It was tested [here](https://github.com/iterative/dvcx/pull/863/files#diff-1ee91a4c5257ef32b59fed5b22bbe2c8896549ac307c49f54ed411f227e6af6eR1203), but this test was deleted and does not exists anymore.
I assume it is ok to remove this line if it is contradicts with docs since all our tests pass.

ilongin (approved):

PR #492, opened 2024-10-04
Making anon access to buckets default

dtulga:
This makes anon access default for buckets (s3, gcs, etc.) when no credentials/custom configuration is provided. As especially on Google Cloud Storage, not specifying any credentials (and also not specifying anonymous access) causes a lookup for default credentials, which includes attempting to access a Google Cloud Metadata Server, which on Studio takes a very long time (20+ seconds on my local Studio and 10+ seconds on the production Studio) and leads to warnings such as: 

```
WARNING google.auth.compute_engine._metadata Compute Engine Metadata server unavailable on attempt 5 of 5. Reason: HTTPConnectionPool(host='metadata.google.internal', port=80): Max retries exceeded with url: /computeMetadata/v1/instance/service-accounts/default/?recursive=true (Caused by NameResolutionError("<urllib3.connection.HTTPConnection object at 0x16206bda0>: Failed to resolve 'metadata.google.internal' ([Errno 8] nodename nor servname provided, or not known)"))
```

Note that this does not change the behavior if credentials are provided in `client_config`. This improves image preview loading times on my local Studio from 20+ seconds to less than one second for an uncached image.

This has been tested locally, with my local Studio, and fixes iterative/studio#10743

### DIFF
--- a/src/datachain/catalog/catalog.py
+++ b/src/datachain/catalog/catalog.py
@@ -554,7 +554,9 @@ class Catalog:
         self.metastore = metastore
         self._warehouse = warehouse
         self.cache = DataChainCache(datachain_dir.cache, datachain_dir.tmp)
-        self.client_config = client_config if client_config is not None else {}
+        self.client_config = (
+            client_config if client_config is not None else {"anon": True}
+        )
         self._init_params = {
             "cache_dir": cache_dir,
             "tmp_dir": tmp_dir,
--- a/src/datachain/lib/dc.py
+++ b/src/datachain/lib/dc.py
@@ -392,6 +392,7 @@ class DataChain:
         object_name: str = "file",
         update: bool = False,
         anon: bool = False,
+        client_config: Optional[dict[str, Any]] = None,
     ) -> "Self":
         """Get data from a storage as a list of file with all file attributes.
         It returns the chain itself as usual.
@@ -412,7 +413,10 @@ class DataChain:
         """
         file_type = get_file_type(type)
 
-        client_config = {"anon": True} if anon else None
+        if anon:
+            if client_config is None:
+                client_config = {}
+            client_config["anon"] = True
 
         session = Session.get(session, client_config=client_config, in_memory=in_memory)
 
--- a/src/datachain/lib/listing.py
+++ b/src/datachain/lib/listing.py
@@ -30,7 +30,7 @@ def list_bucket(uri: str, cache, client_config=None) -> Callable:
     """
 
     def list_func() -> Iterator[File]:
-        config = client_config or {}
+        config = client_config if client_config is not None else {"anon": True}
         client = Client.get_client(uri, cache, **config)  # type: ignore[arg-type]
         _, path = Client.parse_url(uri)
         for entries in iter_over_async(client.scandir(path.rstrip("/")), get_loop()):
@@ -80,7 +80,7 @@ def parse_listing_uri(uri: str, cache, client_config) -> tuple[str, str, str]:
     """
     Parsing uri and returns listing dataset name, listing uri and listing path
     """
-    client_config = client_config or {}
+    client_config = client_config if client_config is not None else {"anon": True}
     client = Client.get_client(uri, cache, **client_config)
     storage_uri, path = Client.parse_url(uri)
     telemetry.log_param("client", client.PREFIX)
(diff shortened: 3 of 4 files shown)


0x2b3bfa0 (approved):
Is there any way of setting `anon=True` in a single place, or the least amount possible of places? 😅 It feels that this pull request goes a bit against the spirit of https://github.com/iterative/dvcx/pull/1411.

0x2b3bfa0:
Tangentially related to https://github.com/iterative/studio/issues/10235

src/datachain/catalog/catalog.py
```diff
@@ -554,7 +554,9 @@ def __init__(
         self.metastore = metastore
         self._warehouse = warehouse
         self.cache = DataChainCache(datachain_dir.cache, datachain_dir.tmp)
-        self.client_config = client_config if client_config is not None else {}
+        self.client_config = (
+            client_config if client_config is not None else {"anon": True}
```
shcheklein, line 558: hmm, either I'm missing something or would it set `anon` to True even if run it locally and credentials are provided via some env or profile config?

0x2b3bfa0: I think that you're right, and that's precisely the point of this pull request. 🤔 Whether it's a good idea or not UX-wise, is a different topic. 😅 

Maybe we can set anonymous-by-default in Studio instead of here, just to fix the image loading bug?

shcheklein (commented):
I'm not sure this is safe tbh (we provide credentials via env vars, probably means in some case client config could be empty, means that we'll set anon True despite having credentials. AFAIR in case of AWS for example it leads to an error in some cases (I can try to reproduce that). I'm not sure about other clouds.

Unless I'm missing something.

If that's ^^ the case I think we just need to finally find a way to disable those managed credentials on pods cc @0x2b3bfa0 . E.g. we can set an env var to a non-existent host instead of the default GC hostname. Let's discuss this.

0x2b3bfa0:
> _If that's ^^ the case I think we just need to finally find a way to disable those managed credentials on pods_

Or — easier — move this patch into the Studio codebase, where it makes more sense than in the command-line tool. In Studio we never have environmental authentication: it's either `client_config` credentials or anonymous access.

> _E.g. we can set an env var to a non-existent host instead of the default GC hostname. Let's discuss this._

Not that easy, sadly. 😅 Let's discuss on https://github.com/iterative/studio/issues/10235 if needed.

shcheklein:
> Not that easy, sadly. 😅 Let's discuss on https://github.com/iterative/studio/issues/10235 if needed.

can you provide details in that ticket? what have you tried so far, etc (I don't see any updates there - so it is hard to say anything). E.g. did we try env vars to block the host?

shcheklein:
> Or — easier — move this patch into the Studio codebase, where it makes more sense than in the command-line tool. In Studio we never have environmental authentication: it's either client_config credentials or anonymous access.

yes, that would require some Studio-specific code / mode? can be done probably, don't know

0x2b3bfa0:
> _yes, that would require some Studio-specific code / mode? can be done probably, don't know_

Yes! It would look exactly like [this](https://github.com/iterative/studio/blob/cb537e416e65f2da2b949e122201c6f836e6d174/backend/dqlapp/views.py#L539) but... in the right place (apparently we're missing it somewhere)

dtulga:
It sounds like the consensus is to move this change into being Studio-only, as that should work, although will require some more debugging/analysis to figure out how to change Studio to do this properly, and may require some anon-default mode here too, perhaps set with an environment variable so that anon is not default for cli users. This should be doable, but will require more debugging/analysis and maybe multiple PR changes. I'll mark this as a draft while I work on that alternative Studio-specific solution instead.

0x2b3bfa0:
@dtulga, are you aware of https://github.com/iterative/studio/pull/10747, by the way? 🤔 If that fix didn't work, is probably because it's using `catalog` before setting the `client_config` and that's where the issue happens.

dtulga:
> By the way, @dtulga, are you aware of [iterative/studio#10747](https://github.com/iterative/studio/pull/10747), by the way? 🤔 If that fix didn't work, is probably because it's using `catalog` before setting the `client_config` and that's where the issue happens.

I'll look into it, thanks for the heads up!

dtulga:
After further analysis / debugging, this issue https://github.com/iterative/studio/issues/10743 is fixed by https://github.com/iterative/studio/pull/10747 as I no longer have this issue on my local Studio. (It appears that it doesn't matter if the `catalog` is used before setting `client_config` as long as `anon` is True.) And as such, closing this PR as it is no longer needed.

PR #494, opened 2024-10-07
Make datachain queries atomic when exception occurs

amritghimire:
With this change, whenever any error or exception is raised when running
the script, this will revert all datachain version and datasets created
during the script.

Studio PR: https://github.com/iterative/studio/pull/10740
Studio Issue: https://github.com/iterative/studio/issues/9875

### DIFF
--- a/src/datachain/catalog/catalog.py
+++ b/src/datachain/catalog/catalog.py
@@ -988,6 +988,14 @@ class Catalog:
         schema = {
             c.name: c.type.to_dict() for c in columns if isinstance(c.type, SQLType)
         }
+
+        job_id = job_id or os.getenv("DATACHAIN_JOB_ID")
+        if not job_id:
+            from datachain.query.session import Session
+
+            session = Session.get(catalog=self)
+            job_id = session.name
+
         dataset = self.metastore.create_dataset_version(
             dataset,
             version,
--- a/src/datachain/query/session.py
+++ b/src/datachain/query/session.py
@@ -1,5 +1,8 @@
 import atexit
+import logging
+import os
 import re
+import sys
 from typing import TYPE_CHECKING, Optional
 from uuid import uuid4
 
@@ -8,6 +11,8 @@ from datachain.error import TableMissingError
 
 if TYPE_CHECKING:
     from datachain.catalog import Catalog
+
+logger = logging.getLogger("datachain")
 
 
 class Session:
@@ -35,6 +40,7 @@ class Session:
 
     GLOBAL_SESSION_CTX: Optional["Session"] = None
     GLOBAL_SESSION: Optional["Session"] = None
+    ORIGINAL_EXCEPT_HOOK = None
 
     DATASET_PREFIX = "session_"
     GLOBAL_SESSION_NAME = "global"
@@ -67,6 +73,9 @@ class Session:
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
+        if exc_type:
+            self._cleanup_created_versions(self.name)
+
         self._cleanup_temp_datasets()
         if self.is_new_catalog:
             self.catalog.metastore.close_on_exit()
@@ -87,6 +96,21 @@ class Session:
         # suppress error when metastore has been reset during testing
         except TableMissingError:
             pass
+
+    def _cleanup_created_versions(self, job_id: str) -> None:
+        versions = self.catalog.metastore.get_job_dataset_versions(job_id)
+        if not versions:
+            return
+
+        datasets = {}
+        for dataset_name, version in versions:
+            if dataset_name not in datasets:
+                datasets[dataset_name] = self.catalog.get_dataset(dataset_name)
+            dataset = datasets[dataset_name]
+            logger.info(
+                "Removing dataset version %s@%s due to exception", dataset_name, version
+            )
+            self.catalog.remove_dataset_version(dataset, version)
 
     @classmethod
     def get(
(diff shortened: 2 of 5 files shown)


dreadatour (approved):
Looks good to me, as long as it’s been tested on both the CLI and SaaS.

tests/test_atomicity.py
```diff
@@ -0,0 +1,58 @@
...
+tests_dir = os.path.dirname(os.path.abspath(__file__))
+
+python_exc = sys.executable or "python3"
+
+E2E_STEP_TIMEOUT_SEC = 90
+
+
+@pytest.mark.e2e
+@pytest.mark.xdist_group(name="tmpfile")
+def test_atomicity_feature_file(tmp_dir, catalog_tmpfile):
```
shcheklein, line 19: can we test this w/o creating a subprocess please?

use Session as a context for example ... (we should try to use these heavy tests as a measure of last resort if possible)

src/datachain/catalog/catalog.py
```diff
@@ -988,6 +988,14 @@ def create_new_dataset_version(
         schema = {
             c.name: c.type.to_dict() for c in columns if isinstance(c.type, SQLType)
         }
+
+        job_id = job_id or os.getenv("DATACHAIN_JOB_ID")
+        if not job_id:
+            from datachain.query.session import Session
+
+            session = Session.get(catalog=self)
```
shcheklein, line 996: it seems a bit strange that we are initializing Session here? why is it needed ... can we not do this (we are spreading logic for Session logic management into lower lever functions which is not good)

here for example, reading this is not enough - there are some assumptions about job_id and or Session that are not clear and it's not clear why this code exits at all tbh

amritghimire: Session.get gives us the global session if it already exists without initializing.

tests/test_atomicity.py
amritghimire: The except hook cannot be tested outside of subprocess because it applies to the python script in general and I wanted to test if it works even when the uncatched exception is handled properly.

src/datachain/query/session.py
```diff
@@ -114,9 +139,23 @@ def get(
                 in_memory=in_memory,
             )
             cls.GLOBAL_SESSION = cls.GLOBAL_SESSION_CTX.__enter__()
+
             atexit.register(cls._global_cleanup)
+            cls.ORIGINAL_EXCEPT_HOOK = sys.excepthook
+            sys.excepthook = cls.except_hook
```
shcheklein, line 145: why does `atexit` not enough?

is it needed for Studio?

amritghimire: atexit doesn't work when unknown exceptions are raised.

src/datachain/catalog/catalog.py
shcheklein: why do we do this only if job_id is not initialized?

what about non-global sessions (query might belong to it)?

src/datachain/query/session.py
shcheklein: what is an "unknown exception"? I see the code above:

```
        if exc_type:
            self._cleanup_created_versions(self.name)
```

that looks like it does the same? no?

src/datachain/catalog/catalog.py
amritghimire: If we already have job id, it is the scenario of Studio. In such case, we will have the job id to identify the dataset versions created during this job. 

I couldn't find a way to find the dataset versions created by a session. So, I used the existing job id field to identify the versions created during current run. We dont have a way to get current local session context in meta store.

src/datachain/query/session.py
amritghimire: This is when the session is used as a context manager only.

src/datachain/catalog/catalog.py
shcheklein: okay, so we have 3 cases:

- job_id explicitly provided
- env var
- one from the global Session

do we need all three?
can we make it always specified explicitly?
what if session is local / created by a user? will it still work?

my point here is that this code looks too complicated, how can we isolate if from here? move it to the session itself, etc, etc? are there any reasonable ways? :)

src/datachain/query/session.py
shcheklein: okay, I see ... it was not obvious that we want to cleanup on a non-global context managed sessions only (vs let users deal with it, e.g. if some exceptions are expected and handled), but it makes sense I guess ... let's keep it!

tests/scripts/feature_class_exception.py
```diff
@@ -0,0 +1,24 @@
...
+logging.basicConfig(level=logging.INFO)
+
+
+class Embedding(BaseModel):
+    value: float
+
+
+ds_name = "feature_class_error"
+ds = (
+    DataChain.from_storage("gs://dvcx-datalakes/dogs-and-cats/")
```
shcheklein, line 17: can we at least use something simpler - like `from_values` - so that we don't do expensive networking, indexing, etc?

src/datachain/catalog/catalog.py
amritghimire: > my point here is that this code looks too complicated, how can we isolate if from here? move it to the session itself, etc, etc? are there any reasonable ways? :)

Here, we are just fetching the session and using its job id. IMO, it is not that complicated or there is anything we can move to the session. We are attaching the job id if we don't have it passed with us either by argument or by env variable.

Issue #318, opened 2024-08-19
Replace deprecated listing / indexing with new `DataChain.from_storage()`


ilongin:
In related PR https://github.com/iterative/datachain/pull/294 we are going to have listing logic inside of `DataChain.from_storage()` itself.

We should replace old listing that is being called from CLI and maybe other places with  `DataChain.from_storage()`.
There is a lot of tests around listing / indexing and we should refactor them as well if needed

As a follow up for this we should remove old legacy listing codebase (maybe it's better to do this in actual separate issue / PR). We should also remove buckets and partials

Note that we also need to replace `Catalog.ls_storages` to use new listing datasets as bucket table will be removed, as well as partials

- [ ] Refactor `datachain.Dataset.is_bucket_listing()`  to not use old listing check and maybe remove this method altogether as low level `Dataset` class should not know about `LISTING_PREFIX` and similar higher level abstractions

ilongin:
I think this should be prioritized as it's sometimes hard to refactor / change codebase as this old indexing part needs to be adopted. It would been much easier if it's just refactored to "new" indexing, not to mention it can only be used for CLI operations .. .e.g when we call `Catalog.index(...)` listing that's been created cannot be used in `DataChain` methods

shcheklein:
Agreed. Moved it to the ready column - let's keep simplifying things there.

PR #496, opened 2024-10-07
Refactor file system source and path

ilongin:
Before, when listing local FS, we had root of the FS always set for `source` field, e.g `file:///` and the rest was in `path`. Idea behind this was to utilize partial indexing and remove obsolete `add_storage` method.
Now, when we are moving away from partial and bucket tables and moving indexing / listing to application level, we don't need this kind of setup as it was not convenient or intuitive.

With this changes when someone does `DataStorage.from_storage("file:///home/ivan/animals/")` we will have something like this in the listing table:
| source | path |
| --- | --- |
| file:///home/ivan/animals | cats.jpg |
| file:///home/ivan/animals | dogs/dog.jpg |

This also adds re-indexing check in `.from_storage()` to avoid re-indexing if not needed.
Old code related to that root of the FS mentioned at beginning of description was removed.

Note that some tests were skipped as it was not possible to refactor them without diving deep into other important task that will be worked on soon as well: https://github.com/iterative/datachain/issues/318

### DIFF
--- a/src/datachain/client/local.py
+++ b/src/datachain/client/local.py
@@ -26,30 +26,6 @@ class FileClient(Client):
     def url(self, path: str, expires: int = 3600, **kwargs) -> str:
         raise TypeError("Signed urls are not implemented for local file system")
 
-    '''
-    @classmethod
-    def get_uri(cls, name) -> StorageURI:
-        """
-        This returns root of FS as uri, e.g
-            Linux & Mac : file:///
-            Windows: file:///C:/
-        """
-        return StorageURI(Path(name).as_uri())
-    '''
-
-    @staticmethod
-    def root_dir() -> str:
-        """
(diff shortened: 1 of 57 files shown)


src/datachain/client/local.py
```diff
@@ -75,23 +57,20 @@ def path_to_uri(cls, path: str) -> str:
 
     @classmethod
     def split_url(cls, url: str) -> tuple[str, str]:
-        """
```
amritghimire, line 78: Should we reword the doc instead of removing this?

ilongin: Just added it in parent class.

rlamy (approved):
LGTM. Removing so many special cases for `cloud_type == "file"` in the tests is nice!

Issue #477, opened 2024-09-24
Clarify ordering semantics


rlamy:
We need to specify clearly how `.order_by()` interacts with other methods.

For instance, in SQL, the order of the results from a SELECT query is undefined unless there is an ORDER BY clause. Currently, `chain.order_by("a").select("a")` generates SQL that look like `SELECT a FROM (SELECT a FROM some_table ORDER BY a)`, which is therefore technically unordered, even though it'll usually return sorted results. While it might be fine to say that `.select()` doesn't preserve ordering, it seems obvious that `chain.order_by("a").collect("a")` should return the "a" column in sorted order, and `chain.collect("a")` is currently implemented as `chain.select("a").collect()`...

dmpetrov:
Good point. We need to keep information about the order and use it in the following statements when it makes sense.
We should use the underlying IDs for that.

mattseddon:
> Good point. We need to keep information about the order and use it in the following statements when it makes sense.
We should use the underlying IDs for that.

It does not seem possible to generate a sequential list of IDs in BigQuery (more details [here](https://github.com/iterative/studio/issues/10635#issuecomment-2381829809)). I am currently stuck on how we can provide deterministic results from all chains in an efficient way.

With that said; Is there a clear use case for ordering an intermediate step in a chain? I can find superficial examples in tests such as this:

```python
    def name_len(path):
        return (len(posixpath.basename(path)),)

    DataChain.from_storage(path, session=session).order_by("file.path").map(
        name_len, params=["file.path"], output={"name_len": int}
    ).save(ds_name)
```

but do we support cumulative values being calculated across rows, wouldn't this be done via `group_by`? What is the use case that we are trying to support? Can we justify the complexity added to support this? Note: I can understand `order_by` preceding `limit` (this should be handled by the generated SQL anyway).

> For instance, in SQL, the order of the results from a SELECT query is undefined unless there is an ORDER BY clause.

We could patch `order_by` before `select` or `select_except` in the following way:

```python
    @detach
    def select(self, *args, **kwargs) -> "Self":
        named_args = [v.label(k) for k, v in kwargs.items()]
        query = self.clone()

        last_step = query.steps[-1] if query.steps else None

        query.steps.append(SQLSelect((*args, *named_args)))

        if isinstance(last_step, SQLOrderBy):
            query.steps.append(SQLOrderBy(last_step.args))

        return query
```

[...]

PR #502, opened 2024-10-08
Fix 'create_result_query' ordering

dreadatour:
Tiny fix for unused code found while working on another PR.

### DIFF
--- a/src/datachain/query/dataset.py
+++ b/src/datachain/query/dataset.py
@@ -666,13 +666,6 @@ class RowGenerator(UDFStep):
     def create_result_query(
         self, udf_table, query: Select
     ) -> tuple[QueryGeneratorFunc, list["sqlalchemy.Column"]]:
-        if not query._order_by_clauses:
-            # if we are not selecting all rows in UDF, we need to ensure that
-            # we get the same rows as we got as inputs of UDF since selecting
-            # without ordering can be non deterministic in some databases
-            c = query.selected_columns
-            query = query.order_by(c.sys__id)
-
         udf_table_query = udf_table.select().subquery()
         udf_table_cols: list[sqlalchemy.Label[Any]] = [
             label(c.name, c) for c in udf_table_query.columns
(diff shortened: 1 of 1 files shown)


src/datachain/query/dataset.py
```diff
@@ -666,13 +666,6 @@ def create_udf_table(self, query: Select) -> "Table":
     def create_result_query(
         self, udf_table, query: Select
     ) -> tuple[QueryGeneratorFunc, list["sqlalchemy.Column"]]:
-        if not query._order_by_clauses:
-            # if we are not selecting all rows in UDF, we need to ensure that
-            # we get the same rows as we got as inputs of UDF since selecting
-            # without ordering can be non deterministic in some databases
-            c = query.selected_columns
-            query = query.order_by(c.sys__id)
```
dreadatour, line 674: This `query` variable is not used after reassigning.

src/datachain/query/dataset.py
```diff
@@ -682,7 +675,15 @@ def q(*columns):
...
             # Columns for the generated table.
             cols = [c for c in udf_table_cols if c.name in names]
-            return sqlalchemy.select(*cols).select_from(udf_table_query)
+            res = sqlalchemy.select(*cols).select_from(udf_table_query)
+
+            if query._order_by_clauses:
+                # if we are not selecting all rows in UDF, we need to ensure that
+                # we get the same rows as we got as inputs of UDF since selecting
+                # without ordering can be non deterministic in some databases
+                res = res.order_by(udf_table_query.c.sys__id)
```
dreadatour, line 684: Move `order_by` here, [same](https://github.com/iterative/datachain/blob/main/src/datachain/query/dataset.py#L633-L636) as in `UDFSignal.create_result_query` method.

mattseddon: I am not against this change but PTAL at this: https://github.com/iterative/datachain/issues/477#issuecomment-2387274525. It looks unlikely we'll be able to support this behaviour in BQ.

ilongin (approved):

src/datachain/query/dataset.py
```diff
@@ -682,7 +675,15 @@ def q(*columns):
             names = {c.name for c in columns}
             # Columns for the generated table.
             cols = [c for c in udf_table_cols if c.name in names]
-            return sqlalchemy.select(*cols).select_from(udf_table_query)
+            res = sqlalchemy.select(*cols).select_from(udf_table_query)
+
+            if query._order_by_clauses:
```
mattseddon, line 680: qq - how come you are switching from `if not query._order_by_clauses:` to `if query._order_by_clauses:`?

dreadatour:
Closing this issue because we decided to do not preserve ordering within dataset.

src/datachain/query/dataset.py
dreadatour: To be honest I am not sure why it was `not` there, may be it is because of previous refactoring attempts.

PR #503, opened 2024-10-08
UDF refactoring

rlamy:
Simplify the logic in `udf.run()` by cleanly separating the 4 different code flows. Part of #40.


dreadatour (approved):
Love it! 😻 Awesome refactoring, thank you! 🙏

PR #504, opened 2024-10-09
Add a way to persist dataset even if exception is thrown

### LINKED ISSUE #500: Add a way to persist dataset even if exception is thrown
After https://github.com/iterative/datachain/pull/494 it would be great to have a way to persist datasets no matter what.

E.g. I'm iterating now on DCLM, here is some code:

```python
import io

import pyarrow as pa
import zstandard as zstd

from datachain.lib.arrow import schema_to_output
from datachain.lib.data_model import dict_to_data_model


def extract(file: File) -> Iterator[tuple[File, str]]:
    with file.open() as f:
        dctx = zstd.ZstdDecompressor()
        stream_reader = dctx.stream_reader(f)
        text_stream = io.TextIOWrapper(stream_reader, encoding="utf-8")
        for line in text_stream:
            yield file, line


datasets = [ds.name for ds in DataChain.datasets().collect("dataset")]

if "index" not in datasets:
   DataChain.from_storage("s3://commoncrawl/contrib/datacomp/DCLM-baseline/").save("index")

if "raw_text" not in datasets:
   (
      DataChain.from_dataset("index")
         .settings(cache=True)
         .limit(1)
         .gen(extract, output={"file": File, "json": dict})
         .save("raw_text")
   )

def infer_schema(json: dict) -> pa.Schema:
    return pa.Table.from_pylist([json]).schema


dc = DataChain.from_dataset("raw_text")
json = list(dc.limit(1).collect("json"))[0]
schema =  infer_schema(json)
output = schema_to_output(schema, None)
model = dict_to_data_model("meta", output)


def explode(json: dict):
  return model.model_validate(json)

dc.map(meta=explode).save("exploded")
```

it is failing somewhere at the very end (on `dc.map(meta=explode).save("exploded")`). But in this case I have:

```python
if "index" not in datasets:
   DataChain.from_storage("s3://commoncrawl/contrib/datacomp/DCLM-baseline/").save("index")
```

to clearly iterate faster. I would love those datasets to be saved (I mean the `index` and `raw_text).

amritghimire:
Refactoring upon the changes at #494 from the comments, this introduces the way to allow a way to persist dataset even if exception is thrown. With this change, the cleanup is now done at the active context. So, we can use the session context as checkpoints now.  Example implementation is as: 

https://github.com/iterative/datachain/blob/029e677e596c84bb8c2ecb56603d164adde03fe8/tests/scripts/feature_class_exception.py#L1-L34

In this file, `local_test_datachain` should only persist since it is the only dataset where the exception doesn't occur in current context. 

Closes #500

### DIFF
--- a/src/datachain/catalog/catalog.py
+++ b/src/datachain/catalog/catalog.py
@@ -72,7 +72,6 @@ from datachain.utils import (
 from .datasource import DataSource
 
 if TYPE_CHECKING:
-    from datachain import Session
     from datachain.data_storage import (
         AbstractIDGenerator,
         AbstractMetastore,
@@ -917,7 +916,6 @@ class Catalog:
         create_rows: Optional[bool] = True,
         validate_version: Optional[bool] = True,
         listing: Optional[bool] = False,
-        session: Optional["Session"] = None,
     ) -> "DatasetRecord":
         """
         Creates new dataset of a specific version.
@@ -965,7 +963,6 @@ class Catalog:
             query_script=query_script,
             create_rows_table=create_rows,
             columns=columns,
-            session=session,
         )
 
     def create_new_dataset_version(
@@ -982,10 +979,7 @@ class Catalog:
         script_output="",
         create_rows_table=True,
         job_id: Optional[str] = None,
-        session: Optional["Session"] = None,
     ) -> DatasetRecord:
-        from datachain.query.session import Session
-
         """
         Creates dataset version if it doesn't exist.
         If create_rows is False, dataset rows table will not be created
@@ -1014,9 +1008,6 @@ class Catalog:
             table_name = self.warehouse.dataset_table_name(dataset.name, version)
             self.warehouse.create_dataset_rows_table(table_name, columns=columns)
             self.update_dataset_version_with_warehouse_info(dataset, version)
-
-        session = Session.get(session=session, catalog=self)
-        session.add_dataset_version(dataset, version)
 
         return dataset
 
@@ -1180,15 +1171,12 @@ class Catalog:
         version: int,
         target_dataset: DatasetRecord,
         target_version: Optional[int] = None,
-        session: Optional["Session"] = None,
     ) -> DatasetRecord:
         """
         Registers dataset version of one dataset as dataset version of another
         one (it can be new version of existing one).
         It also removes original dataset version
         """
-        from datachain.query.session import Session
-
         target_version = target_version or target_dataset.next_version
 
         if not target_dataset.is_valid_next_version(target_version):
@@ -1223,9 +1211,6 @@ class Catalog:
             preview=dataset_version.preview,
             job_id=dataset_version.job_id,
         )
-
-        session = Session.get(session=session, catalog=self)
-        session.add_dataset_version(target_dataset, target_version)
 
         # to avoid re-creating rows table, we are just renaming it for a new version
         # of target dataset
--- a/src/datachain/catalog/catalog.py
+++ b/src/datachain/catalog/catalog.py
@@ -72,6 +72,7 @@ from datachain.utils import (
 from .datasource import DataSource
 
 if TYPE_CHECKING:
+    from datachain import Session
     from datachain.data_storage import (
         AbstractIDGenerator,
         AbstractMetastore,
@@ -916,6 +917,7 @@ class Catalog:
         create_rows: Optional[bool] = True,
         validate_version: Optional[bool] = True,
         listing: Optional[bool] = False,
+        session: Optional["Session"] = None,
     ) -> "DatasetRecord":
         """
         Creates new dataset of a specific version.
@@ -963,6 +965,7 @@ class Catalog:
             query_script=query_script,
             create_rows_table=create_rows,
             columns=columns,
+            session=session,
         )
 
     def create_new_dataset_version(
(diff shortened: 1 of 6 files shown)


tests/scripts/feature_class_exception.py
```diff
@@ -1,11 +1,26 @@
...
+
+try:
+    with Session("local_failure"):
+        # A datachain created in local context. This will not persist since error occur in local context.
+        DataChain.from_values(key=["a", "b", "c"]).save("local_test_datachain_v2")
+        raise ValueError("Local failure class")
+except ValueError:
+    pass
+
+# This will persist, however.
```
shcheklein, line 23: update the comment? It won't persist, right?

src/datachain/catalog/catalog.py
```diff
@@ -1218,6 +1218,10 @@ def register_dataset(
             preview=dataset_version.preview,
             job_id=dataset_version.job_id,
         )
+
+        session = Session.get_current_context()
```
shcheklein, line 1222: `session` can be also passed to DataChain explicitly (check `from_storage`, for example) ... it's probably expected that we pass it directly into all the datacnain factories even when we use contextmanager ... not sure `get_current_context` is the right approach ... at least it might be changing the assumption ... can we check docs / existing tests please?

amritghimire: Updated the logic along with example in the file where the variable session is used. PTAL.

src/datachain/query/session.py
```diff
@@ -64,25 +65,34 @@ def __init__(
 
         session_uuid = uuid4().hex[: self.SESSION_UUID_LEN]
         self.name = f"{name}_{session_uuid}"
-        self.job_id = os.getenv("DATACHAIN_JOB_ID") or str(uuid4())
         self.is_new_catalog = not catalog
         self.catalog = catalog or get_catalog(
             client_config=client_config, in_memory=in_memory
         )
+        self.versions: list[tuple[DatasetRecord, int]] = []
```
dreadatour, line 72: Variable name `versions` is too broad, imo. Should we call it `datasets`, `dataset_versions` or may be `temp_datasets`? 👀

src/datachain/query/session.py
```diff
@@ -64,25 +65,34 @@ def __init__(
...
         self._cleanup_temp_datasets()
         if self.is_new_catalog:
             self.catalog.metastore.close_on_exit()
             self.catalog.warehouse.close_on_exit()
             self.catalog.id_generator.close_on_exit()
 
+        if Session.SESSION_CONTEXTS:
+            Session.SESSION_CONTEXTS.pop()
+
+    def add_created_versions(self, dataset: "DatasetRecord", version: int) -> None:
```
dreadatour, line 93: Simplify to `add_dataset` or `add_dataset_version`?

src/datachain/query/session.py
```diff
@@ -125,33 +129,34 @@ def get(
 
         Parameters:
             session (Session): Optional Session(). If not provided a new session will
-                    be created. It's needed mostly for simplie API purposes.
-            catalog (Catalog): Optional catalog. By default a new catalog is created.
+                    be created. It's needed mostly for simple API purposes.
```
dreadatour, line 132: 👍

dreadatour (approved):
Looks good to me 👍

src/datachain/lib/dc.py
```diff
@@ -1992,6 +1992,7 @@ def from_records(
                 if signal_schema
                 else None
             ),
+            session=session,
```
shcheklein, line 1995: I think we have a few other places where create_dataset is called (and now session is not provided). Some of them might require the session object to be passed?

src/datachain/catalog/catalog.py
```diff
@@ -1178,12 +1180,15 @@ def register_dataset(
         version: int,
         target_dataset: DatasetRecord,
         target_version: Optional[int] = None,
+        session: Optional["Session"] = None,
```
shcheklein, line 1183: Let's think about this. Catalog belong to session, and now we pass back session to catalog - it's like a circular dependency. Something is wrong with the design on these objects to my mind.

Can we call `session.add_dataset_version(target_dataset, target_version)` in places where we run `create_dataset`? (to avoid catalog operating on the Session)

src/datachain/lib/dc.py
amritghimire: The only remaining place where create_dataset is called is `enlist_source` but we don't have  option to specify session variable there. So, we will select the active session there.

shcheklein: I see .. how about pull_dataset? same?

shcheklein (approved):
Approving to move forward, but this design Session -> Catalog -> Sessions looks a bit weird to me.

src/datachain/lib/dc.py
amritghimire: Yes, we cannot pass session there too. The place where we have added session is the direct path from save function, where we save a new dataset and that only seem to have the ability to pass session.  That being said, if we are not passing session, we will still fetch the active context of the session.

amritghimire: Checked for usage of pull_dataset and enlist_source:

- https://github.com/search?q=org%3Aiterative+pull_dataset&type=code
- https://github.com/search?q=org%3Aiterative%20enlist_source&type=code

They seem to be used in tests and CLI only, where we won't need to pass session.

src/datachain/catalog/catalog.py
amritghimire: Thanks @shcheklein. Moved it to dataset query back. It is working as expected.

amritghimire:
> Approving to move forward, but this design Session -> Catalog -> Sessions looks a bit weird to me.

Thanks. Modified the chain logic.

PR #505, opened 2024-10-12
drop DatasetRow outdated class

shcheklein:
Fixes issue with a customer scripts, e.g. we do this now and it fails w/o it:

```python
tcrops = tcrops.rename(columns={'source': 'crop_source'}) # Named "crop_source" to avoid "Source" column duplication
```

I think we don't need this anymore.

Closes https://github.com/iterative/dvcx/issues/1583.


ilongin (approved):
It seems ok. We have some other outdated classes like this one which will be removed soon as well

PR #507, opened 2024-10-13
remove ordering by sys.id

mattseddon:
Related to https://github.com/iterative/datachain/issues/477 and https://github.com/iterative/studio/issues/10635#issuecomment-2381829809 & https://github.com/iterative/studio/issues/10635#issuecomment-2406225017

All the context for this change is in https://github.com/iterative/datachain/issues/477 but the tl;dr is:

This behaviour is currently undocumented, untested and misunderstood. There is no way to support this behaviour with BigQuery and we have to find another way around datasets appearing ordered. One suggestion is to save the order along with the dataset information in the metastore (perhaps save the appropriate select statement as this would be extendable in the future).

### DIFF
--- a/src/datachain/catalog/catalog.py
+++ b/src/datachain/catalog/catalog.py
@@ -1325,8 +1325,6 @@ class Catalog:
         if offset:
             q = q.offset(offset)
 
-        q = q.order_by("sys__id")
-
         return q.to_db_records()
 
     def signed_url(self, source: str, path: str, client_config=None) -> str:
--- a/src/datachain/data_storage/warehouse.py
+++ b/src/datachain/data_storage/warehouse.py
@@ -215,10 +215,6 @@ class AbstractWarehouse(ABC, Serializable):
         limit = query._limit
         paginated_query = query.limit(page_size)
 
-        if not paginated_query._order_by_clauses:
-            # default order by is order by `sys__id`
-            paginated_query = paginated_query.order_by(query.selected_columns.sys__id)
-
         results = None
         offset = 0
         num_yielded = 0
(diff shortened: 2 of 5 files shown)


ilongin:
We had a lot of discussions about this through the history of dvcx / datachain. Basically it boils down to this:
1. Preserving order by making sure inserted rows are in correct order is not how things are **usually** done. That's why we don't have this feature in BigQuery and probably in other DBs. Usually DB don't allow you to "mess" into how they internally store rows.
2. If you want to fetch rows in some order, just use `order by` in query....as you said, we can save ordering in some metadata and use it when doing query.

Why we still end up with 1) ? ... because 2) is not really working well with billions of rows and column oriented DBs like Clickhouse and BigQuery. Having to do `order by` on each select means whole DB must be reshuffled which in Clickhouse also leads to OOM errors if I remember correctly (we need to double check this though). 

Anyway, I'm ok to approve this if product agrees, as this means we are removing feature which was requested explicitly by product some time ago.

mattseddon:
> Preserving order by making sure inserted rows are in correct order is not how things are usually done. That's why we don't have this feature in BigQuery and probably in other DBs. Usually DB don't allow you to "mess" into how they internally store rows.

A small note: Even if we insert rows in the correct order in BQ, there is no way to guarantee the order they are returned when we pull them out. We cannot generate a sequential ID in the same way we do for SQLite and ClickHouse AND ordering is not guaranteed unless an order by statement is provided.

dreadatour (approved):
Finally! Can’t wait for this to be merged; it’ll make things simpler from the code cleanliness perspective and faster in terms of performance.

Might be controversial in terms of user experience, but on the other hand, it is kind of intuitive that you should not rely on default ordering when you’re dealing with big distributed data.

If anyone needs ordering, they can use order_by before using the related query.

shcheklein (approved):
Do we need sys__id at all after this decision (removing ordering)?

mattseddon:
> Do we need sys__id at all after this decision (removing ordering)?

It is used in distributed processing.

PR #513, opened 2024-10-16
Refactor the configuration parsing

amritghimire:
As a part of [#10774 ](https://github.com/iterative/studio/issues/10774), this introduces a process to save the configuration
in local, system and global configuration.

The precedence of the level are as:
- system
- global
- local

Local configuration overrides global and so on.

This borrows the logic of how configuration is managed in DVC.

### DIFF
--- a/src/datachain/config.py
+++ b/src/datachain/config.py
@@ -1,6 +1,6 @@
 from collections.abc import Mapping
 from contextlib import contextmanager
-from typing import Optional
+from typing import Optional, Union
 
 from tomlkit import TOMLDocument, dump, load
 
@@ -36,9 +36,6 @@ class Config:
     def init(self):
         d = DataChainDir(self.get_dir(self.level))
         d.init()
-
-        with open(d.config, "w"):
-            return self
 
     def load_one(self, level: Optional[str] = None) -> TOMLDocument:
         config_path = DataChainDir(self.get_dir(level)).config
@@ -125,10 +122,10 @@ class Config:
         return remote_conf
 
 
-def merge(into, update):
+def merge(into: Union[TOMLDocument, dict], update: Union[TOMLDocument, dict]):
     """Merges second dict into first recursively"""
     for key, val in update.items():
         if isinstance(into.get(key), dict) and isinstance(val, dict):
-            merge(into[key], val)
+            merge(into[key], val)  # type: ignore[arg-type]
         else:
             into[key] = val


src/datachain/config.py
```diff
@@ -1,62 +1,131 @@
...
+
+        remote_type = remote_conf.get("type")
+        if remote_type not in ("local", "http"):
+            raise Exception(
+                f'config section remote.{remote} must have "type" with one of: '
+                '"local", "http"'
+            )
+
+        if remote_type == "http":
+            for key in ["url", "username", "token"]:
```
dreadatour, line 110: Just wonder, should we also have `team_id` as a part of remote config? User can be a member of many teams in Studio 🤔 Or `team_id` will be required as a (required? optional?) param in API calls?

src/datachain/config.py
```diff
@@ -1,62 +1,131 @@
...
+                try:
+                    remote_conf[key]
+                except KeyError:
+                    raise Exception(
+                        f"config section remote.{remote} of type {remote_type} "
+                        f"must contain key {key}"
+                    ) from None
```
dreadatour, line 117: Just a comment: Studio URL may be optional 👀

src/datachain/config.py
```diff
@@ -1,62 +1,131 @@
...
 
 
-def read_config(datachain_root: str) -> Optional["TOMLDocument"]:
-    config_path = os.path.join(datachain_root, "config")
-    try:
-        with open(config_path, encoding="utf-8") as f:
-            return load(f)
-    except FileNotFoundError:
-        return None
+class Config:
```
dreadatour, line 10: Looks good to me, it would be great to add tests for this class.

src/datachain/config.py
```diff
@@ -1,62 +1,131 @@
...
+    SYSTEM_LEVELS = ("system", "global")
+    LOCAL_LEVELS = ("local",)
 
+    # In the order of precedence
+    LEVELS = SYSTEM_LEVELS + LOCAL_LEVELS
+
+    CONFIG = "config"
```
dreadatour, line 17: Not sure if I understand the reason behind this attribute 👀 Can't find any usage of this.

src/datachain/config.py
```diff
@@ -1,62 +1,131 @@
...
+
+    def __init__(
+        self,
+        level: Optional[str] = None,
```
dreadatour, line 21: Just wonder, will it be better from typing prospective to have `level` as enum, let's say? 👀

dreadatour (approved):
Overall, it looks good to me as long as it has been thoroughly tested 👍
It would be great to add more tests, and I’ve left a couple of minor comments below.

src/datachain/config.py
amritghimire: I was not sure if we will continue using this TBH. Keeping this part as it is for now.

src/datachain/config.py
amritghimire: This may be part of future changes when implementing some functionality with the token access.

src/datachain/config.py
amritghimire: Added tests and fixes in https://github.com/iterative/datachain/pull/514

Issue #227, opened 2024-08-02
Introduce window-function


dmpetrov:
I have a dataset with column `Class` which takes one of categorical values like Dog, Cat etc...  

- [ ] I need to select a subset of the records with 5 records in each class.
- [ ] I need the same but the subset has to be randomized.

This is how it looks like in IBIS (with randomization):
```python
# Create a random column for ordering
random_col = table.mutate(random_val=ibis.random())

# Define a window function partitioned by 'Class' and ordered by the random column
window = win.Window(partition_by='Class', order_by=random_col.random_val)

# Add a row number column based on the window function
ranked_table = random_col.mutate(row_number=ibis.row_number().over(window))
```

dreadatour:
Merged. Will continue in the follow-up issues https://github.com/iterative/datachain/issues/522 and https://github.com/iterative/datachain/issues/523

PR #515, opened 2024-10-16
Window functions

dreadatour:
Implement window-functions, see https://github.com/iterative/datachain/issues/227

Implemented functions:
- row_number
- rank
- dense_rank
- first

Let me know if we need to implement more functions 🙏

#### Example

(see [test](https://github.com/iterative/datachain/pull/515/files#diff-7fa3fd7accc76fc7bbc0c5e9b696be6e54cdcc5126b8c1f682b0f75cba1f95baR2449-R2530))

```python
DataChain.from_values(
    col1=["a", "a", "b", "b", "b", "c"],
    col2=[1, 2, 3, 4, 5, 6],
).mutate(
    row_number=func.row_number().over(func.Window(partition_by="col1", order_by="col2"))
)
```

Result:
```
 col1 | col2 | row_number
------+------+------------
 a    | 1    | 1
 a    | 2    | 2
 b    | 3    | 1
 b    | 4    | 2
 b    | 5    | 3
 c    | 6    | 1
```


src/datachain/lib/func/__init__.py
```diff
@@ -1,14 +1,20 @@
 from .aggregate import any_value, avg, collect, concat, count, max, min, sum
-from .func import Func
+from .func import Func, Window
+from .window import dense_rank, first, rank, row_number
 
 __all__ = [
     "Func",
+    "Window",
```
shcheklein, line 7: it should be `window` (all lower case). Even in IBIS, I think it's lower case?:

```
w = ibis.window(group_by="one", preceding=5, following=5)
expr = t.mutate(group_demeaned=t.two - t.two.mean().over(w))
ibis.to_sql(expr)
```
in datachain upper case looks weird

(event though it might be a class underneath - it doesn't matter in this case I think)

src/datachain/lib/func/func.py
```diff
@@ -49,15 +75,31 @@ def get_result_type(self, signals_schema: "SignalSchema") -> "DataType":
...
-            col = Column(self.db_col, python_to_sql(col_type))
+            col = Column(self.db_col, sql_type)
             func_col = self.inner(col)
         else:
             func_col = self.inner()
 
+        if self.is_window:
+            if not self.window:
+                raise DataChainColumnError(
+                    str(self.inner), "Window function requires window"
```
shcheklein, line 90: should it be something like `over() requires a window` ?

src/datachain/lib/func/func.py
```diff
@@ -9,18 +11,44 @@
     from datachain.lib.signal_schema import SignalSchema
 
 
+class Window:
+    def __init__(self, *, partition_by: str, order_by: str, desc: bool = False) -> None:
+        self.partition_by = partition_by
```
shcheklein, line 16: do we cover a scenario like this with this:

```
I need to select a subset of the records with 5 records in each class.
```


?

src/datachain/lib/dc.py
```diff
@@ -1086,7 +1086,7 @@ def mutate(self, **kwargs) -> "Self":
         Example:
         ```py
          dc.mutate(
-                newkey=Column("oldkey")
+            newkey=Column("oldkey")
```
shcheklein, line 1089: should we expand docs here a bit?

src/datachain/lib/func/func.py
```diff
@@ -9,18 +11,44 @@
...
         self.inner = inner
         self.col = col
         self.result_type = result_type
         self.is_array = is_array
+        self.is_window = is_window
+        self.window = window
+
+    def over(self, window: Window) -> "Func":
+        if not self.is_window:
+            raise DataChainColumnError(
```
shcheklein, line 40: why is it a ColumnError?

dtulga (approved):
LGTM, thanks for adding these!

src/datachain/lib/dc.py
dreadatour: I think we have explanation here ([above](https://github.com/iterative/datachain/blob/main/src/datachain/lib/dc.py#L1082-L1090)), I just fix the indentation.

src/datachain/lib/func/func.py
dreadatour: Done 👍

src/datachain/lib/func/func.py
dreadatour: Done 👍

src/datachain/lib/func/func.py
dreadatour: Yes, added test for this scenario: https://github.com/iterative/datachain/pull/515/files#diff-b40ca373e4b7558183b740420a85bc3d95c55a4b3049addcbbc1cf0b04460eecR1421-R1463

src/datachain/lib/dc.py
shcheklein: I mean - do we need to add something about `over` and `window` to docs somewhere?

dreadatour: Oh, that's a good question. Definitely, but if we'll add it here, in `mutate` function docstrings it can become pretty big. Do you have an idea on where to put these docs? Or it is ok to put it here?

tests/func/test_datachain.py
```diff
@@ -1366,3 +1366,98 @@ def file_info(file: File) -> DataModel:
...
+    from datachain import func
+
+    session = cloud_test_catalog.session
+    src_uri = cloud_test_catalog.src_uri
+
+    class FileInfo(DataModel):
+        path: str = ""
+        name: str = ""
+
+    def file_info(file: File) -> DataModel:
```
shcheklein, line 1431: DataModel -> FileInfo

dreadatour: 🫣 Thank you! 🙏

dreadatour: Fixed 🙏

src/datachain/lib/dc.py
dreadatour: I've updated `mutate` function docstring, please, take a look?

src/datachain/lib/func/func.py
shcheklein: I'm not sure we need a test for this tbh (if it doesn't cover any additional window functionality it becomes a functional test / not a unit one, not sure if there is value in this - your call).

What I mean is that we might want to have a simpler syntax for this? It seems that tools like IBIS have `range` that can be passed inside `window`. Can we explore that option? Would it be solving this type of scenarios in a simpler way? wdyt?

dreadatour: Got it. Sounds like a good idea to me.
May be in follow-up issue (I can create one)?

src/datachain/lib/dc.py
```diff
@@ -1069,13 +1069,22 @@ def mutate(self, **kwargs) -> "Self":
            Filename:    name(), parent(), file_stem(), file_ext()
            Array:       length(), sip_hash_64(), euclidean_distance(),
                         cosine_distance()
+           Window:      row_number(), rank(), dense_rank()
```
shcheklein, line 1072: I guess we'll need sum, avg and all those usual suspects as well? (can be a followup)

tests/unit/lib/test_datachain.py
```diff
@@ -2443,3 +2443,106 @@ def test_group_by_error(test_session):
...
+
+    window = func.window(partition_by="col1", order_by="col2")
+
+    dc = DataChain.from_values(
+        col1=["a", "a", "b", "b", "b", "c"],
+        col2=[1, 2, 3, 4, 5, 6],
+        session=test_session,
+    )
+
+    with pytest.raises(DataChainColumnError, match="Window function requires window"):
```
shcheklein, line 2544: not sure I understand this error tbh ... can you explain it please?

tests/unit/lib/test_datachain.py
```diff
@@ -2443,3 +2443,106 @@ def test_group_by_error(test_session):
...
+        dc.mutate(first=func.first("col2"))
+
+    with pytest.raises(DataChainParamsError, match="Over requires a window function"):
```
shcheklein, line 2547: same here - also not sure I understand this error ... (probably I asked to change it? but I don't see a sense now in it :) ) - window is passed right? so why is this an error at all?

src/datachain/lib/dc.py
dreadatour: Right! group_by functions needs to be mentioned as well.

tests/unit/lib/test_datachain.py
shcheklein: ah, first is a window function ... let me think ...

src/datachain/lib/dc.py
shcheklein, line 1072: add `first`?

shcheklein: ah, and I see that we support actually a whole set of functions?

tests/unit/lib/test_datachain.py
dreadatour: Yes, no `over` is applied here.

tests/unit/lib/test_datachain.py
dreadatour: `sum` is not a window function, and we do apply `over` to it.

tests/unit/lib/test_datachain.py
shcheklein: Can we make this error look like:

`Window function `first()` requires `over()` clause with a window spec`

or something like this

tests/unit/lib/test_datachain.py
shcheklein: ah, okay ... any reason it's not a window function?

I would change the error to something like "`sum()` doesn't support window (`over()`)"

src/datachain/lib/dc.py
dreadatour: Ok, so we have group_by functions docs, it is in `group_by` chain method.
I have updated the docs for this: https://github.com/iterative/datachain/pull/515/files#diff-0b1cfd3d89c057cc57811e07eb20298eeaaee60c54cc66e5128f618d144f42eeR1002-R1021

dreadatour: Updated!
I don't think we support a whole set of functions, it is differs for SQLite and ClickHouse and other databases.

Please, let me know if I need to add more functions, I also can do it in the follow-up PRs.

dreadatour: Also updated docs for all functions: https://github.com/iterative/datachain/pull/515/files#diff-c9b298e2d861e6d90972609a2bfcf0043b470ed946543dbf8bd1358cfafd1de7 + https://github.com/iterative/datachain/pull/515/files#diff-3abf137f75868839662512636c606626c4a44f52cafb75ca63b36d00ede0bb75R24-R53

Thank you for a reminder, this is very important 🙏

tests/unit/lib/test_datachain.py
dreadatour: Fixed. Thank you, this was a hardest part, to write a good error message 😅

tests/unit/lib/test_datachain.py
dreadatour: Fixed!

src/datachain/lib/func/func.py
dreadatour: Created a follow-up issue: https://github.com/iterative/datachain/issues/522

shcheklein: thanks!

tests/unit/lib/test_datachain.py
shcheklein: thanks!

tests/unit/lib/test_datachain.py
shcheklein: thanks!

src/datachain/lib/dc.py
shcheklein: what is the usual set of functions? do they all support sum, avg - that type of things? we can create a followup ticket for this I think. Thanks!

shcheklein (approved):
Excellent work, thanks @dreadatour !. Thanks for improving docs along the way.

src/datachain/lib/dc.py
dreadatour: Created a follow-up ticket for that: https://github.com/iterative/datachain/issues/523

Issue #370, opened 2024-08-28
Export to huggingface hub


dberenbaum:
Add method to export datachain to huggingface hub.

Part of #236

dtulga:
Closing this as done with #516 - I will be using #236 to track further research / investigation on structured exports.

Issue #236, opened 2024-08-05
Load from / to Hugging Face ?


lhoestq:
Hi ! I'm Quentin from HF :)

Congrats on the release ! The API is concise and easy, it will be useful to many people

I was wondering if you had plans to support reading / writing from HF datasets ?

If you use fsspec it might work out of the box though, using `hf://` paths (and if you have the `huggingface_hub` lib installed)

dmpetrov:
@lhoestq thank you for the asking this!
Do yo have a specific use case in mind? What exact dataset, and what you want to do with the dataset next?

Let us evaluate this. It seems straightforward and our other project DVC already supports this. We will get back soon.

lhoestq:
Cool ! The main uses cases I imagine are transforming rows / generating more rows with a LLM of existing datasets

dtulga:
This article may be helpful for a future structured export function: https://huggingface.co/docs/datasets/en/repository_structure

shcheklein:
I think this can be closed for now. We have basic `hf://` support. Not sure if we want more complicated integration with Datasets. Can be a separate item.

PR #516, opened 2024-10-16
Adding FSSpec Export for CSV and Parquet

dtulga:
This adds the ability to export / upload to FSSpec filesystems in `to_csv` and `to_parquet` - such as exporting directly to S3 or Hugging Face. This is done by passing the relevant url path to the export functions, such as:
`chain.to_parquet("s3://dtulga-datachain-test/test.parquet")`
`chain.to_csv("hf://datasets/dtulga/datachain-test/test.csv")`

This has been tested manually with S3 and Hugging Face, as seen here: https://huggingface.co/datasets/dtulga/datachain-test/tree/main This is part of #236 and #370

### DIFF
--- a/src/datachain/lib/dc.py
+++ b/src/datachain/lib/dc.py
@@ -1895,10 +1895,14 @@ class DataChain:
         """Save chain to parquet file with SignalSchema metadata.
 
         Parameters:
-            path : Path or a file-like binary object to save the file.
+            path : Path or a file-like binary object to save the file. This supports
+                local paths as well as remote paths, such as s3:// or hf:// with fsspec.
             partition_cols : Column names by which to partition the dataset.
             chunk_size : The chunk size of results to read and convert to columnar
                 data, to avoid running out of memory.
+            fs_kwargs : Optional kwargs to pass to the fsspec filesystem, used only for
+                write, for fsspec-type URLs, such as s3:// or hf:// when
+                provided as the destination path.
         """
         import pyarrow as pa
         import pyarrow.parquet as pq
@@ -1988,8 +1992,12 @@ class DataChain:
         """Save chain to a csv (comma-separated values) file.
 
         Parameters:
-            path : Path to save the file.
+            path : Path to save the file. This supports local paths as well as
+                remote paths, such as s3:// or hf:// with fsspec.
             delimiter : Delimiter to use for the resulting file.
+            fs_kwargs : Optional kwargs to pass to the fsspec filesystem, used only for
+                write, for fsspec-type URLs, such as s3:// or hf:// when
+                provided as the destination path.
         """
         import csv
 
--- a/src/datachain/lib/dc.py
+++ b/src/datachain/lib/dc.py
@@ -1,6 +1,8 @@
 import copy
 import os
+import os.path
 import re
+import sys
 from collections.abc import Iterator, Sequence
 from functools import wraps
 from typing import (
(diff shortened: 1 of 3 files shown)


tests/func/test_datachain.py
```diff
@@ -1366,3 +1372,68 @@ def file_info(file: File) -> DataModel:
...
+
+    df = pd.DataFrame(DF_DATA)
+    dc_to = DataChain.from_pandas(df, session=ctc.session)
+    dc_to.to_csv(path)
+
+    dc_from = DataChain.from_csv(path, session=ctc.session)
+    df1 = dc_from.select("first_name", "age", "city").to_pandas()
+    assert df1.equals(df)
+
+    # Cleanup any written files
```
shcheklein, line 1389: qq - are we using real clouds? do we care about cleanup here? if we care - should it be wrapped into a fixture (so that we do this even in case of a failure)

src/datachain/lib/dc.py
```diff
@@ -1887,6 +1889,7 @@ def to_parquet(
         path: Union[str, os.PathLike[str], BinaryIO],
         partition_cols: Optional[Sequence[str]] = None,
         chunk_size: int = DEFAULT_PARQUET_CHUNK_SIZE,
+        fs_kwargs: Optional[dict[str, Any]] = None,
```
shcheklein, line 1892: I have a concern here that we don't use `fs_kwargs` in all other places (e.g. anon=True, or `from_parquet` I think reads from a file object - file.get_fs() or something). can we do a bit of research on that end and unify or get rid of this additional kwargs?

dtulga: These kwargs are optional and are combined with the Catalog's `client_config` for any unified use cases, such as configuration that applies to read and write. I added this optional kwargs parameter here to be used if users need to specify a write-only custom configuration, such as an access token to be used on write, etc. that is not (or may not apply) on read or for the whole application / chain. For example, a token can be specified for Hugging Face filesystems on write, as described here: https://huggingface.co/docs/huggingface_hub/en/guides/hf_file_system#authentication but users may only want to specify this token on write to Hugging Face, not for other clouds or on read. I can rename this or change as desired, but it seems like having extra write-only kwargs can be useful in some cases.

tests/func/test_datachain.py
dtulga: This uses simulated clouds provided by `pytest-servers` - and the cleanup code is necessary to prevent test failures (as these uploaded files are not automatically cleaned up). I moved the cleanup code into a fixture to keep it in one location.

src/datachain/lib/dc.py
shcheklein: kk. It seems to me that it should be symmetrical (people might need to provide something extra on reads as well eventually) and then the question is - will we be able to do the same easy and w/o changing some logic in those methods (like from_parquet, etc).

should we update the docs here btw?

shcheklein (approved):

rlamy (approved):
The code looks good, but I think the docs should be updated to mention that fsspec URLs work as well.

src/datachain/lib/dc.py
dtulga: Yes, the docs should be updated - I updated them in the latest commit. And shared (read and write) kwargs can be provided in `client_config` on Session or Catalog (as well as in environment variables) and these configuration settings will automatically be used for read and write. This `fs_kwargs` option just provides a way to provide write-only configuration, or override the shared configuration, only if necessary.

dtulga:
> The code looks good, but I think the docs should be updated to mention that fsspec URLs work as well.

Agreed, updated the docs in the latest commit.

PR #517, opened 2024-10-17
Fix CLI to work with `DataChain` new listing

ilongin:
This PR fixes all CLI methods to work with new listing of `DataChain` by using `.from_storage()` method.
Before, CLI was completely detached  which means if you would do `datachain index`, listing created by that command would be invisible by `DataChain` lib, and vice versa.
Now, those are again "one" system.

Note: that this PR removes custom TTL set for listings to expire as we from now on don't utilize `Storage` table any more where we had ability to save this. Instead, we will always use constant TTL time. If this doesn't seem enough, we can think about where to put this and maybe take into account some alternatives (maybe having some global config etc.)

Followups:
* https://github.com/iterative/datachain/issues/536
* https://github.com/iterative/datachain/issues/535
* Replace `Node` with `File`
* Move listing warehouse methods to aplication level

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,7 @@ ci:
   skip: [mypy]
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.6.0
+    rev: v5.0.0
     hooks:
       - id: check-added-large-files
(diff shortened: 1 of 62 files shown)


amritghimire (approved):

amritghimire:
LGTM as long as it is tested

PR #518, opened 2024-10-18
clarify semantics in order_by docstring

### LINKED ISSUE #477: Clarify ordering semantics

mattseddon:
Closes #477

### DIFF
--- a/src/datachain/lib/dc.py
+++ b/src/datachain/lib/dc.py
@@ -981,10 +981,22 @@ class DataChain:
 
     @resolve_columns
     def order_by(self, *args, descending: bool = False) -> "Self":
-        """Orders by specified set of signals.
+        """Orders by specified set of columns.
 
         Parameters:
             descending (bool): Whether to sort in descending order or not.
+
+        Example:
+            ```py
+            dc.order_by("similarity_score", descending=True).limit(10)
+            ```
+
+        Note:
+            Order is not guaranteed when steps are added after an `order_by` statement.
+            I.e. when using `from_dataset` an `order_by` statement should be used if
+            the order of the records in the chain is important.
+            Using `order_by` directly before `limit` will give expected results.
+            See https://github.com/iterative/datachain/issues/477 for further details.
         """
         if descending:
             args = tuple(sqlalchemy.desc(a) for a in args)


src/datachain/lib/dc.py
```diff
@@ -953,10 +953,22 @@ def _extend_to_data_model(self, method_name, *args, **kwargs):
...
+        Example:
+            ```py
+            dc.order_by("similarity_score", descending=True).limit(10)
+            ```
+
+        Note:
+            Order is not guaranteed when steps are added after an `order_by` statement.
+            I.e. when using `from_dataset` an `order_by` statement should be used if
+            the order of the records in the chain is important.
+            Using `order_by` directly before `limit` will give expected results.
```
shcheklein, line 970: I assume `collect`, `show`, `to_parquet` - are not considered as steps and will give also a "desired" output?

mattseddon: `collect` - should but not guranteed (explanation in the original ticket https://github.com/iterative/datachain/issues/477)
`show` - yes
`to_parquet` - should but not guaranteed (uses `collect_flatten`)

shcheklein (approved):

src/datachain/lib/dc.py
mattseddon: Do you mean adding methods like `collect_ordered` OR adding an explicit `order_by` param to those methods or patching those functions? Is there a specific use case we're targeting or are we just trying to avoid confusion?

shcheklein: I think it's a basic expectation that if you do `order_by().collect()` that you are getting a sorted result? collect is a way to iterate on the results I guess.

I mean that `collect` probably should be implemented similar to show() - something like ``order_by().collect()` should be a single SQL query probably

dreadatour (approved):
🙏👍

samran5 (approved):

samran5 (approved):

PR #520, opened 2024-10-18
patch unstructured embeddings gen example

mattseddon:
It is just a patch for the example/test. We should revert once https://github.com/Unstructured-IO/unstructured/issues/3730 is merged and released

### DIFF
--- a/examples/llm_and_nlp/unstructured-embeddings-gen.py
+++ b/examples/llm_and_nlp/unstructured-embeddings-gen.py
@@ -12,11 +12,11 @@ from unstructured.cleaners.core import (
     group_broken_paragraphs,
     replace_unicode_quotes,
 )
-from unstructured.partition.pdf import partition_pdf
-from unstructured_ingest.embed.huggingface import (
+from unstructured.embed.huggingface import (
     HuggingFaceEmbeddingConfig,
     HuggingFaceEmbeddingEncoder,
 )
+from unstructured.partition.pdf import partition_pdf
 
 from datachain import C, DataChain, DataModel, File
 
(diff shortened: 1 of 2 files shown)


examples/llm_and_nlp/unstructured-embeddings-gen.py
```diff
@@ -12,11 +12,11 @@
     group_broken_paragraphs,
     replace_unicode_quotes,
 )
-from unstructured.embed.huggingface import (
+from unstructured.partition.pdf import partition_pdf
+from unstructured_ingest.embed.huggingface import (
```
mattseddon, line 16: [F] From what I can see this is the new package for doing things. There is a new API for ingesting data but I haven't been able to grok it completely.

What I can tell you is that you can no longer instantiate the old `HuggingFaceEmbeddingEncoder` because it is missing abstract methods and the new `embed_documents` expects `list[dict]` instead of `list[Element]` so they are incompatible without these small changes.

mattseddon: I raised https://github.com/Unstructured-IO/unstructured/issues/3731 / https://github.com/Unstructured-IO/unstructured/issues/3730 to fix the issue on their end properly.

mattseddon: it seems simpler to just set the upper limit of unstructured to the version before `0.16.0`

mattseddon:
cc @tibor-mach

dtulga (approved):
LGTM, thanks for investigating this!

shcheklein (approved):

PR #521, opened 2024-10-18
Implement pre-fetching in map() and gen()

rlamy:
This adds a `prefetch` setting which enables async downloading of objects to the cache before running a generator or mapper UDF (see #40). The default is to use 2 workers, but it can be disabled using `.setting(prefetch=0)`. Note that it has no effect if caching isn't enabled (caching is disabled by default).

In order for this to work, `AbstractWarehouse.dataset_select_paginated()` is now required to be thread-safe, so query result pages are now buffered as a list in that function.

### DIFF
--- a/src/datachain/data_storage/warehouse.py
+++ b/src/datachain/data_storage/warehouse.py
@@ -232,7 +232,10 @@ class AbstractWarehouse(ABC, Serializable):
                 if limit < page_size:
                     paginated_query = paginated_query.limit(None).limit(limit)
 
-            results = self.dataset_rows_select(paginated_query.offset(offset))
+            # Ensure we're using a thread-local connection
+            with self.clone() as wh:
+                # Cursor results are not thread-safe, so we convert them to a list
+                results = list(wh.dataset_rows_select(paginated_query.offset(offset)))
 
             processed = False
             for row in results:
--- a/src/datachain/lib/dc.py
+++ b/src/datachain/lib/dc.py
@@ -334,6 +334,7 @@ class DataChain:
         parallel=None,
         workers=None,
         min_task_size=None,
+        prefetch: Optional[int] = None,
         sys: Optional[bool] = None,
     ) -> "Self":
         """Change settings for chain.
@@ -360,7 +361,7 @@ class DataChain:
         if sys is None:
             sys = self._sys
         settings = copy.copy(self._settings)
-        settings.add(Settings(cache, parallel, workers, min_task_size))
+        settings.add(Settings(cache, parallel, workers, min_task_size, prefetch))
         return self._evolve(settings=settings, _sys=sys)
 
     def reset_settings(self, settings: Optional[Settings] = None) -> "Self":
@@ -882,6 +883,8 @@ class DataChain:
             ```
         """
         udf_obj = self._udf_to_obj(Mapper, func, params, output, signal_map)
+        if (prefetch := self._settings.prefetch) is not None:
+            udf_obj.prefetch = prefetch
 
         return self._evolve(
             query=self._query.add_signals(
@@ -919,6 +922,8 @@ class DataChain:
             ```
         """
         udf_obj = self._udf_to_obj(Generator, func, params, output, signal_map)
+        if (prefetch := self._settings.prefetch) is not None:
+            udf_obj.prefetch = prefetch
         return self._evolve(
             query=self._query.generate(
                 udf_obj.to_udf_wrapper(),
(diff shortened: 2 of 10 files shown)


src/datachain/data_storage/warehouse.py
```diff
@@ -227,7 +227,10 @@ def dataset_select_paginated(
                 if limit < page_size:
                     paginated_query = paginated_query.limit(None).limit(limit)
 
-            results = self.dataset_rows_select(paginated_query.offset(offset))
+            # Ensure we're using a thread-local connection
+            with self.clone() as wh:
+                # Cursor results are not thread-safe, so we convert them to a list
```
shcheklein, line 232: Q: why do they have to be thread safe (I mean cursor results)?
Q: are there any implications in terms of memory usage for this?

src/datachain/lib/settings.py
```diff
@@ -66,3 +74,5 @@ def add(self, settings: "Settings"):
         self.parallel = settings.parallel or self.parallel
         self._workers = settings._workers or self._workers
         self.min_task_size = settings.min_task_size or self.min_task_size
+        if settings.prefetch is not None:
```
shcheklein, line 77: is a reason to have a mix of styles here - some protected vars, some not, some ` self._cache = settings._cache or self._cache`, some like `if settings.prefetch is not None:`?

tests/unit/test_asyn.py
```diff
@@ -111,6 +112,37 @@ async def process(row):
         list(mapper.iterate(timeout=4))
 
 
+@pytest.mark.parametrize("create_mapper", [AsyncMapper, OrderedMapper])
+def test_mapper_deadlock(create_mapper):
```
shcheklein, line 116: could you elaborate a bit please? seems to be a regular mapper run, or is there something special about this config that makes it eligible to a deadlock, was there some bug in this configuration?

src/datachain/lib/dc.py
```diff
@@ -325,6 +325,7 @@ def settings(
         parallel=None,
         workers=None,
         min_task_size=None,
+        prefetch: Optional[int] = None,
```
shcheklein, line 328: q: why int? let's update the docs here (do we have some CI to detect these discrepancies btw (missing docs) cc @skshetry )

tests/unit/test_asyn.py
shcheklein, line 116: will it deadlock if we don't wrap producer into a thread? were you trying to test (make sure that producer is wrapped)?

shcheklein (commented):
Looks great. A few question. 

One more general question.

Does this implementation mean that we now won't start (at least the very first row) UDF until file is fetched? Before it was doing this "on-demand" I guess - when file is needed. I wonder how big of an issue it can be in certain scenarios (and especially if we decide to do prefetch for batches (agg, batch mapper)).

src/datachain/data_storage/warehouse.py
rlamy: With sqlite, the cursor holds all the results in memory anyway, so the conversion shouldn't change things much, though we do now temporarily have 2 copies of the full result set.
I'm not sure about the behaviour of the ClickHouse driver, OTOH. But in any case, pagination mitigates the memory issues, as we're dealing with at most 100k rows at a time.

src/datachain/lib/settings.py
rlamy: I don't see the point of using private attributes, but I didn't want to refactor `Settings` as that isn't directly related to this PR.
Using `settings.prefetch or self.prefetch` would be incorrect, as both `0` and `None` are valid values with different meanings (`0` means "disable prefetching", `None` means "use the default value").

rlamy:
> Does this implementation mean that we now won't start (at least the very first row) UDF until file is fetched? Before it was doing this "on-demand" I guess - when file is needed. I wonder how big of an issue it can be in certain scenarios (and especially if we decide to do prefetch for batches (agg, batch mapper)).

Yes, the File object is only passed to the UDF once prefetching is complete, but note that in this PR you need to specify `cache=True` to activate prefetching, which you probably wouldn't do in cases where prefetching is suboptimal.

skshetry (approved):

src/datachain/data_storage/warehouse.py
```diff
@@ -232,7 +232,10 @@ def dataset_select_paginated(
                 if limit < page_size:
                     paginated_query = paginated_query.limit(None).limit(limit)
 
-            results = self.dataset_rows_select(paginated_query.offset(offset))
+            # Ensure we're using a thread-local connection
+            with self.clone() as wh:
```
skshetry, line 236: I guess this `clone()` should be outside the loop? Otherwise, we'll not be reusing the connection.

PR #526, opened 2024-10-21
Implement ls-datasets for studio client

amritghimire:
This fixes the existing studio client to use the configuration from studio login and provides a way to list the datasets available in Studio.

```sh
$ datachain studio datasets
passed_as_argument (v1)
local_test_datachain (v1)
dql-small (v1)

```

```sh
$ datachain studio datasets --team team_name
passed_as_argument (v1)
local_test_datachain (v1)
dql-small (v1)

```

Relates to https://github.com/iterative/studio/issues/10774


ilongin (approved):
LGTM, thanks for the fix

PR #537, opened 2024-10-24
datachain: support mutating existing column

### LINKED ISSUE #336: mutate: overwrite columns
#306 is a workaround that throws an exception. But we need to implement this functionality with a potential schema change.

This command should change my_dist to another value with a new type (int --> float): 
```python
dc.select("file", "my_dist").mutate(my_dist=dc.column("my_dist") * 2* 3.14)
```

skshetry:
Closes #336.

### DIFF
--- a/src/datachain/lib/dc.py
+++ b/src/datachain/lib/dc.py
@@ -1111,14 +1111,6 @@ class DataChain:
         )
         ```
         """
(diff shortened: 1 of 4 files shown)


ilongin (approved):
LGTM

PR #538, opened 2024-10-24
Adding JSON / JSON Lines Export Support

### LINKED ISSUE #533: Add JSON / JSON lines export support
This is to add basic JSON / JSON lines export support with functions like `to_json` and `to_jsonl` - the exported data from these functions should also be importable through the `from_json` and `from_jsonl` functions in DataChain. Feel free to suggest any other relevant or related features (or these can be added in a later issue).

dtulga:
This PR adds the functions `to_json` and `to_jsonl` to export to JSON / JSON lines from DataChain. These functions support remote fsspec filesystems / URLs (such as `s3://`), similar to `to_csv` and `to_parquet`, and implement streaming writing as well, to enable writing data larger than memory. This also adds tests for these functions, tests for the existing `from_json` and `from_jsonl` functions, and tests for both of these on remote filesystems (such as S3).

This has been tested locally and fixes #533

### DIFF
--- a/src/datachain/lib/dc.py
+++ b/src/datachain/lib/dc.py
@@ -2051,6 +2051,81 @@ class DataChain:
             for row in results_iter:
                 writer.writerow(row)
 
+    def to_json(
+        self,
+        path: Union[str, os.PathLike[str]],
+        fs_kwargs: Optional[dict[str, Any]] = None,
+        include_outer_list: bool = True,
+    ) -> None:
+        """Save chain to a JSON file.
+
+        Parameters:
+            path : Path to save the file. This supports local paths as well as
+                remote paths, such as s3:// or hf:// with fsspec.
+            fs_kwargs : Optional kwargs to pass to the fsspec filesystem, used only for
+                write, for fsspec-type URLs, such as s3:// or hf:// when
+                provided as the destination path.
+            include_outer_list : Sets whether to include an outer list for all rows.
+                Setting this to True makes the file valid JSON, while False instead
+                writes in the JSON lines format.
+        """
+        opener = open
+
+        if isinstance(path, str) and "://" in path:
+            from datachain.client.fsspec import Client
+
+            fs_kwargs = {
+                **self._query.catalog.client_config,
+                **(fs_kwargs or {}),
+            }
+
+            client = Client.get_implementation(path)
+
+            fsspec_fs = client.create_fs(**fs_kwargs)
+
+            opener = fsspec_fs.open
+
+        headers, _ = self._effective_signals_schema.get_headers_with_length()
+        column_names = [".".join(filter(None, header)) for header in headers]
+
+        results_iter = self.collect_flatten()
+
+        is_first = True
+
+        with opener(path, "wb") as f:
+            if include_outer_list:
+                # This makes the file JSON instead of JSON lines.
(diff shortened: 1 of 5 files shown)


src/datachain/lib/dc.py
```diff
@@ -2051,6 +2051,81 @@ def to_csv(
...
+            }
+
+            client = Client.get_implementation(path)
+
+            fsspec_fs = client.create_fs(**fs_kwargs)
+
+            opener = fsspec_fs.open
+
+        headers, _ = self._effective_signals_schema.get_headers_with_length()
+        column_names = [".".join(filter(None, header)) for header in headers]
```
shcheklein, line 2089: since JSON supports nesting should we dump it with proper field names? like we do in pandas

to make sure for example (that at least in simple cases) we can read some nested JSON (it creates a nested model) and dump it later in the same way and get the same result

src/datachain/lib/dc.py
```diff
@@ -2051,6 +2051,81 @@ def to_csv(
...
+
+        results_iter = self.collect_flatten()
```
ilongin, line 2091: I would just remove this variable and do `for row in self.collect_flatten()` below to avoid looking at what `results_iter` actually represents ... but this is not as that important, your call.

src/datachain/lib/dc.py
```diff
@@ -2051,6 +2051,81 @@ def to_csv(
...
+                    else:
+                        f.write(b"\n")
+                else:
+                    is_first = False
+                f.write(orjson.dumps(dict(zip(column_names, row))))
+            if include_outer_list:
+                # This makes the file JSON instead of JSON lines.
+                f.write(b"\n]\n")
+
+    def to_jsonl(
```
ilongin, line 2113: It seems little redundant to have this method if the same can be achieved with `to_json()` by setting `include_outer_list` flag. I'm not sure what is the optimal solution though ... maybe move all the logic from `to_json` to some non public private method and make  public `to_json()` and `to_jsonl()` just wrappers around it by calling correct `include_outer_list` flag (similar to current `to_jsonl()`). Another solution is to just remove `to_jsonl()` ... again, this is also not that super important and neither solution seems perfect as well.

ilongin (approved):
LGTM

src/datachain/lib/dc.py
dtulga: Sure, changed to just using `self.collect_flatten()` directly.

src/datachain/lib/dc.py
dtulga: Sure, I changed it to writing a nested dict for JSON / JSON lines, instead of the flattened version, which I was originally doing just to keep consistency with `to_csv` and `to_parquet` - although I can see how nested works well for JSON / JSON lines here.

src/datachain/lib/dc.py
dtulga: Yeah, I wasn't sure what the best approach is either, but this one seems to have the least extra code (as making both `to_json` and `to_jsonl` wrappers adds extra wrapper code), and also the most consistency / user-friendliness as we have `from_json` and `from_jsonl` so I wanted to keep the `to` functions consistent.

shcheklein (approved):

Issue #481, opened 2024-09-27
Add explode and / or dynamic model / schema


shcheklein:
Follow up https://github.com/iterative/dvcx/pull/1368
Based also on this discussion / feedback by @tibor-mach https://iterativeai.slack.com/archives/C04A9RWEZBN/p1727194987119179
Base also on iteration on DCLM - https://github.com/iterative/studio/issues/10596

## Summary

When we have a single file (JSONL or CVS/Parquet with a column with JSONs) we need a way to "explode" those JSONs/dicts into a Pythonic model and store it in DataChain not a single column, but as multiple columns - one per each path in that JSON/dict.

E.g. this is how JSONL looks like after a naive parse:

<img width="1126" alt="Screenshot 2024-09-26 at 6 10 18 PM" src="https://github.com/user-attachments/assets/48cd0da5-26ea-4746-803c-f6ac057afe53">

Or from the CVS file (mind the meta column):

![image (11)](https://github.com/user-attachments/assets/69709909-b0d6-4829-82b4-b031ee43da54)

There is an obvious way to mitigate this - create a Model class and populate it from in the UDF that. But that's seems very annoying and redundant - model description becomes 2x/3x code of the parser.

## Suggestions

- [ ] `DataChain.explode(C("meta"))`. This one is more or less obvious and requires creating an extra table.
- [ ] Make functions like `map`, `gen` dynamically figure out schema and create Pydantic model as it is parsing files. This requires more complicated implementation, but can faster since it can work in a streaming mode:

Imagine something like this:

```python
def extract(file: File) -> Iterator[File, dict]:
    with file.open() as f:
        dctx = zstd.ZstdDecompressor()
        stream_reader = dctx.stream_reader(f)
        text_stream = io.TextIOWrapper(stream_reader, encoding='utf-8')
        for line in text_stream:
            yield file, json.parse(line)


DataChain.from_dataset("index").settings(cache=True).limit(1).gen(extract).save("raw_text")
```

skshetry:
This may be already possible with the combination of `read_meta` and `map` depending on how we want to solve this. (`read_meta` requires a storage_path or a pydantic model passed).

How are we going to determine schema? Is it based on a sample of rows (which is what `read_meta` does, it reads a single row) or by reading all the rows?

PR #542, opened 2024-10-26
fix(parsing): normalize nested column names

shcheklein:
Part of the https://github.com/iterative/datachain/issues/481

Fixes a few issue with nested column names. We had before a [function](https://github.com/iterative/datachain/blob/main/src/datachain/lib/arrow.py#L146-L156) that was normalizing the top level schema names in a single place [here](https://github.com/iterative/datachain/blob/main/src/datachain/lib/arrow.py#L131).

It means that if you have a JSON file to parse with `parse_tabular` (yes, it supports JSONs as well), or any other nested structure where we create Pydantic models for nested fields with [`dict_to_data_model`](https://github.com/iterative/datachain/blob/main/src/datachain/lib/data_model.py#L62-L68) we either had a bunch of non-normalized names except for the top level, or we could end up with a syntax error.

E.g. I had a JSON like this:

```json
{
  "bff_contained_ngram_count_before_dedupe": 0,
  "language_id_whole_page_fasttext": {
    "en": 0.9512948989868164
  },
  "metadata": {
    "Content-Length": "112946",
    "Content-Type": "application/http; msgtype=response",
    "WARC-Block-Digest": "sha1:U462KE2IDQWYDJQDRXIP3UVH465G2I2G",
    "WARC-Concurrent-To": "<urn:uuid:eaf4d817-f140-4a5e-9fa9-64d0dcb32a08>",
    "WARC-Date": "2017-05-27T21:20:09Z",
    "WARC-IP-Address": "198.57.149.47",
    "WARC-Identified-Payload-Type": "text/html",
    "WARC-Payload-Digest": "sha1:KAFZ77S3HPJ7SOABQ4ZMMVT47OCXZR6R",
    "WARC-Record-ID": "<urn:uuid:f8557905-b0dd-47b9-bc60-3106b3c18b4c>",
    "WARC-Target-URI": "http://www.muslimlinkpaper.com/index.php/editors-desk/13-letter-to-the-editor/3054-a-letter-to-president-obama.html",
    "WARC-Truncated": "length",
    "WARC-Type": "response",
    "WARC-Warcinfo-ID": "<urn:uuid:398fb229-e8ed-4f8b-86af-00eb8d6b2594>"
  },
  "previous_word_count": 178,
  "text": "A Letter ...",
  "url": "http://www.muslimlinkpaper.com/index.php/editors-desk/13-letter-to-the-editor/3054-a-letter-to-president-obama.html",
  "warcinfo": "robots: classic\r\nhostname: ip-10-185-224-210.ec2.internal\r\nsoftware: Nutch 1.6 (CC)/CC WarcExport 1.0\r\nisPartOf: CC-MAIN-2017-22\r\noperator: Common Crawl Admin\r\ndescription: Wide crawl of the web for May 2017\r\npublisher: Common Crawl\r\nformat: WARC File Format 1.0\r\nconformsTo: http://bibnum.bnf.fr/WARC/WARC_ISO_28500_version1_latestdraft.pdf",
  "fasttext_openhermes_reddit_eli5_vs_rw_v2_bigram_200k_train_prob": 0.967424213886261
}
```

Particularly, dealing with a field `WARC-Concurrent-To` is problematic. `sqlalchmey` fails to quote bound params in `INSERT ... VALUES ...` and `-TO` triggers a syntax error, as well as `-SELECT` for example.

With this PR, we are applying the same rules to all column names - top level and nested. It requires making an alias on Pydantic model, but I hope that is fine.

## TODO:

- [x] Add more unit tests for the new name normalization function
- [x] To @0x2b3bfa0 point - make it even more strict (make them valid Python identifiers - e.g. they can't start with numbers) - at least in case when we generate a pydantic model out of them.

### DIFF
--- a/src/datachain/lib/utils.py
+++ b/src/datachain/lib/utils.py
@@ -39,8 +39,7 @@ def normalize_col_names(col_names: Sequence[str]) -> dict[str, str]:
 
     for org_column in col_names:
         new_column = org_column.lower()
-        new_column = re.sub("[^0-9a-z_\\s-]+", "_", new_column)
-        new_column = re.sub("[-_\\s]+", "_", new_column)
+        new_column = re.sub("[^0-9a-z]+", "_", new_column)
         new_column = new_column.strip("_")
 
         if (
--- a/src/datachain/lib/utils.py
+++ b/src/datachain/lib/utils.py
@@ -1,4 +1,6 @@
+import re
 from abc import ABC, abstractmethod
+from collections.abc import Sequence
 
 
 class AbstractUDF(ABC):
@@ -28,3 +30,37 @@ class DataChainParamsError(DataChainError):
 class DataChainColumnError(DataChainParamsError):
     def __init__(self, col_name, msg):
         super().__init__(f"Error for column {col_name}: {msg}")
+
+
+def normalize_col_names(col_names: Sequence[str]) -> dict[str, str]:
+    gen_col_counter = 0
+    new_col_names = {}
+    org_col_names = set(col_names)
+
+    for org_column in col_names:
+        new_column = org_column.lower()
+        new_column = re.sub("[^0-9a-z_\\s-]+", "_", new_column)
+        new_column = re.sub("[-_\\s]+", "_", new_column)
+        new_column = new_column.strip("_")
+
+        if (
+            not new_column
+            or new_column[0].isdigit()
+            or (new_column != org_column and new_column in org_col_names)
+            or new_column in new_col_names
+        ):
+            while True:
+                generated_column = f"c{gen_col_counter}"
+                gen_col_counter += 1
+                if new_column:
+                    generated_column = f"{generated_column}_{new_column}"
+                if (
+                    generated_column not in org_col_names
+                    and generated_column not in new_col_names
+                ):
+                    new_column = generated_column
+                    break
+
+        new_col_names[new_column] = org_column
+
+    return new_col_names
--- a/src/datachain/lib/utils.py
+++ b/src/datachain/lib/utils.py
@@ -42,24 +42,19 @@ def normalize_col_names(col_names: Sequence[str]) -> dict[str, str]:
         new_column = re.sub("[^0-9a-z]+", "_", new_column)
         new_column = new_column.strip("_")
 
-        if (
-            not new_column
-            or new_column[0].isdigit()
-            or (new_column != org_column and new_column in org_col_names)
-            or new_column in new_col_names
-        ):
-            while True:
-                generated_column = f"c{gen_col_counter}"
-                gen_col_counter += 1
-                if new_column:
-                    generated_column = f"{generated_column}_{new_column}"
-                if (
-                    generated_column not in org_col_names
-                    and generated_column not in new_col_names
-                ):
-                    new_column = generated_column
-                    break
+        generated_column = new_column
 
-        new_col_names[new_column] = org_column
+        while (
+            not generated_column.isidentifier()
+            or generated_column in new_col_names
+            or (generated_column != org_column and generated_column in org_col_names)
+        ):
+            if new_column:
+                generated_column = f"c{gen_col_counter}_{new_column}"
+            else:
+                generated_column = f"c{gen_col_counter}"
+            gen_col_counter += 1
+
+        new_col_names[generated_column] = org_column
 
     return new_col_names
--- a/src/datachain/lib/arrow.py
+++ b/src/datachain/lib/arrow.py
@@ -1,4 +1,3 @@
-import re
 from collections.abc import Sequence
 from tempfile import NamedTemporaryFile
 from typing import TYPE_CHECKING, Any, Optional
@@ -13,6 +12,7 @@ from datachain.lib.file import ArrowRow, File
 from datachain.lib.model_store import ModelStore
 from datachain.lib.signal_schema import SignalSchema
 from datachain.lib.udf import Generator
+from datachain.lib.utils import normalize_col_names
 
 if TYPE_CHECKING:
     from datasets.features.features import Features
@@ -128,7 +128,7 @@ def schema_to_output(schema: pa.Schema, col_names: Optional[Sequence[str]] = Non
     signal_schema = _get_datachain_schema(schema)
     if signal_schema:
         return signal_schema.values
-    columns = _convert_col_names(col_names)  # type: ignore[arg-type]
+    columns = list(normalize_col_names(col_names).keys())  # type: ignore[arg-type]
     hf_schema = _get_hf_schema(schema)
     if hf_schema:
         return {
@@ -141,19 +141,6 @@ def schema_to_output(schema: pa.Schema, col_names: Optional[Sequence[str]] = Non
             dtype = Optional[dtype]  # type: ignore[assignment]
         output[column] = dtype
     return output
-
-
-def _convert_col_names(col_names: Sequence[str]) -> list[str]:
-    default_column = 0
-    converted_col_names = []
-    for column in col_names:
-        column = column.lower()
-        column = re.sub("[^0-9a-z_]+", "", column)
-        if not column:
-            column = f"c{default_column}"
-            default_column += 1
-        converted_col_names.append(column)
-    return converted_col_names
 
 
 def arrow_type_mapper(col_type: pa.DataType, column: str = "") -> type:  # noqa: PLR0911
--- a/src/datachain/lib/data_model.py
+++ b/src/datachain/lib/data_model.py
@@ -2,9 +2,10 @@ from collections.abc import Sequence
 from datetime import datetime
 from typing import ClassVar, Union, get_args, get_origin
 
-from pydantic import BaseModel, create_model
+from pydantic import BaseModel, Field, create_model
 
 from datachain.lib.model_store import ModelStore
+from datachain.lib.utils import normalize_col_names
 
 StandardType = Union[
     type[int],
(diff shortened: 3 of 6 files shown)


src/datachain/lib/utils.py
```diff
@@ -28,3 +30,19 @@ def __init__(self, message):
...
         super().__init__(f"Error for column {col_name}: {msg}")
+
+
+def normalize_col_names(col_names: Sequence[str]) -> dict[str, str]:
+    default_column = 0
+    converted_col_names = {}
+    for column in col_names:
+        original_column = column
+        column = column.lower()
+        column = re.sub("-+", "_", column)
```
shcheklein, line 41: C: I've changed the rule a bit here as well. It now changes `-` into `_` and multiple `_` into a single one. 

Also checks if the column name after normalization exists it now also applies the `default_column` generator.

0x2b3bfa0: @shcheklein, quick question: what happens if the original column name _begins_ with a number, would it break something?

0x2b3bfa0: Asking just in case we use these identifiers as e.g. Python attributes, where `column1` would be valid but `1column` wouldn't

shcheklein: good question, I'm a bit hesitant to go that far (not always we have to create a pydantic model - e.g. we might not need for top level fields - just a flat parquet / CSV). I feel it's a bit broader topic ... I'll think about it and if we need to solve it in this PR

shcheklein: probably at least in the `dict_to_data_model` we have to ensure this I guess, so can introduce an additional param for that

src/datachain/lib/utils.py
```diff
@@ -28,3 +30,29 @@ def __init__(self, message):
...
+def normalize_col_names(col_names: Sequence[str]) -> dict[str, str]:
+    gen_col_counter = 0
+    new_col_names = {}
+    org_col_names = set(col_names)
+
+    for org_column in col_names:
+        new_column = org_column.lower()
+        new_column = re.sub("[-_\\s]+", "_", new_column)
+        new_column = new_column.strip("_")
+        new_column = re.sub("[^0-9a-z_]+", "", new_column)
```
0x2b3bfa0, line 44:
```suggestion
        new_column = org_column.lower()
        new_column = re.sub("[^0-9a-z]+", "_", new_column)
        new_column = new_column.strip("_")
```

Wouldn't you prefer this alternative normalization, which interprets any non-alphanumeric character as a separator?
* Before: `Old/Man's@Haven 1` → `oldmanshaven_1`
* After: `Old/Man's@Haven@ 1` → `old_man_s_haven_1`

src/datachain/lib/utils.py
```diff
@@ -28,3 +30,29 @@ def __init__(self, message):
...
+
+        if not new_column or (new_column != org_column and new_column in org_col_names):
+            while True:
+                generated_column = f"c{gen_col_counter}"
+                gen_col_counter += 1
+                if new_column:
+                    generated_column = f"{generated_column}_{new_column}"
+                if generated_column not in org_col_names:
+                    new_column = generated_column
+                    break
```
0x2b3bfa0, line 54:
```suggestion
        if not new_column or (new_column != org_column and new_column in new_col_names):
            while True:
                generated_column = f"c{gen_col_counter}"
                gen_col_counter += 1
                if new_column:
                    generated_column = f"{generated_column}_{new_column}"
                if generated_column not in new_col_names:
                    new_column = generated_column
                    break
```

I thought the whole point of this logic was checking for duplicates in the `new_column_names` dictionary?

### Test vector

```python
normalize_col_names(("column+name", "column=name"))
```

#### Before
```python
assert result == {"columnname": "column=name"}
```

#### After
```python
assert result == {"columnname": "column+name", "c0_columnname": "column=name"}
```

src/datachain/lib/utils.py
shcheklein: I'm fixing this (and a bunch of other case) - don't spend time yet on this please :)

src/datachain/lib/utils.py
shcheklein: sorry about that, I'm fixing all of this right now ... give me a few minutes a review again pls

shcheklein: > I thought the whole point of this logic was checking for duplicates in the new_column_names dictionary?

it's not that simple ... we should also ideally check that there is no already a column in the original list and try to avoid naming it the same if possible

src/datachain/lib/utils.py
shcheklein: not sure, I'll think a bit about this ... there could be non ascii literals - should we also change them to _ - not sure tbh.

0x2b3bfa0: You're right; we risk going too far with normalization, although I wonder if it'd be a good idea to create table names containing non-ASCII characters anyway.

We can always normalize diacritics with something like this:

```python
>>> from unicodedata import normalize
>>> normalize("NFKD", "próspera piraña").encode("ascii", "ignore")
"prospera pirana"
```

But, again, thorough normalization is a complex topic, and we'd still not be taking into account e.g. Cyrillic, Han...

0x2b3bfa0: Irrelevant rambling (continued)...

```suggestion
        new_column = org_column.casefold()  # Unicode-friendly .lower()
        new_column = regex.sub("[^\p{L}\p{Nd}]+", "_", new_column)  # Unicode-friendly [^a-z0-9]
        new_column = new_column.strip("_")
```

Requires the `regex` package to use Unicode character classes and crossing your fingers for database support of non-ASCII column names.

0x2b3bfa0: Based on the latest logic, I still believe my original proposal — https://github.com/iterative/datachain/pull/542#discussion_r1817963778 — would be an improvement over the current approach.

0x2b3bfa0: Based on the latest logic, I still believe my original proposal — https://github.com/iterative/datachain/pull/542#discussion_r1817963778 — would be an improvement over the current approach.

src/datachain/lib/utils.py
0x2b3bfa0: Not sure if relevant, but in the same vein as column names beginning with a number aren't valid, reserved words like `return` shouldn't be valied either. 🙈

0x2b3bfa0:
I've been playing a bit with the logic and drafting an alternative implementation, with many debatable choices but some ideas that might be useful:

```python
def normalize_column_names(names: Sequence[str]) -> dict[str, str]:
    collision_counter = Counter(names)
    result = {}

    for original_name in names:
        normalized = original_name.lower()
        normalized = re.sub("[^0-9a-z]+", "_", normalized)
        normalized = normalized.strip("_")

        if not normalized:
            normalized = "_column"

        if not normalized.isidentifier() or iskeyword(normalized):
            normalized = f"_{normalized}"

        collision_counter.update({normalized: 1})
        if (collisions := collision_counter[normalized]) > 1:
            normalized = f"{normalized}_{collisions - 1}"

        result[normalized] = original_name

    return result
```

src/datachain/lib/utils.py
shcheklein: changed to replace with `_` all "noisy" symbols, let's not go into things like unicode normalization, etc 

is there anything left here?

shcheklein:
any other practical improvements left @0x2b3bfa0 ?

src/datachain/lib/utils.py
```diff
@@ -28,3 +30,37 @@ def __init__(self, message):
...
+
+def normalize_col_names(col_names: Sequence[str]) -> dict[str, str]:
+    gen_col_counter = 0
+    new_col_names = {}
+    org_col_names = set(col_names)
+
+    for org_column in col_names:
+        new_column = org_column.lower()
+        new_column = re.sub("[^0-9a-z_\\s-]+", "_", new_column)
+        new_column = re.sub("[-_\\s]+", "_", new_column)
```
0x2b3bfa0, line 43: Can't we simplify all that regular expression soup like this?

```suggestion
        new_column = re.sub("[^0-9a-z]+", "_", new_column)
```

src/datachain/lib/utils.py
```diff
@@ -28,3 +30,37 @@ def __init__(self, message):
...
+                if new_column:
+                    generated_column = f"{generated_column}_{new_column}"
+                if (
+                    generated_column not in org_col_names
+                    and generated_column not in new_col_names
+                ):
+                    new_column = generated_column
+                    break
+
+        new_col_names[new_column] = org_column
```
0x2b3bfa0, line 64:
```suggestion
        generated_column = new_column

        while (
            not generated_column
            or generated_column[0].isdigit()
            or generated_column in new_col_names
            or (generated_column != org_column and generated_column in org_col_names)
        ):
            if new_column:
                generated_column = f"c{gen_col_counter}_{new_column}"
            else:
                generated_column = f"c{gen_col_counter}"
            gen_col_counter += 1

        new_col_names[generated_column] = org_column
```

0x2b3bfa0 (approved):
Made a couple PEP-20 suggestions, but looks good to me!

dreadatour (approved):
Looks like a good improvement to me 👍
I can’t think of any cases where this change would cause issues.

src/datachain/lib/utils.py
0x2b3bfa0: I believe it's the same: we're replacing one or more non-alphanumerical characters with a single underscore, and it's a greedy match.

```python
>>> re.sub("[^0-9a-z]+", "_", "test$-_-$string")
'test_string'
```

0x2b3bfa0:
@shcheklein, are tests failing due to my last changes, or is it an unrelated failure?

shcheklein:
> @shcheklein, are tests failing due to my last changes, or is it an unrelated failure?

no, it's an unfortunate coincidence - pyarrow 18 got released 8 hours ago

test is not reliable, I'll improve it a bit

PR #543, opened 2024-10-28
Finish SQL functions refactoring

dreadatour:
- Move all functions from `datachain.sql.functions` to `datachain.lib.func` (hide `sql` from user).
- Wrap all functions with custom wrapper — this allows us to know the type of output signal
- Add more docstrings for functions


examples/computer_vision/openimage-detect.py
```diff
@@ -4,7 +4,7 @@
 from pydantic import BaseModel
 
 from datachain import C, DataChain, File
-from datachain.sql.functions import path
+from datachain.lib.func import path
```
mattseddon, line 7: [Q] Do we want to import `func` from datachain instead?

mattseddon: ** same for all examples - I don't think it matters which approach we take but consistency between examples would be good

src/datachain/data_storage/warehouse.py
```diff
@@ -20,8 +20,8 @@
 from datachain.data_storage.schema import convert_rows_custom_column_types
 from datachain.data_storage.serializer import Serializable
 from datachain.dataset import DatasetRecord, StorageURI
+from datachain.lib.func.inner import path as pathfunc
```
mattseddon, line 23: [Q] What is the difference between `datachain.lib.func` & `datachain.lib.func.inner`? When would you use one and not the other?

mattseddon (approved):

examples/computer_vision/openimage-detect.py
dmpetrov: right. `lib` needs to be deprecated. User should import like `from datachain.func import path`

dreadatour: Agree! I was thinking to move `func` module to the top-level (into `datachain` directory) to have more "plain" structure.

src/datachain/data_storage/warehouse.py
dreadatour: I have replaced it back from `datachain.lib.func.inner` to `datachain.sql.functions`.

The reason why we use `datachain.sql.functions` is because in some places internally we are using SQLAlchemy directly, and our `Func` wrapper is not compatible with SQLAlchemy :(

examples/computer_vision/openimage-detect.py
dreadatour: Done! 👍

Issue #554, opened 2024-10-30
FileNotFoundError when reading from Hugging Face


lhoestq:
### Description

Hi I'm Quentin from HF :) I wanted to play with datachain after https://github.com/iterative/datachain/pull/375 by @dberenbaum  but I'm getting this error:

```python
from datachain import DataChain

DataChain.from_csv("hf://datasets/infinite-dataset-hub/MobilePlanAssistant/data.csv").show()
```

```
FileNotFoundError                         Traceback (most recent call last)
[<ipython-input-2-1e396698d13d>](https://localhost:8080/#) in <cell line: 3>()
      1 from datachain import DataChain
      2 
----> 3 DataChain.from_csv("hf://datasets/infinite-dataset-hub/MobilePlanAssistant/data.csv").show()

5 frames
[/usr/local/lib/python3.10/dist-packages/datachain/lib/dc.py](https://localhost:8080/#) in from_csv(cls, path, delimiter, header, output, object_name, model_name, source, nrows, session, settings, column_types, **kwargs)
   1860             convert_options=convert_options,
   1861         )
-> 1862         return chain.parse_tabular(
   1863             output=output,
   1864             object_name=object_name,

[/usr/local/lib/python3.10/dist-packages/datachain/lib/dc.py](https://localhost:8080/#) in parse_tabular(self, output, object_name, model_name, source, nrows, **kwargs)
   1743         if col_names or not output:
   1744             try:
-> 1745                 schema = infer_schema(self, **kwargs)
   1746                 output = schema_to_output(schema, col_names)
   1747             except ValueError as e:

[/usr/local/lib/python3.10/dist-packages/datachain/lib/arrow.py](https://localhost:8080/#) in infer_schema(chain, **kwargs)
    112     schemas = []
    113     for file in chain.collect("file"):
--> 114         ds = dataset(file.get_path(), filesystem=file.get_fs(), **kwargs)  # type: ignore[union-attr]
    115         schemas.append(ds.schema)
    116     return pa.unify_schemas(schemas)

[/usr/local/lib/python3.10/dist-packages/pyarrow/dataset.py](https://localhost:8080/#) in dataset(source, schema, format, filesystem, partitioning, partition_base_dir, exclude_invalid_files, ignore_prefixes)
    792 
    793     if _is_path_like(source):
--> 794         return _filesystem_dataset(source, **kwargs)
    795     elif isinstance(source, (tuple, list)):
[...]

PR #555, opened 2024-10-30
fix(hf): use proper source when we create a file entry

shcheklein:
Fixes https://github.com/iterative/datachain/issues/554

Need to understand the reason why we didn't have `source` set properly in the first place. @dberenbaum might know better (please share if you remember anything about this).

TODO:

- [x] Add / improve / review tests

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -152,4 +152,6 @@ jobs:
         run: uv pip install nox --system
 
       - name: Run examples
+        env:
+          HF_TOKEN: ${{ secrets.HF_TOKEN }}
         run: nox -s examples -p ${{ matrix.pyv }} -- -m "${{ matrix.group }}"
--- a/examples/llm_and_nlp/hf-dataset-llm-eval.py
+++ b/examples/llm_and_nlp/hf-dataset-llm-eval.py
@@ -1,7 +1,4 @@
-import os
-
-from dotenv import load_dotenv
-from openai import OpenAI
+from huggingface_hub import InferenceClient
 
 from datachain import C, DataChain, DataModel
 
@@ -10,8 +7,6 @@ Was this dialog successful? Put result as a single word: Success or Failure.
 Explain the reason in a few words.
 """
 
-load_dotenv(".env.test")
-
 
 class DialogEval(DataModel):
     result: str
@@ -19,7 +14,7 @@ class DialogEval(DataModel):
 
 
 def eval_dialog(user_input: str, bot_response: str) -> DialogEval:
-    client = InferenceClient("meta-llama/Llama-3.1-70B-Instruct")  # change model here
+    client = InferenceClient("meta-llama/Llama-3.1-70B-Instruct")
 
     completion = client.chat_completion(
         messages=[
@@ -40,25 +35,21 @@ def eval_dialog(user_input: str, bot_response: str) -> DialogEval:
 
 # Run OpenAI in parallel for each example
 # Get result as Pydantic model that DataChain can understand and serialize
-# Save to HF as CSV
+# Save to HF as Parquet
 (
     DataChain.from_csv(
         "hf://datasets/infinite-dataset-hub/MobilePlanAssistant/data.csv"
     )
     .settings(parallel=10)
     .map(response=eval_dialog)
-    .to_csv(
-        "hf://datasets/dvcorg/test-datachain-llm-eval/data.csv",
-        fs_kwargs={"token": os.environ["HF_API_TOKEN"]},
-    )
+    .to_parquet("hf://datasets/dvcorg/test-datachain-llm-eval/data.parquet")
 )
 
 # Read it back to filter and show
 (
-    DataChain.from_csv(
-        "hf://datasets/dvcorg/test-datachain-llm-eval/data.csv",
-        column_types={"source.file.location": "str"},
+    DataChain.from_parquet(
+        "hf://datasets/dvcorg/test-datachain-llm-eval/data.parquet", source=False
     )
-    .filter(C("response_result") == "Failure")
+    .filter(C("response.result") == "Failure")
     .show(3)
 )
--- a/examples/llm_and_nlp/hf-dataset-llm-eval.py
+++ b/examples/llm_and_nlp/hf-dataset-llm-eval.py
@@ -13,6 +13,8 @@ class DialogEval(DataModel):
     reason: str
 
 
+# DataChain function to evaluate dialog.
+# DataChain is using types for inputs, results to automatically infer schema.
 def eval_dialog(user_input: str, bot_response: str) -> DialogEval:
     client = InferenceClient("meta-llama/Llama-3.1-70B-Instruct")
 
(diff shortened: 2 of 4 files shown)


dberenbaum (approved):
I can't recall if there was a specific reason for excluding the field here, but at the time it was inferred anyway, so it wasn't needed. I'm not sure what has changed since that caused the failure. As noted here and in the original PR, it would be useful to have a test for this client to guard against these regressions, but otherwise this fix LGTM.

shcheklein:
@dberenbaum thanks!

@lhoestq added an example that I'll then wrap as a test - TAL. It is analyzing your datasets using OpenAI, returns pydantic, that is then can be saved back to HF (CVS is not ideal - it's losing structure, I'll try parquet tomorrow since it can preserve Pydantic nested structure that is "native" for datachain). Though, CVS can be also read (with flat normalized fields) and manipulated further.

examples/llm_and_nlp/hf-dataset-llm-eval.py
```diff
@@ -0,0 +1,65 @@
...
+# Save to HF as CSV
+(
+    DataChain.from_csv(
+        "hf://datasets/infinite-dataset-hub/MobilePlanAssistant/data.csv"
+    )
+    .settings(parallel=10)
+    .map(response=eval_dialog)
+    .to_csv(
+        "hf://datasets/dvcorg/test-datachain-llm-eval/data.csv",
+        fs_kwargs={"token": os.environ["HF_API_TOKEN"]},
```
lhoestq, line 53: FYI you don't need this line if you logged in with `huggingface-cli login` :)

lhoestq (commented):
Thanks for the quick fix ! The examples is great as well :)
Let me know once the fix is released ! You should share the demo online as well, I'd be happy to re-share/amplify

btw I added some comments in the example but feel free to ignore

examples/llm_and_nlp/hf-dataset-llm-eval.py
shcheklein: yep, we are running these examples in Studio as well and it won't support login + if we do this as a test - we also can't use cli there ... I'll think and make a second example somewhere (e.g. on README) or a comment that it might be not needed

examples/llm_and_nlp/hf-dataset-llm-eval.py
```diff
@@ -0,0 +1,65 @@
...
+    )
+)
+
+# Read it back to filter and show
+(
+    DataChain.from_csv(
+        "hf://datasets/dvcorg/test-datachain-llm-eval/data.csv",
+        column_types={"source.file.location": "str"},
```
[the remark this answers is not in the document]
shcheklein: good question. Unfortunately, CSV doesn't preserve column types, and this particular column has all nulls inside it, and automatic schema inference breaks. 

Tbh, I think I'll do parquet instead, but I need to look into that ... we support both parquet and CSV and parquet is a bit better (we can preserve types, pydantic model schemas, etc)

examples/llm_and_nlp/hf-dataset-llm-eval.py
```diff
@@ -0,0 +1,65 @@
...
+            },
+        ],
+        response_format=DialogEval,
+    )
+
+    message = completion.choices[0].message
+    if message.parsed:
+        return message.parsed
+
+    return DialogEval(result="Error", reason="Failed to parse response.")
```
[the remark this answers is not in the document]
shcheklein: yes, good ... I'm all in to make it HF specific and less dependencies since it'a HF integration example. Let me play with it a bit ..

examples/llm_and_nlp/hf-dataset-llm-eval.py
shcheklein: remove, (it reads it from HF_TOKEN atuomaticaly I think or from huggingface-cli login)

examples/llm_and_nlp/hf-dataset-llm-eval.py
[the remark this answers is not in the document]
shcheklein: removed it, working with parquet ... need to deal with CSV edge case later (if needed at all)

examples/llm_and_nlp/hf-dataset-llm-eval.py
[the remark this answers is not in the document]
shcheklein: applied, it's now all HF end to end

shcheklein:
@lhoestq simpler and HF end to end now. Any other ideas how we can make it more interesting / powerful? :)

(I still need to wrap this example as a test - it will take a bit of time - thus I keep it as a draft)

.github/workflows/tests.yml
```diff
@@ -152,4 +152,6 @@ jobs:
         run: uv pip install nox --system
 
       - name: Run examples
+        env:
+          HF_TOKEN: ${{ secrets.HF_TOKEN }}
```
shcheklein, line 156: @0x2b3bfa0 could you please review if it's safe (I have a budget limit on the token here, just in case) and if we need to add it to the list of credentials / tokens

0x2b3bfa0: Yes, it is safe (external contributors don't have access), and no, having it listed on GitHub Secrets is enough — more or less — to keep track of it.

examples/llm_and_nlp/hf-dataset-llm-eval.py
```diff
@@ -0,0 +1,59 @@
+from huggingface_hub import InferenceClient
```
mattseddon, line 1: @shcheklein this test looks flaky: https://github.com/iterative/datachain/actions/runs/11655280717/job/32449776121

shcheklein: yep, I'm looking into this ... probably bc we run it in parallel and that's a bit too much for a real e2e test ... I'll if we can run it on a single machine and python

lhoestq:
Great ! I can share the example on social media tomorrow if it sounds good to you :)

What about notebook demos for llm-based dataset operations like generation / filtering  / cleaning ? Let me know if I can help with some ideas or existing datasets / use cases for those

shcheklein:
> Great ! I can share the example on social media tomorrow if it sounds good to you

Yes, absolutely. That would be amazing! Let me know when you do this - we'll promote it through our channels as well.

> What about notebook demos for llm-based dataset operations like generation / filtering / cleaning ? Let me know if I can help with some ideas or existing datasets / use cases for those

yes. I was looking into Datasets a bit more to better understand how tools can complement each other.

Thoughts so far:

- HF datasets (and especially Hub) serves as an excellent collaboration layer for DataChain! Publishing and reading back (while perfectly preserving Pydantic structure, etc) DC datasets as parquet files to / from HF gives versioning, nice visualization, ability for someone else to read it, etc.
- AFAIU datasets is Arrow based and doesn't have it sown engine (I mean SQL in this case) vs DataChain (open source). It means it doesn't have out of core merge, group by? Can we find some interesting scenarios where we use DataChain to merge a few datasets for example?
- Another differentiator AFAIU - DC doesn't embed binary objects, it stores paths to objects in the cloud and can stream them to UDF when it is executed. I wonder if this can help with some use case? Have you see some datasets that have metadata on HF but objects outside?

> What about notebook demos for llm-based dataset operations like generation / filtering / cleaning

any examples from the top of your head btw?

lhoestq:
>  It means `datasets` doesn't have out of core merge, group by? Can we find some interesting scenarios where we use DataChain to merge a few datasets for example?

Yes indeed, `datasets` supports lager-than-memory datasets via memory mapping by default, and there is also a streaming option. So far the only option for group_by/merge has been to to `ds.to_polars()`.

> DC stores paths to objects in the cloud and can stream them to UDF when it is executed. I wonder if this can help with some use case? Have you see some datasets that have metadata on HF but objects outside?

I can help for sure. Most datasets contain the bytes themselves, but some of them have separate metadata and media files (audio/image/video), both on HF. There is also the rare case of storing images URLs.

> any examples from the top of your head btw?

Some examples: LLM for text extraction from scrapped web pages, LLM as a judge to filter chatbot dialogs based on success/failure, dataset generation using a LLM and personas for diversity

lhoestq:
Just shared the example on X and linkedin @shcheklein !
- https://x.com/qlhoest/status/1853826673973575946
- https://www.linkedin.com/posts/quentin-lhoest_dvc-datasets-dataproc-activity-7259595157660909568-uJix

dmpetrov:
thank you, @lhoestq, for sharing and for your valuable feedback!

PR #556, opened 2024-11-03
chore(hf): remove unused arguments

shcheklein:
Minor cleanup in HF integration code.

### DIFF
--- a/src/datachain/lib/hf.py
+++ b/src/datachain/lib/hf.py
@@ -142,11 +142,11 @@ def get_output_schema(features: Features) -> dict[str, DataType]:
     """Generate UDF output schema from huggingface datasets features."""
     fields_dict = {}
     for name, val in features.items():
-        fields_dict[name] = _feature_to_chain_type(name, val)  # type: ignore[assignment]
-    return fields_dict  # type: ignore[return-value]
+        fields_dict[name] = _feature_to_chain_type(name, val)
+    return fields_dict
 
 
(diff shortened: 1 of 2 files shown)


src/datachain/lib/hf.py
```diff
@@ -138,17 +138,15 @@ def convert_feature(val: Any, feat: Any, anno: Any) -> Any:  # noqa: PLR0911
...
-        fields_dict[name] = _feature_to_chain_type(name, val)  # type: ignore[assignment]
-    return fields_dict  # type: ignore[return-value]
+        fields_dict[name] = _feature_to_chain_type(name, val)
+    return fields_dict
 
 
-def _feature_to_chain_type(name: str, val: Any) -> type:  # noqa: PLR0911
+def _feature_to_chain_type(name: str, val: Any) -> DataType:  # noqa: PLR0911
     if isinstance(val, Value):
         return arrow_type_mapper(val.pa_type)
```
dberenbaum, line 151: I see `arrow_type_mapper` also returns `type` but seems like it should also return `DataType`, right? Maybe worth including in the PR. I think this stems from `arrow_type_mapper` predating `DataType`.

dberenbaum (approved):
Good catches @shcheklein!

src/datachain/lib/hf.py
shcheklein: yes, more cleanup / review is needed there with types

e.g. we return `list(list ...` and it's not really `DataType` AFAIU

it's not a complete solution, thus we still have a lot of ignore type pragmas

PR #557, opened 2024-11-03
Add 'BoundingBox', 'Pose2D' and 'Pose3D' models

dreadatour:
These models will be used in SaaS to visualize signals.


src/datachain/lib/file.py
```diff
@@ -20,9 +20,6 @@
 from pyarrow.dataset import dataset
 from pydantic import Field, field_validator
 
-if TYPE_CHECKING:
```
dreadatour, line 23: Removing this because we have exactly the same code below: https://github.com/iterative/datachain/pull/557/files#diff-8a0312f492b4f25b4ebffe0d763f8677e0f07e06e1c4c672fc33d89a5e10bd64R29

shcheklein (approved):
let's add some basic unit tests for this?

ilongin (approved):
LGTM

PR #558, opened 2024-11-04
guarantee order if order_by directly precedes collect in a chain

mattseddon:
Follow up to #518. More specifically [this comment](https://github.com/iterative/datachain/pull/518/files#r1805859400).


shcheklein (approved):
Just to double check - will `show` also work for this (probably it is using one of the collects inside). Do we want to add a test to make sure it is also preserving the order?

mattseddon:
> Just to double check - will `show` also work for this (probably it is using one of the collects inside). Do we want to add a test to make sure it is also preserving the order?

Test is [here](https://github.com/iterative/datachain/pull/558/files#diff-b40ca373e4b7558183b740420a85bc3d95c55a4b3049addcbbc1cf0b04460eecR452)

PR #560, opened 2024-11-04
Dataset pull fixes

ilongin:
Fixing various issues regarding dataset pull after recent feature based schema changes:
1. Adding `UInt32` type as it's needed for Clickhouse random column as `UInt64` is too big for SQLlite
2. Removing casting dataframe data to string - it seems it's not needed anymore
3. Fix parsing empty string as json

### DIFF
--- a/src/datachain/sql/types.py
+++ b/src/datachain/sql/types.py
@@ -187,6 +187,22 @@ class Int32(Int):
         return read_converter(dialect).int32(value)
 
 
+class UInt32(Int):
+    def load_dialect_impl(self, dialect):
+        return converter(dialect).uint32()
+
+    @staticmethod
+    def default_value(dialect):
+        return type_defaults(dialect).uint32()
+
+    @staticmethod
+    def db_default_value(dialect):
+        return db_defaults(dialect).uint32()
+
+    def on_read_convert(self, value, dialect):
+        return read_converter(dialect).uint32(value)
+
+
 class Int64(Int):
     def load_dialect_impl(self, dialect):
         return converter(dialect).int64()
@@ -395,6 +411,9 @@ class TypeReadConverter:
     def int32(self, value):
         return value
 
+    def uint32(self, value):
+        return value
+
     def int64(self, value):
         return value
 
@@ -421,6 +440,8 @@ class TypeReadConverter:
 
     def json(self, value):
         if isinstance(value, str):
+            if value == "":
+                return {}
             return orjson.loads(value)
         return value
 
@@ -444,6 +465,9 @@ class TypeConverter:
         return types.Integer()
 
     def int32(self):
+        return self.int()
+
+    def uint32(self):
         return self.int()
 
     def int64(self):
@@ -487,6 +511,9 @@ class TypeDefaults:
     def int32(self):
         return None
 
+    def uint32(self):
+        return None
+
     def int64(self):
         return None
 
@@ -528,6 +555,9 @@ class DBDefaults:
     def int32(self):
         return self.int()
 
+    def uint32(self):
+        return self.int()
+
     def int64(self):
         return self.int()
 
@@ -561,6 +591,7 @@ TYPES = [
     Boolean,
     Int,
     Int32,
+    UInt32,
     Int64,
     UInt64,
     Float,
--- a/src/datachain/sql/types.py
+++ b/src/datachain/sql/types.py
@@ -187,6 +187,22 @@ class Int32(Int):
         return read_converter(dialect).int32(value)
 
 
+class UInt32(Int):
+    def load_dialect_impl(self, dialect):
+        return converter(dialect).uint32()
+
+    @staticmethod
+    def default_value(dialect):
+        return type_defaults(dialect).uint32()
+
+    @staticmethod
+    def db_default_value(dialect):
+        return db_defaults(dialect).uint32()
+
+    def on_read_convert(self, value, dialect):
+        return read_converter(dialect).uint32(value)
+
+
 class Int64(Int):
     def load_dialect_impl(self, dialect):
         return converter(dialect).int64()
@@ -395,6 +411,9 @@ class TypeReadConverter:
     def int32(self, value):
         return value
 
+    def uint32(self, value):
+        return value
+
     def int64(self, value):
         return value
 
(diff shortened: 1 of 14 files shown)


mattseddon:
`test_row_random` looks like a genuine failure.

amritghimire:
I am getting the error as below when using the pull using `datachain pull ds://02jpg_files -o jpf -vv`

```
Version not specified, pulling the latest one (v1)
Saving dataset ds://02jpg_files@v1 locally: 100%|█████████████████████████████████████████████████████████████████████| 5.00/5.00 [00:00<00:00, 10.2 rows/s]
Dataset ds://02jpg_files@v1 saved locally
_request non-retriable exception: Anonymous caller does not have storage.buckets.get access to the Google Cloud Storage bucket. Permission 'storage.buckets.get' denied on resource (or it may not exist)., 401
  File ".../datachain/.venv/lib/python3.12/site-packages/gcsfs/retry.py", line 126, in retry_request
    return await func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../datachain/.venv/lib/python3.12/site-packages/gcsfs/core.py", line 440, in _request
    validate_response(status, contents, path, args)
  File ".../datachain/.venv/lib/python3.12/site-packages/gcsfs/retry.py", line 113, in validate_response
    raise HttpError(error)
gcsfs.retry.HttpError: Anonymous caller does not have storage.buckets.get access to the Google Cloud Storage bucket. Permission 'storage.buckets.get' denied on resource (or it may not exist)., 401
Error: Dataset lst__gs://datachain-demo/ not found.
  File ".../datachain/src/datachain/cli.py", line 1016, in main
    catalog.pull_dataset(
  File ".../datachain/src/datachain/catalog/catalog.py", line 1454, in pull_dataset
    _instantiate_dataset()
  File ".../datachain/src/datachain/catalog/catalog.py", line 1324, in _instantiate_dataset
    self.cp(
  File ".../datachain/src/datachain/catalog/catalog.py", line 1563, in cp
    node_groups = self.enlist_sources_grouped(
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File .../iterative/datachain/src/datachain/catalog/catalog.py", line 703, in enlist_sources_grouped
    listing = Listing(st, client, self.get_dataset(dataset_name))
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../datachain/src/datachain/catalog/catalog.py", line 1090, in get_dataset
    return self.metastore.get_dataset(name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../src/datachain/data_storage/metastore.py", line 704, in get_dataset
    raise DatasetNotFoundError(f"Dataset {name} not found.")
datachain.error.DatasetNotFoundError: Dataset lst__gs://datachain-demo/ not found.
Telemetry is disabled by environment variable.
```


The datasets were created using:
```
from datachain import DataChain, C

ds = DataChain.from_storage("gs://datachain-demo")
ds1 = ds.filter(C('file.path').glob('*.jpg')).save("jpg_files")
ds2 = ds.filter(C('file.path').glob('*.png')).save("png_files")
ds4 = ds1.union(ds2)
ds5 = ds4.save("jpg_png_files")

dsf1 = ds.filter(C("file.path").glob("*02.jpg")).limit(5).save("02jpg_files")

ds6 = ds5.union(dsf1)
ds6.save("all_files")
```

ilongin:
> datachain pull ds://02jpg_files -o jpf -vv

Thanks for finding the issue.

This is more related to instantiating dataset for which I will create a separate issue as it's not so trivial to fix atm. I would merge current PR to at least have dataset pull without instantiation working fine (which is enough in majority of use cases for clients)

ilongin:
> `test_row_random` looks like a genuine failure.

I've fixed the test. Eventually I decided to use 64 bit for random number but limit it to 2^63 - 1 which is the largest possible value that can be stored in SQLite as it only supports signed numbers. Clickhouse supports unsigned ones, so the largest value is 2^64 but as I mentioned, 2^63 - 1 is largest possible value supported in both DBs. This gives us much larger number "pool" then changing this to 32 bit int.

src/datachain/sql/types.py
```diff
@@ -187,6 +187,22 @@ def on_read_convert(self, value, dialect):
         return read_converter(dialect).int32(value)
 
 
+class UInt32(Int):
```
mattseddon, line 190: [Q] Does this need to be added? (now that it isn't being used - seems unrelated to this PR)

tests/func/test_datasets.py
```diff
@@ -827,10 +827,7 @@ def test_row_random(cloud_test_catalog):
     # Random values are unique
     assert len(set(random_values)) == len(random_values)
 
-    if isinstance(catalog.warehouse, SQLiteWarehouse):
-        RAND_MAX = 2**63  # noqa: N806
-    else:
-        RAND_MAX = 2**64  # noqa: N806
+    RAND_MAX = DataTable.MAX_RANDOM  # noqa: N806
```
mattseddon, line 830: [Q] Won't this test fail when the number drawn by clickhouse is > 2^63?

And if my math is correct the chances of that happening is:

((2^64) - (2^63)) / (2^64) = 0.5 or 50%

src/datachain/sql/types.py
ilongin: It doesn't, but now that I've already added it, I feel bad removing it :D ... I think we actually had the issue to add it anyway (cannot find it atm) as it's kind of missing (we have support for `int32`, `int64` and `uint64`).

tests/func/test_datasets.py
ilongin: In Clickhouse we have limit to 2^63-1 https://github.com/iterative/studio/pull/10860/files#diff-1136c5032c82c12345c3103e18093d3256798daa2d4b13e5f5f2d5e8670ac32bR328

src/datachain/sql/types.py
mattseddon: Can you see how this is unrelated to this PR?

ilongin: Of course I can see, If you really want this to be by PS I will split into 2 PRs, not a problem, but I wanted to avoid that additional work in this relatively small issue...that was my point.

ilongin: Split into separate PR https://github.com/iterative/datachain/pull/568

PR #561, opened 2024-11-04
feat(cli): enhance dataset listing with additional options for studio

amritghimire:
With this change, the following variation are available to list datasets
instead of `datachain ls-datasets`:

- `datachain datasets --local`
- `datachain datasets --studio`
```
Name             Version
-------------  ---------
jpg_files              1
all_files              1
jpg_png_files          1
png_files              1
02jpg_files            1
```

- `datachain datasets --all` (Default option)
- `datachain datasets --team TEAM_NAME`
- `datachain datasets`

```
Name                       Version  Studio    Local
-----------------------  ---------  --------  -------
jpg_files                        1  ✔         ✔
hello                            2  ✔         ✖
02jpg_files                      1  ✔         ✔
embeddings-and-metadata          1  ✔         ✖
```

By default, if the user has logged in to Studio using `datachain studio login`,
it will try to fetch the datasets from both local and studio.
If any specific option (local or studio) is passed, only that option is
passed.

The team name is parsed from config (Set by  `datachain studio team <TEAM_NAME>`).

The same feature is added for `datachain ls` function and added the
error handling for each scenarios.

Relates to #10774

### DIFF
--- a/src/datachain/cli.py
+++ b/src/datachain/cli.py
@@ -4,7 +4,7 @@ import shlex
 import sys
 import traceback
 from argparse import Action, ArgumentParser, ArgumentTypeError, Namespace
-from collections.abc import Iterable, Iterator, Mapping, Sequence
+from collections.abc import Iterable, Iterator, Sequence
 from importlib.metadata import PackageNotFoundError, version
 from itertools import chain
 from multiprocessing import freeze_support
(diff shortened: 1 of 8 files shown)


src/datachain/cli.py
```diff
@@ -789,25 +841,55 @@ def ls_remote(
...
-    config: Optional[Mapping[str, str]] = None,
+    studio: bool = False,
+    local: bool = False,
+    all: bool = False,
+    team: Optional[str] = None,
     **kwargs,
 ):
-    if config is None:
-        from .config import Config
+    token = Config().read().get("studio", {}).get("token")
```
mattseddon, line 850: [I] Both `ls` and `datasets` now run the same logic to determine whether or not to call the local or studio flavour of their underlying commands. Can you pull-out the duplication, please?

[C] I would also say that IMO the default should be to reach out to Studio only if the `studio` or `all` flag is passed. Forget about the auto-token magic. 

I would also say that if `all` is passed that we want to display a table instead of two lists:

| name | version | studio | local |
|--|--|--|--|
| cats | 1 | ✅ | ❌ |
| cats | 2 | ✅ | ❌ | 
| dogs | 1 | ✅ | ✅ | 

That way I can easily check the discrepancies between the two. Instead of manually having to compare the two lists.

PR #570, opened 2024-11-07
pin version of clickhouse to 24.6 (match clickhouse-cloud)

mattseddon:
Pointed out by @dreadatour - Clickhouse cloud currently uses version 24.6 so we should too.


dreadatour (approved):
🙏

PR #572, opened 2024-11-07
Add studio flag for rm-dataset and edit-dataset

amritghimire:
This adds a support for `--studio` flag for edit-dataset and rm-dataset
command. If the --studio flag is passed, it will use the studio client
to process the operation.

Some example are as:
- `datachain rm-dataset "new_test_dataset" --studio --version 1`
- `datachain edit-dataset png_files  --studio --new-name new_dataset_name`

TODO:
- [x] Add test

Studio PR: https://github.com/iterative/studio/pull/10890


shcheklein:
Can be a follow up, but let's rename all these command (similar to `ls-dataset` that we had). Can we make it all part of the `datachain datasets` umbrella for example - at least for now. (make `datasets` alias with `ds` to make it simpler)

datachain ds ls
datachain ds rm
datachain ds rename (better to have an explicit command vs one edit that is not very clear in terms of its semantics).

shcheklein:
We need to figure out docs for this as well.

src/datachain/cli.py
```diff
@@ -403,21 +408,30 @@ def get_parser() -> ArgumentParser:  # noqa: PLR0915
...
         "--labels",
-        default=[],
         nargs="+",
         help="Dataset labels",
     )
+    parse_edit_dataset.add_argument(
+        "--studio",
+        action="store_true",
+        default=False,
+        help="Remove dataset from Studio",
```
mattseddon, line 427: [Q] Is this correct?

src/datachain/cli.py
```diff
@@ -909,8 +935,28 @@ def rm_dataset(
...
+def edit_dataset(
+    catalog: "Catalog",
+    name: str,
+    new_name: Optional[str] = None,
+    description: Optional[str] = None,
+    labels: Optional[list[str]] = None,
+    studio: bool = False,
+    team: Optional[str] = None,
+):
+    if studio:
```
mattseddon, line 956: [Q] Is there a reason for deviating from the behaviour implemented in #561? Why can't a user edit/remove a dataset from both local and remote in the same call?

amritghimire: For list operation, having the option to list the datasets from both local and remote seems like a frequent operation. 

But I am not sure how helpful it will edit or remove dataset from both local and remote in the same call.

 I was also looking at the implementation of `pull` where `ds://` prefix is used to identify if it is remote or if it is local. I don't know if it will be confusing to user though to implement it in edit and remove as well. cc. @shcheklein

src/datachain/remote/studio.py
```diff
@@ -230,6 +222,46 @@ def ls(self, paths: Iterable[str]) -> Iterator[tuple[str, Response[LsData]]]:
...
+            body["new_name"] = new_name
+
+        if description is not None:
+            body["description"] = description
+
+        if labels is not None:
+            body["labels"] = labels  # type: ignore[assignment]
+
+        return self._send_request(
+            "datachain/edit-dataset",
```
shcheklein, line 246: related to the groups of commands, etc - let's please make API also clean

`datachain/datasets/ls` 
`datachain/datasets/rename` 

-etc 

`rm-dataset` and similar look very weird for REST api.

skshetry: It could even be a single endpoint that lists on `GET`, renames via `PUT` (assuming dataset name is not part of a resource URI as `PUT` has to be idempotent), removes via `DELETE`.

Since it's a part of a team, ideally the URI should be `/datachain/<team>/datasets`.

amritghimire: Added this comment to the #576 as well. We can take it as a part of refactor.

src/datachain/cli.py
mattseddon: > But I am not sure how helpful it will edit or remove dataset from both local and remote in the same call.

If you think about this in terms of https://github.com/iterative/datachain/issues/578 users would definitely want to be able to rename both at the same time. We want to avoid users ending up with datasets with the same identifier being named differently between local and remote as much as possible.

amritghimire: Addressed,.

dreadatour (approved):
Looks good to me.

Do we have tests for this on Studio side? 🤔

amritghimire:
> Looks good to me.
> 
> 
> 
> Do we have tests for this on Studio side? 🤔

Yes we have.

PR #577, opened 2024-11-08
docs: add cli docs

yathomasi:
From [comment](https://github.com/iterative/datachain/pull/561#pullrequestreview-2421829495)
> How do we go about docs for these commands? Let's please review and created tickets if needed.

Do we mean sth like this ?

Screenshot 2024-11-08 at 12 36 38

Putting as draft as we need to include all the commands before we merge.

### DIFF
--- a/docs/cli/index.md
+++ b/docs/cli/index.md
@@ -1,0 +1,5 @@
+# CLI Documentation
+
+## Studio
+
+::: datachain.studio.login
--- a/docs/cli/index.md
+++ b/docs/cli/index.md
@@ -1,5 +1,5 @@
 # CLI Documentation
 
-## Studio
-
-::: datachain.studio.login
+::: mkdocs-argparse
+    :module: datachain.cli
+    :function: get_parser
--- a/docs/css/mkdocstrings.css
+++ b/docs/css/mkdocstrings.css
@@ -1,0 +1,5 @@
+@media screen and (min-width: 76.25em) {
+  .md-header__topic {
+      display: none;
+  }
+}
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -63,7 +63,7 @@ theme:
         name: Switch to system preference
 
 extra_css:
-- css/mkdocstrings.css
+  - css/mkdocstrings.css
 
 nav:
   - Home: index.md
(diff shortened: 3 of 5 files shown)


shcheklein:
> Do we mean sth like this ?

No. It looks like Python API docs to me. It should be more like what we have in dvc.org for CLI. I'm not sure if there are good example for this engine.

yathomasi:
I see `mkdocs` support for cli `autodocs` with `mkdocs-click` extension i.e for cli using `click`. But, for us, I see we use `argparse`. 

@iterative/datachain I see we lack proper restructuring for cli commands. So, could you please check the feasibility of using [click](https://realpython.com/python-click/) for proper structure? Also, we can generate the docs using `mkdocs-click`.

If we don't want to take that road, we should basically follow the same procedure as we do for `dvc.org` i.e write them manually for all the commands.

shcheklein:
> So, could you please check the feasibility of using [click](https://realpython.com/python-click/) for proper structure

tbh, I don't like click. To my mind it doesn't provide enough value for an extra dependency, standard lib gives 95% of what we need. Also, `click` is breaking some basic (to my mind) things. E.g. there is no way you could append `-v` at the end of the command to debug it (you have to put in the right position, somewhere in between all args).

> If we don't want to take that road, we should basically follow the same procedure as we do for dvc.org i.e write them manually for all the commands.

@skshetry what is your take on this?

yathomasi:
I could only find way to `autodoc generation` with `mkdocs` for `cli commands` is using `mkdocs-click` or `mkdocs-types` of which we don't use both. ([plugins list](https://github.com/mkdocs/catalog#-api-documentation-building))

 It would be great if there was a solution. If not I don't see alternative of doing them manually. 

cc: @iterative/datachain @shcheklein

0x2b3bfa0:
Maybe someone is crazy enough to prototype[^1] [`mkdocs-argparse`](https://pypi.org/project/mkdocs-argparse/) during a planning session, don't you think? 😈 

[^1]: Still missing a few important things: deduplication of global command-line options, tests...

0x2b3bfa0:
Given that there's already [`mkdocs-typer`](https://github.com/bruce-szalwinski/mkdocs-typer) and [`typer`](https://typer.tiangolo.com) is such a nice library, maybe it's an option worth considering as well.

0x2b3bfa0:
Writing high-quality documentation by hand and maintaining it is also a good option, but it's a lot of work.

yathomasi:
Closing this for now.

PR #579, opened 2024-11-09
Add option to run the datachain query to Studio

amritghimire:
The options are:

positional arguments:
```
  query_file            The query file to run.
```

options:
```
--team TEAM           The team to run a job for. By default, it will use team from config.
  --env-file ENV_FILE   File containing environment variables to set for the job.
  --envs ENVS [ENVS ...]
                        Environment variables to set for the job.
  --workers WORKERS     Number of workers to use for the job.
  --files FILES [FILES ...]
                        Files to include in the job.
  --python-version PYTHON_VERSION
                        Python version to use for the job (e.g. '3.9', '3.10', '3.11').
  --req-file REQ_FILE   File containing Python package requirements.
  --reqs REQS [REQS ...]
                        Python package requirements.
```

Example run:
------------
Example script to run
```sh
$ datachain studio run example_query.py --env-file=env_file.txt --envs="ENV_FROM_ARGS=1" --workers=2 --files file.txt --python-version=3.12 --req-file=reqs.txt --reqs="oneliners"
```

Files:
------
`run/env_file.txt`:

```
ENV_FROM_FILE = 'environments.txt'

```
`run/file.txt`

```
content from file

```

`run/reqs.txt`

```
pyjokes
```

`run/example_query.py`

```py

from os import environ
from oneliners import get_random
import pyjokes

# Define the UDF:
def path_len(path):
    if path.endswith(".json"):
        return (-1,)
    return (len(path),)

if __name__ == "__main__":
    # Run in chain
    print("Environment set from file:", environ["ENV_FROM_FILE"])
    print("Environment set from args:", environ["ENV_FROM_ARGS"])
    print("Oneliners from reqs(args):", get_random())
    print("Joke from pyjokes:(from reqs file)", pyjokes.get_joke())

    print("Content from files(args):", open("file.txt").read())

    DataChain.from_storage(
        uri="gs://datachain-demo/dogs-and-cats/",
    ).map(
        path_len,
        params=["file.path"],
        output={"path_len": int},
    ).show()

```

TODO:

- Rename the argument names to better names
- [x] Add tests

Companion PR: https://github.com/iterative/studio/pull/10897


src/datachain/cli.py
```diff
@@ -228,6 +228,67 @@ def add_studio_parser(subparsers, parent_parser) -> None:
...
+    studio_run_parser.add_argument(
+        "--env-file",
+        action="store",
+        help="File containing environment variables to set for the job.",
+    )
+
+    studio_run_parser.add_argument(
+        "--envs",
+        nargs="+",
+        help="Environment variables to set for the job.",
```
dreadatour, line 262: Do we want to be more verbose on how to set env variables via `--envs` flag? Also consider renaming this to `--env`.

src/datachain/cli.py
```diff
@@ -228,6 +228,67 @@ def add_studio_parser(subparsers, parent_parser) -> None:
...
+    studio_run_parser.add_argument(
+        "--req-file",
+        action="store",
+        help="File containing Python package requirements.",
+    )
+
+    studio_run_parser.add_argument(
+        "--reqs",
+        nargs="+",
+        help="Python package requirements.",
```
dreadatour, line 289: Same here, more information on how to add requirements from command-line flag would be nice to have.

src/datachain/studio.py
```diff
@@ -139,3 +152,69 @@ def save_config(hostname, token):
...
+    team_name: Optional[str],
+    env_file: Optional[str] = None,
+    envs: Optional[list[str]] = None,
+    workers: Optional[int] = None,
+    files: Optional[list[str]] = None,
+    python_version: Optional[str] = None,
+    reqs: Optional[str] = None,
+    req_file: Optional[str] = None,
+):
+    query_type = "PYTHON" if query_file.endswith(".py") else "SHELL"
```
dreadatour, line 168: This might be a problem in future. I think we should let user choose file type either by additional argument or even by separate command.

src/datachain/studio.py
```diff
@@ -139,3 +152,69 @@ def save_config(hostname, token):
...
+    with open(query_file) as f:
+        query = f.read()
```
dreadatour, line 170: Can we check if this is a valid Python file before sending it may be?

src/datachain/studio.py
```diff
@@ -139,3 +152,69 @@ def save_config(hostname, token):
...
+
+    environment = "\n".join(envs) if envs else ""
+    if env_file:
+        with open(env_file) as f:
+            environment = f.read() + "\n" + environment
```
dreadatour, line 175: Empty line in the beginning if no `envs` defined, but that's may be ok.

src/datachain/studio.py
```diff
@@ -139,3 +152,69 @@ def save_config(hostname, token):
...
+
+    requirements = "\n".join(reqs) if reqs else ""
+    if req_file:
+        with open(req_file) as f:
+            requirements = f.read() + "\n" + requirements
```
dreadatour, line 180: Same, empty line in the beginning if no `reqs` defined.

dreadatour (approved):
Looks good to me.

In the follow-up issues/PRs we may want to stream state/logs/progress from Studio.

Also how can we check job execution status? May be first step might be to add `job status` command?

And `job logs` command before streaming may be?

src/datachain/cli.py
shcheklein: agreed, on the naming (and probably `--req`)

PR #583, opened 2024-11-11
Update README.rst, added missing preposition.

gromag:
Minor syntax update to add missing preposition "from"


shcheklein (approved):
Thanks @gromag !

PR #586, opened 2024-11-12
doc(agg): add an example with custom UDF spec

shcheklein:
Add an explicit example on an alternative syntax for `agg` (I spent some time to figure our the appropriate signature).

### DIFF
--- a/src/datachain/lib/dc.py
+++ b/src/datachain/lib/dc.py
@@ -895,12 +895,32 @@ class DataChain:
         2. Group-based UDF function input: Instead of individual rows, the function
            receives a list all rows within each group defined by `partition_by`.
 
-        Example:
+        Examples:
             ```py
             chain = chain.agg(
                 total=lambda category, amount: [sum(amount)],
                 output=float,
                 partition_by="category",
+            )
+            chain.save("new_dataset")
+            ```
+
+            An alternative syntax, when you need to specify a more complex function:
+
+            ```py
+            # It automatically resolves which columns to pass to the function
+            # by looking at the function signature.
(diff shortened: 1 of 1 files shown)


src/datachain/lib/dc.py
```diff
@@ -904,6 +904,23 @@ def agg(
...
+            # It automatically resolves which columns to pass to the function
+            # by looking at the function signature.
+            def agg_sum(file: list[File],
+                        amount: list[float]) -> Iterator[tuple[File, float]]:
+                yield file[0], sum(amount)
+
+            chain = chain.agg(
+                agg_sum,
+                output={"file": File, "total": float},
+                # Alternative syntax is to use `C`
```
ilongin, line 918: I would maybe merge these 2 sentences into one... "Alternative syntax is to use `C` which is useful for nested columns ..."

src/datachain/lib/dc.py
```diff
@@ -904,6 +904,23 @@ def agg(
             )
             chain.save("new_dataset")
             ```
+
+            ```py
+            # It automatically resolves which columns to pass to the function
+            # by looking at the function signature.
+            def agg_sum(file: list[File],
```
ilongin, line 911: I would reformat this to be black "compatible"
```python
def agg_sum(
    file: list[File], amount: list[float]
) -> Iterator[tuple[File, float]]:
    yield file[0], sum(amount)
```

ilongin (approved):
Left 2 small comments

shcheklein:
thanks @ilongin for reviewing it!

PR #595, opened 2024-11-14
Fix 'limit' + 'offset' methods for SQLite warehouse

dreadatour:
This PR fixes `offset` method for DatasetQuery/DataChain for SQLite warehouse (it works fine in SaaS).

+ added tests (fail before fix: https://github.com/iterative/datachain/actions/runs/11829848002)

### DIFF
--- a/tests/func/test_dataset_query.py
+++ b/tests/func/test_dataset_query.py
@@ -459,6 +459,77 @@ def test_order_by_limit(cloud_test_catalog, save, animal_dataset):
     ]
 
 
+@pytest.mark.parametrize("save", [True, False])
+@pytest.mark.parametrize(
+    "cloud_type,version_aware",
+    [("s3", True)],
+    indirect=True,
+)
+def test_limit(cloud_test_catalog, save, animal_dataset):
+    catalog = cloud_test_catalog.catalog
+    q = DatasetQuery(animal_dataset.name, catalog=catalog).limit(2)
+    if save:
+        ds_name = "animals_cats"
+        q.save(ds_name)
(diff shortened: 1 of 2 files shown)


tests/func/test_dataset_query.py
```diff
@@ -459,6 +459,77 @@ def test_order_by_limit(cloud_test_catalog, save, animal_dataset):
     ]
 
 
+@pytest.mark.parametrize("save", [True, False])
+@pytest.mark.parametrize(
+    "cloud_type,version_aware",
+    [("s3", True)],
+    indirect=True,
+)
+def test_limit(cloud_test_catalog, save, animal_dataset):
```
mattseddon, line 468: [Q] Any reason you've added these tests instead of testing the exposes DataChain API?

mattseddon (approved):

tests/func/test_dataset_query.py
dreadatour: I was thinking it is more close to the internals, since exposed DataChain API still uses the internal methods.

tests/func/test_dataset_query.py
```diff
@@ -459,6 +459,77 @@ def test_order_by_limit(cloud_test_catalog, save, animal_dataset):
...
+        result = q.db_results()
+
+    assert len(result) == 2
+    assert [posixpath.basename(r[3]) for r in result] == ["cat1", "cat2"]
+
+
+@pytest.mark.parametrize("save", [True, False])
+@pytest.mark.parametrize(
+    "cloud_type,version_aware",
+    [("s3", True)],
```
skshetry, line 487: Please don't use `s3` fixtures unless you are really testing the clouds-related functionality.

PR #604, opened 2024-11-16
Split Dataset into train/test/val

dreadatour:
See related issue, docstrings and tests.

##### Train-test split

```python
from datachain.toolkit import train_test_split

# Load a DataChain from a storage source (e.g., S3 bucket)
dc = DataChain.from_storage("s3://bucket/dir/")

# Perform a 70/30 train-test split
train, test = train_test_split(dc, [0.7, 0.3])

# Save the resulting splits
train.save("dataset_train")
test.save("dataset_test")
```

##### Train-test-validation split

```python
train, test, val = train_test_split(dc, [0.7, 0.2, 0.1])
train.save("dataset_train")
test.save("dataset_test")
val.save("dataset_val")
```

### TODO

- `seed`
- `shuffle`

### DIFF
--- a/src/datachain/toolkit/split.py
+++ b/src/datachain/toolkit/split.py
@@ -1,0 +1,66 @@
+from datachain import C, DataChain
+
+
+def train_test_split(dc: DataChain, weights: list[float]) -> list[DataChain]:
+    """
+    Splits a DataChain into multiple subsets based on the provided weights.
+
+    This function partitions the rows or items of a DataChain into disjoint subsets,
+    ensuring that the relative sizes of the subsets correspond to the given weights.
+    It is particularly useful for creating training, validation, and test datasets.
+
+    Args:
+        dc (DataChain): The DataChain instance to split.
+        weights (list[float]): A list of positive weights indicating the relative
+                               proportions of the splits.
+
+    Returns:
+        list[DataChain]:
+            A list of DataChain instances, one for each weight in the weights list.
+
+    Examples:
+        Train-test split:
+        ```python
+        from datachain import DataChain
+        from datachain.toolkit import train_test_split
+
+        # Load a DataChain from a storage source (e.g., S3 bucket)
+        dc = DataChain.from_storage("s3://bucket/dir/")
+
+        # Perform a 70/30 train-test split
+        train, test = train_test_split(dc, [0.7, 0.3])
+
+        # Save the resulting splits
+        train.save("dataset_train")
+        test.save("dataset_test")
+        ```
+
+        Train-test-validation split:
+        ```python
+        train, test, val = train_test_split(dc, [0.7, 0.2, 0.1])
+        train.save("dataset_train")
+        test.save("dataset_test")
+        val.save("dataset_val")
+        ```
+
+    Notes:
+        - The splits are random but deterministic, based on Dataset `sys__rand` field.
+        - The weights do not need to sum to 1; they will be normalized internally.
+          For example:
+            - `[0.7, 0.3]` corresponds to a 70/30 split;
+            - `[2, 1, 1]` corresponds to a 50/25/25 split;
+    """
+    if len(weights) < 2:
+        raise ValueError("Weights should have at least two elements")
+
+    weights_normalized = [round(weight * 1000 / sum(weights)) for weight in weights]
+
+    res = []
+    for i in range(len(weights_normalized)):
+        args = []
+        if i > 0:
+            args.append(C("sys__rand") % 1000 >= sum(weights_normalized[:i]))
+        if i < len(weights_normalized) - 1:
+            args.append(C("sys__rand") % 1000 < sum(weights_normalized[: i + 1]))
+        res.append(dc.filter(*args))
+    return res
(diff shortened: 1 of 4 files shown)


src/datachain/toolkit/split.py
```diff
@@ -0,0 +1,69 @@
...
+        - The weights do not need to sum to 1; they will be normalized internally.
+          For example:
+            - `[0.7, 0.3]` corresponds to a 70/30 split;
+            - `[2, 1, 1]` corresponds to a 50/25/25 split;
+    """
+    if len(weights) < 2:
+        raise ValueError("Weights should have at least two elements")
+
+    weights_normalized = [round(weight * 1000 / sum(weights)) for weight in weights]
+    print(weights_normalized)
```
0x2b3bfa0, line 57: Is this `print` intentional?

dreadatour: 🤫

src/datachain/toolkit/split.py
```diff
@@ -0,0 +1,66 @@
...
+        - The splits are random but deterministic, based on Dataset `sys__rand` field.
+        - The weights do not need to sum to 1; they will be normalized internally.
+          For example:
+            - `[0.7, 0.3]` corresponds to a 70/30 split;
+            - `[2, 1, 1]` corresponds to a 50/25/25 split;
+    """
+    if len(weights) < 2:
+        raise ValueError("Weights should have at least two elements")
+
+    weights_normalized = [round(weight * 1000 / sum(weights)) for weight in weights]
```
0x2b3bfa0, line 56: Minor nitpick: consider making these values truly "normalized" in range 0-1, and moving the `* 1000` below, closer to the `%` operation.

0x2b3bfa0 (approved):

src/datachain/toolkit/split.py
0x2b3bfa0: Also, I have been playing a bit with this normalization approach and sometimes I have found unexpected results:

```python
>>> sum(normalize([5, 5]))
1000
>>> sum(normalize([3, 3, 1]))
1001  # What!?
```

Not sure how important is that for this use case (maybe it's not), but looks like a potential subtle bug to me.

dreadatour: I guess this is because of `round`. Have it removed.

dreadatour:
```
In [1]: weights = [3, 3, 1]

In [2]: [weight / sum(weights) for weight in weights]
Out[2]: [0.42857142857142855, 0.42857142857142855, 0.14285714285714285]

In [3]: sum([weight / sum(weights) for weight in weights])
Out[3]: 1.0
```

shcheklein (approved):
Amazing, thanks @dreadatour

src/datachain/toolkit/split.py
```diff
@@ -0,0 +1,67 @@
...
+        raise ValueError("Weights should have at least two elements")
+    if any(weight < 0 for weight in weights):
+        raise ValueError("Weights should be non-negative")
+
+    weights_normalized = [weight / sum(weights) for weight in weights]
+
+    return [
+        dc.filter(
+            C("sys__rand") % 1000 >= round(sum(weights_normalized[:index]) * 1000),
+            C("sys__rand") % 1000 < round(sum(weights_normalized[: index + 1]) * 1000),
```
dmpetrov, line 64: Please do not use double underscore. It's technical detail, not user facing API.

src/datachain/toolkit/split.py
```diff
@@ -0,0 +1,67 @@
...
+    if len(weights) < 2:
+        raise ValueError("Weights should have at least two elements")
+    if any(weight < 0 for weight in weights):
+        raise ValueError("Weights should be non-negative")
+
+    weights_normalized = [weight / sum(weights) for weight in weights]
+
+    return [
+        dc.filter(
+            C("sys__rand") % 1000 >= round(sum(weights_normalized[:index]) * 1000),
```
dmpetrov, line 63: should it be a bigger number like 100K to cover more precise ratios like [99.007 0.001, 0.002]?

dmpetrov (commented):
Amazing!!! Just a couple of comments are inline.

src/datachain/toolkit/split.py
0x2b3bfa0: Yes! 💯 If we know the exact bounds of `sys__rand` (do we?) we should use them to provide as much precision as possible.

dmpetrov: it's a random Int, so IntMax ~2B. I'd not be too crazy about this and make it like 100K or 1M.

0x2b3bfa0: If it's a 32-bit ([un]signed?) integer, we can perfectly cover the entire range. 🤷‍♂️

dmpetrov: yeah, bit it might depend on architecture/DB that can potentially complicate the logic. If not - sure.

0x2b3bfa0: Moved here: https://github.com/iterative/datachain/pull/611

PR #607, opened 2024-11-18
feat(dc): make expand work if some fields are optional

shcheklein:
Adds `schema_sample_size` option to `explode` to specify how many samples it should be analyzing do derive schema. This is needed in case some fields are missing in a single sample.

Fixes a few other things to make it work on the existing fixture.

### DIFF
--- a/src/datachain/lib/arrow.py
+++ b/src/datachain/lib/arrow.py
@@ -116,31 +116,43 @@ def infer_schema(chain: "DataChain", **kwargs) -> pa.Schema:
     return pa.unify_schemas(schemas)
 
 
-def schema_to_output(schema: pa.Schema, col_names: Optional[Sequence[str]] = None):
-    """Generate UDF output schema from pyarrow schema."""
+def schema_to_output(
+    schema: pa.Schema, col_names: Optional[Sequence[str]] = None
+) -> tuple[dict[str, type], list[str]]:
+    """
+    Generate UDF output schema from pyarrow schema.
+    Returns a tuple of output schema and original column names (since they may be
+    normalized in the output dict).
+    """
+    signal_schema = _get_datachain_schema(schema)
+    if signal_schema:
+        return signal_schema.values, list(signal_schema.values)
+
     if col_names and (len(schema) != len(col_names)):
         raise ValueError(
             "Error generating output from Arrow schema - "
             f"Schema has {len(schema)} columns but got {len(col_names)} column names."
         )
     if not col_names:
-        col_names = schema.names
-    signal_schema = _get_datachain_schema(schema)
-    if signal_schema:
-        return signal_schema.values
-    columns = list(normalize_col_names(col_names).keys())  # type: ignore[arg-type]
+        col_names = schema.names or []
+
+    normalized_col_dict = normalize_col_names(col_names)
+    col_names = list(normalized_col_dict)
+
     hf_schema = _get_hf_schema(schema)
     if hf_schema:
         return {
-            column: hf_type for hf_type, column in zip(hf_schema[1].values(), columns)
-        }
+            column: hf_type for hf_type, column in zip(hf_schema[1].values(), col_names)
+        }, list(normalized_col_dict.values())
+
     output = {}
-    for field, column in zip(schema, columns):
-        dtype = arrow_type_mapper(field.type, column)  # type: ignore[assignment]
+    for field, column in zip(schema, col_names):
+        dtype = arrow_type_mapper(field.type, column)
         if field.nullable and not ModelStore.is_pydantic(dtype):
             dtype = Optional[dtype]  # type: ignore[assignment]
         output[column] = dtype
-    return output
+
+    return output, list(normalized_col_dict.values())
 
 
 def arrow_type_mapper(col_type: pa.DataType, column: str = "") -> type:  # noqa: PLR0911
--- a/src/datachain/lib/data_model.py
+++ b/src/datachain/lib/data_model.py
@@ -1,8 +1,8 @@
 from collections.abc import Sequence
 from datetime import datetime
-from typing import ClassVar, Union, get_args, get_origin
+from typing import ClassVar, Optional, Union, get_args, get_origin
 
-from pydantic import BaseModel, Field, create_model
+from pydantic import AliasChoices, BaseModel, Field, create_model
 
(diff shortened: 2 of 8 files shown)


src/datachain/lib/dc.py
```diff
@@ -775,7 +780,7 @@ def print_json_schema(  # type: ignore[override]
             ```py
             uri = "gs://datachain-demo/coco2017/annotations_captions/"
             chain = DataChain.from_storage(uri)
-            chain = chain.show_json_schema()
+            chain = chain.print_json_schema()
```
shcheklein, line 783: super minor docs fix - unrelated

src/datachain/lib/dc.py
```diff
@@ -1836,6 +1841,7 @@ def parse_tabular(
         if isinstance(output, dict):
             model_name = model_name or object_name or ""
             model = dict_to_data_model(model_name, output)
+            output = model
```
shcheklein, line 1844: added to hit the condition below:

 `elif isinstance(output, type(BaseModel)):` 

https://github.com/iterative/datachain/pull/607/files#diff-0b1cfd3d89c057cc57811e07eb20298eeaaee60c54cc66e5128f618d144f42eeR1850

and always create here output with normalized names (`dict_to_data_model` normalizes names)

Mufty4 (approved):

src/datachain/lib/arrow.py
```diff
@@ -116,31 +116,43 @@ def infer_schema(chain: "DataChain", **kwargs) -> pa.Schema:
...
-    """Generate UDF output schema from pyarrow schema."""
+def schema_to_output(
+    schema: pa.Schema, col_names: Optional[Sequence[str]] = None
+) -> tuple[dict[str, type], list[str]]:
+    """
+    Generate UDF output schema from pyarrow schema.
+    Returns a tuple of output schema and original column names (since they may be
+    normalized in the output dict).
+    """
+    signal_schema = _get_datachain_schema(schema)
```
shcheklein, line 127: [C]: cosmetic, moved these three lines above since they are independent (don't depend on input fields or other logic)

src/datachain/lib/arrow.py
```diff
@@ -116,31 +116,43 @@ def infer_schema(chain: "DataChain", **kwargs) -> pa.Schema:
...
-    for field, column in zip(schema, columns):
-        dtype = arrow_type_mapper(field.type, column)  # type: ignore[assignment]
+    for field, column in zip(schema, col_names):
+        dtype = arrow_type_mapper(field.type, column)
         if field.nullable and not ModelStore.is_pydantic(dtype):
             dtype = Optional[dtype]  # type: ignore[assignment]
         output[column] = dtype
-    return output
+
+    return output, list(normalized_col_dict.values())
```
shcheklein, line 155: main change here, besides `output` which is a dict column_name -> type, we also return the list of original column names since returned dict can have already normalized values.

we need the original columns later to set Alias choice in pydantic to be able to parse raw JSON

e.g. imagine we have a json like:

```
{
"Na-Me": "something"
}
```

Pydantic model will have a field like

```python
na_me: srt
```

we need to set an alias for the model to be able to consume the raw JSON

src/datachain/lib/data_model.py
```diff
@@ -60,17 +60,33 @@ def is_chain_type(t: type) -> bool:
...
+            anno,
+            Field(
+                validation_alias=AliasChoices(name, original_names[idx] or name),
+                default=None,
+            ),
+        )
+        for idx, (name, anno) in enumerate(data_dict.items())
     }
+
+    class _DataModelStrict(BaseModel, extra="forbid"):
```
shcheklein, line 85: [C]: make model strict on extra fields. This is primarily to ensure that we crafted the schema right.

E.g. if let's say we have 100 jsons, and all of them have slightly different schema. We need to analyze all of them, not only the first one to get a proper schema, otherwise (if we derive the schema based on the first 1 or 2 values) we might start skipping values silently

src/datachain/lib/hf.py
```diff
@@ -98,7 +98,7 @@ def process(self, split: str = ""):
         with tqdm(desc=desc, unit=" rows") as pbar:
             for row in ds:
                 output_dict = {}
-                if split:
+                if split and "split" in self.output_schema.model_fields:
```
shcheklein, line 101: needed because we added a `extra` forbidden and by some other (and relevant) logic "split" doesn't exist in some cases in the `self.output_schema` model

tests/unit/lib/test_arrow.py
shcheklein, line 1: No real changes in this file, just adjusting based on return values, etc.

tests/unit/lib/test_datachain.py
```diff
@@ -1317,7 +1322,7 @@ def test_to_csv_features_nested(tmp_dir, test_session):
 @pytest.mark.parametrize("object_name", (None, "test_object_name"))
 @pytest.mark.parametrize("model_name", (None, "TestModelNameExploded"))
 def test_explode(tmp_dir, test_session, column_type, object_name, model_name):
-    df = pd.DataFrame(DF_DATA)
+    df = pd.DataFrame(DF_DATA_NESTED_NOT_NORMALIZED)
```
shcheklein, line 1325: Here, we parse a way more complicated JSON (with optional values in JSON, with non-normalized top level values) - that pretty much drove a lot of changes in the code, besides the main one - `schema_sample_size`. `schema_sample_size` didn't work w/o a lot of fixes around

tests/unit/lib/test_hf.py
shcheklein, line 1: No changes in this file, except just adjusting based on the changes in other places

tests/unit/lib/test_datachain.py
```diff
@@ -1327,22 +1332,34 @@ def test_explode(tmp_dir, test_session, column_type, object_name, model_name):
...
+            object_name=object_name,
+            model_name=model_name,
+            schema_sample_size=2,
+        )
     )
 
     object_name = object_name or "content_expl"
     model_name = model_name or "ContentExplodedModel"
 
+    # In CH we have (atm at least) None coтverted to ''
```
dreadatour, line 1346:
```suggestion
    # In CH we have (atm at least) None converted to ''
```

dreadatour (approved):
Looks good to me 👍

PR #592, opened 2024-11-13
Update ultralytics models

dreadatour:
Update Ultralytics YOLO models to be used in SaaS.

This adds `Segments` model and common `YoloBBox`, `YoloBBoxes`, `YoloPose`, `YoloPoses`, `YoloSegment` and `YoloSegments` models to be used to parse ultralytics results.

PR #610, opened 2024-11-19
Revert "Update ultralytics models"

shcheklein:
Reverts iterative/datachain#592

To be able redeploy Studio


mattseddon (approved):

0x2b3bfa0 (approved):
It's a pity we have to revert this instead of merging https://github.com/iterative/studio/pull/10935, but... 😿

shcheklein:
> It's a pity we have to revert this instead of merging https://github.com/iterative/studio/pull/10935, but... 😿

I don't have enough context to judge that on such a short notice tbh ... sorry folks

PR #611, opened 2024-11-19
Increase resolution of `train_test_split`

0x2b3bfa0:
* https://github.com/iterative/datachain/pull/604#discussion_r1846911532

### DIFF
--- a/src/datachain/toolkit/split.py
+++ b/src/datachain/toolkit/split.py
@@ -1,9 +1,6 @@
 from datachain import C, DataChain
 
 
-UINT64_MAX = 2**64-1  # Maximum value for an UInt64
-
-
 def train_test_split(dc: DataChain, weights: list[float]) -> list[DataChain]:
     """
     Splits a DataChain into multiple subsets based on the provided weights.
--- a/src/datachain/toolkit/split.py
+++ b/src/datachain/toolkit/split.py
@@ -1,6 +1,9 @@
 from datachain import C, DataChain
 
 
+UINT64_MAX = 2**64-1  # Maximum value for an UInt64
+
+
 def train_test_split(dc: DataChain, weights: list[float]) -> list[DataChain]:
     """
     Splits a DataChain into multiple subsets based on the provided weights.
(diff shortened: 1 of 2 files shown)


tests/func/test_pull.py
```diff
@@ -78,7 +78,7 @@ def _adapt_row(row):
 def schema():
     return {
         "id": {"type": "UInt64"},
-        "sys__rand": {"type": "Int64"},
+        "sys__rand": {"type": "UInt64"},
```
0x2b3bfa0, line 81: For consistency with `DataTable.sys_columns` here? 🤔 

https://github.com/iterative/datachain/blob/976b5fd1b27d33f40fc25d88a96cc5ddeda378cd/src/datachain/data_storage/schema.py#L279

shcheklein (approved):

src/datachain/toolkit/split.py
```diff
@@ -58,10 +58,14 @@ def train_test_split(dc: DataChain, weights: list[float]) -> list[DataChain]:
 
     weights_normalized = [weight / sum(weights) for weight in weights]
 
+    resolution = 2**31 - 1  # Maximum positive value for a 32-bit signed integer.
```
0x2b3bfa0, line 61: Ideally, we could use `2**64 - 1` and get rid of the `%` operation, but @dmpetrov expressed concerns at https://github.com/iterative/datachain/pull/604#discussion_r1848903487 about differences between the different database implementations.

```suggestion
    resolution = 2**64 - 1  # Maximum positive value for a 64-bit unsigned integer.
```

dreadatour (approved):
Looks good to me, thank you for the fix.
Need to deal with tests now and to be careful with different DB implementations as you've mentioned.

tests/func/test_pull.py
0x2b3bfa0: Oh, by the way: on SQLite it looks like the maximum `INTEGER` is a **signed** 64-bit number. 🙃

src/datachain/toolkit/split.py
0x2b3bfa0, line 61: I noticed after merging the pull request that my comment is a lie! 🤦🏼 Not that it matters, but the logic is slightly flawed: we should be using `2**31` for the modulo operation and `2**31 - 1` to multiply the normalized weights.

PR #614, opened 2024-11-22
test_query_e2e: do not interrupt running script multiple times

skshetry:
The PR makes three changes:

1) The E2E test is updated only to send the interrupt signal once (done in the first commit).
2) Suppresses `usearch`'s `UserWarning`, and
3) Fixes "leaked semaphore issue" (see #615) due to worker process being SIGKILL'd on `KeyboardInterrupt`. 

This happens because, while running script, sending <kbd>Ctrl</kbd> + <kbd>C</kbd> would also send SIGINT to
the worker processes, but then [it would immediately follow up with a SIKILL][1] [after 0.25s][2].

This allowed very little time for workers to do cleanup, due to which, Python was raising `UserWarning`
for leaked semaphores:

```py
.../lib/python3.12/site-packages/multiprocess/resource_tracker.py:257: UserWarning: resource_tracker: There appear to be 6 leaked semaphore objects to clean >
```

Also see python/cpython#70130.

This patch changes worker processes to be run using `subprocess.Popen`
rather than `subprocess.run`, so that SIGKILL never gets sent.
The responsibility is on worker process to cleanup properly and exit. 

[1]: https://github.com/python/cpython/blob/3c770e3f0978d825c5ebea98fcd654660e7e135f/Lib/subprocess.py#L571-L573
[2]: https://github.com/python/cpython/blob/3c770e3f0978d825c5ebea98fcd654660e7e135f/Lib/subprocess.py#L891

### DIFF
--- a/src/datachain/query/dataset.py
+++ b/src/datachain/query/dataset.py
@@ -457,18 +457,15 @@ class UDFStep(Step, ABC):
                 # Run the UDFDispatcher in another process to avoid needing
                 # if __name__ == '__main__': in user scripts
                 exec_cmd = get_datachain_executable()
+                cmd = [*exec_cmd, "internal-run-udf"]
                 envs = dict(os.environ)
                 envs.update({"PYTHONPATH": os.getcwd()})
                 process_data = filtered_cloudpickle_dumps(udf_info)
-                result = subprocess.run(  # noqa: S603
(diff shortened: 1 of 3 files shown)


dreadatour (approved):
👍

PR #617, opened 2024-11-23
Refactor pulling dataset rows

ilongin:
As we are chainging the way we export files to s3 in https://github.com/iterative/studio/pull/10966, we needed to adjust CLI code as well to have better performance.
Now we are exporting chunks to s3 "in order", which means chunks with lower index will be there first so we can use that to make sure we are fetching those chunks with lower indexes first to avoid idling until chunks are ready.


src/datachain/catalog/catalog.py
```diff
@@ -198,6 +194,20 @@ def fix_columns(self, df) -> None:
...
 
+        # id will be autogenerated in DB
+        return df.drop("sys__id", axis=1)
+
+    def get_parquet_content(self, url: str):
+        while True:
+            if self.should_check_for_status():
+                self.check_for_status()
+            r = requests.get(url, timeout=PULL_DATASET_CHUNK_TIMEOUT)
+            if r.status_code == 404:
```
amritghimire, line 205: Sorry if it is a silly question. How likely is it that we will be stuck on forever loop here if the url is indeed incorrect url and returning 404 response?

dreadatour: Nice catch! Do we need retry counter here?

ilongin: Good question, but we don't need retry counter here. 404 is expected as this particular chunk may not be exported yet into s3 (in parallel with this Studio is exporting chunks). If something actually fails and we are not able to export chunk to s3 which leads to 404 forever, export itself will fail in Studio and in this loop we are checking for export (whole export job) status on Studio as well every 20 seconds. When we realize that exporting dataset failed on Studio, we will print an error and end the loop.

PR #621, opened 2024-11-25
Using version `uuid` in `datachain pull`

ilongin:
In this PR we are changing the way to find if the same local dataset already exists when pulling remote dataset from Studio or not. Before we were using `name` + `version` combination which is not stable (as commented by Matt  [here](https://github.com/iterative/datachain/pull/573#discussion_r1833421855)) as someone could have local dataset with the same name and version as remote one, but with totally different data which would lead to not pulling the remote one. Also, someone could pull remote dataset and then rename it which would lead to pulling remote dataset again on another pull.

Now, instead of `name` + `version` combination we use `uuid` which is uniquely identifying specific dataset (version).
If local dataset version with same `uuid` as remote one already exists, regardless of the `name` or `version` mismatch, it is not pulled again.

Also, optional parameters `local-name` and `local-version` are added to CLI of dataset pull to specify the name / version of local dataset to which we want to pull remote one. This is needed in the case if local dataset with same name + version already exists with different data, but we still want to pull remote one - in that case use can enter alternative name or version or both.


src/datachain/catalog/catalog.py
```diff
@@ -170,7 +174,7 @@ def check_for_status(self) -> None:
         Checks are done every PULL_DATASET_CHECK_STATUS_INTERVAL seconds
         """
         export_status_response = self.studio_client.dataset_export_status(
-            self.dataset_name, self.dataset_version
+            self.remote_ds_name, self.remote_ds_version
```
mattseddon, line 177: [Q] Why not use `uuid` instead of `remote_ds_name` + `remote_ds_version` & `local_ds_name` + `local_ds_version`?

ilongin: Yea, there are other APIs that use `name` + `version` as well, but I would leave that to followup issue not to bloat this one too much. In general, user does use explicit name and version when initiating pull, e.g `datachain pull ds://dogs@v3` but the use of `uuid` in the process after initial get dataset info would fix potential issue of someone renaming remote dataset right in the middle of pull process (not sue how likely is this to happen but it's still the potential risk)

mattseddon: None of this needs to go into this PR but I wanted to mention it and this seems like an ok place.

I would advocate for hiding local/remote name/version from users as much as possible. (From experience) I think the concept will only confuse users. Only exposing it when we absolutely need to makes sense to me 

For example:

If the user tries to pull `cat@v1` under the following circumstances we should throw an error stating there is a dataset with that name already and the uuid doesn't match. Ask them to rename the local dataset (as SaaS is the source of truth):

| name | version | uuid | local | remote |
| -- | -- | -- | -- | --| 
| cat | 1 | cf308a6b-0e66-477f-ba76-9c3f8ffd27fe | ✓ | X |
| cat | 1 | d1c1ac59-5763-4069-a328-8292ec6a6205 | X | ✓ |

I think we should update the output of `ls datasets` to have the same format as the above or at least give it the ability to show these discrepancies somehow.

I am now second-guessing myself and wondering whether the local/remote name and version are that bad.

src/datachain/catalog/catalog.py
```diff
@@ -1316,116 +1327,125 @@ def ls(
...
+            # if version is not specified in uri, take the latest one
+            if not version:
+                version = remote_ds.latest_version
+                print(f"Version not specified, pulling the latest one (v{version})")
+                # updating dataset uri with latest version
+                remote_ds_uri = create_dataset_uri(remote_ds_name, version)
+            remote_ds_version = remote_ds.get_version(version)
+        except (DatasetVersionNotFoundError, StopIteration) as exc:
+            raise DataChainError(
+                f"Dataset {remote_ds_name} doesn't have version {version}" " on server"
```
mattseddon, line 1379: [C] Should be a single string

src/datachain/cli.py
```diff
@@ -400,6 +400,18 @@ def get_parser() -> ArgumentParser:  # noqa: PLR0915
         "--edatachain-file",
         help="Use a different filename for the resulting .edatachain file",
     )
+    parse_pull.add_argument(
```
mattseddon, line 403: [C] I still don't know if it is better to expose these or uuid to the user

ilongin: I think we should let client work with name / version. `uuid` is not really practical, and also user cannot really set it.

skshetry:
Test failures are unrelated, and are due to https://github.com/iterative/PyDrive2/issues/361#issuecomment-2507021740.

Issue #576, opened 2024-11-08
Reorganize the cli commands with groups


amritghimire:
Can be a follow up, but let's rename all these command (similar to `ls-dataset` that we had). Can we make it all part of the `datachain datasets` umbrella for example - at least for now. (make `datasets` alias with `ds` to make it simpler)

datachain ds ls
datachain ds rm
datachain ds rename (better to have an explicit command vs one edit that is not very clear in terms of its semantics).

_Originally posted by @shcheklein in https://github.com/iterative/datachain/issues/572#issuecomment-2463030028_

ilongin:
We are talking about this issue literally from the beginning of datachain but it always ended up at the end of the queue somehow :) .. it would be nice if we could finally do it.
Related old issue https://github.com/iterative/dvcx/issues/372

amritghimire:
related to the groups of commands, etc - let's please make API also clean

`datachain/datasets/ls` 
`datachain/datasets/rename` 

-etc 

`rm-dataset` and similar look very weird for REST api.

_Originally posted by @shcheklein in https://github.com/iterative/datachain/pull/572#discussion_r1834853298_

PR #622, opened 2024-11-25
Refactor the cli commands for the datasets

amritghimire:
The changes for the cli commands:
- `datachain datasets ls` from `datachain ls-datasets`
- `datachain datasets edit` from `datachain edit-datasets`
- `datachain datasets rm` from `datachain rm-datasets`
- `datachain datasets pull` from `datachain pull`

Closes https://github.com/iterative/datachain/issues/576

### DIFF
--- a/src/datachain/cli.py
+++ b/src/datachain/cli.py
@@ -419,7 +419,18 @@ def get_parser() -> ArgumentParser:  # noqa: PLR0915
 
     add_studio_parser(subp, parent_parser)
 
-    parse_pull = subp.add_parser(
+    datasets_parser = subp.add_parser(
+        "datasets",
+        aliases=["ds"],
+        parents=[parent_parser],
+        description="Commands for managing datasers",
+    )
+    datasets_subparser = datasets_parser.add_subparsers(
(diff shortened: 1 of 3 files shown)


skshetry:
> * `datachain datasets ls` from `datachain ls-datasets`
> * `datachain datasets edit` from `datachain edit-datasets`
> * `datachain datasets rm` from `datachain rm-datasets`
> * `datachain datasets pull` from `datachain pull`

This feels unnecessarily verbose. Instead, why don't we promote them to top-level commands? 

We should not reserve top-level commands for filesystem operations since that's not the core value that datachain provides.

amritghimire:
> > * `datachain datasets ls` from `datachain ls-datasets`
> > * `datachain datasets edit` from `datachain edit-datasets`
> > * `datachain datasets rm` from `datachain rm-datasets`
> > * `datachain datasets pull` from `datachain pull`
> 
> This feels unnecessarily verbose. Instead, why don't we promote them to top-level commands?


We also have alias for datasets to ds. So, we can use `datachain ds ls` instead.

PR #629, opened 2024-11-26
dataset_select_paginated: do not clone on each paginated query

skshetry:
Let's see if the test passes.

Context: https://github.com/iterative/datachain/pull/521#discussion_r1857724246.


dreadatour (approved):
Great finding 🔥 Thank you! 👍

PR #630, opened 2024-11-26
docs: document prefetch

### LINKED ISSUE #40: pre_fetch option in additional to cache for lib.File

skshetry:
Closes #40.

### DIFF
--- a/src/datachain/lib/dc.py
+++ b/src/datachain/lib/dc.py
@@ -348,6 +348,9 @@ class DataChain:
                 enable all available CPUs (default=1)
             workers : number of distributed workers. Only for Studio mode. (default=1)
             min_task_size : minimum number of tasks (default=1)
+            prefetch: number of workers to use for downloading files in advance.
+                      This is enabled by default and uses 2 worker.
+                      To disable prefetching, set it to 0.
 
         Example:
             ```py
--- a/src/datachain/lib/dc.py
+++ b/src/datachain/lib/dc.py
@@ -349,7 +349,7 @@ class DataChain:
             workers : number of distributed workers. Only for Studio mode. (default=1)
             min_task_size : minimum number of tasks (default=1)
             prefetch: number of workers to use for downloading files in advance.
-                      This is enabled by default and uses 2 worker.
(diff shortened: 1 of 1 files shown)


src/datachain/lib/dc.py
```diff
@@ -348,6 +348,9 @@ def settings(
                 enable all available CPUs (default=1)
             workers : number of distributed workers. Only for Studio mode. (default=1)
             min_task_size : minimum number of tasks (default=1)
+            prefetch: number of workers to use for downloading files in advance.
+                      This is enabled by default and uses 2 worker.
```
0x2b3bfa0, line 352:
```suggestion
                      This is enabled by default and uses 2 workers.
```

ilongin (approved):

src/datachain/lib/dc.py
```diff
@@ -348,6 +348,9 @@ def settings(
                 enable all available CPUs (default=1)
             workers : number of distributed workers. Only for Studio mode. (default=1)
             min_task_size : minimum number of tasks (default=1)
+            prefetch: number of workers to use for downloading files in advance.
```
shcheklein, line 351: workers in this context might sound like actual DC workers (especially in Studio context). Worker in this case is probably an internal term (and also quite nuanced since we use async for this?). Is there a better way to phrase this?

can we just say: 

`number of files to download in advance for `map`, `gen` operations when `File` object is passed to them`

shcheklein: Also, probably better to mention about this option existence in the map / gen also.

PR #628, opened 2024-11-26
remove JSON columns from dataset lists

mattseddon:
Related to https://github.com/iterative/studio/issues/10849

In Studio we're seeing a slow query alerts for the following query:

<details> 
<summary> SQL </summary>

```SQL
    SELECT
        dqlapp_dataset.id,
        dqlapp_dataset.name,
        dqlapp_dataset.description,
        dqlapp_dataset.labels,
        dqlapp_dataset.status,
        dqlapp_dataset.feature_schema,
        dqlapp_dataset.created_at,
        dqlapp_dataset.finished_at,
        dqlapp_dataset.error_message,
        dqlapp_dataset.error_stack,
        dqlapp_dataset.script_output,
        dqlapp_dataset.sources,
        dqlapp_dataset.query_script,
        dqlapp_dataset.schema,
        dqlapp_dataset.team_id,
        dqlapp_dataset.warehouse_id,
        dqlapp_dataset.created_by_id,
        dqlapp_datasetversion.id AS id_1,
        dqlapp_datasetversion.uuid,
        dqlapp_datasetversion.dataset_id,
        dqlapp_datasetversion.version,
        dqlapp_datasetversion.status AS status_1,
        dqlapp_datasetversion.feature_schema AS feature_schema_1,
        dqlapp_datasetversion.created_at AS created_at_1,
        dqlapp_datasetversion.finished_at AS finished_at_1,
        dqlapp_datasetversion.error_message AS error_message_1,
        dqlapp_datasetversion.error_stack AS error_stack_1,
        dqlapp_datasetversion.script_output AS script_output_1,
        dqlapp_datasetversion.num_objects,
        dqlapp_datasetversion.size,
        dqlapp_datasetversion.preview,
        dqlapp_datasetversion.sources AS sources_1,
        dqlapp_datasetversion.query_script AS query_script_1,
        dqlapp_datasetversion.schema AS schema_1,
        dqlapp_datasetversion.job_id,
        dqlapp_datasetversion.created_by_id AS created_by_id_1  
    FROM
        dqlapp_dataset 
    LEFT OUTER JOIN
        dqlapp_datasetversion 
            ON dqlapp_dataset.id = dqlapp_datasetversion.dataset_id  
    WHERE
        dqlapp_dataset.team_id = $1
```

</details>

This statement is being generated by a `catalog.ls_datasets` call via a GraphQL endpoint. 

[...]

PR #634, opened 2024-11-26
add DatasetListRecord and DatasetListVersion dataclasses to only extract required fields when listing datasets

mattseddon:
Related to https://github.com/iterative/studio/issues/10849 and replaces https://github.com/iterative/datachain/pull/628 (see replaced PR for some details)

This PR adds `DatasetListRecord` and `DatasetListVersion` `dataclass`es and uses those classes within `Catalog`'s `ls_datasets` and `list_datasets_versions` methods. Using these new classes means that we can significantly cut down the data being processed when listing datasets. This should give us a performance boost in both the CLI and SaaS and hopefully prevent the slow query alerts in the related issue.

Note: There is some duplication added in this PR, it does not seem practical to extend the original `dataclass`es in this instance.

### DIFF
--- a/src/datachain/data_storage/metastore.py
+++ b/src/datachain/data_storage/metastore.py
@@ -27,6 +27,8 @@ from datachain.data_storage import JobQueryType, JobStatus
 from datachain.data_storage.serializer import Serializable
 from datachain.dataset import (
     DatasetDependency,
+    DatasetListRecord,
+    DatasetListVersion,
     DatasetRecord,
     DatasetStatus,
     DatasetVersion,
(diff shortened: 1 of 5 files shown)


src/datachain/data_storage/metastore.py
```diff
@@ -348,6 +350,14 @@ def _dataset_fields(self) -> list[str]:
             if c.name  # type: ignore [attr-defined]
         ]
 
+    @cached_property
+    def _dataset_list_fields(self) -> list[str]:
+        return [
+            c.name  # type: ignore [attr-defined]
+            for c in self._datasets_columns()
+            if c.name in ["id", "name", "description", "labels", "created_at"]  # type: ignore [attr-defined]
```
mattseddon, line 358: [C] try and get this from the dataclass (same below)

dreadatour (approved):
Looks good to me 👍

Issue #606, opened 2024-11-17
Improve split Dataset into train / test / eval


dreadatour:
See original issue: https://github.com/iterative/datachain/issues/603

Some improvements we may want to be implemented:

**Seed**. Seed is usually implemented by XOR-ing: sys.rand ^ seed if seed is provided or without any XOR if not provided.
if XOR is not implemented yet and it takes time to implement we can implement it without seed for now.

**Shuffle**. Optional, as using sys.rand makes it shuffled by default. If we need shuffle=False at some point we might consider using sorting and sys.id instead. But I'm not sure shuffle=False is very practical in real applications.

dmpetrov:
> **Shuffle**.

It's not needed. See https://github.com/iterative/datachain/issues/603#issuecomment-2479864068

0x2b3bfa0:
Closed with https://github.com/iterative/datachain/pull/678

PR #641, opened 2024-11-28
Update base Func class and tests

dreadatour:
Remove some operators not supported by all DBs by default. Implement bitwise functions. Improve tests.

Related to https://github.com/iterative/datachain/issues/606 (implement `XOR` function).

### DIFF
--- a/src/datachain/func/func.py
+++ b/src/datachain/func/func.py
@@ -163,13 +163,13 @@ class Func(Function):
 
     def __mod__(self, other: Union[ColT, float]) -> "Func":
         if isinstance(other, (int, float)):
-            return Func("mod", lambda a: a % other, [self])
-        return Func("mod", lambda a1, a2: a1 % a2, [self, other])
+            return Func("mod", lambda a: a % other, [self], result_type=int)
+        return Func("mod", lambda a1, a2: a1 % a2, [self, other], result_type=int)
 
     def __rmod__(self, other: Union[ColT, float]) -> "Func":
         if isinstance(other, (int, float)):
-            return Func("mod", lambda a: other % a, [self])
-        return Func("mod", lambda a1, a2: a1 % a2, [other, self])
+            return Func("mod", lambda a: other % a, [self], result_type=int)
+        return Func("mod", lambda a1, a2: a1 % a2, [other, self], result_type=int)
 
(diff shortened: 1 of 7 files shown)


src/datachain/func/func.py
```diff
@@ -96,94 +98,136 @@ def _db_col_type(self, signals_schema: "SignalSchema") -> Optional["DataType"]:
...
+            return Func("div", lambda a: _truediv(a, other), [self], result_type=float)
+        return Func(
+            "div", lambda a1, a2: _truediv(a1, a2), [self, other], result_type=float
+        )
 
     def __rtruediv__(self, other: Union[ColT, float]) -> "Func":
-        return math_truediv(other, self)
+        if isinstance(other, (int, float)):
+            return Func("div", lambda a: _truediv(other, a), [self], result_type=float)
+        return Func(
```
shcheklein, line 140: do we need a test for this?

src/datachain/func/func.py
```diff
@@ -96,94 +98,136 @@ def _db_col_type(self, signals_schema: "SignalSchema") -> Optional["DataType"]:
...
+            "floordiv", lambda a1, a2: _floordiv(a1, a2), [self, other], result_type=int
+        )
 
     def __rfloordiv__(self, other: Union[ColT, float]) -> "Func":
-        return math_floordiv(other, self)
+        if isinstance(other, (int, float)):
+            return Func(
+                "floordiv", lambda a: _floordiv(other, a), [self], result_type=int
+            )
+        return Func(
```
shcheklein, line 158: same here, do we need a test?

shcheklein (approved):
(please check is need a few more tests)

src/datachain/func/func.py
dreadatour: Here we have it for all three cases: https://github.com/iterative/datachain/pull/641/files#diff-1a98b3b24a2c496f079d2c5f5757a55850ca5840a7e53350e39c5d1c10175979R159-R167

src/datachain/func/func.py
dreadatour: Same here: https://github.com/iterative/datachain/pull/641/files#diff-1a98b3b24a2c496f079d2c5f5757a55850ca5840a7e53350e39c5d1c10175979R189-R197 🤗

PR #178, opened 2024-07-25
Optimizations: low-hanging fruits

dreadatour:
# Preparation

Let's profile laion embeddings generation from `gs://dvcx-datacomp-small/metadata` bucket.

List bucket:
```
$ gsutil ls -l gs://dvcx-datacomp-small/metadata
3106844080  2024-01-15T15:37:25Z  gs://dvcx-datacomp-small/metadata/0020f0cbd157d470aa56bea08e304b90.npz
 120845372  2024-01-15T15:28:20Z  gs://dvcx-datacomp-small/metadata/0020f0cbd157d470aa56bea08e304b90.parquet
...
 673764696  2024-01-15T15:29:52Z  gs://dvcx-datacomp-small/metadata/036d6b9ae87a00e738f8fc554130b65b.npz
  26585624  2024-01-15T15:27:58Z  gs://dvcx-datacomp-small/metadata/036d6b9ae87a00e738f8fc554130b65b.parquet
TOTAL: 52 objects, 81707590807 bytes (76.1 GiB)
$
```

Pre-download "small" `643Mb`, `109662` rows `npz` file (just to make it easier to profile):
```
$ mkdir -p dvcx-datacomp-small/metadata
$ gsutil cp -v gs://dvcx-datacomp-small/metadata/036d6b9ae87a00e738f8fc554130b65b.npz dvcx-datacomp-small/metadata/
Copying gs://dvcx-datacomp-small/metadata/036d6b9ae87a00e738f8fc554130b65b.npz...

Created: file://dvcx-datacomp-small/metadata/036d6b9ae87a00e738f8fc554130b65b.npz

Operation completed over 1 objects/642.6 MiB.
$
```

## Script to profile with

`laion_emb.py`:
```python
from datachain.lib.dc import C, DataChain
from datachain.lib.webdataset_laion import process_laion_meta

(
    DataChain.from_storage("dvcx-datacomp-small/metadata/")
    .gen(emd=process_laion_meta)
    .map(stem=lambda file: file.get_file_stem(), params=["emd.file"], output=str)
    .save("laion_emb")
)
```

#### Run time
```
$ time python laion_emb.py
Processed: 1 rows [00:00, 2582.70 rows/s]
Download: 643MB [01:51, 6.02MB/s]
Processed: 1 rows [01:51, 111.84s/ rows]
Generated: 109662 rows [01:51, 981.86 rows/s]
Processed: 109662 rows [00:33, 3250.94 rows/s]
[...]

PR #642, opened 2024-11-28
do not pre-load dataset version preview from string

mattseddon:
This PR ensures that we don't load a dataset version's preview to a dict from a string until it has been requested. This should mean that we can continue to use the pattern of loading all version for a dataset via `catalog.get_dataset` without taking the performance hit of loading all of the dataset version previews via `json.loads`

### DIFF
--- a/src/datachain/dataset.py
+++ b/src/datachain/dataset.py
@@ -2,6 +2,7 @@ import builtins
 import json
 from dataclasses import dataclass, fields
 from datetime import datetime
+from functools import cached_property
 from typing import (
     Any,
     NewType,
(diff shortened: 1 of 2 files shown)


src/datachain/dataset.py
```diff
@@ -260,9 +261,17 @@ def serialized_schema(self) -> dict[str, Any]:
             for c_name, c_type in self.schema.items()
         }
 
+    @cached_property
+    def preview(self) -> Optional[list[dict]]:
+        if isinstance(self._preview_data, str):
+            return json.loads(self._preview_data)
```
dreadatour, line 267: Should we use `orjson` as it is faster?
See https://github.com/iterative/datachain/pull/178

dreadatour (approved):
🔥

PR #648, opened 2024-12-01
add some missing type hint

shibuiwilliam:
# What
- Add some missing type hints.


src/datachain/client/local.py
```diff
@@ -7,6 +7,7 @@
 
 from fsspec.implementations.local import LocalFileSystem
 
+from datachain.cache import DataChainCache
```
shcheklein, line 10: should it be under `if TYPE_CHECKING:` below?

shcheklein (approved):

amritghimire (approved):

Issue #635, opened 2024-11-26
prefetch examples to showcase performance gain


dmpetrov:
### Description

We need examples when it's clear how prefetch helps. I tried in several examples and I don't see any difference.

An examples is below. Note, the library utilizes CPU pretty well (can utilize it 600% in my laptop), no parallelization is needed if prefetch is good.

Results:
- prefetch=0: **35.144s**
- prefetch=2: **35.126s**
- prefetch=8: **34.418s**

Code:
```python
from io import BytesIO
from typing import Optional

from PIL import Image
from ultralytics import YOLO
from datachain import DataChain, C, File, model

from datachain.lib.data_model import DataModel

def process1(yolo: YOLO, file: File) -> model.ultralytics.YoloBBoxes:
    img = Image.open(BytesIO(file.read()))
    return model.ultralytics.YoloBBoxes.from_results(yolo(img, verbose=False))

(
    DataChain
    .from_storage("gs://mpii-human-pose")
    .limit(100)
    .settings(prefetch=0)
    .setup(
        yolo=lambda: YOLO("yolo11n.pt"),
    )
    .map(my_yolo=process1)
    .save("mpii-human-pose-segment")
)
```

### Version Info

```Text
0.7.1
Python 3.9.16
```

dmpetrov:
@skshetry I'd appreciate it if you could take a look

skshetry:
Turns out, `prefetch` is only enabled when caching is enabled, which makes sense.

https://github.com/iterative/datachain/blob/3bd22ad2becf1bdd255a6f75925add8a0aaff5e6/src/datachain/lib/file.py#L271-L274

For me, `cache=True, prefetch=8` reduces runtime to 15s from `cache=True, prefetch=0`.

skshetry:
Can we enable caching by default? What is the reason for not doing so already?

shcheklein:
it might be taking a lot of space by default? But, yes seems reasonable to enable it by default. @dmpetrov wdyt?

dmpetrov:
pre-fetch is not related to caching. We need to decouple these options. Created #647

Caching by default is a separate question - we should not do that.

skshetry:
> Caching by default is a separate question - but I don't see any strong reasons of doing this.


We have to persist them somewhere for the lifetime of the script's run. IIUC, due to the worker processes, we cannot do it in memory.

dmpetrov:
[...]

PR #653, opened 2024-12-02
pytorch dataset: pass cache/prefetch to DataChain instances

### LINKED ISSUE #631: `to_pytorch` prefetching
### Description

TBD

skshetry:
This should enable prefetching on `to_pytorch` API. We were not passing any `settings` to the `DataChain` instances created inside `to_pytorch`, due to which it was not using `cache` or `prefetch`.

I have refrained from setting `workers`, etc. because I believe that is better left to `PytorchDataLoader` for now. 

I have found this PR to increase performance for `torch-loader.py` example by 20-25% when prefetch and cache is enabled. However, this was done on my machine and was not a scientific measurement.

Closes #631.

### DIFF
--- a/src/datachain/lib/dc.py
+++ b/src/datachain/lib/dc.py
@@ -1446,6 +1446,7 @@ class DataChain:
             tokenizer=tokenizer,
             tokenizer_kwargs=tokenizer_kwargs,
             num_samples=num_samples,
+            dc_settings=chain._settings,
         )
 
     def remove_file_signals(self) -> "Self":  # noqa: D102
--- a/src/datachain/lib/pytorch.py
+++ b/src/datachain/lib/pytorch.py
@@ -12,6 +12,7 @@ from tqdm import tqdm
 from datachain import Session
 from datachain.catalog import Catalog, get_catalog
 from datachain.lib.dc import DataChain
+from datachain.lib.settings import Settings
 from datachain.lib.text import convert_text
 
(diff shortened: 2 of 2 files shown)


shcheklein:
@skshetry could you clarify the scope for the #631 based on the previous meetings discussion? (does this PR close it for example or we had more things in mind?). Do we need more testing, research how it works or not?

shcheklein (approved):

dreadatour (approved):
Looks good to me! 👍

skshetry:
> @skshetry could you clarify the scope for the #631 based on the previous meetings discussion? (does this PR close it for example or we had more things in mind?). Do we need more testing, research how it works or not?

I was considering whether we need prefetching at all, given that `pytorch` handles some of this internally. However, in the example I mentioned earlier, a significant amount of time is spent downloading files.

(Even with `prefetch_factor` set, it still is going to download in sequence, downloading while the iterator gets consumed).

When running with `cache=True`, I see that the `DataLoader` processes the dataset (200 files) in under a second, compared to around 30 seconds without caching. So the potential is there for improvement. That said, I haven’t noticed any meaningful improvements with `prefetch` so far.

I’m currently experimenting with a dataset containing a larger number of files to test the hypothesis that prefetching improves performance.

While prefetching should theoretically enhance `DataLoader` performance, I haven’t yet observed any tangible gains (see related issue: #635).

PR #655, opened 2024-12-03
Fix inaccurate comment in `split.py`

0x2b3bfa0:
Continuation of #611, see https://github.com/iterative/datachain/pull/611#discussion_r1850895073.

To do:
* Call `abs` always before comparison, and make sure we're always comparing positive numbers.
* Check that we're using the right data types everywhere [un]signed integers? 64 bits? SQLite doesn't support uint64, but only int64.

Successor of https://github.com/iterative/datachain/pull/612 (branch name was conflicting)


0x2b3bfa0:
Closed in favor of #678

PR #657, opened 2024-12-03
Implement 'seed' for 'train_test_split' + simplify split logic

dreadatour:
1. New `seed` param for `train_test_split` toolkit method was added:

```python
train, test, val = train_test_split(dc, [0.7, 0.2, 0.1], seed=37)
train.save("dataset_train")
test.save("dataset_test")
val.save("dataset_val")
```

2. Split logic was simplified. Instead of running modulo computations in DB, we are now running bitwise operations, it should be faster. See [here](https://github.com/iterative/datachain/compare/func-tests-follow-up...606-improve-split-dataset-into-train-test-eval?expand=1#diff-f3a9af671188ea77937788641b29496befd23d343be02835a5ab1d53d026ffb8R88-R94). It is also respects negative values in `sys__rand` (if any), because in SQLite we do not have unsigned integers.

3. For `seed` we use `int_hash_64(sys__rand ^ seed)` operation, where `int_hash_64` is a function, copied from ClickHouse [int_hash_64](https://clickhouse.com/docs/en/sql-reference/functions/hash-functions#inthash64). This way we can run the same code in both SQLite and ClickHouse.


dreadatour:
Studio tests are failing because we need to merge https://github.com/iterative/studio/pull/10999 first.

src/datachain/toolkit/split.py
```diff
@@ -56,16 +67,28 @@ def train_test_split(dc: DataChain, weights: list[float]) -> list[DataChain]:
...
+    weights_normalized = [weight / sum(weights) for weight in weights]
+    limits = [
+        [
+            scale(sum(weights_normalized[:index])),
+            scale(sum(weights_normalized[: index + 1])),
+        ]
+        for index, _ in enumerate(weights_normalized)
+    ]
 
+    rand_col = "sys__rand" if seed is None else int_hash_64(bit_xor("sys__rand", seed))
```
dmpetrov, line 87: Please avoid using the underscores in column names.

src/datachain/toolkit/split.py
```diff
@@ -1,7 +1,16 @@
-from datachain import C, DataChain
+from typing import Optional
 
+from datachain import DataChain
+from datachain.func import bit_and, bit_xor, int_hash_64
```
dmpetrov, line 4: Is it an universal function among all DBs?

Ideally, dc.split(0.2, seed=20241203) should return the same result for the same dataset in locally as well as in cloud.

src/datachain/toolkit/split.py
dmpetrov, line 87: it would be great to know how beadly we need `int_hash_64`. I'd understand the concern if we would be XORing with `sys.id` which is not uniformally distributed. However, `sys.rand` should not have any skews.

dmpetrov (approved):
Great change. Couple comments are inline. But I'm approving to not to block anyone.

src/datachain/toolkit/split.py
dreadatour: Oops, sorry. Fixed!

src/datachain/toolkit/split.py
dreadatour: > Is it an universal function among all DBs?

`int_hash_64` implementation is specific for ClickHouse. I was able to implement it for SQLite, but I am not 100% sure about other databases :(

Other problem is with int64 in SQLite, it only supports signed int64, this requires additional steps and attention.

One of the possible solution for all databases will be using md5/halfmd5/sha256, but these are slow.

> Ideally, dc.split(0.2, seed=20241203) should return the same result for the same dataset in locally as well as in cloud.

That was my goal, and it works now for both SQLite and ClickHouse.

src/datachain/toolkit/split.py
dreadatour: I understand `inthash64` looks like we don't really need it.

I was trying XORing `sys.rand` with `seed` and different big prime numbers in different combinations, but I can not make it works to get good enough distribution. In two words: using `seed` gives not much difference with default "no seed" split.

I was also trying to use `murmurhash`, `siphash` and other hash algorithms, as they are pretty common, implemented in ClickHouse and can be added to SQLite, but they requires installing additional packages which I was trying to avoid (and I can, but I don't really want to implement those by myself). Also they might not be available in other databases.

If anyone have suggestions on good algorithm using XOR only, I'm all in. I can also spent more time on this.

src/datachain/toolkit/split.py
```diff
@@ -1,7 +1,16 @@
-from datachain import C, DataChain
+from typing import Optional
```
0x2b3bfa0, line 1:
```suggestion
from random import Random
from typing import Optional
```

src/datachain/toolkit/split.py
0x2b3bfa0, line 4:
```suggestion
from datachain.func import bit_and, bit_xor
```

src/datachain/toolkit/split.py
```diff
@@ -56,16 +67,28 @@ def train_test_split(dc: DataChain, weights: list[float]) -> list[DataChain]:
...
+    weights_normalized = [weight / sum(weights) for weight in weights]
+    limits = [
+        [
+            scale(sum(weights_normalized[:index])),
+            scale(sum(weights_normalized[: index + 1])),
+        ]
+        for index, _ in enumerate(weights_normalized)
+    ]
 
+    rand_col = "sys.rand" if seed is None else int_hash_64(bit_xor("sys.rand", seed))
```
0x2b3bfa0, line 87:
```suggestion
    uniform_seed = Random(seed).randint(0, MAX_SIGNED_INT64)
    rand_col = "sys.rand" if seed is None else bit_xor("sys.rand", uniform_seed)
```

0x2b3bfa0 (commented):
I wonder if we could just derive an uniform seed from the user input, and just `XOR` that with `sys__rand` to "apply" the seed.

Wouldn't running <code>xor(sys__rand, hash(seed)) % MAX_SIGNED_INT<b>32</b></code>[^1] be "indistinguishable enough" from running `hash(xor(sys__rand, seed))` for our purposes? 🤔 

> [!NOTE]
> Code suggestions are whiteboard-quality; if they work on the first try, it oughta be a miracle.

[^1]: Yes, we're back to running a modulo operation, and reducing our field to 31 bits again.

src/datachain/toolkit/split.py
```diff
@@ -56,16 +67,28 @@ def train_test_split(dc: DataChain, weights: list[float]) -> list[DataChain]:
...
         raise ValueError("Weights should be non-negative")
 
-    weights_normalized = [weight / sum(weights) for weight in weights]
+    def scale(multiplier: float) -> int:
+        if multiplier < 0:
+            multiplier = 0.0
+        if multiplier > 1:
+            multiplier = 1.0
+        numerator, denominator = float(multiplier).as_integer_ratio()
+        return MAX_SIGNED_INT64 * numerator // denominator
```
0x2b3bfa0, line 76: Wasn't the previous approach with round/multiply good enough?

0x2b3bfa0 (commented):
A couple nitpicks; can be disregarded.

src/datachain/toolkit/split.py
0x2b3bfa0, line 1:
```suggestion
from random import Random
from typing import Optional
```

src/datachain/toolkit/split.py
0x2b3bfa0, line 4:
```suggestion
from datachain.func import bit_and, bit_xor
```

src/datachain/toolkit/split.py
```diff
@@ -1,7 +1,16 @@
-from datachain import C, DataChain
+from typing import Optional
 
+from datachain import DataChain
+from datachain.func import bit_and, bit_xor, int_hash_64
 
-def train_test_split(dc: DataChain, weights: list[float]) -> list[DataChain]:
+MAX_SIGNED_INT64 = (1 << 63) - 1  # Maximum positive value for a 64-bit signed integer.
```
0x2b3bfa0, line 6:
```suggestion
MAX_SIGNED_INT64 = (1 << 63) - 1  # Maximum positive value for a 64-bit signed integer.
MAX_SIGNED_INT32 = (1 << 31) - 1  # Maximum positive value for a 64-bit signed integer.
```

src/datachain/toolkit/split.py
0x2b3bfa0, line 76:
```suggestion
        return MAX_SIGNED_INT32 * numerator // denominator
```

src/datachain/toolkit/split.py
0x2b3bfa0, line 87:
```suggestion
    uniform_seed = Random(seed).randint(0, MAX_SIGNED_INT64)
	rand_col = "sys.rand" if seed is None else bit_xor("sys.rand", uniform_seed)
```

src/datachain/toolkit/split.py
```diff
@@ -56,16 +67,28 @@ def train_test_split(dc: DataChain, weights: list[float]) -> list[DataChain]:
...
 
+    rand_col = "sys.rand" if seed is None else int_hash_64(bit_xor("sys.rand", seed))
     return [
         dc.filter(
-            C("sys__rand") % resolution
-            >= round(sum(weights_normalized[:index]) * resolution),
-            C("sys__rand") % resolution
-            < round(sum(weights_normalized[: index + 1]) * resolution),
+            bit_and(rand_col, MAX_SIGNED_INT64) >= min_,
+            bit_and(rand_col, MAX_SIGNED_INT64) < max_,
```
0x2b3bfa0, line 91:
```suggestion
            bit_and(rand_col, MAX_SIGNED_INT64) % MAX_SIGNED_INT32 >= min_,
            bit_and(rand_col, MAX_SIGNED_INT64) % MAX_SIGNED_INT32 < max_,
```

0x2b3bfa0 (dismissed):
I wonder if we could just derive an uniform seed from the user input, and just `XOR` that with `sys__rand` to "apply" the seed.

Wouldn't running <code>xor(sys__rand, hash(seed)) % MAX_SIGNED_INT<b>32</b></code>[^1] be "indistinguishable enough" from running `hash(xor(sys__rand, seed))` for our purposes? 🤔

Using the _modulo_ operation to reduce our field to "only" 31 bits, we have the desired avalanche effect, and even in cases like `weights = [0.5, 0.5]` the distribution should change for every seed value.

(Without the modulo operation, there wouldn't be avalanche effect: half of the seeds would produce a fixed distribution, and the other half would produce another fixed distribution)

src/datachain/toolkit/split.py
```diff
@@ -56,16 +67,28 @@ def train_test_split(dc: DataChain, weights: list[float]) -> list[DataChain]:
     if any(weight < 0 for weight in weights):
         raise ValueError("Weights should be non-negative")
 
-    weights_normalized = [weight / sum(weights) for weight in weights]
+    def scale(multiplier: float) -> int:
+        if multiplier < 0:
+            multiplier = 0.0
+        if multiplier > 1:
+            multiplier = 1.0
```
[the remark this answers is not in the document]
dreadatour: It does not in any case I can imagine, but I've added it just in case, otherwise SQLite will throw an exception (let's say if we pass `9223372036854775808` (`2^63`) instead of `9223372036854775807` (`2^63-1`), — this is, by the way, why we need this function, because in Python `int(9223372036854775807 * 1.0) == 9223372036854775808` (same for `round`) and SQLite is going crazy about this.

0x2b3bfa0 (approved):
After a failed attempt at finding a simple[r] solution, the MurmurHash finalizer looks like poetry to my eyes. 🙈

src/datachain/toolkit/split.py
amritghimire, line 74:
```suggestion
        multiplier = min(max(0, multiplier),1.0)
```

0x2b3bfa0 (commented):
I haven't found any better & simpler solution than `(random * seed) % MAXIMUM` with a locally computed seed; it has avalanche effect and no self-evident distribution flaws.

src/datachain/toolkit/split.py
dreadatour: Oh, awesome, thank you! 🙏 Always forget about this trick 😅

dreadatour:
@0x2b3bfa0 I made couple experiments meanwhile.

## Seed pseudo-randomness visualization

#### With `inthash64`

```python
from PIL import Image
from datachain.sql.sqlite.base import sqlite_int_hash_64


SEED = 123456

width, height = 512, 512
im = Image.new("RGB", (width, height), "black")
pixels = im.load()

sys__rand = 0
for y in range(height):
    for x in range(int(width / 64)):
        value = sqlite_int_hash_64(sys__rand ^ SEED) & 0xFFFFFFFFFFFFFFFF
        sys__rand += 1
        for i in range(64):
            if value & (1 << i):
                pixels[x * 64 + i, y] = (255, 255, 255)

im.save("random_pattern.png", "PNG")
```

Result:

random_pattern

#### With `(random * seed) % MAXIMUM`

```python
import random
from PIL import Image


MAXIMUM = 2**64 - 1
SEED = 123456
uniform_seed = random.Random(SEED).randrange(MAXIMUM)

width, height = 512, 512
im = Image.new("RGB", (width, height), "black")
pixels = im.load()

sys__rand = 0
for y in range(height):
    for x in range(int(width / 64)):
        value = (sys__rand * uniform_seed) % MAXIMUM
        sys__rand += 1
        for i in range(64):
            if value & (1 << i):
                pixels[x * 64 + i, y] = (255, 255, 255)

im.save("random_pattern.png", "PNG")
```

Result:

random_pattern


## Check with Random Test Tool (RTT)

See https://github.com/xmco/random_test_tool

#### With `inthash64`

```python
import sys
from datachain.sql.sqlite.base import sqlite_int_hash_64


seed = int(sys.argv[1])
num_opts = int(sys.argv[2])

for sys__rand in range(0, num_opts):
    print(sqlite_int_hash_64(sys__rand ^ seed) & 0xFFFFFFFFFFFFFFFF)
```

Test run:

```
$ python inthash.py 123456 10
13379111408315310133
7443788450574447110
3595878185051936035
12667204699528734827
10291846290167640951
11613734228742057041
7218166708332674651
3686523795734826198
3804965084228910159
15009888239023274627
$
```

Check:

```
$ python inthash.py 123456 1000000 > inthash.txt
$ random-test-tool -i inthash.txt
Launching tests on 1 batches...

Tests launched:
- binary_matrix
- block_chi2
- chi2
- compression
- linear_complexity
- run_of_ones
- run
- serial
- sign
- spectral


100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 10/10 [00:45<00:00,  4.56s/it]
╒═════════════════════╤════════════╤═══════════╤══════════════════╤════════════════════════════════════════════════════════════╤══════════╤═════════════════╕
│ test_name           │   n_sample │   p_value │ p_value_limits   │ criterias                                                  │ status   │   sample number │
╞═════════════════════╪════════════╪═══════════╪══════════════════╪════════════════════════════════════════════════════════════╪══════════╪═════════════════╡
│ Binary Rank         │   31523121 │ 0.540907  │ (0.01, 0.05)     │ 0 -KO- 0.01 -SUSPECT- 0.05 -OK- 0.95 -SUSPECT- 0.99 -KO- 1 │ OK       │               1 │
├─────────────────────┼────────────┼───────────┼──────────────────┼────────────────────────────────────────────────────────────┼──────────┼─────────────────┤
│ Block Chi2          │   31523121 │ 0.0817816 │ (0.01, 0.05)     │ 0 -KO- 0.01 -SUSPECT- 0.05 -OK- 0.95 -SUSPECT- 0.99 -KO- 1 │ OK       │               1 │
├─────────────────────┼────────────┼───────────┼──────────────────┼────────────────────────────────────────────────────────────┼──────────┼─────────────────┤
│ Chi2                │    1000000 │ 1         │ (0.01, 0.05)     │ 0 -KO- 0.01 -SUSPECT- 0.05 -OK- 0.95 -SUSPECT- 0.99 -KO- 1 │ KO       │               1 │
├─────────────────────┼────────────┼───────────┼──────────────────┼────────────────────────────────────────────────────────────┼──────────┼─────────────────┤
│ Compression         │   31523121 │ 0.355664  │ (0.01, 0.05)     │ 0 -KO- 0.01 -SUSPECT- 0.05 -OK- 0.95 -SUSPECT- 0.99 -KO- 1 │ OK       │               1 │
├─────────────────────┼────────────┼───────────┼──────────────────┼────────────────────────────────────────────────────────────┼──────────┼─────────────────┤
│ Linear Complexity   │   31523121 │ 0.0667993 │ (0.01, 0.05)     │ 0 -KO- 0.01 -SUSPECT- 0.05 -OK- 0.95 -SUSPECT- 0.99 -KO- 1 │ OK       │               1 │
├─────────────────────┼────────────┼───────────┼──────────────────┼────────────────────────────────────────────────────────────┼──────────┼─────────────────┤
│ Longest Run of Ones │   31523121 │ 0.214657  │ (0.01, 0.05)     │ 0 -KO- 0.01 -SUSPECT- 0.05 -OK- 0.95 -SUSPECT- 0.99 -KO- 1 │ OK       │               1 │
├─────────────────────┼────────────┼───────────┼──────────────────┼────────────────────────────────────────────────────────────┼──────────┼─────────────────┤
│ Run                 │    1000000 │ 0.639778  │ (0.01, 0.05)     │ 0 -KO- 0.01 -SUSPECT- 0.05 -OK- 0.95 -SUSPECT- 0.99 -KO- 1 │ OK       │               1 │
├─────────────────────┼────────────┼───────────┼──────────────────┼────────────────────────────────────────────────────────────┼──────────┼─────────────────┤
│ Serial              │   31523121 │ 0.474092  │ (0.01, 0.05)     │ 0 -KO- 0.01 -SUSPECT- 0.05 -OK- 0.95 -SUSPECT- 0.99 -KO- 1 │ OK       │               1 │
├─────────────────────┼────────────┼───────────┼──────────────────┼────────────────────────────────────────────────────────────┼──────────┼─────────────────┤
│ Sign                │    1000000 │ 1         │ (0.01, 0.05)     │ 0 -KO- 0.01 -SUSPECT- 0.05 -OK- 0.95 -SUSPECT- 0.99 -KO- 1 │ KO       │               1 │
├─────────────────────┼────────────┼───────────┼──────────────────┼────────────────────────────────────────────────────────────┼──────────┼─────────────────┤
│ Spectral            │   31523121 │ 0.727808  │ (0.01, 0.05)     │ 0 -KO- 0.01 -SUSPECT- 0.05 -OK- 0.95 -SUSPECT- 0.99 -KO- 1 │ OK       │               1 │
╘═════════════════════╧════════════╧═══════════╧══════════════════╧════════════════════════════════════════════════════════════╧══════════╧═════════════════╛
╒═════════════════════╤════════════╤═════════════════╤════════════╤══════════════════╤════════════════════╕
│ test_name           │   OK_count │   SUSPECT_count │   KO_count │ KO_interval      │ SUSPECT_interval   │
╞═════════════════════╪════════════╪═════════════════╪════════════╪══════════════════╪════════════════════╡
│ Binary Rank         │          1 │               0 │          0 │ Sample too Small │ Sample too Small   │
├─────────────────────┼────────────┼─────────────────┼────────────┼──────────────────┼────────────────────┤
│ Block Chi2          │          1 │               0 │          0 │ Sample too Small │ Sample too Small   │
├─────────────────────┼────────────┼─────────────────┼────────────┼──────────────────┼────────────────────┤
│ Chi2                │          0 │               0 │          1 │ Sample too Small │ Sample too Small   │
├─────────────────────┼────────────┼─────────────────┼────────────┼──────────────────┼────────────────────┤
│ Compression         │          1 │               0 │          0 │ Sample too Small │ Sample too Small   │
├─────────────────────┼────────────┼─────────────────┼────────────┼──────────────────┼────────────────────┤
│ Linear Complexity   │          1 │               0 │          0 │ Sample too Small │ Sample too Small   │
├─────────────────────┼────────────┼─────────────────┼────────────┼──────────────────┼────────────────────┤
│ Longest Run of Ones │          1 │               0 │          0 │ Sample too Small │ Sample too Small   │
├─────────────────────┼────────────┼─────────────────┼────────────┼──────────────────┼────────────────────┤
│ Run                 │          1 │               0 │          0 │ Sample too Small │ Sample too Small   │
├─────────────────────┼────────────┼─────────────────┼────────────┼──────────────────┼────────────────────┤
│ Serial              │          1 │               0 │          0 │ Sample too Small │ Sample too Small   │
├─────────────────────┼────────────┼─────────────────┼────────────┼──────────────────┼────────────────────┤
│ Sign                │          0 │               0 │          1 │ Sample too Small │ Sample too Small   │
├─────────────────────┼────────────┼─────────────────┼────────────┼──────────────────┼────────────────────┤
│ Spectral            │          1 │               0 │          0 │ Sample too Small │ Sample too Small   │
╘═════════════════════╧════════════╧═════════════════╧════════════╧══════════════════╧════════════════════╛
$
```

#### With `(random * seed) % MAXIMUM`

```python
import random
import sys


MAXIMUM = 2**64 - 1

seed = int(sys.argv[1])
num_opts = int(sys.argv[2])

uniform_seed = random.Random(seed).randrange(MAXIMUM)
for sys__rand in range(0, num_opts):
    print((sys__rand * uniform_seed) % MAXIMUM)
```

Test run:

```
$ python randseed.py 123456 10
0
15565657522269164640
12684570970828777665
9803484419388390690
6922397867948003715
4041311316507616740
1160224765067229765
16725882287336394405
13844795735896007430
10963709184455620455
$
```

Check:

```
$ python randseed.py 123456 1000000 > randseed.txt
$ random-test-tool -i randseed.txt
Launching tests on 1 batches...

Tests launched:
- binary_matrix
- block_chi2
- chi2
- compression
- linear_complexity
- run_of_ones
- run
- serial
- sign
- spectral


100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 10/10 [00:44<00:00,  4.46s/it]
╒═════════════════════╤════════════╤═════════════╤══════════════════╤════════════════════════════════════════════════════════════╤══════════╤═════════════════╕
│ test_name           │   n_sample │     p_value │ p_value_limits   │ criterias                                                  │ status   │   sample number │
╞═════════════════════╪════════════╪═════════════╪══════════════════╪════════════════════════════════════════════════════════════╪══════════╪═════════════════╡
│ Binary Rank         │   31500063 │ 0.928675    │ (0.01, 0.05)     │ 0 -KO- 0.01 -SUSPECT- 0.05 -OK- 0.95 -SUSPECT- 0.99 -KO- 1 │ OK       │               1 │
├─────────────────────┼────────────┼─────────────┼──────────────────┼────────────────────────────────────────────────────────────┼──────────┼─────────────────┤
│ Block Chi2          │   31500063 │ 0           │ (0.01, 0.05)     │ 0 -KO- 0.01 -SUSPECT- 0.05 -OK- 0.95 -SUSPECT- 0.99 -KO- 1 │ KO       │               1 │
├─────────────────────┼────────────┼─────────────┼──────────────────┼────────────────────────────────────────────────────────────┼──────────┼─────────────────┤
│ Chi2                │    1000000 │ 1           │ (0.01, 0.05)     │ 0 -KO- 0.01 -SUSPECT- 0.05 -OK- 0.95 -SUSPECT- 0.99 -KO- 1 │ KO       │               1 │
├─────────────────────┼────────────┼─────────────┼──────────────────┼────────────────────────────────────────────────────────────┼──────────┼─────────────────┤
│ Compression         │   31500063 │ 3.25604e-28 │ (0.01, 0.05)     │ 0 -KO- 0.01 -SUSPECT- 0.05 -OK- 0.95 -SUSPECT- 0.99 -KO- 1 │ KO       │               1 │
├─────────────────────┼────────────┼─────────────┼──────────────────┼────────────────────────────────────────────────────────────┼──────────┼─────────────────┤
│ Linear Complexity   │   31500063 │ 0.570419    │ (0.01, 0.05)     │ 0 -KO- 0.01 -SUSPECT- 0.05 -OK- 0.95 -SUSPECT- 0.99 -KO- 1 │ OK       │               1 │
├─────────────────────┼────────────┼─────────────┼──────────────────┼────────────────────────────────────────────────────────────┼──────────┼─────────────────┤
│ Longest Run of Ones │   31500063 │ 0.37952     │ (0.01, 0.05)     │ 0 -KO- 0.01 -SUSPECT- 0.05 -OK- 0.95 -SUSPECT- 0.99 -KO- 1 │ OK       │               1 │
├─────────────────────┼────────────┼─────────────┼──────────────────┼────────────────────────────────────────────────────────────┼──────────┼─────────────────┤
│ Run                 │    1000000 │ 0           │ (0.01, 0.05)     │ 0 -KO- 0.01 -SUSPECT- 0.05 -OK- 0.95 -SUSPECT- 0.99 -KO- 1 │ KO       │               1 │
├─────────────────────┼────────────┼─────────────┼──────────────────┼────────────────────────────────────────────────────────────┼──────────┼─────────────────┤
│ Serial              │   31500063 │ 0.999995    │ (0.01, 0.05)     │ 0 -KO- 0.01 -SUSPECT- 0.05 -OK- 0.95 -SUSPECT- 0.99 -KO- 1 │ KO       │               1 │
├─────────────────────┼────────────┼─────────────┼──────────────────┼────────────────────────────────────────────────────────────┼──────────┼─────────────────┤
│ Sign                │    1000000 │ 1           │ (0.01, 0.05)     │ 0 -KO- 0.01 -SUSPECT- 0.05 -OK- 0.95 -SUSPECT- 0.99 -KO- 1 │ KO       │               1 │
├─────────────────────┼────────────┼─────────────┼──────────────────┼────────────────────────────────────────────────────────────┼──────────┼─────────────────┤
│ Spectral            │   31500063 │ 0           │ (0.01, 0.05)     │ 0 -KO- 0.01 -SUSPECT- 0.05 -OK- 0.95 -SUSPECT- 0.99 -KO- 1 │ KO       │               1 │
╘═════════════════════╧════════════╧═════════════╧══════════════════╧════════════════════════════════════════════════════════════╧══════════╧═════════════════╛
╒═════════════════════╤════════════╤═════════════════╤════════════╤══════════════════╤════════════════════╕
│ test_name           │   OK_count │   SUSPECT_count │   KO_count │ KO_interval      │ SUSPECT_interval   │
╞═════════════════════╪════════════╪═════════════════╪════════════╪══════════════════╪════════════════════╡
│ Binary Rank         │          1 │               0 │          0 │ Sample too Small │ Sample too Small   │
├─────────────────────┼────────────┼─────────────────┼────────────┼──────────────────┼────────────────────┤
│ Block Chi2          │          0 │               0 │          1 │ Sample too Small │ Sample too Small   │
├─────────────────────┼────────────┼─────────────────┼────────────┼──────────────────┼────────────────────┤
│ Chi2                │          0 │               0 │          1 │ Sample too Small │ Sample too Small   │
├─────────────────────┼────────────┼─────────────────┼────────────┼──────────────────┼────────────────────┤
│ Compression         │          0 │               0 │          1 │ Sample too Small │ Sample too Small   │
├─────────────────────┼────────────┼─────────────────┼────────────┼──────────────────┼────────────────────┤
│ Linear Complexity   │          1 │               0 │          0 │ Sample too Small │ Sample too Small   │
├─────────────────────┼────────────┼─────────────────┼────────────┼──────────────────┼────────────────────┤
│ Longest Run of Ones │          1 │               0 │          0 │ Sample too Small │ Sample too Small   │
├─────────────────────┼────────────┼─────────────────┼────────────┼──────────────────┼────────────────────┤
│ Run                 │          0 │               0 │          1 │ Sample too Small │ Sample too Small   │
├─────────────────────┼────────────┼─────────────────┼────────────┼──────────────────┼────────────────────┤
│ Serial              │          0 │               0 │          1 │ Sample too Small │ Sample too Small   │
├─────────────────────┼────────────┼─────────────────┼────────────┼──────────────────┼────────────────────┤
│ Sign                │          0 │               0 │          1 │ Sample too Small │ Sample too Small   │
├─────────────────────┼────────────┼─────────────────┼────────────┼──────────────────┼────────────────────┤
│ Spectral            │          0 │               0 │          1 │ Sample too Small │ Sample too Small   │
╘═════════════════════╧════════════╧═════════════════╧════════════╧══════════════════╧════════════════════╛
$
```

---

To be honest I an not sure if this is really a big deal in our case, just wanted to share these results.

0x2b3bfa0:
Those tests look amazing! 🤩

Note that your test methodology might be slightly flawed, because on your tests `sys__rand` is a counter, but my formula relies on `sys__rand` having actual random values.

dreadatour:
> Those tests look amazing! 🤩

❤️

> Note that your test methodology might be slightly flawed, because on your tests `sys__rand` is a counter, but my formula relies on `sys__rand` having actual random values.

I know 👀 My idea was to check generated numbers distribution only. By having `sys__rand` as a counter I am relying on pseudo-random generator only and can check if generated numbers are evenly distributed all across the interval (`int64` in our case).

Was thinking how to test distribution rely on `sys__rand` having actual random values and made another test:

#### With inthash64

```python
import random

from PIL import Image
from datachain.sql.sqlite.base import sqlite_int_hash_64


width, height = 512, 512
im = Image.new("RGB", (width, height), "black")
pixels = im.load()

sys__rand = random.randint(0, 0xFFFFFFFFFFFFFFFF)
seed = 0
for y in range(height):
    for x in range(int(width / 64)):
        value = sqlite_int_hash_64(sys__rand ^ seed) & 0xFFFFFFFFFFFFFFFF
        seed += 1
        for i in range(64):
            if value & (1 << i):
                pixels[x * 64 + i, y] = (255, 255, 255)

im.save("random_pattern.png", "PNG")
```

Result:

random_pattern

#### With `(random * seed) % MAXIMUM`

```python
import random
from PIL import Image


MAXIMUM = 2**64 - 1

width, height = 512, 512
im = Image.new("RGB", (width, height), "black")
pixels = im.load()

sys__rand = random.randint(0, 0xFFFFFFFFFFFFFFFF)
seed = 0
for y in range(height):
    for x in range(int(width / 64)):
        uniform_seed = random.Random(seed).randrange(MAXIMUM)
        seed += 1
        value = (sys__rand * uniform_seed) % MAXIMUM
        for i in range(64):
            if value & (1 << i):
                pixels[x * 64 + i, y] = (255, 255, 255)

im.save("random_pattern.png", "PNG")
```

Result:

random_pattern

---

This looks much more promising I guess.

dreadatour:
Finally found an issue with tests [here](https://github.com/iterative/datachain/pull/678/files#diff-906ffda534f667d94b9d1473a61aeebdce354da65fe7c240dc07f785597c845cR31-R36).

It is because of SQLite math.

Python:

```
In [1]: 8822966792743105186 * 1730071088
Out[1]: 15264359758508934493641462368

In [2]: (8822966792743105186 * 1730071088) % 2147483647
Out[2]: 581439120
```

SQLite:

```
sqlite> select 8822966792743105186 * 1730071088;
1.52643597585089e+28
sqlite> select (8822966792743105186 * 1730071088) % 2147483647;
1.0
```

dreadatour:
I'll continue in https://github.com/iterative/datachain/pull/678 (switch to the `(random * seed) % MAXIMUM` as it will be easier to implement in all databases).

src/datachain/toolkit/split.py
0x2b3bfa0: Resolved with #678

src/datachain/toolkit/split.py
0x2b3bfa0: Resolved with #678

Issue #637, opened 2024-11-26
select_except() does not support sub-classes


dmpetrov:
### Description

For some reason I cannot select_except() a class (`yolo`) while a sub-class was created (`my_poses`).

Note, if I won't have a subclass my_poses - it will except the whole yolo just fine. But we still need to support the case when only a subclass left.

```python
from io import BytesIO

from PIL import Image
from ultralytics import YOLO
from datachain import DataChain, DataModel, C, File, model


class MyYolo(DataModel):
    bbox: model.ultralytics.YoloBBoxes
    segm: model.ultralytics.YoloSegments
    poses: model.ultralytics.YoloPoses


def process(yolo:YOLO, yolo_segm: YOLO, yolo_pose: YOLO, file: File) -> MyYolo:
    img = Image.open(BytesIO(file.read()))
    return MyYolo(
        bbox=model.ultralytics.YoloBBoxes.from_results(yolo(img, verbose=False)),
        segm=model.ultralytics.YoloSegments.from_results(yolo_segm(img, verbose=False)),
        poses=model.ultralytics.YoloPoses.from_results(yolo_pose(img, verbose=False))
    )


(
    DataChain
    .from_storage("gs://mpii-human-pose")
    .limit(10)
    .setup(
        yolo=lambda: YOLO("yolo11n.pt"),
        yolo_segm=lambda: YOLO("yolo11n-seg.pt"),
        yolo_pose=lambda: YOLO("yolo11n-pose.pt")
    )
    .map(yolo=process)
  	.mutate(my_poses=C("yolo.poses"))
  	.select_except("yolo")
    .save("yolo_mpii")
)
```

Error:
```
Preparing: 10 rows [00:00, 3952.42 rows/s]
Download: 5.46MB [00:05, 1.02MB/s]
Processed: 10 rows [00:04,  2.49 rows/s]
Cleanup: 2 tables [00:00, 3278.08 tables/s]
Traceback (most recent call last):
  File "/Users/dmitry/src/datachain/yolo_test.py", line 24, in <module>
    DataChain
  File "/Users/dmitry/venv/venv-py39/lib/python3.9/site-packages/datachain/lib/dc.py", line 818, in save
    query=self._query.save(
  File "/Users/dmitry/venv/venv-py39/lib/python3.9/site-packages/datachain/query/dataset.py", line 1577, in save
    query = self.apply_steps()
[...]

PR #658, opened 2024-12-03
Allow adding new signal from existing signal field

dreadatour:
See original issue with full example: https://github.com/iterative/datachain/issues/637
Also fix for the https://github.com/iterative/studio/issues/10989

---

Let's say we have a complex model with other models used as signals:

```python
from datachain.model.ultralytics import YoloBBoxes, YoloPoses, YoloSegments

class MyYolo(DataModel):
    bbox: YoloBBoxes
    segm: YoloSegments
    poses: YoloPoses
```

this PR allow us to use these sub-signals in `mutate` (assuming `yolo` is signal with `MyYolo` type):

```python
dc.mutate(my_poses=C("yolo.poses"))
```

then `my_poses` will be a signal with `YoloPoses` type.

### DIFF
--- a/src/datachain/lib/signal_schema.py
+++ b/src/datachain/lib/signal_schema.py
@@ -402,9 +402,12 @@ class SignalSchema:
             if ModelStore.is_pydantic(finfo.annotation):
                 SignalSchema._set_file_stream(getattr(obj, field), catalog, cache)
 
-    def get_column_type(self, col_name: str) -> DataType:
+    def get_column_type(self, col_name: str, leaf_only: bool = True) -> DataType:
         for path, _type, has_subtree, _ in self.get_flat_tree():
-            if not has_subtree and DEFAULT_DELIMITER.join(path) == col_name:
+            if (
+                not (leaf_only and has_subtree)
+                and DEFAULT_DELIMITER.join(path) == col_name
+            ):
                 return _type
         raise SignalResolvingError([col_name], "is not found")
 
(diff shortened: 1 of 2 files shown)


src/datachain/lib/signal_schema.py
```diff
@@ -402,9 +402,12 @@ def _set_file_stream(
             if ModelStore.is_pydantic(finfo.annotation):
                 SignalSchema._set_file_stream(getattr(obj, field), catalog, cache)
 
-    def get_column_type(self, col_name: str) -> DataType:
+    def get_column_type(self, col_name: str, leaf_only: bool = True) -> DataType:
```
dreadatour, line 405: Don't really like `leaf_only` name, if anyone have better suggestions, please, suggest 🙏

dreadatour: I'll add a comment 👍

In two words: this function return type of the column if this is not a subtree column (check [here](https://github.com/iterative/datachain/pull/658/files/9cc79abc3590404f459b13a8ef2007160fb818bd#diff-69a1fc2d77ff106392ad7649ed8cf7ec344dad230af2e4d0ea9c64b816cee15bL407)). It will return `int` type for `file.size`, but will raise `SignalResolvingError` exception for `file` signal because it have subtree.

I was thinking about adding separate function, but I think it will be a code duplication :(

dreadatour: `allow_subtree` instead of `leaf_only` may be?

dreadatour: > is it user facing in some cases?

As far as I know users don't need it, we use it internally.

dreadatour: Renamed `leaf_only` param to `with_subtree`.

Added docstring:
```python
    def get_column_type(self, col_name: str, with_subtree: bool = False) -> DataType:
        """
        Returns column type by column name.

        If `with_subtree` is True, then it will return the type of the column
        even if it has a subtree (e.g. model with nested fields), otherwise it will
        return the type of the column (standard type field, not the model).

        If column is not found, raises `SignalResolvingError`.
        """
```

amritghimire (approved):

PR #664, opened 2024-12-05
`to_pytorch`: enable prefetching

### LINKED ISSUE #631: `to_pytorch` prefetching

skshetry:
#653 did not really enable prefetching. Prefetch was only implemented for `map()`, so the example gave me a false impression that the prefetching was working, but it was not.

Now, `to_pytorch` uses `AsyncMapper` to prefetch the data. The number of workers is set to 2 by default, but it can be changed by setting the `prefetch` in the settings.

For me, this dropped the time to download the data by 90%, from ~300s to now ~35s.

Closes #631.

### DIFF
--- a/src/datachain/lib/pytorch.py
+++ b/src/datachain/lib/pytorch.py
@@ -10,6 +10,7 @@ from torchvision.transforms import v2
 from tqdm import tqdm
 
 from datachain import Session
+from datachain.asyn import AsyncMapper
 from datachain.catalog import Catalog, get_catalog
 from datachain.lib.dc import DataChain
 from datachain.lib.settings import Settings
@@ -31,6 +32,8 @@ def label_to_int(value: str, classes: list) -> int:
 
 
 class PytorchDataset(IterableDataset):
+    prefetch: int = 2
+
     def __init__(
         self,
         name: str,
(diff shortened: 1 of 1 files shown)


src/datachain/lib/pytorch.py
```diff
@@ -31,6 +32,8 @@ def label_to_int(value: str, classes: list) -> int:
 
 
 class PytorchDataset(IterableDataset):
+    prefetch: int = 2
```
skshetry, line 35: Keeping it same as the default.

https://github.com/iterative/datachain/blob/911c22f37306d118d42ff7d95d8d7731c4faaf71/src/datachain/lib/udf.py#L293

src/datachain/lib/pytorch.py
```diff
@@ -85,51 +92,58 @@ def _get_catalog(self) -> "Catalog":
...
         return Catalog(ms, wh, **self._catalog_params)
 
-    def __iter__(self) -> Iterator[Any]:
-        if self.catalog is None:
-            self.catalog = self._get_catalog()
-        session = Session.get(catalog=self.catalog)
-        total_rank, total_workers = self.get_rank_and_workers()
+    def _rows_iter(self, total_rank: int, total_workers: int):
+        catalog = self._get_catalog()
+        session = Session("PyTorch", catalog=catalog)
```
skshetry, line 97: We need to create a new session here, since the `AsyncMapper` runs this on a separate thread.

dreadatour (approved):
Looks good to me! 👍

PR #666, opened 2024-12-06
Adding `DataChain.diff(...)`

ilongin:
Adding new `DataChain.diff(...)`. method that returns difference between two datasets / chains.
Result is new chain that has additional column with possible values: `A`, `D`, `M`, `U` representing added, deleted, modified and unchanged row. 
Note that if only one "status"  is asked, by setting proper flags, this additional column is not created as it would have only one value for all rows. Beside additional diff column, new chain has schema of the chain on which method was called.

Example usage:
```python
diff = persons.diff(
      new_persons,
      on=["id"],
      right_on=["other_id"],
      compare=["name"],
      added=True,
      deleted=True,
      modified=True,
      unchanged=True,
      status_col="diff"
)
```


src/datachain/lib/dc.py
```diff
@@ -1624,6 +1625,208 @@ def subtract(  # type: ignore[override]
             )
         return self._evolve(query=self._query.subtract(other._query, signals))  # type: ignore[arg-type]
 
+    def diff(  # noqa: PLR0912, C901
```
dmpetrov, line 1628: please rename to `compare()` or something similar

src/datachain/lib/dc.py
```diff
@@ -1624,6 +1625,208 @@ def subtract(  # type: ignore[override]
...
+        on: Union[str, Sequence[str]],
+        right_on: Optional[Union[str, Sequence[str]]] = None,
+        compare: Optional[Union[str, Sequence[str]]] = None,
+        right_compare: Optional[Union[str, Sequence[str]]] = None,
+        added: bool = True,
+        deleted: bool = True,
+        modified: bool = True,
+        unchanged: bool = False,
+        status_col: Optional[str] = None,
+    ) -> "Self":
```
dmpetrov, line 1640: It would be great if you can extract these 100 lines of logic outside of dc.py while keep the method with the docs here 🙂

src/datachain/lib/dc.py
```diff
@@ -1624,6 +1625,208 @@ def subtract(  # type: ignore[override]
...
+            )
+
+        if num_statuses == 0:
+            raise ValueError(
+                "At least one of added, deleted, modified, unchanged flags must be set"
+            )
+
+        # we still need status column for internal implementation even if not
+        # needed in output
+        status_col = status_col or "sys__diff"
```
dmpetrov, line 1737: `sys__diff` - it might be better to use some random column name like `diff_37a345`

src/datachain/lib/dc.py
```diff
@@ -1624,6 +1625,208 @@ def subtract(  # type: ignore[override]
...
+            right_compare = _to_list(right_compare)
+            if len(compare) != len(right_compare):
+                raise ValueError(
+                    "'compare' and 'right_compare' must be have the same length"
+                )
+
+        num_statuses = sum(1 if s else 0 for s in [added, deleted, modified, unchanged])
+
+        if num_statuses > 1 and not status_col:
+            raise ValueError(
```
dmpetrov, line 1726: it's not necessary. these columns are completely independent

src/datachain/lib/dc.py
```diff
@@ -1624,6 +1625,208 @@ def subtract(  # type: ignore[override]
...
+                )
+            )
+        )
+
+        if not deleted:
+            res = left_right_merge
+        elif deleted and not any([added, modified, unchanged]):
+            res = right_left_merge
+        else:
+            res = left_right_merge.union(right_left_merge)
```
dmpetrov, line 1820: please double check that it does not have duplicates after the joins 😅

src/datachain/lib/dc.py
```diff
@@ -1624,6 +1625,208 @@ def subtract(  # type: ignore[override]
...
+            compare = self.signals_schema.resolve(*compare).db_signals()  # type: ignore[assignment]
+            right_compare = other.signals_schema.resolve(*right_compare).db_signals()  # type: ignore[assignment]
+        else:
+            compare = [c for c in cols if c in right_cols]  # type: ignore[misc, assignment]
+            right_compare = compare
+
+        diff_cond = []
+
+        if added:
+            added_cond = sqlalchemy.and_(
```
dmpetrov, line 1757: Can we replace all sqlalchamy function to our built-in functions.
It can look like this:

```python
from datachain.func import _case, _if

l = dc1.mutate(ldiff=1)
r = dc2.mutate(rdiff=1)

dc_diff = (
 l.join(r, on="id")
 .mutate(d1=_case(_isnon(ldiff), "A", "D") )
 .mutate(d2=_case(_isnon(rdiff), "A", "D") )
 .mutate(diff=_case(d1!=d2, "A", "D") )
 .select_except("d1", "d2", "ldiff", "rdiff")
)
```

dmpetrov (approved):
Great change! Added a few comment.
Approving to not to block.

src/datachain/lib/dc.py
ilongin: Done

src/datachain/lib/dc.py
ilongin: Extracted to `datachain.lib.diff.py`

src/datachain/lib/dc.py
ilongin: It shouldn't have any duplicated.

src/datachain/lib/dc.py
ilongin: As we talked, I will create a followup issue for this as improvement, not to block additional work on delta updated etc.

src/datachain/lib/dc.py
ilongin: Done. Made `status_col` optional even when there are multiple statuses requested.

PR #669, opened 2024-12-07
Add 'bit_hamming_distance' and 'byte_hamming_distance' SQLite functions

dreadatour:
New `bit_hamming_distance` function will allow us to calculate [hamming distance](https://en.wikipedia.org/wiki/Hamming_distance) between to numbers. This will be useful on image deduplication (see [this article](https://fullstackml.com/wavelet-image-hash-in-python-3504fdd282b5) for more details).

Also added `byte_hamming_distance` function, just because it is similar and it is worth it to add in the same PR 👀

### DIFF
--- a/src/datachain/func/numeric.py
+++ b/src/datachain/func/numeric.py
@@ -172,9 +172,9 @@ def bit_hamming_distance(*args: Union[ColT, int]) -> Func:
     in the integer indicate higher dissimilarity.
 
     Args:
-        args (str | literal): Two integers to compute the Hamming distance between.
-            If a string is provided, it is assumed to be the name of the column.
-            If a literal is provided, it is assumed to be an integer literal.
+        args (str | int): Two integers to compute the Hamming distance between.
+            If a str is provided, it is assumed to be the name of the column.
+            If an int is provided, it is assumed to be an integer literal.
 
     Returns:
         Func: A Func object that represents the Hamming distance function.
--- a/src/datachain/func/numeric.py
+++ b/src/datachain/func/numeric.py
@@ -160,3 +160,49 @@ def int_hash_64(col: Union[ColT, int]) -> Func:
     return Func(
         "int_hash_64", inner=numeric.int_hash_64, cols=cols, args=args, result_type=int
     )
+
+
+def bit_hamming_distance(*args: Union[ColT, int]) -> Func:
+    """
+    Computes the Hamming distance between the bit representations of two integer values.
+
+    The Hamming distance is the number of positions at which the corresponding bits
+    are different. This function returns the dissimilarity between the integers,
+    where 0 indicates identical integers and values closer to the number of bits
+    in the integer indicate higher dissimilarity.
+
+    Args:
+        args (str | literal): Two integers to compute the Hamming distance between.
+            If a string is provided, it is assumed to be the name of the column.
+            If a literal is provided, it is assumed to be an integer literal.
+
+    Returns:
+        Func: A Func object that represents the Hamming distance function.
+
+    Example:
+        ```py
+        dc.mutate(
+            ham_dist=func.bit_hamming_distance("embed1", 123456s),
+        )
+        ```
+
+    Notes:
+        - Result column will always be of type int.
+    """
+    cols, func_args = [], []
+    for arg in args:
+        if isinstance(arg, int):
+            func_args.append(arg)
+        else:
+            cols.append(arg)
+
+    if len(cols) + len(func_args) != 2:
+        raise ValueError("bit_hamming_distance() requires exactly two arguments")
+
+    return Func(
+        "bit_hamming_distance",
+        inner=numeric.bit_hamming_distance,
+        cols=cols,
+        args=func_args,
+        result_type=int,
+    )
--- a/src/datachain/func/numeric.py
+++ b/src/datachain/func/numeric.py
@@ -182,7 +182,7 @@ def bit_hamming_distance(*args: Union[ColT, int]) -> Func:
     Example:
         ```py
         dc.mutate(
-            ham_dist=func.bit_hamming_distance("embed1", 123456s),
+            ham_dist=func.bit_hamming_distance("embed1", 123456),
         )
         ```
 
--- a/src/datachain/func/string.py
+++ b/src/datachain/func/string.py
@@ -174,7 +174,7 @@ def byte_hamming_distance(*args: Union[str, Func]) -> Func:
     Example:
         ```py
         dc.mutate(
-            ham_dist=func.byte_hamming_distance("embed1", "embed2"),
+            ham_dist=func.byte_hamming_distance("file.phash", literal("hello")),
         )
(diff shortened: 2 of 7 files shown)


src/datachain/func/string.py
```diff
@@ -152,3 +152,49 @@ def inner(arg):
...
+
+    The Hamming distance is the number of positions at which the corresponding
+    characters are different. This function returns the dissimilarity between
+    the strings, where 0 indicates identical strings and values closer to the length
+    of the strings indicate higher dissimilarity.
+
+    Args:
+        args (str | literal): Two strings to compute the Hamming distance between.
+            If a string is provided, it is assumed to be the name of the column.
+            If a literal is provided, it is assumed to be a string literal.
```
0x2b3bfa0, line 169: I don't understand the difference between string (column) and literal (string). Can you find a better wording?

dreadatour: Hm... This is a good example: https://github.com/iterative/datachain/pull/669/files#diff-1a98b3b24a2c496f079d2c5f5757a55850ca5840a7e53350e39c5d1c10175979R640
but let me think how to update the docstrings.

0x2b3bfa0: Oh! Maybe I was too picky; now I understand it.

dreadatour: No, you're right, it should be clear. May be if I'll update an example below: https://github.com/iterative/datachain/pull/669/files/27108dc1862f92f500482fd1f2457bbdaf10002d#diff-cd56a8e471c52456b3e8e5492b585f075a6534b01d2fcc3af4a82b7ead109323R177 it will be enough?

dreadatour: Like this?
```python
        dc.mutate(
            ham_dist=func.byte_hamming_distance("embed1", literal("hello")),
        )
```

src/datachain/func/numeric.py
```diff
@@ -160,3 +160,49 @@ def int_hash_64(col: Union[ColT, int]) -> Func:
...
+            If a string is provided, it is assumed to be the name of the column.
+            If a literal is provided, it is assumed to be an integer literal.
+
+    Returns:
+        Func: A Func object that represents the Hamming distance function.
+
+    Example:
+        ```py
+        dc.mutate(
+            ham_dist=func.bit_hamming_distance("embed1", 123456s),
```
0x2b3bfa0, line 185: Is <code>123456s</code> a valid integer?

src/datachain/func/string.py
0x2b3bfa0: For the sake of comparing pears with pears, what if you do this?

```diff
- If a string is provided, it is assumed to be the name of the column.
+ If a str is provided, it is assumed to be the name of the column.
- If a literal is provided, it is assumed to be a string literal.
+ If a Literal is provided, it is assumed to be a string literal.
```

I also tend to prefer full words over abbreviations, but in this case using the `str` abbreviation and the title case `Literal` makes clear we're referring to Python/SQLAlchemy types, and not just vague concepts.

0x2b3bfa0: And yes, the new example is awesome, thanks! 🤩

src/datachain/func/numeric.py
dreadatour: 😳🫣🤫

src/datachain/func/numeric.py
```diff
@@ -160,3 +160,49 @@ def int_hash_64(col: Union[ColT, int]) -> Func:
...
+
+    The Hamming distance is the number of positions at which the corresponding bits
+    are different. This function returns the dissimilarity between the integers,
+    where 0 indicates identical integers and values closer to the number of bits
+    in the integer indicate higher dissimilarity.
+
+    Args:
+        args (str | literal): Two integers to compute the Hamming distance between.
+            If a string is provided, it is assumed to be the name of the column.
+            If a literal is provided, it is assumed to be an integer literal.
```
0x2b3bfa0, line 177: * Needs same fixes as https://github.com/iterative/datachain/pull/669#discussion_r1874395258?

src/datachain/sql/sqlite/base.py
```diff
@@ -191,6 +193,26 @@ def sqlite_int_hash_64(x: int) -> int:
...
+
+def sqlite_byte_hamming_distance(a: str, b: str) -> int:
+    """Calculate the Hamming distance between two strings."""
+    diff = 0
+    if len(a) < len(b):
+        diff = len(b) - len(a)
+        b = b[: len(a)]
+    elif len(b) < len(a):
+        diff = len(a) - len(b)
+        a = a[: len(b)]
```
0x2b3bfa0, line 212: I thought that Hamming distance wasn't defined for strings of unequal length. 🤔

If we want to "define" it, I wonder if we should account for `a` being a substring of `b` with an arbitrary offset (e.g. use `str.index()` and catch `ValueError` to find out).

(I'm just being picky)

src/datachain/func/numeric.py
dreadatour: Yes 👍 but I will do it in separate PR

src/datachain/sql/sqlite/base.py
dreadatour: I was trying to mimic ClickHouse implementation so CLI and SaaS will behave the same way:

```
18af6774c4ea :) select byteHammingDistance('foo', 'foobar');

SELECT byteHammingDistance('foo', 'foobar')

Query id: 36942bee-72cd-4496-8235-0ff2fee3d6c0

   ┌─byteHammingDistance('foo', 'foobar')─┐
1. │                                    3 │
   └──────────────────────────────────────┘

1 row in set. Elapsed: 0.003 sec.

18af6774c4ea :) select byteHammingDistance('foo', 'barfoo');

SELECT byteHammingDistance('foo', 'barfoo')

Query id: 453df233-a481-4e9e-8b50-a86386f77da6

   ┌─byteHammingDistance('foo', 'barfoo')─┐
1. │                                    6 │
   └──────────────────────────────────────┘

1 row in set. Elapsed: 0.004 sec.

18af6774c4ea :)
```

And yes, unfortunatelly it might be different from strict implementations, as we saw for `MurmurHash` and `SipHash` examples before 😢

dreadatour: > I'm just being picky

And thank you for that! Love these discussions and reviews ❤️

0x2b3bfa0 (approved):
Awesome! 🙏 

I didn't get why would you fix one docstring but not the other, but that's minor; see https://github.com/iterative/datachain/pull/669#discussion_r1874629141. 🤔

dreadatour:
> I didn't get why would you fix one docstring but not the other, but that's minor; see [#669 (comment)](https://github.com/iterative/datachain/pull/669#discussion_r1874629141). 🤔

I was thinking to go through all docstrings in `func` module to update them all in one go, but you're right, it should be done in this PR. Updated!

PR #673, opened 2024-12-09
chore: follow up docs navigation update

yathomasi:
- Remove icon from site title
- Use GH style link icon for headers
- Use left sidebar only for `API References`


dreadatour (approved):
❤️

PR #676, opened 2024-12-09
ci: add a consolidated `check` step for required PR checks

skshetry:
This commit introduces a check step that acts as a single point of failure for specified jobs in a workflow. The step fails if any of the specified jobs fail and passes only if they succeed. This is useful for adding a single step to "Required Checks" for PR.

This avoids the need to manually add each job to "Required Checks", and reduces maintenance overhead when renaming jobs.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -170,3 +170,13 @@ jobs:
         env:
           HF_TOKEN: ${{ secrets.HF_TOKEN }}
         run: nox -s examples -p ${{ matrix.pyv }} -- -m "${{ matrix.group }}"
+
+  check:
+    if: always()
+    needs: [lint, datachain, examples]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: re-actors/alls-green@release/v1
+        with:
+          allowed-failures: examples
+          jobs: ${{ toJSON(needs) }}


.github/workflows/tests.yml
```diff
@@ -170,3 +170,13 @@ jobs:
         env:
           HF_TOKEN: ${{ secrets.HF_TOKEN }}
         run: nox -s examples -p ${{ matrix.pyv }} -- -m "${{ matrix.group }}"
+
+  check:
+    if: always()
+    needs: [lint, datachain, examples]
```
skshetry, line 176: This PR will block merging until `examples` completes. That’s probably what we want, but I wanted to clarify in case it wasn’t obvious.

0x2b3bfa0: Have you considered the Studio approach, that doesn't need any third-party action?

skshetry: I do see a note in [the action's README that says the following](https://github.com/re-actors/alls-green#why):

> **you can make a single check job listing all of the other jobs in its needs list.** How about that? 🤯 Now you can add only that single job to your branch protection settings and you won't have to maintain that list manually anymore.
>
> Right..? Wrong 🙁 — apparently, in case when something fails, the check job's result is set to skipped and not failed as one would expect. This is problematic and requires extra work to get it right. Some of it is iteration over the needed jobs and checking that all their results are set to success but it's no fun to maintain this sort of thing in many different repositories. Also, it is mandatory to make the check job run always, not only when all of the needed jobs succeed.

I am not familiar about the Studio approach, and reading action's README makes me thing it's a bit more involved (?). Feel free to create a PR if you think it's straightforward.

PR #677, opened 2024-12-09
ci: drop macos example test run

skshetry:
GHA only provides 5 concurrent macOS jobs for a team. Running 8 macOS jobs for "examples" (for a single PR) is excessive, and frankly unnecessary here, since they are more or less covered in `ubuntu-latest`.

This makes us wait a lot of time before the PR succeeds, especially when GHA is overloaded.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -136,7 +136,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


mattseddon:
This will drop examples from being tested - https://github.com/iterative/datachain/blob/de6b1c683194de06e725bec23d6be481451abde0/tests/examples/test_examples.py#L24-L28

PR #678, opened 2024-12-09
Implement 'seed' for 'train_test_split' (take two)

dreadatour:
Testing simple approach discussed in https://github.com/iterative/datachain/pull/657

---

New `seed` param for `train_test_split` toolkit method was added:

```python
train, test, val = train_test_split(dc, [0.7, 0.2, 0.1], seed=37)
train.save("dataset_train")
test.save("dataset_test")
val.save("dataset_val")
```


dreadatour:
Test this implementation with this simple script:

```python
import random
import sys

from PIL import Image


def sys__rand():
    return random.randint(0, MAX_INT_64) - 2**63


MAX_INT_64 = 2**64 - 1
COLORS = [(0, 0, 0), (255, 255, 255), (255, 0, 0), (0, 255, 0), (0, 0, 255)]

seed = int(sys.argv[1])
weights = [float(val) for val in sys.argv[2:]]

resolution = 2**31 - 1  # Maximum positive value for a 32-bit signed integer.
uniform_seed = random.Random(seed).randrange(1, resolution)
weights_normalized = [weight / sum(weights) for weight in weights]

width, height = 512, 512
im = Image.new("RGB", (width, height), "black")
pixels = im.load()
for y in range(height):
    for x in range(int(width)):
        sys_rand_col = sys__rand()
        rand_col = sys_rand_col if seed < 0 else (sys_rand_col % resolution) * uniform_seed
        rand_col = rand_col % resolution

        for index, _ in enumerate(weights_normalized):
            if (
                rand_col >= round(sum(weights_normalized[:index]) * resolution) and
                rand_col < round(sum(weights_normalized[: index + 1]) * resolution)
            ):
                pixels[x, y] = COLORS[index]
                break

im.save("random_pattern.png", "PNG")
```

### No seed, 1:1 distribution

```
python test-train-split-img.py -1 1 1
```

random_pattern


### No seed, 10:1 distribution

```
python test-train-split-img.py -1 10 1
```

random_pattern

### No seed, 10:5:1 distribution

```
python test-train-split-img.py -1 10 5 1
```

random_pattern


### With seed, 1:1 distribution

```
python test-train-split-img.py 123456 1 1
```

random_pattern

### With seed, 10:1 distribution

```
python test-train-split-img.py 123456 10 1
```

random_pattern

---

Unless proven otherwise, I will assume this to be a uniform distribution and consider the task complete.

src/datachain/toolkit/split.py
```diff
@@ -58,14 +69,16 @@ def train_test_split(dc: DataChain, weights: list[float]) -> list[DataChain]:
 
     weights_normalized = [weight / sum(weights) for weight in weights]
 
-    resolution = 2**31 - 1  # Maximum positive value for a 32-bit signed integer.
+    rand_col = C("sys.rand")
+    if seed is not None:
+        uniform_seed = random.Random(seed).randrange(1, RESOLUTION)  # noqa: S311
+        rand_col = (rand_col % RESOLUTION) * uniform_seed  # type: ignore[assignment]
```
dreadatour, line 75: `(rand_col % RESOLUTION)` is needed because of [this](https://github.com/iterative/datachain/pull/657#issuecomment-2530115157):

Python:

```
In [1]: 8822966792743105186 * 1730071088
Out[1]: 15264359758508934493641462368

In [2]: (8822966792743105186 * 1730071088) % 2147483647
Out[2]: 581439120
```

SQLite:

```
sqlite> select 8822966792743105186 * 1730071088;
1.52643597585089e+28

sqlite> select (8822966792743105186 * 1730071088) % 2147483647;
1.0
```

src/datachain/toolkit/split.py
```diff
@@ -58,14 +69,16 @@ def train_test_split(dc: DataChain, weights: list[float]) -> list[DataChain]:
 
     weights_normalized = [weight / sum(weights) for weight in weights]
 
-    resolution = 2**31 - 1  # Maximum positive value for a 32-bit signed integer.
+    rand_col = C("sys.rand")
+    if seed is not None:
+        uniform_seed = random.Random(seed).randrange(1, RESOLUTION)  # noqa: S311
```
0x2b3bfa0, line 74: Good catch with the lower bound! I was relying on statistics and hoping it would never be `0` but better to make sure. 🙈 ❤️

0x2b3bfa0: Note that here `RESOLUTION` should be `RESOLUTION + 1`[^1] because `randrange` behaves like range, and does not include the last element.

```suggestion
        uniform_seed = random.Random(seed).randrange(1, RESOLUTION + 1)  # noqa: S311
```

[^1]: Because you declared `RESOLUTION` as `2**31 - 1` above.

src/datachain/toolkit/split.py
0x2b3bfa0: My bad, thanks! I assumed that SQLite was displaying "scientific notation" but it wouldn't be an issue. 🙈

0x2b3bfa0: On _modulo_ operations, `RESOLUTION` should be `RESOLUTION + 1` as well.

```suggestion
        rand_col = (rand_col % (RESOLUTION + 1)) * uniform_seed  # type: ignore[assignment]
```

src/datachain/toolkit/split.py
```diff
@@ -58,14 +69,16 @@ def train_test_split(dc: DataChain, weights: list[float]) -> list[DataChain]:
...
+        rand_col = (rand_col % RESOLUTION) * uniform_seed  # type: ignore[assignment]
+    rand_col = rand_col % RESOLUTION  # type: ignore[assignment]
```
0x2b3bfa0, line 76: On _modulo_ operations, `RESOLUTION` should be `RESOLUTION + 1` as well.

```suggestion
    rand_col = rand_col % (RESOLUTION + 1)  # type: ignore[assignment]
```

tests/conftest.py
```diff
@@ -654,18 +654,20 @@ def studio_datasets(requests_mock):
 
 @pytest.fixture
 def not_random_ds(test_session):
+    # `sys__rand` column is carefully crafted to ensure that `train_test_split` func
```
0x2b3bfa0, line 657: Thanks for documenting this! ❤️

0x2b3bfa0 (approved):
Thanks for this awesome work, @dreadatour! 🤩 

Just a small note with regard to half-open intervals: should `RESOLUTION` be `RESOLUTION + 1` for `%` and `randrange` but not for `*`?

src/datachain/toolkit/split.py
```diff
@@ -58,14 +69,16 @@ def train_test_split(dc: DataChain, weights: list[float]) -> list[DataChain]:
 
     weights_normalized = [weight / sum(weights) for weight in weights]
 
-    resolution = 2**31 - 1  # Maximum positive value for a 32-bit signed integer.
+    rand_col = C("sys.rand")
+    if seed is not None:
+        uniform_seed = random.Random(seed).randrange(1, RESOLUTION + 1)  # noqa: S311
+        rand_col = (rand_col % (RESOLUTION + 1)) * uniform_seed  # type: ignore[assignment]
+    rand_col = rand_col % RESOLUTION  # type: ignore[assignment]
```
dreadatour, line 76: @0x2b3bfa0 wait. `RESOLUTION` Should NOT be power of 2 for modulo. It should be a prime number for even distribution. `RESOLUTION = 2^31-1` is a prime (Mersenne prime to be exact), and `RESOLUTION + 1` is a power of 2, so it is much better to use `RESOLUTION` for modulo.

Since max reminder will be `RESOLUTION - 1`, we should use `RESOLUTION + 1` for multiplication below.

0x2b3bfa0: 1. Unless proven otherwise, we don't need `RESOLUTION` to be prime; this is just multiplication of big numbers[^1] _modulo_ **anything**.
2. If you use `2**31 - 1` for _modulo_ you should be using `2**31 - 2` for multiplication, not `2**31 + 2` lest leaving some of the items unreachable.

[^1]: More specifically two big, pseudo-random numbers: a constant `seed` and a variable a sys.rand`

dreadatour: 🤔

```
In [1]: import math

In [2]: RESOLUTION = 2**31 - 1

In [3]: math.log((RESOLUTION - 1) * (RESOLUTION + 1), 2)
Out[3]: 61.99999999865639
```

0x2b3bfa0:
```python 
>>> one = 2**31 * 2**31
4611686018427387904
>>> two = (2**31 - 1) * (2**31 + 1)
4611686018427387903
>>> one - two
1
```

dreadatour: The best conversation ever 🤣
```
In [1]: (2**31 - 1) * (2**31 + 1) == 2**62 - 1
Out[1]: True
```

dreadatour: Argh, of course, you're right, should be `RESOLUTION - 1` in multiplication below.

src/datachain/toolkit/split.py
```diff
@@ -58,14 +69,16 @@ def train_test_split(dc: DataChain, weights: list[float]) -> list[DataChain]:
...
 
     return [
         dc.filter(
-            C("sys__rand") % resolution
-            >= round(sum(weights_normalized[:index]) * resolution),
-            C("sys__rand") % resolution
-            < round(sum(weights_normalized[: index + 1]) * resolution),
+            rand_col >= round(sum(weights_normalized[:index]) * (RESOLUTION + 1)),
+            rand_col < round(sum(weights_normalized[: index + 1]) * (RESOLUTION + 1)),
```
dreadatour, line 81:
```suggestion
            rand_col >= round(sum(weights_normalized[:index]) * (RESOLUTION - 1)),
            rand_col < round(sum(weights_normalized[: index + 1]) * (RESOLUTION - 1)),
```

src/datachain/toolkit/split.py
```diff
@@ -58,14 +69,16 @@ def train_test_split(dc: DataChain, weights: list[float]) -> list[DataChain]:
...
+    rand_col = rand_col % RESOLUTION  # type: ignore[assignment]
 
     return [
         dc.filter(
-            C("sys__rand") % resolution
-            >= round(sum(weights_normalized[:index]) * resolution),
-            C("sys__rand") % resolution
-            < round(sum(weights_normalized[: index + 1]) * resolution),
+            rand_col >= round(sum(weights_normalized[:index]) * (RESOLUTION - 1)),
+            rand_col < round(sum(weights_normalized[: index + 1]) * (RESOLUTION - 1)),
```
0x2b3bfa0, line 81:
```suggestion
            rand_col >= round(sum(weights_normalized[:index]) * RESOLUTION),
            rand_col < round(sum(weights_normalized[: index + 1]) * RESOLUTION),
```

0x2b3bfa0: @dreadatour, https://github.com/iterative/datachain/pull/678/commits/24cd3b573d95dedfad49e5444621abb5f4c0035a was not needed.

src/datachain/toolkit/split.py
```diff
@@ -1,7 +1,16 @@
+import random
+from typing import Optional
+
 from datachain import C, DataChain
 
+RESOLUTION = 2**31 - 1  # Maximum positive value for a 32-bit signed integer.
```
0x2b3bfa0, line 6: Off the top of my head, distribution doesn't get any "better" by using "the @dreadatour prime"[^1] 😄 for the _modulo_ operation.

We could just use `2**31`; thus, considering only the least significant 31 bits of `sys.rand` and discarding the rest; but it's nicer to use your approach and take into account all the bits.

```suggestion
RESOLUTION = 2**31
```

[^1]: A particular Mersenne prime: `2**32 - 1`

0x2b3bfa0 (approved):

PR #680, opened 2024-12-09
Add log streaming to the created job

amritghimire:
When running a job from studio, this adds a log streaming which checks
for the updates from the Studio.

Related PR: https://github.com/iterative/studio/pull/11068

image


dreadatour (approved):
Looks good to me! Thank you for this update 🥰

src/datachain/remote/studio.py
```diff
@@ -231,6 +235,38 @@ def _unpacker_hook(code, data):
...
+        Follow job logs via websocket connection.
+
+        Args:
+            job_id: ID of the job to follow logs for
+
+        Yields:
+            Dict containing either job status updates or log messages
+        """
+        parsed_url = urlparse(self.url)
+        ws_url = urlunparse(parsed_url._replace(scheme="ws"))
```
0x2b3bfa0, line 249: Let's not use WebSockets without TLS, please. At least, let's match the scheme from `self.url` converting `http` to `ws` and `https` to `wss` to simplify local development.

```suggestion
        ws_url = urlunparse(parsed_url._replace(scheme="wss"))
```

0x2b3bfa0 (changes requested):

src/datachain/remote/studio.py
```diff
@@ -231,6 +235,40 @@ def _unpacker_hook(code, data):
...
+            job_id: ID of the job to follow logs for
+
+        Yields:
+            Dict containing either job status updates or log messages
+        """
+        parsed_url = urlparse(self.url)
+        ws_url = urlunparse(
+            parsed_url._replace(scheme="wss" if parsed_url.scheme == "https" else "ws")
+        )
+        ws_url = f"{ws_url}/logs/follow/?job_id={job_id}&team_name={self.team}"
```
0x2b3bfa0, line 252: Why do we need to pass the team name?

0x2b3bfa0 (approved):
Still not sure WebSockets are a good idea, especially for log streaming. I'm not even talking about SSE here, but even about plain old polling.

0x2b3bfa0 (commented):
It works, but I wonder if it should be closer to real-time; the following example seems to be buffering the whole output and takes ages to print the logs: 🧐 

#### `datachain studio run test.py`
```python
import time

for number in range(60 * 60):
  time.sleep(0.1)
  print(number)
```

amritghimire:
> It works, but I wonder if it should be closer to real-time; the following example seems to be buffering the whole output and takes ages to print the logs: 🧐 
> 
> 
> 
> #### `datachain studio run test.py`
> 
> ```python
> 
> import time
> 
> 
> 
> for number in range(60 * 60):
> 
>   time.sleep(0.1)
> 
>   print(number)
> 
> ```

Found an possible cause in Studio side. Working on it alongside the test cases for this.

src/datachain/remote/studio.py
amritghimire: It seems to be an existing approach for Studio endpoints that takes the team name to check for permission before any other tasks. I reused the same decorator and approach to make sure the permission is correct

0x2b3bfa0: This doesn't feel right to me. 🤔 In theory, the job identifier and the access token should be the only values needed to check for permission; anything else is a potential footgun in the best case, or an actual footgun in the worst case.

0x2b3bfa0: Resolving as off-topic, but pinging @dreadatour just to make noise.

PR #691, opened 2024-12-11
torch-loader(example): use prefetch and try to run example in linux

skshetry:


examples/get_started/torch-loader.py
```diff
@@ -64,8 +66,9 @@ def forward(self, x):
 
     train_loader = DataLoader(
         ds.to_pytorch(transform=transform),
-        batch_size=16,
-        num_workers=2,
+        batch_size=25,
+        num_workers=4,
+        multiprocessing_context=multiprocessing.get_context("spawn"),
```
skshetry, line 71: `fsspec`'s `loop` is not fork-safe. Even though we create a new loop for each forked processes, s3fs or other filesystems may not be fork-safe.

See https://s3fs.readthedocs.io/en/latest/#multiprocessing.

This causes future from a `run_coroutine_threadsafe` to hang forever on prefetch.

I could contribute a fix to the first problem, but the second problem still remains. 
Also, Python 3.14 is changing default start method for posix systems (except macOS which uses `'spawn'`) to `'forkserver'`. See https://github.com/python/cpython/issues/84559.

So, I think it's better to recommend to use a different start method.

mattseddon (approved):

PR #693, opened 2024-12-12
cli: improve startup time

skshetry:
This PR halves the total startuptime from 2.8s to 1.4s, with minimal effort.

There are two more imports that can be reduced: requests, and numpy, that will reduce total time by .3 sec.

The total import time spent is now 1.251s. It's difficult to fix remaining imports because we expose them on top-level imports, and fixing them will require too much duplication/effort or incur runtime costs.

The import trace looks like follows after this PR:

Screenshot 2024-12-12 at 08 40 47

### DIFF
--- a/src/datachain/client/__init__.py
+++ b/src/datachain/client/__init__.py
@@ -1,4 +1,3 @@
 from .fsspec import Client
-from .s3 import ClientS3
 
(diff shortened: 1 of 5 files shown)


shcheklein (approved):
thanks!

PR #694, opened 2024-12-12
torch-loader(example): use persistent workers to reduce test time

skshetry:


examples/get_started/torch-loader.py
```diff
@@ -69,6 +69,7 @@ def forward(self, x):
         ds.to_pytorch(transform=transform),
         batch_size=25,
         num_workers=4,
+        persistent_workers=True,
```
skshetry, line 72: With `persistent_workers` set, it takes much less time to run 3 epochs than before.

noxfile.py
```diff
@@ -81,6 +81,8 @@ def examples(session: nox.Session) -> None:
     session.install(".[examples]")
     session.run(
         "pytest",
+        "--durations=0",
```
skshetry, line 84: This will be useful to know how much each tests take.

examples/get_started/torch-loader.py
```diff
@@ -69,6 +69,7 @@ def forward(self, x):
         ds.to_pytorch(transform=transform),
         batch_size=25,
         num_workers=4,
```
skshetry, line 71:
```suggestion
        num_workers=max(4, os.cpu_count() or 2),
```

mattseddon (approved):

PR #695, opened 2024-12-12
Implement cancel job operation for CLI

amritghimire:
This introduces a cancel job operation for CLI that can be used to
cancel a job running in studio.

Command:
`datachain studio cancel <JOB_ID`

### DIFF
--- a/src/datachain/cli.py
+++ b/src/datachain/cli.py
@@ -294,6 +294,28 @@ def add_studio_parser(subparsers, parent_parser) -> None:
         help="Python package requirement. Can be specified multiple times.",
     )
 
+    studio_cancel_help = "Cancel a job in Studio"
+    studio_cancel_description = "This command cancels a job in Studio."
+
+    studio_cancel_parser = studio_subparser.add_parser(
+        "cancel",
+        parents=[parent_parser],
+        description=studio_cancel_description,
+        help=studio_cancel_help,
+    )
+
+    studio_cancel_parser.add_argument(
+        "job_id",
+        action="store",
+        help="The job ID to cancel.",
+    )
+    studio_cancel_parser.add_argument(
+        "--team",
+        action="store",
+        default=None,
+        help="The team to cancel a job for. By default, it will use team from config.",
+    )
+
 
 def get_parser() -> ArgumentParser:  # noqa: PLR0915
     try:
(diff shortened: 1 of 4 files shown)


dreadatour (approved):
Looks good to me!

I am not sure `datachain studio cancel` is a good command name, it is too general and not verbose enough IMO. But I have no better options.

amritghimire:
> thanks @amritghimire !
> 
> 
> 
> `datachain studio cancel <JOB_ID` 
> 
> 
> 
> let's rename it please though.
> 
> 
> 
> It should be:
> 
> 
> 
> `datachain job cancel --studio` (similar to `ds`)
> 

I think that it may be confusing to users since we don't have non studio version of job in local. The reason I included the cancel inside studio is that the operation is a concept totally located in Studio.

amritghimire:
> same with `run`, same with API - it should be consistent + we should be using PUT, POST, DELETE, etc 
> 
> 
> 
> 

https://github.com/iterative/studio/pull/11077#discussion_r1885529578

shcheklein:
> I think that it may be confusing to users since we don't have non studio version of job in local. The reason I included the cancel inside studio is that the operation is a concept totally located in Studio.

I see the point.

But, I don't think it justifies the inconsistency tbh (datasets, jobs, etc - all should be the same in API and in CLI to my mind). Also the name "datachain studio cancel" is strange a bit anymways. It should be then `datachain studio job cancel`

or let's find some other options please ...

shcheklein:
> https://github.com/iterative/studio/pull/11077#discussion_r1885529578

is it PUT / POST though then?

amritghimire:
> > I think that it may be confusing to users since we don't have non studio version of job in local. The reason I included the cancel inside studio is that the operation is a concept totally located in Studio.
> 
> I see the point.
> 
> But, I don't think it justifies the inconsistency tbh (datasets, jobs, etc - all should be the same in API and in CLI to my mind). Also the name "datachain studio cancel" is strange a bit anymways. It should be then `datachain studio job cancel`
> 
> or let's find some other options please ...

WDYT about `datachain studio cancel-job`?

>is it PUT / POST though then?

It is POST currently. We already have POST request for update-job.

shcheklein:
> WDYT about datachain studio cancel-job?

I think it is still not good, since it's not consistent with datasets. We should be using the same general schema for all types of entities. This is quite important I think.

PR #713, opened 2024-12-13
Optimize UDF with parallel execution

dreadatour:
## Highlights

In this PR I am:

- Passing IDs only instead of whole rows into parallel processes UDFs. This will prevent multiple data conversion (pickling) along the way
- Write to DB right from parallel process instead of passing results pack into main process

This will prevent multiple types conversion.

Before it was:

1. Read rows from DB -> convert them from raw DB types into Python types (whole raw)
2. Pass rows into parallel processes via `multiprocess.Queue` -> convert rows from Python types with msgpack
3. Read rows in parallel process from `multiprocess.Queue` -> convert rows to Python types (maspack)
4. Process UDF
5. Send rows back to the main process via `multiprocess.Queue` -> convert from Python type (msgpack)
6. In main process read rows from `multiprocess.Queue` -> convert them back into Python type (msgpack)
7. Write rows into DB -> convert them from Python types into DB raw types

After:

1. Read rows IDs from DB -> convert DB `int`s into Python `int`s is quick, stable, predictable
2. Pass rows IDs to parallel processes in batches via `multiprocess.Queue` -> convert list of IDs with pickle (by default)
3. In parallel process for each batch of IDs read rows from DB -> convert them from raw DB types into Python types
4. Process UDF
5. Write result rows into DB right from the parallel process -> convert them from Python types into DB raw types

In the end:

- no multiple type conversion
- no more single point reading/writing from/to warehouse DB: read and write from parallel processes (worker machines in future)

## Test scenario

Simple script to check raw `parallel` setting only:

```python
import sys


def path_len(file: File) -> int:
    return len(file.path)


parallel = int(sys.argv[1]) if len(sys.argv) > 1 else None


DataChain \
    .from_dataset("50k-laion-files") \
    .settings(parallel=parallel, prefetch=0) \
    .map(path_len=path_len) \
    .save("50k-laion-files-len")
```

This is very simple and basic scenario, but it helps us to test `parallel` setting only, without any overheads.

**Note** `prefetch` is off in this case to measure `parallel` only

## Overview

On the chart below there are two series: before optimization (blue) and after (green). On the X axis is parallel processes count, on the Y axis is number of total rows processed by UDF in parallel. This is valid for SQLite warehouse on my local machine.

image

As we can see, "before" `parallel` option does not affect performance at all, there is a strict limit on performance and it does not depends on number of parallel processes.

The reason is because we pass rows into UDF for each parallel process via `multiprocess.Queue` and get results back the same way. `Queue` performance is very limited. I wrote [a simple script](https://gist.github.com/dreadatour/59508ce92010ebe6b4bc5ee97e430d21) to test `Queue` only and it is limited indeed. I have tried different ways for IPC (Pipes, ZeroMQ) and they all have this limit. This can be solved introducing external dependencies (Redis, RabbitMQ, etc), but it is not what we want for CLI tool.

"After" performs much better, `Queue` is used only to pass IDs in batches and is performant enough to show the performance boost depending on parallel processes count. It is not linear, because performance of SQLite warehouse is now the limit, but it is much better, stable and predictable.

Also note "1 parallel process" performance is ~2.15 times slower than clean "no parallel processes" and this is basically overhead for using parallel processes and queues to read and pass IDs. on 2-3 parallel processes performance is the same as on "no parallel" and it is increasing over parallel processes count increasing.

**Next** I am going to measure the same numbers on ClickHouse DB warehouse, I suppose it is going to be much better and linear.

##### Not parallel (for reference)

```
$ time python path-len.py
Preparing: 129136 rows [00:01, 117496.00 rows/s]
Processed: 129136 rows [00:04, 30999.24 rows/s]
Cleanup: 2 tables [00:00, 297.27 tables/s]
python path-len.py  7.14s user 3.01s system 131% cpu 7.728 total
```

##### Parallel = 1 (edge case)

```
$ time python path-len.py 1
Preparing: 129136 rows [00:01, 114113.84 rows/s]
Processed: 129136 rows [00:11, 11689.08 rows/s]
Cleanup: 2 tables [00:00, 310.41 tables/s]
python path-len.py 1  20.59s user 11.62s system 202% cpu 15.872 total
```

##### Parallel = 8

```
$ time python path-len.py 8
Preparing: 129136 rows [00:01, 117677.17 rows/s]
Processed: 129136 rows [00:10, 12675.34 rows/s]
Cleanup: 2 tables [00:00, 331.29 tables/s]
python path-len.py 8  32.53s user 15.31s system 306% cpu 15.630 total
```

##### Not parallel (for reference)

```
$ time python path-len.py
Preparing: 129136 rows [00:01, 113141.62 rows/s]
Processed: 129136 rows [00:04, 30850.59 rows/s]
Cleanup: 2 tables [00:00, 275.74 tables/s]
python path-len.py  7.06s user 3.38s system 135% cpu 7.701 total
```

##### Parallel = 1 (edge case)

```
$ time python path-len.py 1
Preparing: 129136 rows [00:01, 115298.96 rows/s]
Processed: 129137 rows [00:08, 15122.98 rows/s]
Cleanup: 2 tables [00:00, 317.81 tables/s]
python path-len.py 1  14.27s user 9.30s system 173% cpu 13.569 total
```

##### Parallel = 8

```
$ time python path-len.py 8
Preparing: 129136 rows [00:01, 115127.01 rows/s]
Processed: 129144 rows [00:03, 39761.77 rows/s]
Cleanup: 2 tables [00:00, 295.43 tables/s]
python path-len.py 8  27.65s user 13.99s system 452% cpu 9.206 total
```


src/datachain/lib/udf.py
```diff
@@ -85,7 +85,6 @@ def run(
         udf_fields: "Sequence[str]",
         udf_inputs: "Iterable[RowsOutput]",
         catalog: "Catalog",
-        is_generator: bool,
```
dreadatour, line 88: Not used anywhere

src/datachain/query/dispatch.py
```diff
@@ -54,54 +56,49 @@ def udf_entrypoint() -> int:
...
-        # Use default number of CPUs (cores)
-        n_workers = None
+        n_workers = None  # Use default number of CPUs (cores)
+
+    wh_cls, wh_args, wh_kwargs = udf_info["warehouse_clone_params"]
+    warehouse: AbstractWarehouse = wh_cls(*wh_args, **wh_kwargs)
 
     with contextlib.closing(
-        batching(warehouse.dataset_select_paginated, query)
+        batching(warehouse.db.execute, query, ids_only=True)
```
dreadatour, line 85: Not sure yet, but it looks like we don't need pagination here since we are only selecting IDs.
Should be tested on bigger scale and confirmed.

src/datachain/query/dispatch.py
```diff
@@ -54,54 +56,49 @@ def udf_entrypoint() -> int:
...
         try:
-            udf_results = dispatch.run_udf_parallel(
-                marshal(udf_inputs),
+            dispatch.run_udf_parallel(
+                udf_inputs,
                 n_workers=n_workers,
                 processed_cb=processed_cb,
                 download_cb=download_cb,
             )
-            process_udf_outputs(warehouse, table, udf_results, udf, cb=generated_cb)
```
dreadatour, line 97: We are processing results and inserting them into DB in parallel processes now.

src/datachain/query/dispatch.py
```diff
@@ -224,6 +219,9 @@ def run_udf_parallel(  # noqa: C901, PLR0912
             input_finished = False
 
             if not streaming_mode:
+                if not self.is_batching:
+                    input_rows = batched(flatten(input_rows), DEFAULT_BATCH_SIZE)
```
dreadatour, line 223: Batching and flatten IDs for better performance so each parallel process will receive batch of IDs and can select them all at once instead of many single selects.

Here is it better to analyze input dataset size (`count`) and adjust batch size because on small input datasets it can be one parallel process who get all the rows IDs. Item for the follow-up issue, I guess.

src/datachain/query/dispatch.py
```diff
@@ -238,19 +236,17 @@ def run_udf_parallel(  # noqa: C901, PLR0912
...
+                if processed := result.get("processed"):
+                    processed_cb.relative_update(processed)
+
                 status = result["status"]
-                if status == NOTIFY_STATUS:
-                    if downloaded := result.get("downloaded"):
-                        download_cb.relative_update(downloaded)
-                    if processed := result.get("processed"):
-                        processed_cb.relative_update(processed)
+                if status in (OK_STATUS, NOTIFY_STATUS):
```
dreadatour, line 246: We are doing updates above now for all types of signals, no need to process these signals here.

src/datachain/query/dispatch.py
```diff
@@ -325,31 +323,57 @@ def _default_callback(self) -> WorkerCallback:
...
+            self.notify_and_process(udf_results, processed_cb),
+            self.udf,
+            cb=processed_cb,
+        )
+        warehouse.insert_rows_done(self.table)
+
+        put_into_queue(
+            self.done_queue,
+            {"status": FINISHED_STATUS, "processed": processed_cb.processed_rows},
+        )
```
dreadatour, line 349: Do not pass results into main process, write them right into DB here.

src/datachain/query/dispatch.py
```diff
@@ -325,31 +323,57 @@ def _default_callback(self) -> WorkerCallback:
...
+
+    def notify_and_process(self, udf_results, processed_cb):
+        for row in udf_results:
             put_into_queue(
                 self.done_queue,
-                {"status": NOTIFY_STATUS, "processed": processed_cb.processed_rows},
+                {"status": OK_STATUS, "processed": processed_cb.processed_rows},
             )
-        put_into_queue(self.done_queue, {"status": FINISHED_STATUS})
+            yield row
```
dreadatour, line 357: Helper function to notify process before writing results into DB.

amritghimire:
> Also note "1 parallel process" performance is ~2.15 times slower than clean "no parallel processes" and this is basically overhead for using parallel processes and queues to read and pass IDs. on 2-3 parallel processes performance is the same as on "no parallel" and it is increasing over parallel processes count increasing.

Is it possible to treat 1 parallel process as no parallel processes or raise error when only one parallel is specified?

skshetry:
@dreadatour, while you are working on this, could you please also take a look at this example test:

https://github.com/iterative/datachain/actions/runs/12326251928/job/34406754039?pr=713#step:7:29

> 862.81s call     tests/examples/test_examples.py::test_get_started_examples[examples/get_started/udfs/parallel.py]

Does this PR improve that test? Should it take that long?

dreadatour:
> Is it possible to treat 1 parallel process as no parallel processes or raise error when only one parallel is specified?

Sure, sounds reasonable 👍

> @dreadatour, while you are working on this, could you please also take a look at this example test:
> 
> https://github.com/iterative/datachain/actions/runs/12326251928/job/34406754039?pr=713#step:7:29
> 
> > 862.81s call     tests/examples/test_examples.py::test_get_started_examples[examples/get_started/udfs/parallel.py]
> 
> Does this PR improve that test? Should it take that long?

Nice catch, let me take a look 🙏

dreadatour:
> @dreadatour, while you are working on this, could you please also take a look at this example test:
> 
> https://github.com/iterative/datachain/actions/runs/12326251928/job/34406754039?pr=713#step:7:29
> 
> > 862.81s call     tests/examples/test_examples.py::test_get_started_examples[examples/get_started/udfs/parallel.py]
> 
> Does this PR improve that test? Should it take that long?

Found an issue. This is because of [this](https://github.com/iterative/datachain/pull/713#discussion_r1884797977) (basically everything runs in single process because batch size is 10k and number of records is 400).

Couple tests:

##### `main` branch

```
$ time python examples/get_started/udfs/parallel.py
Preparing: 400 rows [00:00, 75699.21 rows/s]
Processed: 400 rows [00:35, 11.34 rows/s]
Cleanup: 2 tables [00:00, 10525.23 tables/s]
python examples/get_started/udfs/parallel.py  391.85s user 15.58s system 948% cpu 42.941 total
```

##### this branch

Basically single process

```
$ time python examples/get_started/udfs/parallel.py
Preparing: 400 rows [00:00, 107926.77 rows/s]
Processed: 401 rows [04:01,  1.66 rows/s]
Cleanup: 2 tables [00:00, 10512.04 tables/s]
python examples/get_started/udfs/parallel.py  260.37s user 15.32s system 110% cpu 4:09.27 total
```

##### this branch but batch size is set to `10`

```
$ time python examples/get_started/udfs/parallel.py
Preparing: 400 rows [00:00, 110894.41 rows/s]
Processed: 412 rows [00:38, 10.64 rows/s]
Cleanup: 2 tables [00:00, 10356.31 tables/s]
python examples/get_started/udfs/parallel.py  391.61s user 18.10s system 885% cpu 46.277 total
```

dreadatour:
> @dreadatour, while you are working on this, could you please also take a look at this example test:
> 
> https://github.com/iterative/datachain/actions/runs/12326251928/job/34406754039?pr=713#step:7:29
> 
> > 862.81s call     tests/examples/test_examples.py::test_get_started_examples[examples/get_started/udfs/parallel.py]
> 
> Does this PR improve that test? Should it take that long?

Fixed: https://github.com/iterative/datachain/actions/runs/12372015291/job/34529325219?pr=713 (409 sec)
Current main for reference: https://github.com/iterative/datachain/actions/runs/12367070520/job/34514761451 (same 410 sec)

src/datachain/query/dispatch.py
ilongin: our `sys__id` is 8 bytes, so on 1B scale this will take 8GB of memory by my calculation. I would still maybe leave it paginated.

src/datachain/query/dispatch.py
```diff
@@ -326,30 +302,54 @@ def _default_callback(self) -> WorkerCallback:
...
     def get_inputs(self):
-        while (batch := get_from_queue(self.task_queue)) != STOP_SIGNAL:
-            yield batch
+        warehouse = self.catalog.warehouse.clone()
+        col_id = get_query_id_column(self.query)
+
+        if self.is_batching:
+            while (batch := get_from_queue(self.task_queue)) != STOP_SIGNAL:
+                ids = [row[0] for row in batch.rows]
+                rows = warehouse.dataset_rows_select(self.query.where(col_id.in_(ids)))
```
ilongin, line 343: One thing that I'm worried about is this query in Clickhouse. There we have rows sorted and "packed" in granules by `sys__id` which is primary key. It would be ideal if these `id` batches are all sorted and "close" to each other, as otherwise we could end up in situation where big chunk of DB is read for every batch just because one id ended up in first granule / part, other id was in second one etc. This is because CH reads whole part / granule even if there is we need only one record from it.

ilongin (changes requested):
Good work! I'\ve added a couple of small comments and a few thoughts / questions.

src/datachain/query/dataset.py
```diff
@@ -464,8 +465,8 @@ def populate_udf_table(self, udf_table: "Table", query: Select) -> None:
 
                 with subprocess.Popen(cmd, env=envs, stdin=subprocess.PIPE) as process:  # noqa: S603
                     process.communicate(process_data)
-                    if process.poll():
-                        raise RuntimeError("UDF Execution Failed!")
+                    if ret := process.poll():
```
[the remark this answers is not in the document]
dreadatour: `retval` may be? 👀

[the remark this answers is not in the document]
skshetry: Alternatively, it could be renamed `retval` (as you have proposed), `exitcode`, `retcode`, etc.

0x2b3bfa0 (approved):

0x2b3bfa0 (approved):

ilongin (approved):

0x2b3bfa0:
Amazing work! 🤩 

Unless I missed something, this sets the foundations for truly scalable distributed processing, where, instead of having a central process batch + pass the rows inefficiently, each worker can query its own batch of records using the same algorithm as `train_test_split` but based on `sys__id`

https://github.com/iterative/datachain/blob/60256d618a79d53c1d18872cd93c6095e442aa1c/src/datachain/toolkit/split.py#L79-L83

Workers would only need to know two variables: the total number of workers, and the current worker's index.

0x2b3bfa0:
Also, please consider — as a follow-up — making sure that this approach is also efficient on ClickHouse, as mentioned on https://github.com/iterative/datachain/pull/713#discussion_r1890063875

src/datachain/query/dispatch.py
shcheklein: Q: is there a way to get in CH ids in the order they are stored? (to guarantee that we read sequentially then)?

src/datachain/query/dispatch.py
```diff
@@ -326,30 +302,54 @@ def _default_callback(self) -> WorkerCallback:
...
+                yield RowsOutputBatch(list(rows))
+        else:
+            while (batch := get_from_queue(self.task_queue)) != STOP_SIGNAL:
+                yield from warehouse.dataset_rows_select(
+                    self.query.where(col_id.in_(batch))
+                )
+
+
+class RepeatTimer(Timer):
```
shcheklein, line 352: is it used anywhere?

src/datachain/query/dispatch.py
```diff
@@ -326,30 +302,54 @@ def _default_callback(self) -> WorkerCallback:
...
-        while (batch := get_from_queue(self.task_queue)) != STOP_SIGNAL:
-            yield batch
+        warehouse = self.catalog.warehouse.clone()
+        col_id = get_query_id_column(self.query)
+
+        if self.is_batching:
+            while (batch := get_from_queue(self.task_queue)) != STOP_SIGNAL:
+                ids = [row[0] for row in batch.rows]
+                rows = warehouse.dataset_rows_select(self.query.where(col_id.in_(ids)))
+                yield RowsOutputBatch(list(rows))
```
shcheklein, line 344: Q: should / can it be covered with a test?

src/datachain/query/dispatch.py
```diff
@@ -189,26 +161,27 @@ def create_input_queue(self):
...
         download_cb: Callback = DEFAULT_CALLBACK,
-    ) -> Iterator[Sequence[UDFResult]]:
+    ) -> None:
         n_workers = get_n_workers_from_arg(n_workers)
 
+        input_batch_size = total_rows // n_workers
+        if input_batch_size == 0:
+            input_batch_size = 1
+        elif input_batch_size > DEFAULT_BATCH_SIZE:
+            input_batch_size = DEFAULT_BATCH_SIZE
```
shcheklein, line 176: should it be then MAX_BATCH_SIZE?

PR #718, opened 2024-12-17
Added `DataChain.diff()`

ilongin:
This is wrapper method around more generic `DataChain.compare()` which is used to calculate diff between file based chains (those which have `File` objects in it).
For matching same rows file signals `source` and `path` are used.
For comparing if row is modified / unchanged, file signals `version` and `etag` are used.


src/datachain/lib/dc.py
```diff
@@ -1700,6 +1700,77 @@ def compare(
...
+            modified (bool): Whether to return modified rows in resulting chain.
+            unchanged (bool): Whether to return unchanged rows in resulting chain.
+            status_col (str): Name of the new column that is created in resulting chain
+                representing diff status.
+
+        Example:
+            ```py
+            diff = images.diff(
+                new_images,
+                on=["file"],
```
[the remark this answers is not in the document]
ilongin: Nice catch, fixed!

tests/unit/lib/test_datachain.py
```diff
@@ -3350,3 +3350,108 @@ def test_compare_right_compare_wrong_length(test_session):
     assert str(exc_info.value) == (
         "'compare' and 'right_compare' must be have the same length"
     )
+
+
+@pytest.mark.parametrize("added", (True, False))
+@pytest.mark.parametrize("deleted", (True, False))
+@pytest.mark.parametrize("modified", (True, False))
+@pytest.mark.parametrize("unchanged", (True, False))
+@pytest.mark.parametrize("status_col", ("diff", None))
```
dmpetrov, line 3359: The test looks overcomplicated 🙂

Is it possible to separate this to small tests instead of large parametrized one?

Also, I don't see value in all modified, deleted, unchanged statuses since compare() tests have to cover this, not diff(). The same for `status_col`. 2-3 simple tests should cover diff() pretty well.

tests/unit/lib/test_datachain.py
```diff
@@ -3350,3 +3350,108 @@ def test_compare_right_compare_wrong_length(test_session):
...
+        collect_fields = collect_fields[1:]
+
+    assert list(diff.order_by("file.source").collect(*collect_fields)) == expected
+
+
+@pytest.mark.parametrize("added", (True, False))
+@pytest.mark.parametrize("deleted", (True, False))
+@pytest.mark.parametrize("modified", (True, False))
+@pytest.mark.parametrize("unchanged", (True, False))
+@pytest.mark.parametrize("status_col", ("diff", None))
```
dmpetrov, line 3409: Same here - to complicated to unit test and it test mostly compare(), not diff().
I'd appreciate it if you could simplify it.

tests/unit/lib/test_datachain.py
```diff
@@ -3350,3 +3350,108 @@ def test_compare_right_compare_wrong_length(test_session):
...
+
+    expected = []
+    if modified:
+        expected.append(("M", fs1_updated, 1))
+    if added:
+        expected.append(("A", fs2, 2))
+    if deleted:
+        expected.append(("D", fs3, 3))
+    if unchanged:
+        expected.append(("U", fs4, 4))
```
dmpetrov, line 3395: 🤔  I was sure that we decided to get rid of `U` and use `S` (Same) instead, didn't we?

dmpetrov (approved):
Amazing ✨ 
Please see comments inline. I'm approving to not to block it.

tests/unit/lib/test_datachain.py
```diff
@@ -3350,3 +3350,108 @@ def test_compare_right_compare_wrong_length(test_session):
     assert str(exc_info.value) == (
```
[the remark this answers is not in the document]
ilongin: Done

src/datachain/lib/dc.py
```diff
@@ -1700,6 +1700,77 @@ def compare(
...
+        other: "DataChain",
+        on: str = "file",
+        right_on: Optional[str] = None,
+        added: bool = True,
+        deleted: bool = True,
+        modified: bool = True,
+        unchanged: bool = False,
+        status_col: Optional[str] = None,
+    ) -> "DataChain":
+        """Similar as .compare() but for file based chains, i.e. those that have
```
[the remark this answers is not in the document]
ilongin: Added better docs

tests/unit/lib/test_datachain.py
ilongin: Simplified it

tests/unit/lib/test_datachain.py
ilongin: Changed to `S` .. also changed flag name from `unchanged` to `same`

PR #719, opened 2024-12-17
Added toolkit `compare`

ilongin:
Added `compare` function to toolkit. It uses `datachain.lib.diff.compare()` to compare `DataChain`s and returns dictionary with values for each of `added`, `removed`, `modified` and `unchanged` chains. Each chain consists only of those changes that are related to it (`added` only has added fields, `deleted` only `deleted fields etc.)
Keys of dicts are shortcuts for each status: `A`, `D`, `M` and `U`. 
Status column is removed from each resulting chain as it's not needed.

Also added `CompareStatus` enum and replaced hardcoded status letters (`A`, `D` ...) across the codebasej


shcheklein:
@skshetry @dreadatour can we please review this folks?

shcheklein:
@iterative/datachain a reminder, please take a look team.

src/datachain/lib/diff.py
```diff
@@ -112,25 +125,27 @@ def _to_list(obj: Union[str, Sequence[str]]) -> list[str]:
                 for c in [f"{_rprefix(c, rc)}{rc}" for c, rc in zip(on, right_on)]
             ]
         )
-        diff_cond.append((added_cond, "A"))
+        diff_cond.append((added_cond, CompareStatus.ADDED))
```
dreadatour, line 128: ❤️

src/datachain/toolkit/diff.py
```diff
@@ -0,0 +1,111 @@
...
+    unchanged: bool = False,
+) -> dict[str, "DataChain"]:
+    """Comparing two chains by identifying rows that are added, deleted, modified
+    or unchanged. Result is the new chain that has additional column with possible
+    values: `A`, `D`, `M`, `U` representing added, deleted, modified and unchanged
+    rows respectively. Note that if only one "status" is asked, by setting proper
+    flags, this additional column is not created as it would have only one value
+    for all rows. Beside additional diff column, new chain has schema of the chain
+    on which method was called.
+
```
dreadatour, line 34: It looks like these lines are outdated and needs to be removed from docstring?

dreadatour (approved):
Looks good to me!
Not related to this PR, but it is also would be nice to have an example of real-world usage 👀

src/datachain/toolkit/diff.py
```diff
@@ -0,0 +1,111 @@
...
+    left: "DataChain",
+    right: "DataChain",
+    on: Union[str, Sequence[str]],
+    right_on: Optional[Union[str, Sequence[str]]] = None,
+    compare: Optional[Union[str, Sequence[str]]] = None,
+    right_compare: Optional[Union[str, Sequence[str]]] = None,
+    added: bool = True,
+    deleted: bool = True,
+    modified: bool = True,
+    unchanged: bool = False,
```
skshetry, line 25: Do we need these arguments? Can we leave this up to the user to filter? 
```python
dc.compare(...).filter(C("col") == "added")
```

src/datachain/toolkit/diff.py
skshetry, line 1: Could you please explain the difference between a `toolkit` and a `lib`?

src/datachain/lib/diff.py
```diff
@@ -16,6 +17,21 @@
 C = Column
 
 
+def get_status_col_name() -> str:
```
skshetry, line 20: Does the column name need to be random? Can we have a default column name that can be changed by users?

Eg:
```
def compare(col="status"):
   pass


dc.compare(col=...)
```

src/datachain/lib/diff.py
```diff
@@ -72,13 +88,10 @@ def _to_list(obj: Union[str, Sequence[str]]) -> list[str]:
...
-    # we still need status column for internal implementation even if not
-    # needed in output
     need_status_col = bool(status_col)
-    status_col = status_col or "diff_" + "".join(
-        random.choice(string.ascii_letters)  # noqa: S311
-        for _ in range(10)
-    )
+    # we still need status column for internal implementation even if not
+    # needed in the output
+    status_col = status_col or get_status_col_name()
```
dmpetrov, line 94: Please make sure you drop this random column if it was None.

src/datachain/toolkit/diff.py
```diff
@@ -0,0 +1,111 @@
...
+            right_on=["other_id"],
+            compare=["name"],
+            added=True,
+            deleted=True,
+            modified=True,
+            unchanged=True,
+        )
+        ```
+    """
+    status_col = get_status_col_name()
```
dmpetrov, line 79: User can define it, can't they?

src/datachain/toolkit/diff.py
```diff
@@ -0,0 +1,111 @@
...
+    if deleted:
+        chains[CompareStatus.DELETED.value] = filter_by_status(CompareStatus.DELETED)
+    if modified:
+        chains[CompareStatus.MODIFIED.value] = filter_by_status(CompareStatus.MODIFIED)
+    if unchanged:
+        chains[CompareStatus.UNCHANGED.value] = filter_by_status(
+            CompareStatus.UNCHANGED
+        )
+
+    return chains
```
dmpetrov, line 111: I don't see how status column is cleaned up. or I'm missing something.

src/datachain/toolkit/diff.py
dmpetrov: The idea was to move "heavy" code from dc.py to somewhere else and keep only simple wrapper function in dc.py. 

If it's implemented in lib/diff.py - it's enough and we don't need toolkit.

dmpetrov (approved):
Some comments are inline. Please take a look.

I'm approving to not to block.

src/datachain/toolkit/diff.py
dmpetrov: Yes, that looks like the best option.

@ilongin what do you think?

src/datachain/toolkit/diff.py
ilongin: `DataChain.compare()`  returns new chain with that column which can be filtered and this new toolkit method does exactly what you described, so it saves the user that one filter step. Now, we can discuss if toolkit method is even needed in the first place.

src/datachain/toolkit/diff.py
ilongin: I'm also for top-level modules. I will move this to `datachain.diff`

src/datachain/lib/diff.py
ilongin: Status column is dropped with `select_except(...)` before returning to the user.

src/datachain/toolkit/diff.py
ilongin: In this toolkit method status column is not returned to the user so it's created only for our internal implementation and removed before returning to the user. User can define status column in core `DataChain.compare()` which returns one chain with all statuses written in that status column

src/datachain/toolkit/diff.py
ilongin: It's cleaned in `filter_by_status()` method with `select_except()`

src/datachain/toolkit/diff.py
ilongin: I think with this new toolkit we maybe have too many functions, although 2) was meant to be "private"

dmpetrov: It looks like the 3rd should have a different name and should follow similar pattern as the compare() - it should be in dc.py if there is not much code or in lib.diff otherwise with a wrapper in dc.py

dmpetrov: Anyway, lib.diff is a better place for the code than a new toolkit.

PS: I might be the person who proposed the toolkit file but it does not seem a good idea in this case 😅

skshetry (approved):

src/datachain/toolkit/diff.py
ilongin: I think 3) should not be in `dc.py` as it returns multiple instances of `DataChain` so it should be in util file.
Also, the question is should we use `src.datachain.diff.py` or `src.datachain.lib.diff.py` for public util functions?

Issue #725, opened 2024-12-20
`datachain.lib.dc.DatasetPrepareError` when saving, loading, and saving parquet datasets using `to_parquet` and `from_parquet`


AlamHasabie:
## Background
I'm following this tutorial [here](https://github.com/shcheklein/example-datachain-dvc) to produce the `.parquet` representation of a dataset. While I'm able to generate the `.parquet` file and deserialize it, there's a problem when trying to save and load repeatedly in certain orders.

## Setup
1. Create a folder containing some random files
2. Use `from_storage` method and export the dataset using `to_json`. Call this script `save_parquet.py`
```python
from datachain import DataChain
dataset_chain = DataChain.from_storage(
    LOCAL_DATASET_URI,
    update=True
)

dataset_chain.to_parquet(PARQUET_PATH)
```

3. Now create a script to load the parquet. Let's call this `load_parquet.py`
```python
dataset = DataChain.from_parquet(PARQUET_PATH)
```

4. In order, call `save_parquet.py`, `load_parquet.py`, `save_parquet.py` and then `load_parquet.py`

## Expected Behavior
The second call to `load_parquet.py` converts the `.parquet` files correctly.

## Additional Notes
1. This error seems to happen only if the `.parquet` file is mixed up with other files, for example mixed with above sample scripts. I haven't been able to replicate the issue if I were to isolate the `.parquet` file under single directory as in the tutorial e.g.  under `data/dataset.parquet`. I decided to keep reporting this in case the behavior described in this issue is still unexpected.

## Error
```
Traceback (most recent call last):
  File "~/miniconda3/envs/datachain/lib/python3.10/site-packages/datachain/lib/dc.py", line 1877, in parse_tabular
    schema = infer_schema(self, **kwargs)
  File "~/miniconda3/envs/datachain/lib/python3.10/site-packages/datachain/lib/arrow.py", line 116, in infer_schema
    return pa.unify_schemas(schemas)
  File "pyarrow/types.pxi", line 3648, in pyarrow.lib.unify_schemas
  File "pyarrow/error.pxi", line 155, in pyarrow.lib.pyarrow_internal_check_status
  File "pyarrow/error.pxi", line 92, in pyarrow.lib.check_status
pyarrow.lib.ArrowInvalid: Must provide at least one schema to unify.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "~/Documents/datachain/test_load_parquet.py", line 5, in <module>
[...]

PR #726, opened 2024-12-20
fix(listing): pick exact or create new one on update

shcheklein:
Partially addresses and stabilizes https://github.com/iterative/datachain/issues/725

Fixes a few situations in the way we find and suggest an existing cached listing when we do an `update=True`.

## 1. There was a bug in the origin code:

```python
# choosing the smallest possible one to minimize update time
listing = sorted(listings, key=lambda ls: len(ls.name))[0]
```

This code actually was picking the "biggest" one instead. There should have been `reverse=True`.

So, in a situation like:

`file:///whatever/dir1` exists and cached
`file:///whatever/` exists and cached

and we were trying to update the `file:///whatever/dir1` - it was updating the `file:///whatever/` instead.

## 2. Then, in a situation like:

`file:///whatever/` exists and cached
`file:///whatever/dir1` doesn't yet exist

and we do "update=True" for the `file:///whatever/dir1`

it was updating the whole `file:///whatever/` which can be arbitrary large. And we don't want to do this.

## So, bottom line is:

 Generalizing a bit code for both of those cases led to the code in this PR.

### DIFF
--- a/src/datachain/lib/dc.py
+++ b/src/datachain/lib/dc.py
@@ -410,8 +410,9 @@ class DataChain:
         """Returns correct listing dataset name that must be used for saving listing
         operation. It takes into account existing listings and reusability of those.
         It also returns boolean saying if returned dataset name is reused / already
-        exists or not, and it returns correct listing path that should be used to find
-        rows based on uri.
+        exists or not (on update it always returns False - just because there was no
+        reason to complicate it so far). And it returns correct listing path that should
+        be used to find rows based on uri.
         """
         catalog = session.catalog
         cache = catalog.cache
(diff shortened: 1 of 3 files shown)


tests/func/test_catalog.py
```diff
@@ -840,13 +840,17 @@ def test_listing_stats(cloud_test_catalog):
 
     catalog.enlist_source(f"{src_uri}/dogs/", update=True)
     stats = listing_stats(src_uri, catalog)
+    assert stats.num_objects == 7
```
shcheklein, line 843: [C]: not sure why we expected here any change to the `src_uri` listing, after listing the subset. It seems it is better to create a second listing for `/dogs` in this case (or at least update the whole `src_uri`).

shcheklein: cc @ilongin you might know this better

src/datachain/lib/dc.py
```diff
@@ -430,7 +430,11 @@ def parse_uri(
         if listings:
             if update:
                 # choosing the smallest possible one to minimize update time
-                listing = sorted(listings, key=lambda ls: len(ls.name))[0]
+                listing = sorted(listings, key=lambda ls: len(ls.name), reverse=True)[0]
```
shcheklein, line 433: @ilongin why `parse_uri` is exposed externally? also it's a bit of a strange name tbh. Could you please give more context?

ilongin: Yes, this should not be exposed. Probably this can be removed from `DataChain` completely in the first place ...

shcheklein: Kk, I'll move it

tests/func/test_catalog.py
ilongin: I think you are right and this was a mistake / bug in the listing. Creating new separate `/dogs` seems like correct thing to do here.

src/datachain/lib/dc.py
```diff
@@ -427,17 +427,18 @@ def parse_uri(
...
+        # if no need to update - choosing the most recent one;
+        # otherwise, we'll using the exact original `ds_name`` in this case:
+        # - if a "bigger" listing exists, we don't want to update it, it's better
+        #   to create a new "smaller" one on "update=True"
+        # - if an exact listing exists it will have the same name as `ds_name`
+        #   anyway below
+        if listings and not update:
+            listing = sorted(listings, key=lambda ls: ls.created_at)[-1]
+
+        # For local file system we need to fix listing path / prefix
```
shcheklein, line 439: @ilongin Q: why do we need to do this only for the local file system?

tests/unit/test_listing.py
```diff
@@ -38,20 +39,43 @@ def _tree_to_entries(tree: dict, path=""):
 @pytest.fixture
 def listing(test_session):
     catalog = test_session.catalog
-    dataset_name, _, _, _ = DataChain.parse_uri("s3://whatever", test_session)
+    dataset_name, _, _, _ = DataChain.parse_uri("file:///whatever", test_session)
```
shcheklein, line 42: [C]: minor semi-related change - there is not reason to use `s3`, tests were failing since it was hitting actual token update w/o proper fixtures. cc @ilongin

dreadatour (approved):
👍

amritghimire (approved):

PR #727, opened 2024-12-22
cleanup a few from_json methods from DC

shcheklein:
I don't think methods like `print_json_schema` belong to the main DC API.

It drops methods:

`print_json_schema` (a helper `gen_datamodel_code` can be used)
`print_jsonl_schema` (`format` can be specified `jsonl` in the `gen_datamodel_code`)
`from_jsonl` (`format` can be specified to `jsonl`)

It drops args:

`print_schema` (it is weird to have top level API just to toggle a single print. logger or something else should be used)

It renames args:

`meta_type` -> `format` (similar to some other places).

There will be more changes to all this JSON logic.

## TODO:

Tutorial to be fixed:

https://colab.research.google.com/github/iterative/datachain-examples/blob/main/formats/json-metadata-tutorial.ipynb

### DIFF
--- a/examples/get_started/json-csv-reader.py
+++ b/examples/get_started/json-csv-reader.py
@@ -4,6 +4,7 @@ from pydantic import BaseModel
 
 from datachain import C, DataChain
 from datachain.lib.data_model import ModelStore
+from datachain.lib.meta_formats import gen_datamodel_code
 
 
 # Sample model for static JSON model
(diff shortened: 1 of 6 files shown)


examples/get_started/json-csv-reader.py
```diff
@@ -53,18 +54,15 @@ def main():
...
-    print("Reading JSON schema from main COCO annotation")
+    print("Print JSON schema in Pydantic format from main COCO annotation")
     print("========================================================================")
-    chain = (
-        DataChain.from_storage(uri)
-        .filter(C("file.path").glob("*.json"))
-        .print_json_schema(jmespath="@", model_name="Coco")
-    )
-    chain.save()
+    chain = DataChain.from_storage(uri).filter(C("file.path").glob("*.json"))
```
shcheklein, line 59: [C]: main change in this file

examples/get_started/json-csv-reader.py
```diff
@@ -45,54 +46,55 @@ def main():
...
+        uri, schema_from=schema_uri, jmespath="@", model_name="OpenImage", anon="True"
     )
     json_pairs_ds.show()
 
     uri = "gs://datachain-demo/coco2017/annotations_captions/"
 
     print()
     print("========================================================================")
-    print("Reading JSON schema from main COCO annotation")
+    print("Print JSON schema in Pydantic format from main COCO annotation")
```
mattseddon, line 57: [Q] Do we need these print statements? Can they just be comments?

mattseddon (approved):

examples/get_started/json-csv-reader.py
shcheklein: tbh, not that important at this point .. i'm fine with either

src/datachain/lib/meta_formats.py
```diff
@@ -39,37 +39,37 @@ def process_json(data_string, jmespath):
 
 
 # Print a dynamic datamodel-codegen output from JSON or CSV on stdout
```
mattseddon, line 41: I think this is out of date

examples/get_started/json-csv-reader.py
mattseddon: My gripe is that this is the only example with this style of `print`s in it

Issue #724, opened 2024-12-20
`TypeError: Cannot recognize type <class 'NoneType'>` error when calling `Datachain.from_json()` for a file produced by `from_storage().to_json()` chain


AlamHasabie:
## Background
I'm following this tutorial [here]( https://github.com/shcheklein/example-datachain-dvc) to produce the JSON representation of a dataset. While I'm able to generate the JSON file, I'm unable to consume it back as a dataset using `from_json` method 

## Setup
1. Create a folder containing some random files
2. Use `from_storage` method and export the dataset using `to_json`.
```python
from datachain import DataChain
dataset_chain = DataChain.from_storage(
    LOCAL_DATASET_URI,
    update=True
)
dataset_chain.to_json(JSON_PATH)
```

3. Now load the json file using `from_json`, where the error is triggered
```python
dataset = DataChain.from_json(JSON_PATH, type="text")
```

## Expected Behavior
I expect that the the json file is correctly deserialized into a DataChain instance and I can use the chain functions on it, e.g. `export_files()` or `filter()` 

## Error
```
Traceback (most recent call last):
  File "~/Documents/datachain/test_load_json.py", line 5, in <module>
    dataset = DataChain.from_json(json_path, type="text")
  File "~/miniconda3/envs/datachain/lib/python3.10/site-packages/datachain/lib/dc.py", line 610, in from_json
    return chain.gen(**signal_dict)  # type: ignore[misc, arg-type]
  File ~"/miniconda3/envs/datachain/lib/python3.10/site-packages/datachain/lib/dc.py", line 960, in gen
    udf_obj.to_udf_wrapper(),
  File "~/miniconda3/envs/datachain/lib/python3.10/site-packages/datachain/lib/udf.py", line 218, in to_udf_wrapper
    self.output.to_udf_spec(),
  File "~/miniconda3/envs/datachain/lib/python3.10/site-packages/datachain/lib/signal_schema.py", line 335, in to_udf_spec
    res[db_name] = python_to_sql(type_)
  File "~miniconda3/envs/datachain/lib/python3.10/site-packages/datachain/lib/convert/python_to_sql.py", line 82, in python_to_sql
    raise TypeError(f"Cannot recognize type {typ}")
TypeError: Cannot recognize type <class 'NoneType'>
```

Here's the JSON output (path is partially redacted) : 
```json
[
[...]

PR #728, opened 2024-12-22
fix(session): keep cached successful listings

shcheklein:
As found during debugging this use case https://github.com/iterative/datachain/issues/724 - it's better to keep already cached listing, otherwise some small operations become slow to iterate until you fully sure that script is running, and the script on every run does the same indexing, every single time. There is no reason for that to my mind.

## TODO:

- [x] See if we need tests


dreadatour (approved):
Sounds reasonable 👍

amritghimire (approved):

Issue #661, opened 2024-12-04
Cache for HF file system is broken


shcheklein:
![Image](https://github.com/user-attachments/assets/a1084fa0-fd06-4c0f-bf6e-46705fed7587)

Query to reproduce:

```python
from datachain import DataChain, C

from datachain.lib.tar import process_tar

(
  DataChain.from_storage("hf://datasets/mozilla-foundation/common_voice_17_0")
    .settings(parallel=4, cache=True, prefetch=4)
  	.filter(C("file.path").glob("*/en/train/*.tar"))
  	.gen(file=process_tar)
    .save("common-voice-index")
)
```

shcheklein:
@skshetry can this be closed?

skshetry:
Cache is still broken. The above script might work because #730 stopped prefetching for HF.

shcheklein:
@skshetry thanks! does require an upstream fix? could you please summarize (or copy from the previous tickets / discussion).

skshetry:
The base `Client` implementation expects an async filesystem. PTAL at `scandir` and other async methods in `Client`.

HF filesystem does not offer an asynchronous API, so there are two issues in our implementation:

1. some fsspec methods that `Client` expects does not exist (eg: `_info`, etc). 
2. We have some async APIs implemented in HFClient that calls synchronous function.
 
   https://github.com/iterative/datachain/blob/376717347b57a72664397367c4f065be674fc3a0/src/datachain/client/hf.py#L34-L35
  
    But this is a bad thing to do, as it blocks the event loop. (Similar stuff is done in `FileClient`, and I am surprised that this thing even passed reviews).

Luckily for caching, it seems we are using `client.download` and `client.put_in_cache` which are both synchronous function, so we could reimplement them in `HFClient`. And that would fix usages from `File` API.

For usages in `AsyncMapper` and other places where we need to concurrently perform operations, unfortunately, there are no good solutions. Either we'll be blocking event-loop by running a synchronous function, or we'll need to run the operation in another thread.

`fsspec` can do the latter automatically, it can generate `AsyncFileSystem` from a synchronous filesystem, where the operations are run in another thread.

Issue #647, opened 2024-12-01
pre-fetch should work without caching as well


dmpetrov:
pre-fetch and caching should be independent settings.

See discussion in #635

skshetry:
I have a PR in #730 to use a separate directory for caching (by default it uses `.datachain/tmp` directory that we already have). That directory gets cleaned up after the prefetching is completed.

However, for `pytorch`, auto-cleanup looks tricky. We cannot do so with `PytorchDataset` instance, so either we'd need to provide a custom `DataLoader`, or provide an API to cleanup the dataset. 

I went with latter in #730 that provides `PytorchDataset.close()` API to cleanup cache.
This will also avoid synchronization issues that may arise due to `DataLoader` creating multiple processes.

Example usage: 
```python
dataset = ds.to_pytorch(...)
with closing(dataset):
  ...
```

skshetry:
Closed by #730.

PR #730, opened 2024-12-23
prefetch: use a separate temporary cache for prefetching

skshetry:
This PR will use a separate temporary cache for prefetching that resides in `.datachain/tmp/prefetch-<random>` directory when `prefetch=` is set but `cache` is not.
The temporary directory will be automatically deleted after the prefetching is done.

For `cache=True`, the cache will be reused and won't be deleted.

Please note that auto-cleanup does not work for PyTorch datasets because there is no way to invoke cleanup from the `Dataset` side. The `DataLoader` may still have cached data or rows even after the `Dataset` instance has finished iterating. As a result, values associated with a `catalog`/`cache` instance can outlive the `Dataset` instance. 

One potential solution is to implement a custom dataloader or provide a user-facing API.
In this PR, I have implemented the latter. The `PytorchDataset` now includes a `close()` method, which can be used to clean up the temporary prefix cache.

Eg:
```python
with closing(dataset):
    pass
```


src/datachain/asyn.py
```diff
@@ -179,6 +180,7 @@ def iterate(self, timeout=None) -> Generator[ResultT, None, None]:
             self.shutdown_producer()
             if not async_run.done():
                 async_run.cancel()
+                wait([async_run])
```
skshetry, line 183: `.cancel()` does not immediately cancel the task the underlying `asyncio` task.

We could add a `.result()` to wait for the future, but that does not seem to work either for the cancelled future from `run_coroutine_threadsafe()`. See https://github.com/python/cpython/issues/105836.

So, I have added `wait(...)` as it seems to wait the cancelled future, and wait for underlying asyncio task.

Alternatively, we could add an `asyncio.Event` and wait for it.

src/datachain/asyn.py
```diff
@@ -167,6 +167,7 @@ def iterate(self, timeout=None) -> Generator[ResultT, None, None]:
         init = asyncio.run_coroutine_threadsafe(self.init(), self.loop)
         init.result(timeout=1)
         async_run = asyncio.run_coroutine_threadsafe(self.run(), self.loop)
+
```
skshetry, line 170:
```suggestion
```

src/datachain/lib/file.py
```diff
@@ -269,10 +269,26 @@ def ensure_cached(self) -> None:
...
+
+        catalog = catalog or self._catalog
+        download_cb = download_cb or self._download_cb
+        if catalog is None:
+            raise RuntimeError("cannot prefetch file because catalog is not setup")
+
+        client = catalog.get_client(self.source)
+        if client.protocol == HfClient.protocol:
+            self._set_stream(catalog, self._caching_enabled, download_cb=download_cb)
+            return False
```
skshetry, line 287: `prefetch` is disabled for huggingface. See #746.

src/datachain/lib/pytorch.py
```diff
@@ -104,16 +122,33 @@ def _rows_iter(self, total_rank: int, total_workers: int):
...
         yield from ds.collect()
 
-    def __iter__(self) -> Iterator[Any]:
+    def _iter_with_prefetch(self) -> Generator[tuple[Any], None, None]:
         total_rank, total_workers = self.get_rank_and_workers()
+        download_cb = CombinedDownloadCallback()
+        if os.getenv("DATACHAIN_SHOW_PREFETCH_PROGRESS"):
+            download_cb = get_download_callback(
+                f"{total_rank}/{total_workers}", position=total_rank
+            )
```
skshetry, line 131: This shows a prefetch download progressbar for each worker which will be useful for debugging.

We cannot enable this by default, as this will mess up user's progressbar due to multiprocessing.

src/datachain/progress.py
```diff
@@ -132,8 +133,24 @@ def format_dict(self):
...
     def set_size(self, size):
         # This is a no-op to prevent fsspec's .get_file() from setting the combined
         # download size to the size of the current file.
         pass
+
+    def increment_file_count(self, n: int = 1) -> None:
+        pass
+
+
+class TqdmCombinedDownloadCallback(CombinedDownloadCallback, TqdmCallback):
```
skshetry, line 146: I have modified the callback to also show file counts on prefetching.
This will not show up on pytorch however.

Eg:
```
Download: 1.03MB [00:01, 605kB/s, 50 files]
```

src/datachain/lib/prefetcher.py
```diff
@@ -0,0 +1,65 @@
...
+    cache: Optional["Cache"] = None,
+    download_cb: Callback = DEFAULT_CALLBACK,
+) -> Generator[T, None, None]:
+    cache_ctx: AbstractContextManager[Cache]
+    if cache:
+        cache_ctx = nullcontext(cache)
+    else:
+        tmp_dir = catalog.cache.tmp_dir
+        assert tmp_dir
+        cache_ctx = temporary_cache(tmp_dir, prefix="prefetch-")
```
skshetry, line 59: We'll create a temporary cache if the `cache` is not passed or set to `None`. (The caller will only pass `cache` object if `cache=True` is set in settings).

In that case, `rows_prefetcher` is responsible for removing it at the end, which it does below.

To use a temporary cache, we have to clone a catalog and update `cache` to the temporary cache object that we just created. Not only that, we need to update `File` objects to use the new `catalog` instead of the old one.

Admittedly, the approach in this file is a bit messy.

src/datachain/lib/pytorch.py
```diff
@@ -75,6 +80,17 @@ def __init__(
         if (prefetch := dc_settings.prefetch) is not None:
             self.prefetch = prefetch
 
+        if self.cache or not self.prefetch:
+            self._cache = catalog.cache
+        else:
+            tmp_dir = catalog.cache.tmp_dir
+            assert tmp_dir
+            self._cache = get_temp_cache(tmp_dir, prefix="prefetch-")
```
skshetry, line 88: In `PytorchDataset`, we create a temporary cache ourselves and pass it to `rows_prefetcher` instead of letting it do that for us.

This is done because of multiprocessing - each worker would end up creating a separate cache directory.

The downside of this approach is that we'll have to remove the temporary cache ourselves. But we cannot do it at the end of the iteration because `DataLoader` might be keeping a buffer of rows.
So this has to be done manually by users. For this, I have added a `close()` API below.

src/datachain/asyn.py
```diff
@@ -64,11 +66,13 @@ def start_task(self, coro: Coroutine) -> asyncio.Task:
         return task
 
     def _produce(self) -> None:
-        for item in self.iterable:
-            if self._shutdown_producer.is_set():
-                return
-            fut = asyncio.run_coroutine_threadsafe(self.work_queue.put(item), self.loop)
-            fut.result()  # wait until the item is in the queue
+        with safe_closing(self.iterable):
```
skshetry, line 69: If the `self.iterable` is not exhausted, the finalizer for `self.iterable` will run during garbage collection, which could happen in another thread.

So, `AsyncMapper` should completely manage the iterable and close when it's done.

[PEP 533](https://peps.python.org/pep-0533/) would be a proper fix for this.

We need this especially for sqlite3, because we may execute query on one thread, and try to close on another thread. `sqlite3` does not allow sharing connection between threads, so it fails with the following message:
```pytb
sqlite3.ProgrammingError: SQLite objects created in a thread can only be used in that same thread. The object was created in thread id 123145580888064 and this is thread id 140704648571712.
Exception ignored in: <generator object AbstractWarehouse.dataset_select_paginated at 0x13151d440>
  File "/Users/saugat/Projects/iterative/datachain/src/datachain/data_storage/warehouse.py", line 223, in dataset_select_paginated
    with self.clone() as wh:
         ^^^^^^^^^^^^
  File "/Users/saugat/Projects/iterative/datachain/src/datachain/data_storage/sqlite.py", line 426, in __exit__
    self.close()
  File "/Users/saugat/Projects/iterative/datachain/src/datachain/data_storage/warehouse.py", line 168, in close
    self.db.close()
  File "/Users/saugat/Projects/iterative/datachain/src/datachain/data_storage/sqlite.py", line 242, in close
    self.db.close()
sqlite3.ProgrammingError: SQLite objects created in a thread can only be used in that same thread. The object was created in thread id 123145580888064 and this is thread id 140704648571712.
```


Alternatively, we could make the `dataset_select_paginated` threadsafe by using thread-local connection, but that takes effort.

src/datachain/lib/file.py
```diff
@@ -269,10 +269,19 @@ def ensure_cached(self) -> None:
...
-            await client._download(self, callback=self._download_cb)
+    async def _prefetch(self, download_cb: Optional["Callback"] = None) -> bool:
+        from datachain.client.hf import HfClient
+
+        if self._catalog is None:
+            raise RuntimeError("cannot prefetch file because catalog is not setup")
+
+        client = self._catalog.get_client(self.source)
+        if client.protocol == HfClient.protocol:
+            return False
```
skshetry, line 280: See:
- https://github.com/iterative/datachain/issues/661

src/datachain/lib/udf.py
```diff
@@ -300,24 +337,29 @@ def run(
...
-                output = [{"sys__id": id_} | dict(zip(self.signal_names, udf_output))]
-                processed_cb.relative_update(1)
-                yield output
+
+        def _prepare_rows(udf_inputs) -> "abc.Generator[Sequence[Any], None, None]":
+            with safe_closing(udf_inputs):
+                for row in udf_inputs:
+                    yield self._prepare_row_and_id(
+                        row, udf_fields, catalog, cache, download_cb
+                    )
```
skshetry, line 347: Using `safe_closing` here and below so that the `udf_inputs` gets properly closed when this `_prepare_rows` gets `.close()`d.

We were using generator expression before:
```python
     prepared_inputs: abc.Generator[Sequence[Any], None, None] = (
            self._prepare_row_and_id(row, udf_fields, cache, download_cb)
            for row in udf_inputs
        )
```

Generator expression/function does not send `close` or `throw` to the subgenerator (`udf_inputs` in this case), so we have to add this contextmanager and change it to a generator function from a generator expression.

Note that this function runs in another thread in case of prefetching. So, we want `.close()` for `udf_inputs` to run in the same thread, not in a different thread when it gets garbage-collected.

src/datachain/asyn.py
```diff
@@ -64,11 +67,16 @@ def start_task(self, coro: Coroutine) -> asyncio.Task:
...
+        try:
+            with safe_closing(self.iterable):
+                for item in self.iterable:
+                    if self._shutdown_producer.is_set():
+                        return
+                    coro = self.work_queue.put(item)
+                    fut = asyncio.run_coroutine_threadsafe(coro, self.loop)
+                    fut.result()  # wait until the item is in the queue
+        finally:
+            self._producer_is_shutdown.set()
```
skshetry, line 79: This `_produce()` is run via `loop.run_in_executor`. Since this is running in a different thread, and is a synchronous function, `.cancel()` does not really cancel the task/function, and it'll keep running in the background.

We do ask it to shut down, by setting `self._shutdown_producer` event, but this function is running in the background.
We want to wait for the `self.iterable` to be closed, but that function may not run immediately when we `.close()` the `mapper.iterate()` iterator.

Eg:
```python
mapper = AsyncMapper(func, iterable) 
with closing(mapper.iterate()) as it:
    next(it)
```

In this case, the `iterable.close()` may be called much later than `mapper.iterate.close()`. We want to ensure the iterable gets closed when `mapper.iterate()` gets closed.

So, for this, we are setting another event that gets set after `iterable.close()` gets called.

dreadatour (approved):
Looks good to me overall, great job! 🥰
Not sure if I was able to go through every single line though, it is quite a big change.

src/datachain/lib/file.py
```diff
@@ -269,10 +269,19 @@ def ensure_cached(self) -> None:
...
+
+        await client._download(self, callback=download_cb or self._download_cb)
+        self._download_cb = DEFAULT_CALLBACK
```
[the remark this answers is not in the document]
skshetry: The file gets pre-downloaded, so we don't need to use the callback later.

src/datachain/lib/pytorch.py
```diff
@@ -75,6 +80,18 @@ def __init__(
         if (prefetch := dc_settings.prefetch) is not None:
             self.prefetch = prefetch
 
+        if self.cache or not self.prefetch:
+            self._cache = catalog.cache
```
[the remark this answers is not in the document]
skshetry: It's the `cache` that we are going to use the `PytorchDataset`. (We use a temporary cache if `prefetch>0` and `cache=False`, otherwise we use the default cache).

So, I am not sure if this needs to be explained.

shcheklein:
@skshetry are we creating a single prefetch cache for the whole session or per UDF? E.g. if I'm processing a very large dataset at once (doesn't fit my machine) - how will it work?

skshetry:
> E.g. if I'm processing a very large dataset at once (doesn't fit my machine) - how will it work?

We are creating cache per-udf. For the whole session, the default cache can be repurposed by using `cache=True, prefetch=N`.

> E.g. if I'm processing a very large dataset at once (doesn't fit my machine) - how will it work?

prefetching doesn't really help with this at the moment. We don't have cache pruning _yet_.

shcheklein:
> prefetching doesn't really help with this at the moment. We don't have cache pruning yet.

hmm, how is this implementation different from what we have before? correct me if I'm wrong, but I think we wanted to decouple cache and prefetch specifically because we didn't want to keep the full copy of data or are there any other use cases for this?

skshetry:
> > prefetching doesn't really help with this at the moment. We don't have cache pruning yet.
> 
> hmm, how is this implementation different from what we have before? correct me if I'm wrong, but I think we wanted to decouple cache and prefetch specifically because we didn't want to keep the full copy of data or are there any other use cases for this?

Before this PR, prefetching only worked when `cache=True` was set, and used the default cache directory. So it was not the default.

This PR is a fix for #647, which enables prefetching by default, and uses a temporary cache directory, i.e. decouples cache and prefetch settings.

shcheklein:
But essentially it is still using cache underneath, right? Again, I'm thinking more in terms of the use case, not the API. My concern is that we are still caching the whole dataset pretty much.

skshetry:
> But essentially it is still using cache underneath, right? Again, I'm thinking more in terms of the use case, not the API. My concern is that we are still caching the whole dataset pretty much.

Yes, it is still using cache underneath, although I think of that as an implementation detail.

I understand your usecase, and we've discussed this in relation to other training dataloader tools that have LRU cache pruning. IIRC, we did not decide on doing cache pruning.

See https://github.com/iterative/datachain/issues/635#issuecomment-2510213124, where it was discussed to keep objects' for the lifetime of the udf. But #647 is more of a motivation for this PR.

Technical implementation wise, cache pruning could be considered as a next iteration of this - We can still use cache for prefetching, we only need to figure out how to preserve `last-used` information for pruning.
(or, whatever alternative strategy we end up using).

shcheklein:
yep, my only question here is could have we used just the same cache (decouple settings, but keep using the same cache location). It would be simpler, right? and would work the same in those scenarios, unless I'm missing something. We could have asked for now people delete .datachain directory after the run or put cache on some ephemeral cluster storage ...


let me rephrase this - it seems complicated solution (at least PR makes look like this), while we don't solve fully the issue. I wonder if could have solved most of the stuff that this PR solves faster ...

skshetry:
> yep, my only question here is could have we used just the same cache (decouple settings, but keep using the same cache location). It would be simpler, right?

Yes, that was my first question too in https://github.com/iterative/datachain/issues/635#issuecomment-2505996432. But 

> 
> let me rephrase this - it seems complicated solution (at least PR makes look like this), while we don't solve fully the issue. I wonder if could have solved most of the stuff that this PR solves faster ...

Most of the changes in this PR are for making things _correct_, and cleaning things properly.
I know it looks complicated, but it's stuff that we need regardless of whether we implement a separate prefetching cache or not. I needed all this to handle lifecycle of the cache properly.
Also, ran into lots of issues with asyncio where coroutines would not get cancelled and just continue running in the background. :(

But I agree that it took a lot of time.

dmpetrov:
@skshetry it looks like most of the issues are related to pytorch when we have limited control.

How about regular UDFs? If user runs a regular UDF with pre-fetch and without caching, what prevents us from caching and deleting data properly, right after a files was processed?

PR #732, opened 2024-12-23
fix(dc): move parse uri and rename out of main class

shcheklein:
Moves `parse_uri` out of public DC methods + renames it a bit to make move obvious hopefully.

### DIFF
--- a/src/datachain/catalog/catalog.py
+++ b/src/datachain/catalog/catalog.py
@@ -52,6 +52,7 @@ from datachain.error import (
     QueryScriptCancelError,
     QueryScriptRunError,
 )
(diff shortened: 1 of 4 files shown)


0x2b3bfa0 (approved):
Fine by me

src/datachain/catalog/catalog.py
```diff
@@ -697,11 +698,9 @@ def _row_to_node(d: dict[str, Any]) -> Node:
                 )
                 indexed_sources = []
                 for source in dataset_sources:
-                    from datachain.lib.dc import DataChain
```
dreadatour, line 700: 👍

PR #734, opened 2024-12-23
fix(from_storage): no listing / cache on a single file

shcheklein:
Adjusts `from_storage` to avoid creating a listing / or using a listing cache when we need a single file. In this case we are creating a single entry DC right away.

This speedups a lot simple cases when you need a single file from a repo and it stars indexing all files around. It won't do this again.

Also, it won't be failing by hitting the previously cached index when we just write and then read a single file.

### DIFF
--- a/src/datachain/lib/dc.py
+++ b/src/datachain/lib/dc.py
@@ -32,7 +32,7 @@ from datachain.lib.data_model import DataModel, DataType, DataValue, dict_to_dat
 from datachain.lib.dataset_info import DatasetInfo
 from datachain.lib.file import ArrowRow, File, FileType, get_file_type
 from datachain.lib.file import ExportPlacement as FileExportPlacement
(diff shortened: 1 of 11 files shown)


dreadatour (commented):
Fix looks good to me, but some tests were failing now :(

src/datachain/lib/dc.py
```diff
@@ -438,6 +438,18 @@ def from_storage(
             uri, session, update=update
         )
 
+        # ds_name is None if object is a file, we don't want to use cache
+        # or do listing in that case - just read that single object
+        if not list_ds_name:
```
shcheklein, line 443: [C]: main change here (along with `get_listing` that now returns empty ds_name in case of a single file). Everything else is pretty much fixing things to work with this change.

dreadatour (approved):
👍

PR #735, opened 2024-12-23
Fix Mistral example in quick-start

dmpetrov:
1. New mistral lib does not pick up env variable by default - fix.
2. With new async enabled it hits to mistral limit very often - remove parallel and cache. It also simplifies the code!


shcheklein (approved):
Are there examples in the `examples/` dir that use Mistral?

dmpetrov:
@shcheklein very likely - I'll double check

dmpetrov:
fyi: I was not able to find Mistral examples in examples/ 
nothing to fix but we can consider adding it.

PR #745, opened 2024-12-24
from_parquet: use virtual filesystem to preserve partition information when using cache

skshetry:
Alternative proposal to #744, that uses a `ReferenceFileSystem` (which acts as a pointer) to preserve partitioned information.

### DIFF
--- a/src/datachain/lib/arrow.py
+++ b/src/datachain/lib/arrow.py
@@ -1,9 +1,11 @@
 from collections.abc import Sequence
-from tempfile import NamedTemporaryFile
+from itertools import islice
 from typing import TYPE_CHECKING, Any, Optional
 
+import fsspec.implementations.reference
 import orjson
 import pyarrow as pa
+from fsspec.core import split_protocol
 from pyarrow.dataset import CsvFileFormat, dataset
 from tqdm import tqdm
 
@@ -25,7 +27,18 @@ if TYPE_CHECKING:
 DATACHAIN_SIGNAL_SCHEMA_PARQUET_KEY = b"DataChain SignalSchema"
 
 
+class ReferenceFileSystem(fsspec.implementations.reference.ReferenceFileSystem):
+    def _open(self, path, mode="rb", *args, **kwargs):
+        # overriding because `fsspec`'s `ReferenceFileSystem._open`
+        # reads the whole file in-memory.
+        (uri,) = self.references[path]
+        protocol, _ = split_protocol(uri)
+        return self.fss[protocol]._open(uri, mode, *args, **kwargs)
+
+
 class ArrowGenerator(Generator):
+    DEFAULT_BATCH_SIZE = 2**17  # same as `pyarrow._dataset._DEFAULT_BATCH_SIZE`
+
     def __init__(
         self,
         input_schema: Optional["pa.Schema"] = None,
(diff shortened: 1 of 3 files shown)


shcheklein:
@skshetry so, which PR is better to review? :)

skshetry:
> so, which PR is better to review? :)

I have closed the other PR as I think using a virtual fs is a much cleaner and less fragile solution.

src/datachain/lib/arrow.py
```diff
@@ -190,18 +225,6 @@ def arrow_type_mapper(col_type: pa.DataType, column: str = "") -> type:  # noqa:
     raise TypeError(f"{col_type!r} datatypes not supported, column: {column}")
 
 
-def _nrows_file(file: File, nrows: int) -> str:
```
skshetry, line 193: I have removed this hack. If `nrows` is a significant number, this is likely to be slower. 

We now iterate up to `self.nrows` with the pyarrow dataset.
I tried running `json-csv-reader` example, and the new method is slightly faster.

src/datachain/lib/dc.py
```diff
@@ -557,7 +557,9 @@ def jmespath_to_name(s: str):
                 nrows=nrows,
             )
         }
-        return chain.gen(**signal_dict)  # type: ignore[misc, arg-type]
+        # disable prefetch if nrows is set
+        settings = {"prefetch": 0} if nrows else {}
```
skshetry, line 561: disabled `prefetch` when `nrows` is set.

src/datachain/lib/dc.py
```diff
@@ -1976,8 +1981,6 @@ def from_csv(
             else:
                 msg = f"error parsing csv - incompatible output type {type(output)}"
                 raise DatasetPrepareError(chain.name, msg)
-        elif nrows:
-            nrows += 1
```
skshetry, line 1980: This was a workaround to make `nrows` works if the data contains headers. This is no longer needed as we count the rows internally.

src/datachain/lib/arrow.py
```diff
@@ -25,7 +27,17 @@
 DATACHAIN_SIGNAL_SCHEMA_PARQUET_KEY = b"DataChain SignalSchema"
 
 
+class ReferenceFileSystem(fsspec.implementations.reference.ReferenceFileSystem):
+    def _open(self, path, mode="rb", *args, **kwargs):
+        # `fsspec`'s `ReferenceFileSystem._open` reads the whole file in-memory.
```
shcheklein, line 32: @skshetry what are the practical implications of this? does it mean we can't handle large files anymore?

src/datachain/lib/arrow.py
```diff
@@ -55,57 +67,80 @@ def __init__(
     def process(self, file: File):
```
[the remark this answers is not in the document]
shcheklein: thanks!

src/datachain/lib/dc.py
shcheklein: should we just let users disable it though?

src/datachain/lib/arrow.py
skshetry: The comment is explaining the rationale why I am implementing `_open` method.
The `fsspec.ReferenceFileSystem._open` reads the whole file in memory and gives you a `io.BytesIO()` object. I have _fixed_ it to return a file object. 

So this should now be able to handle large files.

dreadatour (approved):
Looks good to me overall 👍 Went through all the changes carefully.
Same time I am not really familiar yet with this part of the codebase 😢

shcheklein (approved):

src/datachain/lib/dc.py
skshetry: The fact that we are doing pre-downloading/pre-fetching is not going to be obvious to the users. So, finding the right settings to disable/enable them is going to be difficult.

Similarly, the default settings with prefetch is going to be suboptimal when you only want to read a few rows. In fact, I'd have to adjust example tests to disable `prefetch` as well, as we use `laion` dataset to read just a couple of rows.

That said, I do not have any strong opinion, though — please let me know what you think.

PR #748, opened 2024-12-25
Return url for file in GCP public bucket instead of error

dreadatour:
When retrieving URL for file in GCP bucket if this bucket is public and we have no GCP credentials set, error will be raised.

In this PR we are returning public download URL in this case instead of exception.

Note: I think we need to do the same for other types of storages.

### DIFF
--- a/src/datachain/client/gcs.py
+++ b/src/datachain/client/gcs.py
@@ -32,6 +32,14 @@ class GCSClient(Client):
 
         return cast(GCSFileSystem, super().create_fs(**kwargs))
 
+    def url(self, path: str, expires: int = 3600, **kwargs) -> str:
+        try:
+            return self.fs.sign(self.get_full_path(path), expiration=expires, **kwargs)
+        except AttributeError as exc:
+            if "you need a private key to sign credentials" in str(exc):
+                return f"https://storage.googleapis.com/{self.name}/{path}"
+            raise
+
     @staticmethod
     def parse_timestamp(timestamp: str) -> datetime:
         """
--- a/src/datachain/client/gcs.py
+++ b/src/datachain/client/gcs.py
@@ -36,7 +36,8 @@ class GCSClient(Client):
         try:
             return self.fs.sign(self.get_full_path(path), expiration=expires, **kwargs)
         except AttributeError as exc:
-            if "you need a private key to sign credentials" in str(exc):
+            is_anon = self.fs.storage_options.get("token") == "anon"
+            if is_anon and "you need a private key to sign credentials" in str(exc):
                 return f"https://storage.googleapis.com/{self.name}/{path}"
             raise
 
--- a/tests/unit/test_client_gcs.py
+++ b/tests/unit/test_client_gcs.py
@@ -1,0 +1,6 @@
+from datachain.client import Client
+
+
+def test_anon_url():
+    client = Client.get_client("gs://foo", None, anon=True)
+    assert client.url("bar") == "https://storage.googleapis.com/foo/bar"
--- a/tests/unit/test_client_gcs.py
+++ b/tests/unit/test_client_gcs.py
@@ -1,6 +1,17 @@
 from datachain.client import Client
 
 
-def test_anon_url():
+def test_anon_url(mocker):
+    def sign(*args, **kwargs):
+        raise AttributeError(
+            "you need a private key to sign credentials."
+            "the credentials you are currently using"
+            " <class 'google.oauth2.credentials.Credentials'> just contains a token."
+            " see https://googleapis.dev/python/google-api-core/latest/auth.html"
+            "#setting-up-a-service-account for more details."
+        )
+
+    mocker.patch("gcsfs.GCSFileSystem.sign", side_effect=sign)
+
     client = Client.get_client("gs://foo", None, anon=True)
     assert client.url("bar") == "https://storage.googleapis.com/foo/bar"


0x2b3bfa0 (approved):

src/datachain/client/gcs.py
```diff
@@ -32,6 +32,14 @@ def create_fs(cls, **kwargs) -> GCSFileSystem:
 
         return cast(GCSFileSystem, super().create_fs(**kwargs))
 
+    def url(self, path: str, expires: int = 3600, **kwargs) -> str:
```
0x2b3bfa0, line 35: What's calling this function?

tests/unit/test_client_gcs.py
```diff
@@ -0,0 +1,6 @@
+from datachain.client import Client
+
+
+def test_anon_url():
+    client = Client.get_client("gs://foo", None, anon=True)
+    assert client.url("bar") == "https://storage.googleapis.com/foo/bar"
```
skshetry, line 6: This test is going to hit gcs. I'd suggest mocking/stubbing `fs.sign` it.

src/datachain/client/gcs.py
```diff
@@ -32,6 +32,14 @@ def create_fs(cls, **kwargs) -> GCSFileSystem:
...
+        try:
+            return self.fs.sign(self.get_full_path(path), expiration=expires, **kwargs)
+        except AttributeError as exc:
+            if "you need a private key to sign credentials" in str(exc):
```
skshetry, line 39: Is there really this message in `AttributeError`, and what raises this?

Could you please share stacktrace?

src/datachain/client/gcs.py
dreadatour: Entrypoint I came from is SaaS calling this function: https://github.com/iterative/datachain/blob/main/src/datachain/catalog/catalog.py#L1199-L1202

src/datachain/client/gcs.py
dreadatour: Sure!

```
In [1]: from datachain.catalog import get_catalog

In [2]: catalog = get_catalog()

In [3]: url = catalog.signed_url('gs://datachain-demo', 'aspset510/test/videos/1e28/1e28-0026-right.mp4', {'anon': True})
> /Users/vlad/work/iterative/datachain/src/datachain/client/gcs.py(40)url()
     39             import ipdb; ipdb.set_trace()
---> 40             print(exc)
     41

ipdb> exc
AttributeError("you need a private key to sign credentials.the credentials you are currently using <class 'google.oauth2.credentials.Credentials'> just contains a token. see https://googleapis.dev/python/google-api-core/latest/auth.html#setting-up-a-service-account for more details.")
```

src/datachain/client/gcs.py
0x2b3bfa0: Oh, I see! 🙏

src/datachain/client/gcs.py
skshetry: Where is this exception is being raised? The above stacktrace is just your `print(exc)` function.

dreadatour: Sorry, it was wrong code I was copied. Updated the stacktrace.

skshetry: Also, what is the expectation here for `url()` method? Should it return `https://storage.googleapis.com/...` url when it cannot create a signed url, or should that only be the case for public bucket?

If the latter, how can we be sure that this exception only gets raised for public buckets? (sorry too many questions).

dreadatour: These are good question, please, feel free to ask any 🤗
What if we will return public URL only if [`anon` param](https://github.com/iterative/datachain/blob/main/src/datachain/client/gcs.py#L30-L31) is set? 🤔

dreadatour: Like this, for example, should be more accurate:
```diff
diff --git a/src/datachain/client/gcs.py b/src/datachain/client/gcs.py
index 99e3c14..6eaaf1c 100644
--- a/src/datachain/client/gcs.py
+++ b/src/datachain/client/gcs.py
@@ -36,7 +36,8 @@ class GCSClient(Client):
         try:
             return self.fs.sign(self.get_full_path(path), expiration=expires, **kwargs)
         except AttributeError as exc:
-            if "you need a private key to sign credentials" in str(exc):
+            is_anon = self.fs.storage_options.get("token") == "anon"
+            if is_anon and "you need a private key to sign credentials" in str(exc):
                 return f"https://storage.googleapis.com/{self.name}/{path}"
             raise
```

dreadatour: Or may be even like this?
```python
    def url(self, path: str, expires: int = 3600, **kwargs) -> str:
        if self.fs.storage_options.get("token") == "anon":
            return f"https://storage.googleapis.com/{self.name}/{path}"
        return self.fs.sign(self.get_full_path(path), expiration=expires, **kwargs)
```

dreadatour: Choose the first option just to be on the safe side and to do not break any existing flows. Although second option looks like a better one.

tests/unit/test_client_gcs.py
dreadatour: Thank you for suggestion! Done 👍

src/datachain/client/gcs.py
```diff
@@ -32,6 +32,15 @@ def create_fs(cls, **kwargs) -> GCSFileSystem:
 
         return cast(GCSFileSystem, super().create_fs(**kwargs))
 
+    def url(self, path: str, expires: int = 3600, **kwargs) -> str:
+        try:
+            return self.fs.sign(self.get_full_path(path), expiration=expires, **kwargs)
+        except AttributeError as exc:
+            is_anon = self.fs.storage_options.get("token") == "anon"
+            if is_anon and "you need a private key to sign credentials" in str(exc):
```
shcheklein, line 40: a few questions come to my mind: :)

can we return a public link right away if anon is set to True?

is there any guarantee that HTTP link this always work for public buckets?

do we do anything like this for S3? if not, why?

when does this error happen?

dreadatour: > a few questions come to my mind: :)

These are good questions! 👍

> can we return a public link right away if anon is set to True?

Yes, we can. I doubted about chosen option [here](https://github.com/iterative/datachain/pull/748#discussion_r1897985463) and still not sure, but on the other side I just wanted to be sure I will not break any existing use case.

> is there any guarantee that HTTP link this always work for public buckets?

Yes: https://cloud.google.com/storage/docs/access-public-data#api-link

> do we do anything like this for S3? if not, why?

We definitely should do the same for other types of storages. I am not doing it here for some reasons:

- I wanted to check first if this will solve all the issues in SaaS
- it will requires more work and proper tests, I am not sure we have capacity now for this (fix for GCP was quick and simple)

> when does this error happen?

Original issue is here: https://github.com/iterative/studio/issues/11118

shcheklein: > Original issue is here: https://github.com/iterative/studio/issues/11118

yep, I understand. I guess my question is why it is happening. If anon it passed AFAIU it should not be even trying to use credentials (I assume it is failing on a cluster because some default Google cred are present, right?)

dreadatour: > yep, I understand. I guess my question is why it is happening. If anon it passed AFAIU it should not be even trying to use credentials (I assume it is failing on a cluster because some default Google cred are present, right?)

Ah, got it. It is failing not in a workers cluster, but on the backend (API request). It is interesting, I can not find any default Google credentials on backend 🤔

PR #753, opened 2024-12-27
fix(cache, client): support versioned files

shcheklein:
On top of https://github.com/iterative/datachain/pull/734

Fixes versioning. So far it seems it was broken when we were trying to download a previous version of a file from version aware fs.

This is breaking basic examples of versioning with DataChain.


dreadatour (approved):
👍

PR #754, opened 2024-12-27
Return url for file in GCP public bucket instead of error

dreadatour:
Follow-up for the https://github.com/iterative/datachain/pull/748

If no Google Cloud credentials were set we still get an error in SaaS. This fix will allow us to return public URL without trying to sign it if `anon=True` is set.

I also can confirm this will works for public AWS S3 buckets without any changes. It will return public URL right out of the box.

### DIFF
--- a/src/datachain/client/gcs.py
+++ b/src/datachain/client/gcs.py
@@ -35,19 +35,12 @@ class GCSClient(Client):
     def url(self, path: str, expires: int = 3600, **kwargs) -> str:
         """
         Generate a signed URL for the given path.
-        If the client is anonymous, a public URL is returned instead.
+        If the client is anonymous, a public URL is returned instead
+        (see https://cloud.google.com/storage/docs/access-public-data#api-link).
         """
-        fs = cast(GCSFileSystem, self.fs)
-        if fs.storage_options.get("token") == "anon":
-            return self.public_url(path)
-        return fs.sign(self.get_full_path(path), expiration=expires, **kwargs)
-
-    def public_url(self, path: str) -> str:
-        """
-        Generate a public URL for the given path.
-        See https://cloud.google.com/storage/docs/access-public-data#api-link
-        """
-        return f"https://storage.googleapis.com/{self.name}/{path}"
+        if self.fs.storage_options.get("token") == "anon":
+            return f"https://storage.googleapis.com/{self.name}/{path}"
+        return self.fs.sign(self.get_full_path(path), expiration=expires, **kwargs)
 
     @staticmethod
     def parse_timestamp(timestamp: str) -> datetime:
--- a/src/datachain/client/gcs.py
+++ b/src/datachain/client/gcs.py
@@ -33,13 +33,21 @@ class GCSClient(Client):
         return cast(GCSFileSystem, super().create_fs(**kwargs))
 
     def url(self, path: str, expires: int = 3600, **kwargs) -> str:
-        try:
-            return self.fs.sign(self.get_full_path(path), expiration=expires, **kwargs)
-        except AttributeError as exc:
-            is_anon = self.fs.storage_options.get("token") == "anon"
-            if is_anon and "you need a private key to sign credentials" in str(exc):
-                return f"https://storage.googleapis.com/{self.name}/{path}"
-            raise
+        """
+        Generate a signed URL for the given path.
+        If the client is anonymous, a public URL is returned instead.
+        """
+        fs = cast(GCSFileSystem, self.fs)
+        if fs.storage_options.get("token") == "anon":
+            return self.public_url(path)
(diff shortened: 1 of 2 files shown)


src/datachain/client/gcs.py
```diff
@@ -33,13 +33,21 @@ def create_fs(cls, **kwargs) -> GCSFileSystem:
...
+        """
+        Generate a signed URL for the given path.
+        If the client is anonymous, a public URL is returned instead.
+        """
+        fs = cast(GCSFileSystem, self.fs)
+        if fs.storage_options.get("token") == "anon":
+            return self.public_url(path)
+        return fs.sign(self.get_full_path(path), expiration=expires, **kwargs)
+
+    def public_url(self, path: str) -> str:
```
skshetry, line 45: Please, let's not extend `Client` - not every `path` has a `public_url`, and it feels like it's just a special condition that could be handled by `url()` itself.

Also, do you mind opening a bug report on `gcsfs`? Maybe they will be able to suggest a better alternative?

tests/unit/test_client_gcs.py
```diff
@@ -1,17 +1,6 @@
...
-            "you need a private key to sign credentials."
-            "the credentials you are currently using"
-            " <class 'google.oauth2.credentials.Credentials'> just contains a token."
-            " see https://googleapis.dev/python/google-api-core/latest/auth.html"
-            "#setting-up-a-service-account for more details."
-        )
-
-    mocker.patch("gcsfs.GCSFileSystem.sign", side_effect=sign)
-
+def test_anon_url():
```
skshetry, line 4: As mentioned before, this will hit google cloud.

src/datachain/client/gcs.py
dreadatour: > Please, let's not extend `Client` - not every `path` has a `public_url`, and it feels like it's just a special condition that could be handled by `url()` itself.

I am not extending `Client` here (yet), I am adding extra method to `GCSClient` for better readability.

Also the idea in general (for future) is to have `public_url` method in client so in case we are sure we have no credentials and bucket is public we can call `public_url` right away, without running all complex logic and checks inside clients (e.g. `gcsfs`).

> Also, do you mind opening a bug report on `gcsfs`? Maybe they will be able to suggest a better alternative?

I am considering to open a PR in `gcsfs`, but this will require much more work and I am still not sure, to be honest, if this is a way to go for the `gcsfs` or this is only specific for our use case, since `fs.sign` method is about signing URL, and if it will return a public URL instead of an error it may not be something user is waiting for.

In case of AWS S3 it is how `s3` library works and `fsspec` just proxying an error, in case of GCP Storage it is how gcloud library works and `fsspec` just proxying public URL. It is kind of complicated in general case and needed much more work comparing to the changes in this PR where we do know we do need public URL for sure.

dreadatour: > Please, let's not extend `Client` - not every `path` has a `public_url`, and it feels like it's just a special condition that could be handled by `url()` itself.

Updated the code to remove extra method.

tests/unit/test_client_gcs.py
dreadatour: 🤔 Not sure if I understand how it can hit google cloud.
I also have tried it with debugger and can not find a way this to hit google cloud in current implementation.
If I miss something, please, help me to understand? 🙏

src/datachain/client/gcs.py
skshetry: Thank you, excellent points. I am okay with adding public_url.

---

Btw, have you seen this: [GCSFileSystem.url](https://github.com/fsspec/gcsfs/blob/378cae6f56f3dbca1b909d58f21a8d61ed5b0a45/gcsfs/core.py#L1008)? Maybe we can use this instead of hardcoding url?

tests/unit/test_client_gcs.py
skshetry: Ah, I see — My apologies for the oversight. 😅

src/datachain/client/gcs.py
dreadatour: > Thank you, excellent points. I am okay with adding public_url.

❤️ Will be back to this in the follow-up.

> Btw, have you seen this: [GCSFileSystem.url](https://github.com/fsspec/gcsfs/blob/378cae6f56f3dbca1b909d58f21a8d61ed5b0a45/gcsfs/core.py#L1008)? Maybe we can use this instead of hardcoding url?

I saw it but I am not sure what kind of an URL is this.

I might want to use https://github.com/fsspec/gcsfs/blob/378cae6f56f3dbca1b909d58f21a8d61ed5b0a45/gcsfs/core.py#L321-L323 for the domain URL if I'll prepare a PR to the `gcsfs`.

skshetry (approved):

PR #755, opened 2024-12-28
Fix signed url versioning

shcheklein:
Fixes signed URL for versioned objects (download button is broken atm in Studio for versioned buckets, video previews are probably broken since we are also generating singed URLs for them). 

https://github.com/iterative/studio/pull/11128 - Studio part

## TODO

- [x] Add more tests of possible (versioned signed URL is not an easy one to test)
- [x] Create Studio PR
- [x] Add Azure support
- [X] Create a ticket for the followup - https://github.com/iterative/datachain/issues/761

## Followup

Public S3 and Azure `client.url()` code. Similar to GS that already has a check for `anon` in it - we need to generate and return direct URL to the cloud storage.

Make sure along the way:

- endpoint URLs are supported, especially for AWS
- on the Studio side pass `ms` header to signed URL to get a public URL that actually works (see some SO discussions)
- add tests


dreadatour (approved):
Good catch about version! 👍
Code looks good to me, even if it is still draft 👀

PR #757, opened 2024-12-28
Allow empty session name

0x2b3bfa0:
We're supposed to accept empty session names as per the documentation string and as per the surrounding logic, but the regular expression was broken.


shcheklein:
It would be nice to have a test for this. Thanks @0x2b3bfa0 !

shcheklein (approved):

PR #758, opened 2024-12-28
File upload

dreadatour:
Implementing file upload.

Usage example:

```python
import io
from typing import Iterator

import imageio
import imageio.v3 as iio
from ultralytics import YOLO

from datachain.catalog import get_catalog
from datachain.client import Client
from datachain.model.ultralytics import YoloPoses


def pose_estimation(client: Client, yolo: YOLO, file: File) -> Iterator[tuple[File, int, float, YoloPoses]]:
    stem = file.get_file_stem()
    ext = file.get_file_ext()

    reader = imageio.get_reader(io.BytesIO(file.read()), format=ext)
    fps = reader.get_meta_data()["fps"]

    for frame, img in enumerate(reader):
        filename = f"{stem}_{frame:06d}.jpg"
        img_encoded = iio.imwrite("<bytes>", img, extension=".jpeg")
        f = client.upload(filename, img_encoded)

        timestamp = frame / fps
        results = yolo(img)

        yield f, frame, timestamp, YoloPoses.from_results(results)


(
    DataChain.from_dataset("videos")
        .limit(1)
        .setup(
            client=lambda: get_catalog().get_client("gs://bucket/videos/frames"),
            yolo=lambda: YOLO("yolo11n-pose.pt"),
        )
        .gen(pose_estimation, output=("file", "frame", "timestamp", "poses"))
        .save("videos-frames-poses")
)
```

### DIFF
--- a/src/datachain/client/fsspec.py
+++ b/src/datachain/client/fsspec.py
@@ -364,6 +364,12 @@ class Client(ABC):
         assert not file.location
         return FileWrapper(self.fs.open(self.get_full_path(file.path)), cb)  # type: ignore[return-value]
 
+    def upload(self, path: str, data: bytes) -> "File":
+        full_path = self.get_full_path(path)
+        self.fs.pipe_file(full_path, data)
+        file_info = self.fs.info(full_path)
+        return self.info_to_file(file_info, path)
+
     def download(self, file: "File", *, callback: Callback = DEFAULT_CALLBACK) -> None:
         sync(get_loop(), functools.partial(self._download, file, callback=callback))
 


dreadatour:
> Is `Client` part of an API? 

Not really. I set it up via:

```
setup(
    client=lambda: get_catalog().get_client("gs://bucket/videos/frames"),
)
```

(see example in PR description).

It is an open question of simplifying API (making `upload` to be a part of `File` may be?)

src/datachain/client/fsspec.py
```diff
@@ -364,6 +364,12 @@ def open_object(
         assert not file.location
         return FileWrapper(self.fs.open(self.get_full_path(file.path)), cb)  # type: ignore[return-value]
 
+    def upload(self, path: str, data: bytes) -> "File":
+        full_path = self.get_full_path(path)
+        self.fs.pipe_file(full_path, data)
```
shcheklein, line 369: I wonder if we can make fsspec return File result object. It's very sad that we have to do the second call here to get info - it's an additional API call and slow.

src/datachain/client/fsspec.py
```diff
@@ -364,6 +364,12 @@ def open_object(
         assert not file.location
         return FileWrapper(self.fs.open(self.get_full_path(file.path)), cb)  # type: ignore[return-value]
 
+    def upload(self, path: str, data: bytes) -> "File":
```
shcheklein, line 367: why can't we add a tests here? or do you plan to rewrite it?

I know that it is a thin wrapper, but it still hits multiple apis + path, etc

shcheklein (approved):

ilongin (approved):

dmpetrov:
Created #771 as the next step for the upload.

PR #760, opened 2024-12-29
Fix Storage abstract Client 'info_to_file' signature

dreadatour:
In Storage `Client` abstract class we do have `info_to_file` method. All it implementation have `path` param, but in abstract class this is `parent` param. This PR is going to fix this error.

### DIFF
--- a/src/datachain/client/fsspec.py
+++ b/src/datachain/client/fsspec.py
@@ -211,7 +211,7 @@ class Client(ABC):
         info = await self.fs._info(
             self.get_full_path(file.path, file.version), **kwargs
         )
-        return self.info_to_file(info, "").etag
+        return self.info_to_file(info, file.path).etag
 
     def get_file_info(self, path: str, version_id: Optional[str] = None) -> "File":
         info = self.fs.info(self.get_full_path(path, version_id), version_id=version_id)
@@ -339,7 +339,7 @@ class Client(ABC):
         return self.version_path(f"{self.PREFIX}{self.name}/{rel_path}", version_id)
 
     @abstractmethod
-    def info_to_file(self, v: dict[str, Any], parent: str) -> "File": ...
+    def info_to_file(self, v: dict[str, Any], path: str) -> "File": ...
 
     def fetch_nodes(
         self,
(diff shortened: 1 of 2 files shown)


src/datachain/client/fsspec.py
```diff
@@ -211,7 +211,7 @@ async def get_current_etag(self, file: "File") -> str:
         info = await self.fs._info(
             self.get_full_path(file.path, file.version), **kwargs
         )
-        return self.info_to_file(info, "").etag
+        return self.info_to_file(info, file.path).etag
```
dreadatour, line 214: This param is not used here, since we are only taking `etag`, but for consistency why not to pass file path here?

src/datachain/lib/file.py
```diff
@@ -364,7 +364,7 @@ def resolve(self) -> "Self":
 
         try:
             info = client.fs.info(client.get_full_path(self.path))
-            converted_info = client.info_to_file(info, self.source)
+            converted_info = client.info_to_file(info, self.path)
```
dreadatour, line 367: Same, path is not used here, but passing `source` is an error.

src/datachain/client/fsspec.py
```diff
@@ -339,7 +339,7 @@ def get_full_path(self, rel_path: str, version_id: Optional[str] = None) -> str:
         return self.version_path(f"{self.PREFIX}{self.name}/{rel_path}", version_id)
 
     @abstractmethod
-    def info_to_file(self, v: dict[str, Any], parent: str) -> "File": ...
+    def info_to_file(self, v: dict[str, Any], path: str) -> "File": ...
```
dreadatour, line 342: See all implementations:
- [azure.py](https://github.com/iterative/datachain/blob/main/src/datachain/client/azure.py#L16)
- [gcs.py](https://github.com/iterative/datachain/blob/main/src/datachain/client/gcs.py#L125)
- [hf.py](https://github.com/iterative/datachain/blob/main/src/datachain/client/hf.py#L24)
- [local.py](https://github.com/iterative/datachain/blob/main/src/datachain/client/local.py#L126)
- [s3.py](https://github.com/iterative/datachain/blob/main/src/datachain/client/s3.py#L158)

shcheklein (approved):

0x2b3bfa0 (approved):

Issue #762, opened 2024-12-30
Empty-named files / directories need special care


dmpetrov:
### Description

There is a bucket that contains directory `examples` with files. It looks like this directory was created in cloud console UI - it has a empty-name record for this directory as far as I can see. This is a common way of creating empty dirs in such storages.

To handle this correctly, we need to treat these empty-named files in a special way - they should look like dirs. This will help prevent issues when working with files and folders in the bucket.

**Issue 1**. When I list this directory I see only this empty-name record. It looks like we need to handle these empty files / directories in a special way.

```python
from datachain import DataChain

dc = DataChain.from_storage("gs://mybucket/examples").save("myds")
dc.show()
print("Files: ", dc.count())
```
```
                         file      file file              file              file  \
                       source      path size           version              etag
0  gs://mybucket  examples    0  1735506238856683  COu7/cbwzYoDEAE=
       file                             file     file
  is_latest                    last_modified location
0         1 2024-12-29 21:03:58.859000+00:00     None
Files:  1
```

The result is expected if I list it as a dir:
```python
from datachain import DataChain

dc = DataChain.from_storage("gs://mybucket/examples/").save("myds")
dc.show()
print("Files: ", dc.count())
```
```
                          file                                               file  \
                        source                                               path
0   gs://mybucket  examples/videos/HoldingPen_...
1   gs://mybucket  examples/videos/HoldingPen_...
...
Files:  171
```

**Issue 2.** System signals seem broken for this special file / dir

```python
dc = DataChain.from_storage("gs://mybucket/examples").save("myds")
dc.shuffle().show()
```

```
datachain.lib.signal_schema.SignalResolvingError: cannot resolve signal name 'sys.rand': is not found
```

### Version Info

```Text
0.8.3
Python 3.10.8
```

shcheklein:
Relevant link: https://cloud.google.com/storage/docs/objects#simulated-folders

These directories indeed were created using console UI.

PR #767, opened 2025-01-02
fix(listing): ignore special 'dir' files

shcheklein:
Fixes https://github.com/iterative/datachain/issues/762 

We override `fsspec`'s `info` to also analyze if an object listed / returned is a special case of a file like this: https://cloud.google.com/storage/docs/objects#simulated-folders

## TODO:

- [X] Find a better way / replace `isfile` / `get_info`  with something else to avoid full listing in `fsspec` (can be a followup) https://github.com/fsspec/gcsfs/issues/656 or fix GCSF
- [X] ~Add a test~ No easy way to create these special files via fsspec. Probably not worth researching this.
- [x] See if we should make `_isfile` part of the Client

### DIFF
--- a/src/datachain/lib/listing.py
+++ b/src/datachain/lib/listing.py
@@ -85,6 +85,24 @@ def ls(
     return dc.filter(pathfunc.parent(_file_c("path")) == path.lstrip("/").rstrip("/*"))
 
 
+def _isfile(client: "Client", path: str) -> bool:
+    """
+    Returns True if uri points to a file
+    """
+    try:
+        info = client.fs.info(path)
+        name = info.get("name")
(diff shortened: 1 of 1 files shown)


amritghimire (approved):

skshetry:
We use this method also to determine `dataset_name`, which this `None` thing breaks.

https://github.com/iterative/datachain/blob/e45a62f43cad47a162345f01e3a2f9a1fff8f47a/src/datachain/dataset.py#L94

Also, a URI parsing utility should not do contact remotes, at least not in this place.
I came to this PR when I was investigating a unit-test being slow, and it turns out it was trying to connect to s3 and failing.

PR #770, opened 2025-01-02
Added `case()` function

ilongin:
Adding `case()` function.
Example:
```python
  dc.mutate(
      res=func.case((C("num") > 0, "P"), (C("num") < 0, "N"), else_="Z"),
  )
```


amritghimire:
Looks like test cases are failing

amritghimire (approved):

PR #777, opened 2025-01-04
feat(signal schema): serialize base classes for custom types

shcheklein:
Fixes https://github.com/iterative/studio/issues/11140 (private repo).

If we have a class hierarchy: 

```python
from typing import Iterator

dc = DataChain.from_storage("gs://mybucket")

class VideoInterval(File):
    start: int
    end: int


def get_actions(file: File) -> Iterator[VideoInterval]:
    yield VideoInterval(start=6, end=15, **file.model_dump())
    yield VideoInterval(start=39, end=51, **file.model_dump())
    yield VideoInterval(start=145, end=166, **file.model_dump())


dc = dc.gen(action=get_actions).save("res")
```

We want for the `VideoInterval` class to have enough information about its base class `File` to be available after we serialize  / deserialize it. For downstream tools to be able to work with it like a `File` (e.g. preview in Studio).

Also, this PR fixes an existing issue with schema - complex types like `list[MyType]` were not properly handled during serialization / deserialization (and it was returning `list[Any]`).

## Fix Description

It adds `bases` field to the `_custom_types` dict that we serialize to preserve fields and their types. `bases` includes all base classes, their names, etc. When we deserialize we take the first registered in the ModelStore and pass it as a base to Pydantic create model:

## Demo

https://github.com/user-attachments/assets/9e204d9d-f0b1-465c-9dcd-99c9ca677049

## TODO:

- [x] Add more tests
- [x] Refactor a bit to reduce the complexity (remove noqa)

### DIFF
--- a/src/datachain/lib/signal_schema.py
+++ b/src/datachain/lib/signal_schema.py
@@ -13,13 +13,14 @@ from typing import (  # noqa: UP035
     Final,
     List,
     Literal,
+    Mapping,
     Optional,
     Union,
     get_args,
     get_origin,
 )
 
-from pydantic import BaseModel, create_model
+from pydantic import BaseModel, Field, create_model
 from sqlalchemy import ColumnElement
 from typing_extensions import Literal as LiteralEx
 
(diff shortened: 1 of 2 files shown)


dreadatour (approved):
Thank you for this update, looks good, tests are awesome! 👍

PR #784, opened 2025-01-05
use windows-2025 image for testing examples

mattseddon:
I am not sure if this will fix the error but you can see one [here](https://github.com/iterative/datachain/actions/runs/12623960984/job/35173552972?pr=784#step:7:212) and the list of available images is [here](https://github.com/actions/runner-images) (2025 is beta).

We can always try and then revert.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -136,7 +136,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest]
+        os: [ubuntu-latest, windows-2025]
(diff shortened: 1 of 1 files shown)


0x2b3bfa0 (approved):
1. This "error" seems to be just a warning. 🤔
2. Are you sure bumping to a new Windows Server version will fix the issue?, see https://github.com/microsoft/onnxruntime/blob/2a16ad0215a8e0fb38218b8def1f364ef34829ac/onnxruntime/python/onnxruntime_validation.py#L25

Still, there's nothing blatantly wrong with merging this.

PR #787, opened 2025-01-06
Implements if else function

ilongin:
This PR implements `ifelse` function uses already implemented `case` function.
Example of usage:
```python
  dc.mutate(
      res=func.ifelse(C("num") > 0, "P", "N"),
  )
```

### DIFF
--- a/src/datachain/func/conditional.py
+++ b/src/datachain/func/conditional.py
@@ -8,6 +8,8 @@ from datachain.sql.functions import conditional
 
 from .func import ColT, Func
 
+CaseT = Union[int, float, complex, bool, str]
+
 
 def greatest(*args: Union[ColT, float]) -> Func:
     """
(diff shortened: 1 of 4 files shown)


dreadatour (approved):
Looks great 👍

src/datachain/func/conditional.py
```diff
@@ -108,26 +108,48 @@ def case(
...
+
+
+def ifelse(condition: BinaryExpression, if_val: CaseT, else_val: CaseT) -> Func:
+    """
+    Returns the ifelse function that produces if expression which has a condition
+    and values for true and false outcome. Results can only be python primitives
+    like string, numbes or booleans. Result type is inferred from the values.
+
+    Args:
+        condition: BinaryExpression - condition which is evaluated
```
shcheklein, line 140: what is BinaryExpression - is it a regular thing for our API to specify inputs as SQLalchemy expressions? ... can we hide it? can we make it clear what is allowed here?

PR #794, opened 2025-01-07
git ignore files all pytorch files

mattseddon:
When running the `ultralytics` examples files like `yolo11n-seg.pt` are produced in the root directory. I have come close to committing these a couple of times.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -140,3 +140,6 @@ cython_debug/
 .vscode/
 .datachain/
 .dvcx/
+
+# pt files produced by ultralytics examples
+*.pt


.gitignore
```diff
@@ -140,3 +140,6 @@ cython_debug/
 .vscode/
 .datachain/
 .dvcx/
+
+# pt files produced by ultralytics examples
+yolo*.pt
```
dmpetrov, line 145: We can ignore all pt file

dmpetrov (approved):
Looks good!

PR #796, opened 2025-01-07
fix(signed url): handle content disposition

yathomasi:
- Allow to pass kwargs to generate signed url
- More context/follow up in Studio: https://github.com/iterative/studio/pull/11169


src/datachain/client/s3.py
```diff
@@ -51,6 +51,21 @@ def create_fs(cls, **kwargs) -> S3FileSystem:
...
         return cast(S3FileSystem, super().create_fs(**kwargs))
 
+    def url(self, path: str, expires: int = 3600, **kwargs) -> str:
+        """
+        Generate a signed URL for the given path.
+        """
+        version_id = kwargs.pop("version_id", None)
+        content_disposition = kwargs.pop("content_disposition", None)
+        if content_disposition:
+            kwargs["ResponseContentDisposition"] = content_disposition
```
shcheklein, line 61: can it be covered by a test? could you experiment a bit the `test_signed_url` and `test_signed_url_versioned` test in the `test_catalog`? add a similar test with content disposition? (it might happen that our mocks for the clouds don't support this, but I would still try it)

shcheklein: it's not a blocker / not critical

shcheklein (approved):

PR #801, opened 2025-01-07
Added `isnone()` function

ilongin:
`isnone(...)` function returns `True` if column or literal value is `None` / `NULL`, otherwise `False`

Note that this doesn't currently work for Clickhouse where we don't have nullable columns. We should decide if we want to introduce nullable columns and if yes, this will work there as well.

UPDATE:
In addition, this PR extends `case()` function and related codebase to accept nested functions in:
1. Case condition - value (both places)
2. Case `else_` value

This will allow usage of functions inside other functions that are based on case like `ifelse` or `isnone` as well. We also will have ability to do nested cases etc.


src/datachain/func/conditional.py
```diff
@@ -153,3 +154,28 @@ def ifelse(condition: BinaryExpression, if_val: CaseT, else_val: CaseT) -> Func:
...
+        col (str | Column | literal): Column or literal to check if None.
+            If a string is provided, it is assumed to be the name of the column.
+            If a literal is provided, it is assumed to be a string literal.
+
+    Returns:
+        Func: A Func object that represents the conditional to check if column is None.
+
+    Example:
+        ```py
+        dc.mutate(test=isnone("value"))
```
shcheklein, line 172: can we make a more reasonable example - filter, or some if condition ...

src/datachain/func/conditional.py
```diff
@@ -153,3 +154,28 @@ def ifelse(condition: BinaryExpression, if_val: CaseT, else_val: CaseT) -> Func:
...
     return case((condition, if_val), else_=else_val)
+
+
+def isnone(col: Union[str, Column]) -> Func:
+    """
+    Returns True if column value or literal is None, otherwise False
+    Args:
+        col (str | Column | literal): Column or literal to check if None.
+            If a string is provided, it is assumed to be the name of the column.
+            If a literal is provided, it is assumed to be a string literal.
```
shcheklein, line 165: do we need to specify literals separately? why?

src/datachain/func/conditional.py
```diff
@@ -153,3 +154,28 @@ def ifelse(condition: BinaryExpression, if_val: CaseT, else_val: CaseT) -> Func:
         ```
     """
     return case((condition, if_val), else_=else_val)
+
+
+def isnone(col: Union[str, Column]) -> Func:
+    """
+    Returns True if column value or literal is None, otherwise False
```
shcheklein, line 161: `or literal` - what does it mean? could you give an example please?

shcheklein: can we pass an expression btw ? e.f. isnone(ifelse(...))? if not, why not?

ilongin: Added ability to use nested functions inside `case` or `ifelse` (which is based on `case`)

src/datachain/func/conditional.py
ilongin: Added better example

src/datachain/func/conditional.py
ilongin: Removed it. Technically someone can provide just string literal but that doesn't make much sense, except for unit tests. Columns are the one to be used here

src/datachain/func/conditional.py
```diff
@@ -87,17 +88,18 @@ def least(*args: Union[ColT, float]) -> Func:
...
 
 
-def case(*args: tuple[BinaryExpression, CaseT], else_=None) -> Func:
+def case(
+    *args: tuple[Union[ColumnElement, Func], CaseT], else_: Optional[CaseT] = None
+) -> Func:
     """
     Returns the case function that produces case expression which has a list of
     conditions and corresponding results. Results can only be python primitives
     like string, numbes or booleans. Result type is inferred from condition results.
```
shcheklein, line 97: typo: numbers ...

(can we install a spellchecker please cc @skshetry ?)

src/datachain/func/conditional.py
```diff
@@ -111,45 +113,83 @@ def case(*args: tuple[BinaryExpression, CaseT], else_=None) -> Func:
...
+    type_ = _get_type(else_) if else_ is not None else None
+
     for arg in args:
-        if type_ and not isinstance(arg[1], type_):
-            raise DataChainParamsError("Statement values must be of the same type")
-        type_ = type(arg[1])
+        arg_type = _get_type(arg[1])
+        if type_ and arg_type != type_:
+            raise DataChainParamsError(
+                f"Statement values must be of the same type, got {type_} amd {arg_type}"
```
shcheklein, line 131: amd -> and (please run a spell checker, install it in IDE)

src/datachain/func/conditional.py
```diff
@@ -87,17 +88,18 @@ def least(*args: Union[ColT, float]) -> Func:
     )
 
 
-def case(*args: tuple[BinaryExpression, CaseT], else_=None) -> Func:
+def case(
+    *args: tuple[Union[ColumnElement, Func], CaseT], else_: Optional[CaseT] = None
+) -> Func:
     """
     Returns the case function that produces case expression which has a list of
     conditions and corresponding results. Results can only be python primitives
```
shcheklein, line 96: is it still true? can result now be a Func?

src/datachain/func/conditional.py
```diff
@@ -87,17 +88,18 @@ def least(*args: Union[ColT, float]) -> Func:
...
     like string, numbes or booleans. Result type is inferred from condition results.
 
     Args:
-        args (tuple(BinaryExpression, value(str | int | float | complex | bool):
-            - Tuple of binary expression and values pair which corresponds to one
-            case condition - value
-        else_ (str | int | float | complex | bool): else value in case expression
+        args (tuple((ColumnElement, Func), (str | int | float | complex | bool, Func))):
+            - Tuple of condition and values pair
```
shcheklein, line 101: please be consistent in docs ... check other places. Do we usually start with an upper case or lower? end with period or not, etc ...

src/datachain/func/conditional.py
```diff
@@ -111,45 +113,83 @@ def case(*args: tuple[BinaryExpression, CaseT], else_=None) -> Func:
...
+            condition outcome
+        else_val: (str | int | float | complex | bool, Func): value for false condition
+            outcome
 
     Returns:
         Func: A Func object that represents the ifelse function.
 
     Example:
         ```py
         dc.mutate(
```
shcheklein, line 166: let's add another example with a column expression?

also an examples with a results as an expression?

src/datachain/func/conditional.py
```diff
@@ -111,45 +113,83 @@ def case(*args: tuple[BinaryExpression, CaseT], else_=None) -> Func:
...
         ```
     """
     return case((condition, if_val), else_=else_val)
+
+
+def isnone(col: Union[str, Column]) -> Func:
+    """
+    Returns True if column value is None, otherwise False
+    Args:
+        col (str | Column): Column to check if it's None or not.
```
shcheklein, line 178: consistency

src/datachain/func/conditional.py
```diff
@@ -111,45 +113,83 @@ def case(*args: tuple[BinaryExpression, CaseT], else_=None) -> Func:
...
+        dc.mutate(test=ifelse(isnone("col"), "NONE", "NOT_NONE"))
+        ```
+    """
+    from datachain import C
+
+    if isinstance(col, str):
+        # if string, it is assumed to be the name of the column
+        col = C(col)
+
+    return case((col == None, True), else_=False)  # type: ignore [arg-type]  # noqa: E711
```
shcheklein, line 195: why do we need `# type: ignore [arg-type]  # noqa: E711`?

will users who will use `case` also have this issue with types?

tests/unit/test_func.py
```diff
@@ -678,3 +733,31 @@ def test_ifelse_mutate(dc, if_val, else_val, type_):
         [if_val, else_val, else_val, else_val, else_val]
     )
     assert res.schema["test"] == type_
+
+
+@pytest.mark.parametrize("col", ["val", C("val")])
+@skip_if_not_sqlite
```
shcheklein, line 739: does it mean that we can't even run the operation itself on CH - it won't work at all? or will it run, just returning always False?

tests/unit/test_func.py
```diff
@@ -678,3 +733,31 @@ def test_ifelse_mutate(dc, if_val, else_val, type_):
...
+    assert list(res.order_by("test").collect("test")) == sorted(
+        [False, False, False, True, True]
+    )
+    assert res.schema["test"] is bool
+
+
+@pytest.mark.parametrize("col", [C("val"), "val"])
+@skip_if_not_sqlite
+def test_isnone_with_ifelse_mutate(col):
+    dc = DataChain.from_values(
```
shcheklein, line 756: can we run these tests in memory?

src/datachain/func/func.py
```diff
@@ -423,7 +430,7 @@ def get_db_col_type(signals_schema: "SignalSchema", col: ColT) -> "DataType":
         return sql_to_python(col)
 
     return signals_schema.get_column_type(
-        col.name if isinstance(col, ColumnElement) else col
+        col.name if isinstance(col, ColumnElement) else col  # type: ignore[arg-type]
```
shcheklein, line 433: why do we have this ignore here?

src/datachain/func/conditional.py
ilongin: We do have spellchecker and it works for me, but I speculate `numbes` could be a valid word (although strange one) so it's not marked by it.

src/datachain/func/conditional.py
ilongin: Yes, all methods here return `Func` which is essentially a wrapper around sqlalchemy functions like [case](https://docs.sqlalchemy.org/en/20/core/sqlelement.html#sqlalchemy.sql.expression.case) which produces `case` construct

src/datachain/func/conditional.py
ilongin: Do we need that many examples? All other functions have exactly one example even though there can be more as well like the version with column expression as you stated.

src/datachain/func/conditional.py
ilongin: I'm trying to figure out but not sure what is wrong in this part of docs?

src/datachain/func/conditional.py
ilongin: `arg-type` is not needed indeed... probably was needed in some previous iterations and stopped being needed after some refactoring so it's a leftover.
Regarding `E711` ... `ruff` is complaining about comparing with `None` with `==` and says it should be `is None`, but since this is SQL construct it should actually be like this (I've tried with `is None` but doesn't work)

tests/unit/test_func.py
ilongin: I will double check but from what I remember it always return `False` so wrong result, which is even worse than failing IMO. Dmitry created task in Studio to allow nullable columns in CH so this should not be a problem in the future. 
I'm not 100% sure if I should merge this PR now though, as I don't like to leave it in review for so long due conflicts etc, but on the other hand it doesn't really work for CH...so I guess it will idle here until we allow nullable in CH after all.

tests/unit/test_func.py
ilongin: All tests are run in memory btw. There are more "low level" tests in `tests/unit/sql/test_conditional.py` but here I wanted to test with `DataChain` and `mutate` itself as there it will be mostly used.

src/datachain/func/conditional.py
shcheklein: I mean this part `Results can only be python primitives ...`, sorry

src/datachain/func/conditional.py
shcheklein: to my mind yes, examples is the most valuable part usually since you can get an idea of what is actually possible. We need more simple examples everywhere.

src/datachain/func/conditional.py
shcheklein: e.e. in some cases I saw it starts with lower case, sometimes I saw `-` before the descriptions, sometimes there was no period at the end, sometimes it's a new line, etc

and all of this within this single PR

and it also applies to the description, e.g. we don't have period here - is it always the case?

src/datachain/func/conditional.py
shcheklein: is there an analog of our `isnone` for this case? I wonder how should people use it raises an obvious warning ...

tests/unit/test_func.py
shcheklein: I think it's fine to merge ... it always returns False for now - and I would test for it (so that function can run). We just don't support nullables there at all ...

src/datachain/func/conditional.py
```diff
@@ -87,17 +88,19 @@ def least(*args: Union[ColT, float]) -> Func:
...
+    like string, numbers or booleans. Result type is inferred from condition results.
 
     Args:
-        args (tuple(BinaryExpression, value(str | int | float | complex | bool):
-            - Tuple of binary expression and values pair which corresponds to one
-            case condition - value
-        else_ (str | int | float | complex | bool): else value in case expression
+        args (tuple((ColumnElement, Func), (str | int | float | complex | bool, Func))):
+            Tuple of condition and values pair.
+        else_ (str | int | float | complex | bool, Func): else value in case
```
shcheklein, line 102: is it optional? can we say `optional else value`. What happens by default?

(no `in case expression` since it's clear from the context)

src/datachain/func/conditional.py
```diff
@@ -111,45 +114,84 @@ def case(*args: tuple[BinaryExpression, CaseT], else_=None) -> Func:
...
-def ifelse(condition: BinaryExpression, if_val: CaseT, else_val: CaseT) -> Func:
+def ifelse(
+    condition: Union[ColumnElement, Func], if_val: CaseT, else_val: CaseT
+) -> Func:
     """
     Returns the ifelse function that produces if expression which has a condition
-    and values for true and false outcome. Results can only be python primitives
-    like string, numbes or booleans. Result type is inferred from the values.
+    and values for true and false outcome. Results can be one of python primitives
+    like string, numbes or booleans, but can also be nested functions.
```
shcheklein, line 152: numbes -> numbers

src/datachain/func/conditional.py
```diff
@@ -1,14 +1,15 @@
...
-from sqlalchemy.sql.elements import BinaryExpression
 
 from datachain.lib.utils import DataChainParamsError
+from datachain.query.schema import Column
 from datachain.sql.functions import conditional
 
 from .func import ColT, Func
 
-CaseT = Union[int, float, complex, bool, str]
+CaseT = Union[int, float, complex, bool, str, Func]
```
shcheklein, line 12: are there tests for Func values btw? can we add them?

src/datachain/func/conditional.py
ilongin: I've seen people putting `# noqa` but it seems that there is also `is_()` and `isnone()` which I will use here.

src/datachain/func/conditional.py
ilongin: `else_` is optional correct. If omitted value is set to `None` (if it doesn't satisfies any of the case conditions). I've added tests for this and updated docs

src/datachain/func/conditional.py
ilongin: Yes, I've added tests for all possible places `Func` can be found - case condition, case value and case else. There are in `test_func.py`

src/datachain/func/func.py
ilongin: Removed it . ..added special check that `tuple` which has bee nadded to `ColT` doesn't end up here

shcheklein (approved):

PR #804, opened 2025-01-08
fix(schema): reconnect closed connection

### LINKED ISSUE #660: After some failures datachain is getting stuck in closed database error
Script to reproduce:

```python
from datachain import DataChain, File, DataModel, C


def add_video_metadata(file: File) -> str:
    raise Exception("This is a test exception")
    return file.path
    
DATA_PATH = "gs://datachain-kinetics-700-2020/val/"

video_dc = (
    DataChain.from_storage(DATA_PATH, anon=True).filter(C("file.path").glob("*.mp4"))
)

try:
    video_dc.limit(1).map(video=lambda file: add_video_metadata(file), output={'video': str}).exec()
except Exception as e:
    print(e)

try:
    video_dc.limit(1).map(video=lambda file: add_video_metadata(file), output={'video': str}).exec()
except Exception as e:
    print(e)

try:
    video_dc.limit(1).map(video=lambda file: add_video_metadata(file), output={'video': str}).exec()
except Exception as e:
    print(e)
```

It makes it the DataChain very hard to use in Jupyter. Since it's getting stuck in this state and the kernel restart is needed.

https://github.com/user-attachments/assets/0028d143-aa41-4c12-be49-50076fec672b

shcheklein:
Fixes #660 

Refactors `DataTable` by removing direct access to sqlachemy (and thus requiring direct access to sa.Engine, etc). This allows us to do a check and reconnect if connection was already closed.

### DIFF
--- a/src/datachain/data_storage/schema.py
+++ b/src/datachain/data_storage/schema.py
@@ -16,7 +16,6 @@ from datachain.sql.functions import path as pathfunc
 from datachain.sql.types import Int, SQLType, UInt64
 
 if TYPE_CHECKING:
-    from sqlalchemy import Engine
     from sqlalchemy.engine.interfaces import Dialect
     from sqlalchemy.sql.base import (
         ColumnCollection,
@@ -24,6 +23,8 @@ if TYPE_CHECKING:
         ReadOnlyColumnCollection,
     )
     from sqlalchemy.sql.elements import ColumnElement
+
+    from datachain.data_storage.db_engine import DatabaseEngine
 
 
 DEFAULT_DELIMITER = "__"
@@ -150,14 +151,12 @@ class DataTable:
     def __init__(
         self,
         name: str,
-        engine: "Engine",
-        metadata: Optional["sa.MetaData"] = None,
+        engine: "DatabaseEngine",
         column_types: Optional[dict[str, SQLType]] = None,
         object_name: str = "file",
     ):
         self.name: str = name
         self.engine = engine
-        self.metadata: sa.MetaData = metadata if metadata is not None else sa.MetaData()
         self.column_types: dict[str, SQLType] = column_types or {}
         self.object_name = object_name
 
@@ -211,12 +210,7 @@ class DataTable:
         return sa.Table(name, metadata, *columns)
 
     def get_table(self) -> "sa.Table":
-        table = self.metadata.tables.get(self.name)
-        if table is None:
-            sa.Table(self.name, self.metadata, autoload_with=self.engine)
-            # ^^^ This table may not be correctly initialised on some dialects
-            # Grab it from metadata instead.
-            table = self.metadata.tables[self.name]
+        table = self.engine.get_table(self.name)
 
         column_types = self.column_types | {c.name: c.type for c in self.sys_columns()}
         # adjusting types for custom columns to be instances of SQLType if possible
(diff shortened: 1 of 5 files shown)


src/datachain/data_storage/schema.py
```diff
@@ -213,7 +215,9 @@ def new_table(
     def get_table(self) -> "sa.Table":
         table = self.metadata.tables.get(self.name)
         if table is None:
-            sa.Table(self.name, self.metadata, autoload_with=self.engine)
+            if self.engine.is_closed:  # type: ignore[attr-defined]
+                self.engine._reconnect()  # type: ignore[attr-defined]
+            sa.Table(self.name, self.metadata, autoload_with=self.engine.engine)
```
shcheklein, line 220: [C]: the only (?) place where we had SA engine directly used (usually we would use our wrapper that does reconnect on execute).

Internally `sa.Table` does some logic to query and fails if connection is closed.

src/datachain/data_storage/schema.py
```diff
@@ -150,14 +151,12 @@ class DataTable:
     def __init__(
         self,
         name: str,
-        engine: "Engine",
-        metadata: Optional["sa.MetaData"] = None,
```
shcheklein, line 154: I'm not sure what was the case for it to be `None` and for this code below:

`self.metadata: sa.MetaData = metadata if metadata is not None else sa.MetaData()` 

that initializes its own copy of `MetaData()`

I haven't found any use case so far. It is simpler this way when internals of the `DatabaseEngine` are not exposed into DataTable.

Let's see if tests pass.

src/datachain/data_storage/sqlite.py
```diff
@@ -186,6 +186,12 @@ def _reconnect(self) -> None:
         self.db_file = db_file
         self.is_closed = False
 
+    def get_table(self, name: str) -> Table:
+        if self.is_closed:
+            # Reconnect in case of being closed previously.
+            self._reconnect()
+        return super().get_table(name)
```
shcheklein, line 193: `get_table` might trigger a connection and that triggers closed database

dreadatour (approved):
👍

src/datachain/data_storage/warehouse.py
```diff
@@ -220,7 +219,7 @@ def dataset_select_paginated(
         num_yielded = 0
 
         # Ensure we're using a thread-local connection
-        with self.clone() as wh:
+        with self.clone(use_new_connection=True) as wh:
```
skshetry, line 222: Was this added because `clone()` could reuse the closed connection by default? Maybe this could be fixed on clickhouse backend instead of having to add this flag everywhere?

shcheklein: this flag was added for CH as far as I remember, good catch @skshetry I'm not sure it is needed in this place

(not sure it worth refactoring this flag atm, I'll just revert this flag in this place)

src/datachain/data_storage/sqlite.py
```diff
@@ -186,6 +186,12 @@ def _reconnect(self) -> None:
         self.db_file = db_file
         self.is_closed = False
 
+    def get_table(self, name: str) -> Table:
+        if self.is_closed:
+            # Reconnect in case of being closed previously.
+            self._reconnect()
```
skshetry, line 192: Ideally, this should be automatic. 

```
@property
def db(self):
  if self._db.is_closed:
      self._reconnect()
  return self._db
```

I see `executemany`, `execute_str` and other methods that are not reconnecting.
Although `engine` and `metadata` do complicate this.

PR #811, opened 2025-01-13
docs(merge): add examples with Func object

shcheklein:
Improve docs a bit on merge. I keep forgetting how to put function into merge condition.


dmpetrov (approved):
Looks good!

Issue #780, opened 2025-01-05
Progress bars are broken in Jupyter notebooks


shcheklein:
### Description

https://github.com/user-attachments/assets/92c75f62-a191-40d9-9481-0168d6ac8d4c
https://github.com/user-attachments/assets/9931e7d5-9ab4-49da-87de-672061bfbdc4

### Version Info

```Text

```

shcheklein:
@P3st83 there were some improvements made, you can give a try and see if it can be fixed further or not. Thanks.

shcheklein:
No easy fix here for parallel mode, closing for now.

PR #812, opened 2025-01-13
Import tqdm to support jupyter better

shcheklein:
Partially addresses #780 

https://github.com/user-attachments/assets/911f8bff-502c-411c-aa7b-90e9e3778e6e

What I don't know how to fix yet is progress bars when we use `parallel` and we run the whole thing in the subprocess. In this case child process doesn't have access to Ipython widgets (or whatever the mechanics is behind these nice progress bars).

### DIFF
--- a/src/datachain/cache.py
+++ b/src/datachain/cache.py
@@ -9,8 +9,6 @@ from dvc_objects.fs.local import LocalFileSystem
 from dvc_objects.fs.utils import remove
 from fsspec.callbacks import Callback, TqdmCallback
 
-from .progress import Tqdm
-
 if TYPE_CHECKING:
     from datachain.client import Client
     from datachain.lib.file import File
@@ -86,9 +84,11 @@ class DataChainCache:
         size = file.size
         if size < 0:
             size = await client.get_size(from_path, version_id=file.version)
+        from tqdm.auto import tqdm
+
         cb = callback or TqdmCallback(
             tqdm_kwargs={"desc": odb_fs.name(from_path), "bytes": True, "leave": False},
-            tqdm_cls=Tqdm,
+            tqdm_cls=tqdm,
             size=size,
         )
         try:
--- a/src/datachain/progress.py
+++ b/src/datachain/progress.py
@@ -1,136 +1,14 @@
 """Manages progress bars."""
 
 import logging
-import sys
 from threading import RLock
-from typing import Any, ClassVar
 
 from fsspec import Callback
 from fsspec.callbacks import TqdmCallback
-from tqdm import tqdm
-
-from datachain.utils import env2bool
+from tqdm.auto import tqdm
 
 logger = logging.getLogger(__name__)
 tqdm.set_lock(RLock())
-
-
-class Tqdm(tqdm):
-    """
-    maximum-compatibility tqdm-based progressbars
-    """
-
-    BAR_FMT_DEFAULT = (
-        "{percentage:3.0f}% {desc}|{bar}|"
-        "{postfix[info]}{n_fmt}/{total_fmt}"
-        " [{elapsed}<{remaining}, {rate_fmt:>11}]"
-    )
-    # nested bars should have fixed bar widths to align nicely
-    BAR_FMT_DEFAULT_NESTED = (
-        "{percentage:3.0f}%|{bar:10}|{desc:{ncols_desc}.{ncols_desc}}"
-        "{postfix[info]}{n_fmt}/{total_fmt}"
-        " [{elapsed}<{remaining}, {rate_fmt:>11}]"
-    )
-    BAR_FMT_NOTOTAL = "{desc}{bar:b}|{postfix[info]}{n_fmt} [{elapsed}, {rate_fmt:>11}]"
-    BYTES_DEFAULTS: ClassVar[dict[str, Any]] = {
-        "unit": "B",
-        "unit_scale": True,
-        "unit_divisor": 1024,
-        "miniters": 1,
-    }
-
-    def __init__(
-        self,
-        iterable=None,
-        disable=None,
-        level=logging.ERROR,
-        desc=None,
-        leave=False,
-        bar_format=None,
-        bytes=False,
-        file=None,
-        total=None,
-        postfix=None,
-        **kwargs,
-    ):
-        """
-        bytes   : shortcut for
-            `unit='B', unit_scale=True, unit_divisor=1024, miniters=1`
-        desc  : persists after `close()`
-        level  : effective logging level for determining `disable`;
-            used only if `disable` is unspecified
-        disable  : If (default: None) or False,
-            will be determined by logging level.
-            May be overridden to `True` due to non-TTY status.
-            Skip override by specifying env var `DATACHAIN_IGNORE_ISATTY`.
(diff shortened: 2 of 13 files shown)


src/datachain/cache.py
```diff
@@ -86,9 +84,11 @@ async def download(
         size = file.size
         if size < 0:
             size = await client.get_size(from_path, version_id=file.version)
+        from tqdm.auto import tqdm
+
         cb = callback or TqdmCallback(
             tqdm_kwargs={"desc": odb_fs.name(from_path), "bytes": True, "leave": False},
-            tqdm_cls=Tqdm,
```
shcheklein, line 91: @skshetry do we really need that whole custom class to support this?

src/datachain/progress.py
```diff
@@ -1,138 +1,16 @@
...
-from tqdm import tqdm
-
-from datachain.utils import env2bool
+from tqdm.auto import tqdm
 
 logger = logging.getLogger(__name__)
 tqdm.set_lock(RLock())
 
 
-class Tqdm(tqdm):
```
shcheklein, line 18: @skshetry this class ..

src/datachain/cache.py
skshetry: This was copied from `dvc-objects` which in turn came from `dvc`. There are some minor improvements (see how it sets bar_format, and fills the `ncols`, etc.), but it is debatable if it's needed. 

I believe we should have a uniform Tqdm progress bars, so I do think we should have a small class or a function that gives us similar looking progress bars.

skshetry:
> What I don't know how to fix yet is progress bars when we use `parallel` and we run the whole thing in the subprocess. In this case child process doesn't have access to Ipython widgets (or whatever the mechanics is behind these nice progress bars).

We should not use progress bars inside `udf` at all. This just does not work with `Tqdm`, and it will make progress bars jump constantly, and leave gaps in between. The problem gets worse when the user's code creates progress bars in the main thread. `tqdm` has `position` value that does not get shared in multiple processes - it is essentially copied to different processes once, and one will try to overwrite another.

src/datachain/cache.py
shcheklein: then I suggest we come to it later when we have actual demand / improvements

it's strange to have a large class that is used in a single internal place

src/datachain/query/dataset.py
```diff
@@ -366,12 +367,16 @@ def get_download_callback(suffix: str = "", **kwargs) -> CombinedDownloadCallbac
 
 
 def get_processed_callback() -> Callback:
-    return TqdmCallback({"desc": "Processed", "unit": " rows", "leave": False})
+    return TqdmCallback(
+        {"desc": "Processed", "unit": " rows", "leave": False}, tqdm_cls=tqdm
```
skshetry, line 371: Note that `TqdmCallback` (or, `fsspec.Callback` to be specific) can open nested progress bars in certain case. One example is `copy` where one progress bar is about no. of files copied, and the second is the progress bar for individual files being copied.

I am not sure how `notebook_tqdm` works in that case.

shcheklein:
> We should not use progress bars inside udf at all. This just does not work with Tqdm, and it will make progress bars jump constantly, and leave gaps in between. The problem gets worse when the user's code creates progress bars in the main thread. tqdm has position value that does not get shared in multiple processes - it is essentially copied to different processes once, and one will try to overwrite another.

we have a single progress bar there (per udf), we don't pass it to all the subprocesses - so it works fine in CLI. We run a wrapper process though to avoid __main__, that process controls progress (not processes that actually run UDFs) and that process doesn't have access to JUpyter.

src/datachain/query/dataset.py
shcheklein: nested bars work in notebook tqdm (see the demo in the description). Not 100% about nested bars inside copy.

Anyways, we should be better now compared to what it was before. Nesting doesn't work in text mode anyways, the whole experience is broken.

skshetry:
> > We should not use progress bars inside udf at all. This just does not work with Tqdm, and it will make progress bars jump constantly, and leave gaps in between. The problem gets worse when the user's code creates progress bars in the main thread. tqdm has position value that does not get shared in multiple processes - it is essentially copied to different processes once, and one will try to overwrite another.
> 
> 
> 
> we have a single progress bar there (per udf), we don't pass it to all the subprocesses - so it works fine in CLI. We run a wrapper process though to avoid __main__, that process controls progress (not processes that actually run UDFs) and that process doesn't have access to JUpyter.

We do have some UDFs that create progress bars - such as ArrowGenerator and maybe Listing?

shcheklein:
> We do have some UDFs that create progress bars - such as ArrowGenerator and maybe Listing?

yep, may be. But even in that case this PR is most likely an improvement. We can decide and review and remove some internal progress bars. I'm not sure why we use / need them.

src/datachain/progress.py
```diff
@@ -1,138 +1,16 @@
...
 
 from fsspec import Callback
 from fsspec.callbacks import TqdmCallback
-from tqdm import tqdm
-
-from datachain.utils import env2bool
+from tqdm.auto import tqdm
 
 logger = logging.getLogger(__name__)
 tqdm.set_lock(RLock())
```
skshetry, line 11: Let's remove this too when we can.

Maybe too much detail, but [`tqdm` does locking by default which is multiprocess-safe](https://github.com/tqdm/tqdm/blob/0ed5d7f18fa3153834cbac0aa57e8092b217cc16/tqdm/std.py#L76). But since we overwrite this, this is only thread-safe.

[`dvc-data` does reset it, so we'll be overriding it anyway as we are importing it someplace](https://github.com/iterative/dvc-data/blob/8e28434fd4f07f948f018350abe86928094312ea/src/dvc_data/callbacks.py#L12). But I hope we can remove it someday there. I have no idea why this was done in `dvc`.

skshetry: (I have a suspicion that this is what causes progress to jump in dvc)

skshetry (approved):

src/datachain/progress.py
shcheklein: yep, makes sense

src/datachain/progress.py
```diff
@@ -148,6 +26,8 @@ def __init__(self, tqdm_kwargs=None, *args, **kwargs):
         self.files_count = 0
         tqdm_kwargs = tqdm_kwargs or {}
         tqdm_kwargs.setdefault("postfix", {}).setdefault("files", self.files_count)
+        kwargs = kwargs or {}
+        kwargs["tqdm_cls"] = tqdm
```
skshetry, line 30:
```suggestion
        kwargs.setdefault("tqdm_cls", tqdm)
```

shcheklein: it's not the same though - it will raise if kwargs is None, right?

skshetry: kwargs would never be None, as it's a variable keyword arguments. It is always going to be a dict.

shcheklein: yep, makes sense ... feel free to do a separate minor PR :)

src/datachain/progress.py
skshetry: `dvc-data` fixed in https://github.com/iterative/dvc-data/pull/592.

PR #813, opened 2025-01-13
Add CSV parsing options

skirdey:
Adds support for CSV parsing options (via `pyarrow` parser). Among them: for CSV files where values can span several lines.


src/datachain/lib/dc.py
```diff
@@ -1955,6 +1956,7 @@ def from_csv(
             column_types : Dictionary of column names and their corresponding types.
                 It is passed to CSV reader and for each column specified type auto
                 inference is disabled.
+            newlines_in_values: Tells pareser that values in csv can span multiple lines.
```
shcheklein, line 1959: `pareser` -> `parser`

csv -> CSV (for consistency)

src/datachain/lib/dc.py
```diff
@@ -1934,6 +1934,7 @@ def from_csv(
         session: Optional[Session] = None,
         settings: Optional[dict] = None,
         column_types: Optional[dict[str, "Union[str, ArrowDataType]"]] = None,
+        newlines_in_values: bool = False,
```
shcheklein, line 1937: a suggestion: should we introduce a "parse_options` dict as a parameter that we can just pass into the parser then? (I'm a bit concerned about adding each parameter one by one here). WDYT?

shcheklein (commented):
Thanks a lot @skirdey ! Please read a few comments with some fixes and suggestions on how to improve it a bit.

skirdey:
I've added a simple dict as parse options config. 
Unit tests gave the same results as on main branch, but there are quite a few failures on both when running locally. 

I've kept it simple. Let me know if you have preference to see ParseOptions as more defined dataclass. 

I've left out callable parse option as it seems to add complexity, maybe in the next PR.

src/datachain/lib/dc.py
```diff
@@ -1943,6 +1943,7 @@ def from_csv(
             path : Storage URI with directory. URI must start with storage prefix such
                 as `s3://`, `gs://`, `az://` or "file:///".
             delimiter : Character for delimiting columns.
+                Leaving for backwards compatibility.
```
shcheklein, line 1954: How about "Takes precedence if also specified in `parse_options_config`"?

(I think it's fine to keep both, delimiter it probably commonly used and it's fine to keep it as an explicit option as a syntax sugar (or we can review it later)).

src/datachain/lib/dc.py
```diff
@@ -1934,7 +1934,7 @@ def from_csv(
         session: Optional[Session] = None,
         settings: Optional[dict] = None,
         column_types: Optional[dict[str, "Union[str, ArrowDataType]"]] = None,
-        newlines_in_values: bool = False,
+        parse_options_config: Optional[dict[str, "Union[str, bool]"]] = None,
```
shcheklein, line 1945: (minor, just to brainstorm a bit, feel free to ignore)

How about: `parse_options` or `parse_config` - to make it less verbose and smaller in size (client code will look better a bit potentially)

shcheklein:
@skirdey seems CI failed, could you please take a look?

src/datachain/lib/dc.py
```diff
@@ -1943,14 +1943,16 @@ def from_csv(
         session: Optional[Session] = None,
         settings: Optional[dict] = None,
         column_types: Optional[dict[str, "Union[str, ArrowDataType]"]] = None,
+        parse_options: Optional[dict[str, "Union[str, bool]"]] = None,
```
dreadatour, line 1946: According to https://arrow.apache.org/docs/python/generated/pyarrow.csv.ParseOptions.html for `invalid_row_handler` key in parse_options value type will be `callable`.

dreadatour (approved):
Looks good to me! Thank you for this improvement! 👍

Issue #797, opened 2025-01-07
Support Video file and Video clip, Video frame models and operations with them


shcheklein:
- It should include methods like:
  - getting metadata
  - split into subvideos
  - split into frames

We need to enable visualizations - e.g. when we have a frame we want to open the video on that frame and show bounding boxes, etc.

@dmpetrov has some examples in one of the teams + @dreadatour was doing some of this already + we have video pose detection tutorial in datachain-examples. We need to consolidate and do proper APIs + Studio support.

APIs should work in streaming mode.

shcheklein:
First step in this ticket is describe the API.

dreadatour:
# API

Already implemented here: https://github.com/iterative/datachain/pull/814

I was trying to describe API without implementation and it was quite hard without real usage examples.

### Video Meta

```python
class VideoFile(File):
    """`DataModel` for reading video files."""


class VideoMeta(DataModel):
    """`DataModel` for video file meta information."""

    width: int
    height: int
    fps: float
    duration: float
    frames_count: int
    codec: str
```

```python
def video_meta(file: "VideoFile") -> VideoMeta:
    """
    Returns video file meta information.

    Args:
        file (VideoFile): VideoFile object.

    Returns:
        VideoMeta: Video file meta information.
    """
```

Usage example:

```python
from datachain import DataChain
from datachain.lib.video import video_meta

DataChain.from_dataset("videos").map(meta=video_meta).save("videos-meta")
```

### Get video frame

```python
def video_frame_np(file: "VideoFile", frame: int) -> "ndarray":
    """
    Reads video frame from a file.

    Args:
        file (VideoFile): VideoFile object.
        frame (int): Frame number to read.

    Returns:
        ndarray: Video frame.
    """


def video_frame(file: "VideoFile", frame: int, format: str = "jpeg") -> bytes:
    """
    Reads video frame from a file and returns as image bytes.

    Args:
        file (VideoFile): VideoFile object.
        frame (int): Frame number to read.
        format (str): Image format (default: 'jpeg').

    Returns:
        bytes: Video frame image as bytes.
    """


def save_video_frame(
    file: "VideoFile",
    frame: int,
    output_file: Union[str, pathlib.Path],
    format: str = "jpeg",
) -> None:
    """
    Saves video frame as an image file.

    Args:
        file (VideoFile): VideoFile object.
        frame (int): Frame number to read.
        output_file (Union[str, pathlib.Path]): Output file path.
[...]

PR #814, opened 2025-01-13
Add video models + functions

dreadatour:
See https://github.com/iterative/datachain/issues/797

### Video models added

```python
class VideoFile(File):
    """`DataModel` for reading video files."""

    def get_info(self) -> "Video":
        """Returns video file information."""

    def get_frame_np(self, frame: int) -> "ndarray":
        """Reads video frame from a file."""

    def get_frame(self, frame: int, format: str = "jpg") -> bytes:
        """Reads video frame from a file and returns as image bytes."""

    def save_frame(self, frame: int, output_file: str, format: Optional[str] = None) -> "VideoFrame":
        """Saves video frame as an image file."""

    def get_frames_np(self, start_frame: int = 0, end_frame: Optional[int] = None, step: int = 1) -> "Iterator[ndarray]":
        """Reads video frames from a file."""

    def get_frames(self, start_frame: int = 0, end_frame: Optional[int] = None, step: int = 1, format: str = "jpg") -> "Iterator[bytes]":
        """Reads video frames from a file and returns as bytes."""

    def save_frames(self, output_dir: str, start_frame: int = 0, end_frame: Optional[int] = None, step: int = 1, format: str = "jpg") -> "Iterator[VideoFrame]":
        """Saves video frames as image files."""

    def save_fragment(self, start_time: float, end_time: float, output_file: str) -> "VideoFragment":
        """Saves video interval as a new video file."""

    def save_fragments(self, intervals: list[tuple[float, float]], output_dir: str) -> "Iterator[VideoFragment]":
        """Saves video intervals as new video files."""


class VideoFragment(VideoFile):
    """`DataModel` for reading video fragments."""

    start: float = Field(default=-1.0)
    end: float = Field(default=-1.0)


class VideoFrame(VideoFile):
    """`DataModel` for reading video frames."""

    frame: int = Field(default=-1)
    timestamp: float = Field(default=-1.0)
```

### Meta models added

```python
class Image(DataModel):
    """`DataModel` for image file meta information."""

    width: int = Field(default=-1)
    height: int = Field(default=-1)
    format: str = Field(default="")


class Video(DataModel):
    """`DataModel` for video file meta information."""

    width: int = Field(default=-1)
    height: int = Field(default=-1)
    fps: float = Field(default=-1.0)
    duration: float = Field(default=-1.0)
    frames: int = Field(default=-1)
    format: str = Field(default="")


class Frame(DataModel):
    """`DataModel` for video frame image meta information."""

    frame: int = Field(default=-1)
    timestamp: float = Field(default=-1.0)
    width: int = Field(default=-1)
    height: int = Field(default=-1)
    format: str = Field(default="")
```

### Usage examples

Can be found here: https://github.com/iterative/datachain-examples/pull/28


src/datachain/lib/file.py
```diff
@@ -454,6 +458,53 @@ def save(self, destination: str):
         self.read().save(destination)
 
 
+class ImageMeta(DataModel):
+    """`DataModel` for image file meta information."""
+
+    width: int
+    height: int
+    format: str
```
dmpetrov, line 466: How about EXIF and XMP? :)

dmpetrov (changes requested):
Amazing PR!

It would be great to use concise and minimalistic naming and API because we are going to have many file types for multiple domains.

1. **Naming**

Keywords like Meta will make it hard for user to remember and use the classes - user have their own meta 🙂

How about this renaming:
VideoFile -> BaseVideo (I assume people won't use this often)
VideoMeta -> Video (the most used class)
VideoClip -> Clip (also, shouldn't it be based on Video with meta?)
VideoFrame -> FrameBase
VideoFrameMeta -> Frame

start_time --> start
end_time --> end
frames_count --> count

Image -> BaseImage
ImageMeta -> Image

FileTypes can be also extended: `image` (read meta), `base_image` (do not read meta), `video` (read meta), `base_video` (do not read meta), `video_clip`, `base_video_clip `, ...

2. **Do we need dummy classes?**

I assume that people prefer working with meta information while dealing with images and videos. A followup question - do we really need `BaseImages` and `BaseVideo` without any logic? Why don't we clean up API and keep only Meta-enrich version in the API? User still can work with videos as File if meta is not needed.

3. **Do we need singular methods?**

`save_video_clips()` and `save_video_clip()` How much extra code user needs to get rid of singular form. If one method - let's avoid the singular version.

The same question for `video_frames()` and `video_frames_np()`

I assume, we can add the method and classes later if there is a need. But I'd not start with such rich API for now and try my best to keep in minimalistic.


WDYT?

pyproject.toml
```diff
@@ -77,6 +77,12 @@ hf = [
   "numba>=0.60.0",
   "datasets[audio,vision]>=2.21.0"
 ]
+video = [
+  "av<14",
```
shcheklein, line 81: Q: why `<14`?

src/datachain/lib/file.py
```diff
@@ -489,5 +540,11 @@ def get_file_type(type_: FileType = "binary") -> type[File]:
         file = TextFile
     elif type_ == "image":
         file = ImageFile  # type: ignore[assignment]
+    elif type_ == "video":
+        file = VideoFile
+    elif type_ == "video_clip":
```
shcheklein, line 545: how will it initialize start  / end?

src/datachain/lib/file.py
```diff
@@ -489,5 +540,11 @@ def get_file_type(type_: FileType = "binary") -> type[File]:
...
+        file = VideoClip
+    elif type_ == "video_frame":
```
shcheklein, line 547: how will it initialize frame details?

in general we distinguish files types AFAIR only because they have different behavior for open / read, and other methods ... probably they don't have any additional info

e.g. for VideoClip we can make read / open return us the clip (instead of using an external helper or DataChain uitils)

src/datachain/lib/video.py
```diff
@@ -0,0 +1,273 @@
...
+        if frame < start_frame:
+            continue
+        if (frame - frame_shift) % step != 0:
+            continue
+        if end_frame is not None and frame > end_frame:
+            break
+        yield img
+
+
+def video_frames(
```
shcheklein, line 148: can a lot of these helpers become part of the `Video*` classes?

pyproject.toml
dreadatour: On v14 it is an error: `AttributeError: 'av.video.codeccontext.VideoCodecContext' object has no attribute 'close'`, thread: https://github.com/PyAV-Org/PyAV/discussions/1700

src/datachain/lib/file.py
dreadatour: Good question 👍 Let me think.

pyproject.toml
shcheklein: is it temporary? can we put a link here to it as a comment then please?

src/datachain/lib/file.py
dreadatour: Yes, you're right. Let me think 🤔

src/datachain/lib/file.py
```diff
@@ -454,6 +458,53 @@ def save(self, destination: str):
...
 
+class ImageMeta(DataModel):
+    """`DataModel` for image file meta information."""
+
+    width: int
+    height: int
+    format: str
+
+
+class VideoFile(File):
```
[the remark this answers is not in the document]
dreadatour: It is for typings, so we can handle file type, for example, in Studio. Also I was thinking to add some methods from `video.py` here, but I am not sure yet.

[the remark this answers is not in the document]
shcheklein: okay, similar to Image ... makes sense, and probably we need to move a lot of methods inside

also, do we do `ImgeFile` and `TextFile` or just `Image` (just check consistency)

src/datachain/lib/video.py
dreadatour: Good question 👍 I was thinking about this and tried to implement it this way, but in the end I've checked other types and files in `lib` module (images, hf) and make it the same way.

I was also thinking and trying to move all the models to the `datachain.model` module, but it turns out it needs more work and may be not backward compatible with `File` model. In is a subject for a separate PR.

pyproject.toml
dreadatour: Sure! Good comment, thank you 👍

src/datachain/lib/file.py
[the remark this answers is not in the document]
dreadatour: We do have `ImageFile` and `TextFile`, so we may want to rename these too (in separate PR, I think).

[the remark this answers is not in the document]
shcheklein: that's fine I think to have it for now ... I was just asking to double check the consistency. I'm totally fine with VideoFile

src/datachain/lib/video.py
```diff
@@ -0,0 +1,273 @@
...
+    Args:
+        file (VideoFile): VideoFile object.
+
+    Returns:
+        VideoMeta: Video file meta information.
+    """
+    props = iio.improps(file.stream(), plugin="pyav")
+    frames_count, width, height, _ = props.shape
+
+    meta = iio.immeta(file.stream(), plugin="pyav")
```
dreadatour, line 38: I don't like this part, it looks like we are reading video file twice here. Need to check the other way to get video meta information.

dreadatour:
> 1. **Naming**
> 
> Keywords like Meta will make it hard for user to remember and use the classes - user have their own meta 🙂

👍

> How about this renaming: VideoFile -> BaseVideo (I assume people won't use this often) VideoMeta -> Video (the most used class) VideoClip -> Clip (also, shouldn't it be based on Video with meta?) VideoFrame -> FrameBase VideoFrameMeta -> Frame

For now we have naming with `File`: `TextFile`, `ImageFile` and `File` itself. I left `VideoFile` for now, but rename others:

- `ImageMeta` -> `Image`
- `VideoClipFile` -> `VideoClip` (I can rename it to `Clip` as you suggested, just not sure yet, because see next line)
- `VideoFrameFile` -> `VideoFrame` (I can rename it to `Frame` to be consistent with `Clip`, also `Frame` is already busy, see below)
- `VideoMeta` -> `Video`
- `VideoFrameMeta` -> `Frame`

> start_time --> start end_time --> end frames_count --> count

Done. Only `frames_count` became `frames`, because I am not sure about `count`, too general, IMO.

> Image -> BaseImage ImageMeta -> Image

We don't have `Image` model, we have `ImageFile` model, left it as is for now. `ImageMeta` -> `Image` done.

> FileTypes can be also extended: `image` (read meta), `base_image` (do not read meta), `video` (read meta), `base_video` (do not read meta), `video_clip`, `base_video_clip `, ...

That's good suggestion, only we use `FileTypes` for now only in `from_storage` method. I am not sure we we want to change it to download files and read meta 🤔 Even with additional param.

> 2. **Do we need dummy classes?**
> 
> I assume that people prefer working with meta information while dealing with images and videos. A followup question - do we really need `BaseImages` and `BaseVideo` without any logic? Why don't we clean up API and keep only Meta-enrich version in the API? User still can work with videos as File if meta is not needed.

Good question. I've added `VideoFile` only because we already have `ImageFile`, just to be consistent. Also it is useful when we use `from_storage` with `type=video`, and then we can use `VideoFile` type in mappers, like this:

```python
def video_meta(file: "VideoFile") -> Video:
    """
    Returns video file meta information.

    Args:
        file (VideoFile): VideoFile object.

    Returns:
        Video: Video file meta information.
    """
```

> 3. **Do we need singular methods?**
> 
> `save_video_clips()` and `save_video_clip()` How much extra code user needs to get rid of singular form. If one method - let's avoid the singular version.
> The same question for `video_frames()` and `video_frames_np()`

Sounds reasonable to me 👍 Will update the code (not done yet).

4. **Default values**

Done.

> WDYT?

Those are great comments! Love the discussion ❤️

src/datachain/lib/video.py
```diff
@@ -0,0 +1,273 @@
...
+    from numpy import ndarray
+
+    from datachain.lib.file import VideoFile
+
+try:
+    import imageio.v3 as iio
+    from moviepy.video.io.VideoFileClip import VideoFileClip
+except ImportError as exc:
+    raise ImportError(
+        "Missing dependencies for processing video:\n"
```
shcheklein, line 19: `video:` -> `video.`

src/datachain/lib/video.py
shcheklein: yep, also are we reading the whole file to get meta?

src/datachain/lib/file.py
```diff
@@ -447,13 +451,60 @@ class ImageFile(File):
...
+    width: int = Field(default=0)
+    height: int = Field(default=0)
+    format: str = Field(default="")
+
+
+class VideoFile(File):
+    """`DataModel` for reading video files."""
+
+
+class VideoClip(VideoFile):
```
shcheklein, line 473: so, how are these all modes connected with the helpers? how do I instantiate them? do I have to write my own UDFs to do that (just instantiate these classes?)

src/datachain/lib/file.py
```diff
@@ -447,13 +451,60 @@ class ImageFile(File):
...
         fobj = super().read()
-        return Image.open(BytesIO(fobj))
+        return PilImage.open(BytesIO(fobj))
 
     def save(self, destination: str):
         """Writes it's content to destination"""
         self.read().save(destination)
 
 
+class Image(DataModel):
```
shcheklein, line 461: why do we need this separate model?

src/datachain/lib/file.py
```diff
@@ -447,13 +451,60 @@ class ImageFile(File):
...
+
+
+class VideoFrame(VideoFile):
+    """`DataModel` for reading video frames."""
+
+    frame: int = Field(default=0)
+    timestamp: float = Field(default=0)
+
+
+class Video(DataModel):
```
shcheklein, line 487: Should it be a subclass of VideoFile?

src/datachain/lib/video.py
dreadatour: Fixed. Also updated same error [here](https://github.com/iterative/datachain/pull/814/files#diff-9a896e0e8e9e7843bca99bb9a91564ef3e60a0221a7aec69f6f66afdfb59c4aeR23).

src/datachain/lib/video.py
```diff
@@ -0,0 +1,275 @@
...
+    file_stem = file.get_file_stem()
+
+    for i, img in enumerate(video_frames_np(file, start_frame, end_frame, step)):
+        frame = start_frame + i * step
+        output_file = os.path.join(output_dir, f"{file_stem}_{frame:06d}.{format}")
+        iio.imwrite(output_file, img, extension=f".{format}")
+        yield output_file
+
+
+def save_video_clip(
```
dmpetrov, line 205: It looks like it needs to be renamed to `save_subvideo()`
In the class names, we use term `Clip` for virtual videos (start-end) while in this case you are creating just another Video, not clip.

So, it needs to be renamed or we need to avoid this Clip-as-virtual-reference terminology.

src/datachain/lib/video.py
```diff
@@ -0,0 +1,275 @@
...
+    file: "VideoFile",
+    start_time: float,
+    end_time: float,
+    output_file: Union[str, pathlib.Path],
+    codec: str = "libx264",
+    audio_codec: str = "aac",
+) -> None:
```
dmpetrov, line 212: It would be great to generalize the single and plural methods. We just need to come up with output format like `output="{name}{:06d}.{ext}")` and provide a string in case of a single file.

Also, this method will require generalization for writing to cloud like `output={source}/tmp/{name}{:06d}.{ext}`

dmpetrov (changes requested):
Great improvements.
A few followup questions about moving the methods to Video class and a plural-singular method.

shcheklein:
@dreadatour what is the status of this? :) any updates, any ETA?

dreadatour:
> @dreadatour what is the status of this? :) any updates, any ETA?

I am working on examples here: https://github.com/iterative/datachain-examples/pull/28 to be able to answer the questions from this PR comments and use implemented models and functions in close-to-real-life examples.

src/datachain/lib/video.py
dreadatour: For now I have rewritten this code to use `ffmpeg-python` package, it is using file underneath. This is the fastest way I know to get video file metadata and more robust, since it is using `ffmpeg`. I am open to discussion here.

src/datachain/lib/file.py
dreadatour: Same as for video info (`Video` model). I can remove it from this PR 🤔

src/datachain/lib/file.py
dreadatour: Good question. This is video meta information only. Do you think it will be used with `VideoFile` model only? What is the best name for this model then? 🤔

src/datachain/lib/file.py
```diff
@@ -447,13 +451,61 @@ class ImageFile(File):
...
+
+    start: float = Field(default=0)
+    end: float = Field(default=0)
+
+
+class VideoFrame(VideoFile):
+    """`DataModel` for reading video frames."""
+
+    frame: int = Field(default=0)
+    timestamp: float = Field(default=0)
```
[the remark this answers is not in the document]
dreadatour: Done ✅

src/datachain/lib/video.py
```diff
@@ -0,0 +1,275 @@
...
+    from moviepy.video.io.VideoFileClip import VideoFileClip
+except ImportError as exc:
+    raise ImportError(
+        "Missing dependencies for processing video.\n"
+        "To install run:\n\n"
+        "  pip install 'datachain[video]'\n"
+    ) from exc
+
+
+def video_meta(file: "VideoFile") -> Video:
```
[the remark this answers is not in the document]
dreadatour: Renamed to `video_info`, also added `get_info` method to `VideoFile` model.

src/datachain/lib/video.py
```diff
@@ -0,0 +1,275 @@
...
+    Returns:
+        ndarray: Video frame.
+    """
+    if frame < 0:
+        raise ValueError("frame must be a non-negative integer.")
+
+    return iio.imread(file.stream(), index=frame, plugin="pyav")
+
+
+def video_frame(file: "VideoFile", frame: int, format: str = "jpeg") -> bytes:
```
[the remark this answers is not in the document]
dreadatour: Done ✅

src/datachain/lib/video.py
```diff
@@ -0,0 +1,275 @@
...
+    """
+    img = video_frame_np(file, frame)
+    return iio.imwrite("<bytes>", img, extension=f".{format}")
+
+
+def save_video_frame(
+    file: "VideoFile",
+    frame: int,
+    output_file: Union[str, pathlib.Path],
+    format: str = "jpeg",
```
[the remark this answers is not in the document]
dreadatour: Done ✅

src/datachain/lib/video.py
```diff
@@ -0,0 +1,275 @@
...
+        bytes: Video frame image as bytes.
+    """
+    img = video_frame_np(file, frame)
+    return iio.imwrite("<bytes>", img, extension=f".{format}")
+
+
+def save_video_frame(
+    file: "VideoFile",
+    frame: int,
+    output_file: Union[str, pathlib.Path],
```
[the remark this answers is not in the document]
dreadatour: Not really. Removed 👍

src/datachain/lib/video.py
```diff
@@ -0,0 +1,275 @@
...
+            break
+        yield img
+
+
+def video_frames(
+    file: "VideoFile",
+    start_frame: int = 0,
+    end_frame: Optional[int] = None,
+    step: int = 1,
+    format: str = "jpeg",
```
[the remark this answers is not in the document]
dreadatour: Done ✅

src/datachain/lib/video.py
dreadatour: Renamed to `save_video_fragment` (very common name), same for `VideoClip` model — renamed to `VideoFragment`. Open to discussion here.

src/datachain/lib/video.py
dreadatour: Agree, not done yet 🤔

dreadatour:
I've updated the code.

- `VideoFile` model now have all the methods to work with video files: `get_info`, `get_frame`, `save_frame`, `save_fragment`, etc.
- `VideoClip` model renamed to `VideoFragment`.
- Simplify `save_fragment` and `save_fragments` methods: use raw `ffmpeg` now without re-encoding video files — much faster and robust.
- Remove `moviepy` dependency — it was outdated version (`moviepy<2`), using raw `ffmpeg` now (`ffmpeg-python` dependency).
- Other small changes.

Also updated video example notebook here: https://github.com/iterative/datachain-examples/pull/28, please, check. It is using most of the new features from this PR.

src/datachain/lib/file.py
```diff
@@ -206,6 +207,19 @@ def upload(
...
 
+    @classmethod
+    def from_local_path(cls, path: str, catalog: Optional["Catalog"] = None) -> "File":
+        if catalog is None:
+            from datachain.catalog.loader import get_catalog
+
+            catalog = get_catalog()
+
+        parent, name = posixpath.split(path)
+        client = catalog.get_client(parent)
```
mattseddon, line 218: [C] The top of this method is the same as `upload` could have a `get_client_from_path` helper

src/datachain/lib/file.py
```diff
@@ -206,6 +207,19 @@ def upload(
...
+
+        file_info = client.fs.info(path)
+        return client.info_to_file(file_info, name)
```
mattseddon, line 221: If you want to be able to call `open` on this `File` you need to `_set_stream` with a catalog.

src/datachain/lib/file.py
```diff
@@ -486,13 +504,223 @@ class ImageFile(File):
...
+class VideoFile(File):
+    """`DataModel` for reading video files."""
+
+    def get_info(self) -> "Video":
+        """Returns video file information."""
+        from .video import video_info
+
+        return video_info(self)
+
+    def get_frame_np(self, frame: int) -> "ndarray":
```
mattseddon, line 531: thinking out-loud here but should a frame be an `ImageFile` and `ImageFile` have a `to_ndarray` method?

mattseddon: I'll take a look at the notebook but my thought would be that you would want to be able to call something like `video.split_to_frame` with an optional start/end frame + optional destination path and DataChain would split the video into frames and upload them all to a bucket as images.

mattseddon: The frames use-case could end up looking something like:

(
    DataChain.from_storage("gs://datachain-demo/some-desc/videos")
    .limit(20)
    .gen(frame=file.split_to_frame, params="file", output={"frame": ImageFile})
    .setup(yolo=lambda: YOLO("yolo11n.pt"))
    .map(boxes=process_bboxes)
    .show()
)

dreadatour: This also can be done by the `save_frames` method below, or we can add new method `upload_frames` to upload images to the storage instead of saving them.

mattseddon: `save_frames` without upload breaks the promise of dataset reproducibility

shcheklein (commented):
seems we need quite a lot of tests for this?

shcheklein:
@dreadatour is the PR description up to date?

pyproject.toml
```diff
@@ -77,6 +77,14 @@ hf = [
   "numba>=0.60.0",
   "datasets[audio,vision]>=2.21.0"
 ]
+video = [
+  # Use 'av<14' because of incompatibility with imageio
+  # See https://github.com/PyAV-Org/PyAV/discussions/1700
+  "av<14",
+  "imageio[ffmpeg]",
+  "moviepy<2",
```
shcheklein, line 85: why this is restricted?

src/datachain/lib/video.py
shcheklein: > Done ✅
should this file be removed?

src/datachain/lib/file.py
```diff
@@ -248,6 +262,10 @@ def read_text(self):
         with self.open(mode="r") as stream:
             return stream.read()
 
+    def stream(self) -> BytesIO:
+        """Returns file contents as BytesIO stream."""
+        return BytesIO(self.read())
```
shcheklein, line 267: hmm, looks weird a bit? 

is `self.open` gives a stream?

otherwise this method is not streaming - I think it will read the whole file and just wraps it? it is then a confusing name for this

src/datachain/lib/file.py
shcheklein: it's just a bit weird that we have ImageFile and Image (that contains only some basic metadata) 🤔

src/datachain/lib/file.py
```diff
@@ -486,13 +504,223 @@ class ImageFile(File):
...
+
+        return video_frame(self, frame, format)
+
+    def save_frame(self, frame: int, output_file: str) -> "VideoFrame":
+        """
+        Saves video frame as an image file.
+
+        Args:
+            frame (int): Frame number to read.
+            output_file (str): Output file path.
```
shcheklein, line 566: what format is default? does it support different formats?

src/datachain/lib/file.py
```diff
@@ -486,13 +504,223 @@ class ImageFile(File):
...
+            format (str): Image format (default: 'jpg').
+
+        Returns:
+            bytes: Video frame image as bytes.
+        """
+        from .video import video_frame
+
+        return video_frame(self, frame, format)
+
+    def save_frame(self, frame: int, output_file: str) -> "VideoFrame":
```
shcheklein, line 560: does VideoFrame has information that is needed it to be File?

src/datachain/lib/vfile.py
shcheklein, line 1: this file is needed?

src/datachain/lib/file.py
```diff
@@ -206,6 +207,19 @@ def upload(
         file._set_stream(catalog)
         return file
 
+    @classmethod
+    def from_local_path(cls, path: str, catalog: Optional["Catalog"] = None) -> "File":
```
shcheklein, line 211: what does this method do? why do we need it?

src/datachain/lib/file.py
dreadatour: It was `VideoMeta` (and `ImageMeta`) before, but Dmitry was asked to rename these models [here](https://github.com/iterative/datachain/pull/814#pullrequestreview-2547435306). I agree having `Video` (`Image`) model with just meta looks odd. I think you're right and we should inherit this model from `VideoFile` (`ImageFile`) to extend files with meta, than it will make sense. If no, what do you think about `VideoInfo` (and `ImageInfo`)?

pyproject.toml
dreadatour: Forgot to remove it

src/datachain/lib/file.py
dreadatour: 🤔 You're right. I was confused with `Any` type.

src/datachain/lib/file.py
dreadatour: Output format is taken from output file extension. See [here](https://github.com/iterative/datachain/pull/814/files#diff-281c712b90d682b4f700b6dcb351cd72e823440f3f19bae1d93d652e86279719R143-R146).

src/datachain/lib/file.py
dreadatour: This method returns `File` object from local file. We need it to DRY when splitting video into frames/fragments (see examples). It is for local use only, since we support local fs as a storage, in real life we will use `upload` method.

dreadatour:
> @dreadatour is the PR description up to date?

It is now 👍

src/datachain/lib/file.py
shcheklein: so, extension determines it? I wonder if we need to clarify or will be kinda expected by end users 🤔

src/datachain/lib/file.py
shcheklein: can we use `upload` for the local storage as well? (it should be the same to my mind, same code that works in local and for storage)

src/datachain/lib/file.py
```diff
@@ -486,13 +490,223 @@ class ImageFile(File):
...
+    width: int = Field(default=-1)
+    height: int = Field(default=-1)
+    fps: float = Field(default=-1.0)
+    duration: float = Field(default=-1.0)
+    frames: int = Field(default=-1)
+    format: str = Field(default="")
+    codec: str = Field(default="")
+
+
+class Frame(DataModel):
```
shcheklein, line 700: I would suggest to rename those back to SomethingMeta if we keep this approach

it is super confusing - VideoFrame and Frame - which one is the main class?

shcheklein: cc @dmpetrov

src/datachain/lib/file.py
```diff
@@ -486,13 +490,223 @@ class ImageFile(File):
...
+    end: float = Field(default=-1.0)
+    orig: File
+
+
+class VideoFrame(VideoFile):
+    """`DataModel` for reading video frames."""
+
+    frame: int = Field(default=-1)
+    timestamp: float = Field(default=-1.0)
+    orig: File
```
shcheklein, line 685: what is `orig`?

shcheklein: I mean - do we really need it? Considering that is a subclass of a file already? Or should it be a subclass of an ImageFile? is orig then a link to the video? should it be VideoFile then? (here and above in VideoFragment)

src/datachain/lib/file.py
dreadatour: 🤔 Let me check 👍

src/datachain/lib/file.py
dreadatour: Since `VideoFrame` is an image file and `VideoFragment` is a video file fragment (clip), orig is a link to the video file from which these frame and fragment were taken. I took this from a @dmpetrov example.

I’m still not sure if we need this in the base library models. It looks like overhead in the general case, and if anyone needs it, they can easily add the original file either as a separate signal or by creating a custom model.

shcheklein: yep, let's remove it for now for simplicity

src/datachain/lib/file.py
dreadatour: Updated:
```
    def save_frame(
        self,
        frame: int,
        output_file: str,
        format: Optional[str] = None,
    ) -> "VideoFrame":
        """
        Saves video frame as an image file.

        Args:
            frame (int): Frame number to read.
            output_file (str): Output file path.
            format (str): Image format (default: use output file extension).

        Returns:
            VideoFrame: Video frame model.
        """
```

src/datachain/lib/file.py
dreadatour: Removed

dreadatour:
In this Video Models PR, we now have models for:
- `VideoFile` – based on the `File` model with additional methods added to work with video: get info, get frames, get fragments.
- `VideoFragment` – based on the `VideoFile` model with additional fields: start time and end time.
- `VideoFrame` – based on `ImageFile` with additional fields: frame number and timestamp.

There are some questions about this implementation:

1. When we are splitting `VideoFile` into fragments, we are uploading fragment video clips into storage and creating a new `VideoFragment` model with this uploaded video file. In this model, we have start time and end time signals, but there is no link by default to the original video file. Do we need this link (for example, an `orig` field with `VideoFile` type)? Do we always need this link? Do we need these signals (start and end time) without a link to the original video file?

2. Same for frames: when we are splitting `VideoFile` into frames, we are uploading frame images into storage and creating a new `VideoFrame` model with this uploaded image file. We do have frame and timestamp signals, but no link to the original video. Do we need these signals? Do we need the link to the original video?

3. What about virtual video fragments and lightweight frame models? It is an original video file model with additional signals (start and end time for fragments and timestamp for frames), but the original file is still the same—no physical file split and no upload happens. It looks like these virtual fragment and virtual frame models are required and will be used often, but we haven’t implemented them yet. Do we need them? How should we organize an API to work with these models? Do we need a set of additional methods in the `VideoFile` model class?

dreadatour:
Continue work in https://github.com/iterative/datachain/pull/890

PR #818, opened 2025-01-14
prefetching: remove prefetched item after use in udf

skshetry:
This PR removes the prefetched item after use in the UDF. This is enabled by default on `prefetch>0`, unless `cache=True` is set in the UDF, in which case the prefetched item is not removed.

For pytorch dataset, this is not enabled by default, but can be enabled by setting `remove_prefetched=True` in the `PytorchDataset` class. This is done so because the dataset can be used in multiple epochs, and removing the prefetched item after use can cause it to redownload again in the next epoch.

The exposed `remove_prefetched=True|False` setting could be renamed to
some better option. Feedbacks are welcome.
(I think the UX around prefetch needs to be improved and made obvious).

### DIFF
--- a/src/datachain/lib/pytorch.py
+++ b/src/datachain/lib/pytorch.py
@@ -50,6 +50,7 @@ class PytorchDataset(IterableDataset):
         tokenizer_kwargs: Optional[dict[str, Any]] = None,
         num_samples: int = 0,
         dc_settings: Optional[Settings] = None,
+        remove_prefetched: bool = False,
     ):
         """
         Pytorch IterableDataset that streams DataChain datasets.
@@ -84,6 +85,7 @@ class PytorchDataset(IterableDataset):
 
         self._cache = catalog.cache
         self._prefetch_cache: Optional[Cache] = None
+        self._remove_prefetched = remove_prefetched
         if prefetch and not self.cache:
             tmp_dir = catalog.cache.tmp_dir
             assert tmp_dir
@@ -147,7 +149,7 @@ class PytorchDataset(IterableDataset):
             rows,
             self.prefetch,
             download_cb=download_cb,
-            after_prefetch=download_cb.increment_file_count,
+            remove_prefetched=self._remove_prefetched,
         )
 
         with download_cb, closing(rows):
--- a/src/datachain/lib/udf.py
+++ b/src/datachain/lib/udf.py
@@ -16,6 +16,7 @@ from datachain.lib.convert.flatten import flatten
 from datachain.lib.data_model import DataValue
 from datachain.lib.file import File
 from datachain.lib.utils import AbstractUDF, DataChainError, DataChainParamsError
+from datachain.progress import CombinedDownloadCallback
 from datachain.query.batch import (
     Batch,
     BatchingStrategy,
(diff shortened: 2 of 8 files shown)


src/datachain/lib/pytorch.py
```diff
@@ -50,6 +50,7 @@ def __init__(
         tokenizer_kwargs: Optional[dict[str, Any]] = None,
         num_samples: int = 0,
         dc_settings: Optional[Settings] = None,
+        remove_prefetched: bool = False,
```
skshetry, line 53: I have added this setting for `PytorchDataset`, so prefetched items won't be removed after use by default. This is done so because this dataset could be used in multiple epochs, so I don't think it should be enabled by default.

src/datachain/lib/udf.py
```diff
@@ -301,20 +302,42 @@ async def _prefetch_input(
...
-        prepared_inputs = AsyncMapper(f, prepared_inputs, workers=prefetch).iterate()  # type: ignore[assignment]
-    yield from prepared_inputs
+    if not prefetch:
+        yield from prepared_inputs
+        return
+
+    if after_prefetch is None:
+        after_prefetch = noop
+        if isinstance(download_cb, CombinedDownloadCallback):
+            after_prefetch = download_cb.increment_file_count
```
skshetry, line 330: Minor enhancement that will show no. of files downloaded in the progressbar. 
```console
Download: 575kB [00:01, 369kB/s, 153 files]
```

src/datachain/lib/udf.py
```diff
@@ -301,20 +302,42 @@ async def _prefetch_input(
...
+
+    f = partial(_prefetch_input, download_cb=download_cb, after_prefetch=after_prefetch)
+    mapper = AsyncMapper(f, prepared_inputs, workers=prefetch)
+    with closing(mapper.iterate()) as row_iter:
+        for row in row_iter:
+            try:
+                yield row  # type: ignore[misc]
+            finally:
+                if remove_prefetched:
+                    _remove_prefetched(row)
```
skshetry, line 340: We'll remove the prefetched items right after we `yield` it.

src/datachain/lib/udf.py
```diff
@@ -351,7 +374,13 @@ def _prepare_rows(udf_inputs) -> "abc.Generator[Sequence[Any], None, None]":
                     )
 
         prepared_inputs = _prepare_rows(udf_inputs)
-        prepared_inputs = _prefetch_inputs(prepared_inputs, self.prefetch)
+        prepared_inputs = _prefetch_inputs(
+            prepared_inputs,
+            self.prefetch,
+            download_cb=download_cb,
+            remove_prefetched=bool(self.prefetch) and not cache,
```
skshetry, line 381: The prefetched items are only removed when `prefetch` is greater than 0, and `cache` is not set in the config.

src/datachain/lib/pytorch.py
shcheklein: can it be determined by cache True / False though?

skshetry: `cache=True` will persist prefetched items to the cache. `PytorchDataset` can remove prefetched files on `.close()` or when it gets garbage collected. 

(I think we need a setting in Datachain itself for more control.)

src/datachain/lib/udf.py
```diff
@@ -429,15 +458,22 @@ def _prepare_rows(udf_inputs) -> "abc.Generator[Sequence[Any], None, None]":
                         row, udf_fields, catalog, cache, download_cb
                     )
 
+        def _process_row(row):
+            with safe_closing(self.process_safe(row)) as result_objs:
+                for result_obj in result_objs:
+                    udf_output = self._flatten_row(result_obj)
+                    yield dict(zip(self.signal_names, udf_output))
```
skshetry, line 465: We need to close the iterable properly. This is because the `process_safe()` could be a generator, and might still have a file open that prevents us from removing that file.

We cannot rely on garbage collection as it might happen at a later time, so we need to close it explicitly.

src/datachain/query/dataset.py
```diff
@@ -336,15 +336,16 @@ def process_udf_outputs(
...
-            ):
-                for row_chunk in batched(rows, batch_size):
-                    warehouse.insert_rows(udf_table, row_chunk)
-                rows.clear()
+        with safe_closing(udf_output):
+            for row in udf_output:
+                cb.relative_update()
+                rows.append(adjust_outputs(warehouse, row, udf_col_types))
+                if len(rows) >= batch_size or (
+                    len(rows) % 10 == 0 and psutil.virtual_memory().percent > 80
```
ilongin, line 344: Maybe we can put these exact numbers as constants at the top of the file? Also, would it be beneficial if we could configure this numbers in SaaS (through env variables) or you think it wouldn't make much of a difference of tweaking them?

ilongin (approved):
LGTM as long as it's been tested. Added just a small comment.

src/datachain/query/dataset.py
skshetry: FYI, the only change I made here is that I nested the existing block in a `with safe_closing(udf_output):` statement.

PR #819, opened 2025-01-15
file: support exporting files as a symlink

### LINKED ISSUE #807: Symlink support
### Description

Functionality that will allow the export file command to support symlinks. 
*Ivan knows more details about the case

skshetry:
Closes #807.

### Example Usage:

```python
dc.export_files("output_dir", link_type="symlink")
```

### DIFF
--- a/src/datachain/lib/dc.py
+++ b/src/datachain/lib/dc.py
@@ -11,6 +11,7 @@ from typing import (
     BinaryIO,
     Callable,
     ClassVar,
+    Literal,
     Optional,
     TypeVar,
     Union,
@@ -2415,11 +2416,22 @@ class DataChain:
     def export_files(
         self,
         output: str,
-        signal="file",
+        signal: str = "file",
         placement: FileExportPlacement = "fullpath",
         use_cache: bool = True,
+        link_type: Literal["copy", "symlink"] = "copy",
     ) -> None:
-        """Method that exports all files from chain to some folder."""
+        """Export files from a specified signal to a directory.
+
+        Args:
+            output: Path to the target directory for exporting files.
+            signal: Name of the signal to export files from.
+            placement: The method to use for naming exported files.
+                The possible values are: "filename", "etag", "fullpath", and "checksum".
+            use_cache: If `True`, cache the files before exporting.
+            link_type: Method to use for exporting files.
+                Falls back to `'copy'` if symlinking fails.
+        """
         if placement == "filename" and (
             self._query.distinct(pathfunc.name(C(f"{signal}__path"))).count()
             != self._query.count()
@@ -2427,7 +2439,7 @@ class DataChain:
             raise ValueError("Files with the same name found")
 
         for file in self.collect(signal):
-            file.export(output, placement, use_cache)  # type: ignore[union-attr]
+            file.export(output, placement, use_cache, link_type=link_type)  # type: ignore[union-attr]
 
     def shuffle(self) -> "Self":
         """Shuffle the rows of the chain deterministically."""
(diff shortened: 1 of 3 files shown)


src/datachain/lib/file.py
```diff
@@ -249,6 +265,13 @@ def export(
         dst_dir = os.path.dirname(dst)
         os.makedirs(dst_dir, exist_ok=True)
 
+        if link_type == "symlink":
+            try:
+                return self._symlink_to(dst)
+            except OSError as exc:
+                if exc.errno not in (errno.ENOTSUP, errno.EXDEV, errno.ENOSYS):
+                    raise
```
skshetry, line 273: What should we do in case of overwrite? 

Looks like `export` allows you to overwrite files over and over again, but this `_symlink_to` will fail with `FileExistsError`.

src/datachain/lib/file.py
```diff
@@ -236,11 +237,26 @@ def save(self, destination: str):
...
 
+    def _symlink_to(self, destination: str):
+        if self.location:
+            raise OSError(errno.ENOTSUP, "Symlinking virtual file is not supported")
+
+        if self._caching_enabled:
+            self.ensure_cached()
+            source = self.get_local_path()
+            assert source, "File was not cached"
+        elif self.source.startswith("file://"):
```
skshetry, line 248: Using `self.startswith("file://")` to avoid creating a filesystem.

tests/unit/lib/test_file.py
```diff
@@ -379,3 +380,18 @@ def test_get_local_path(tmp_path, catalog):
...
+    path.write_text("some text")
+
+    file = File(path=path.name, source=f"file://{tmp_path}")
+    file._set_stream(catalog, use_cache)
+
+    file.export(tmp_path / "dir", link_type="symlink", use_cache=use_cache)
+    assert (tmp_path / "dir" / "myfile.txt").is_symlink()
+
+    dst = Path(file.get_local_path()) if use_cache else path
+    assert (tmp_path / "dir" / "myfile.txt").resolve() == dst
```
skshetry, line 397: Using `resolve()` instead of `readlink()` because `WindowsPath.readlink()` returns an extended path that fails to compare.

```diff
-- WindowsPath('C:/Users/runneradmin/AppData/Local/Temp/pytest-of-runneradmin/pytest-0/popen-gw2/default_working_dir0/.datachain/cache/29/89fe8cd7eda0a1e2e28bf837e137f5bdd8d20c506348f67f5671f5a036d529')
++ WindowsPath('//?/C:/Users/runneradmin/AppData/Local/Temp/pytest-of-runneradmin/pytest-0/popen-gw2/default_working_dir0/.datachain/cache/29/89fe8cd7eda0a1e2e28bf837e137f5bdd8d20c506348f67f5671f5a036d529')
```

src/datachain/lib/dc.py
```diff
@@ -2418,6 +2419,7 @@ def export_files(
         signal="file",
         placement: FileExportPlacement = "fullpath",
         use_cache: bool = True,
+        link_type: Literal["copy", "symlink"] = "copy",
```
shcheklein, line 2422: can we update / add docs while we do this (or as a followup)

shcheklein (approved):
while we on this - any ideas about a better name for the export_file? + it would be great to update docs. Thanks for a quick turnaround @skshetry

src/datachain/lib/dc.py
skshetry, line 2422: Another thing I forgot to mention - if files fail to symlink, it will always fall back to `copy`.

skshetry: done in 2d9547f

PR #822, opened 2025-01-15
Added full outer join

ilongin:
Adds full outer join support to our interface `DataChain.merge(...)` method and lower level `DatasetQuery.join(...)`.
Up until now we only had left outer join and inner join.

Note that since lower SQLite versions (< 3.39.0) doesn't support full outer joins out of the box there is a workaround with 2 left joins + union.

Example:
```python
merged  = dogs_cats.merge(cats, "file__path", full=True)
```


tests/func/test_dataset_query.py
```diff
@@ -708,6 +708,130 @@ def test_union(cloud_test_catalog, cats_dataset, dogs_dataset):
...
+    cloud_test_catalog, dogs_cats_dataset, cats_dataset, multiple_predicates
+):
+    catalog = cloud_test_catalog.catalog
+    string_default = String.default_value(catalog.warehouse.db.dialect)
+
+    dogs_cats = DatasetQuery(name=dogs_cats_dataset.name, version=1, catalog=catalog)
+    cats = DatasetQuery(name=cats_dataset.name, version=1, catalog=catalog)
+
+    if multiple_predicates:
+        predicates = sqlalchemy.and_(
```
shcheklein, line 763: can we do this w/o sqlalchemy?

tests/func/test_dataset_query.py
```diff
@@ -708,6 +708,130 @@ def test_union(cloud_test_catalog, cats_dataset, dogs_dataset):
...
+            ((r["file__path"], r["file__path_right"]) for r in res),
+            key=lambda x: x[0] if x[0] is not None else x[1],
+        )
+        == expected
+    )
+
+
+@pytest.mark.parametrize(
+    "cloud_type,version_aware",
+    [("s3", True)],
```
shcheklein, line 749: why only s3? can we run with files or even some in-memory values?

tests/unit/lib/test_datachain_merge.py
shcheklein, line 1: why is in lib, btw?

tests/func/test_dataset_query.py
```diff
@@ -708,6 +708,130 @@ def test_union(cloud_test_catalog, cats_dataset, dogs_dataset):
     assert count == 6
 
 
+@pytest.mark.parametrize(
+    "cloud_type,version_aware",
+    [("s3", True)],
+    indirect=True,
+)
+def test_full_outer_join_no_common_rows(cloud_test_catalog, dogs_dataset, cats_dataset):
```
shcheklein, line 716: why do we test in dataset query? seems these test should just hit datachain directly

tests/func/test_dataset_query.py
ilongin: This is internal API which works with SQLAlchemy. In our public API (`DataChain`) there is no SQLAlchemy constructs. This tests are for `DatasetQuery`

tests/func/test_dataset_query.py
ilongin: There are a lot of these tests with only `s3`. I think we put this as type of cloud doesn't really change anything in the test, it just slows it down.  In other hand we want to use existing fixtures so this is a workaround. Maybe we can create new fixtures that are only with `s3` to avoid repeating this all around. I would do it in separate task though.

tests/unit/lib/test_datachain_merge.py
ilongin: I'm guessing because it actually tests `DataChain` which is in `lib` but it was extracted to separate file from `tests/unit/lib/test_datachain.py` to decreasing number of lines in that "main" test file. 
We should move the whole `DataChain` codebase from `lib` and then it's tests respectively as well (I think we spoke about that couple of times already)

tests/func/test_dataset_query.py
ilongin: We still have "low level" tests that are testing `DatasetQuery` so I decided to put couple there as well. I've added tests for `DataChain` as well

tests/func/test_dataset_query.py
shcheklein: okay, why testing it this way though - an internal API, with internal details (like sqlalchemy)? It would be fine if it was a unit test at least - but it seems a functional one

tests/func/test_dataset_query.py
shcheklein: testing DatasetQuery 9with functional tests) seems excessive, I'm not sure I understand the reason tbh

tests/func/test_dataset_query.py
ilongin: I think the general issue here is having `DatasetQuery` class and it's corresponding unit tests in the first place. I would say those are unit tests, regardless if they are going to memory DB, as they are testing one particular method.
Anyway, i removed those for now and added one more case in unit tests for `DataChain` which now cover all the cases of those removed `DatasetQuery` ones so it should be fine. In future we need to rethink tests and does it make sense to have 2 levels of classes and tests.

tests/func/test_dataset_query.py
ilongin: Removed those tests as mentioned in [above comment](https://github.com/iterative/datachain/pull/822#discussion_r1920005881)

tests/func/test_dataset_query.py
ilongin: Removed those tests as mentioned in [above comment](https://github.com/iterative/datachain/pull/822#discussion_r1920005881)

tests/unit/lib/test_datachain_merge.py
```diff
@@ -77,7 +78,53 @@ def test_merge_objects(test_session):
...
+        ch = ch1.merge(
+            ch2,
+            ["emp.person.name", "emp.person.name"],
+            ["team.player", "team.player"],
+            full=True,
+        )
+    else:
+        ch = ch1.merge(ch2, "emp.person.name", "team.player", full=True)
+
+    str_default = String.default_value(test_session.catalog.warehouse.db.dialect)
```
shcheklein, line 98: are those value differ across implementations?

what happens then if I pull a dataset from Studio to sqlite?

ilongin: Yes, those are different. Actually, right now I'm working on a task to fix this as issue is on CH side https://github.com/iterative/studio/issues/11161 
If you pull dataset from studio, instead of `NULL` (or `None` in the code) it will have `""` for string, `0` for int etc.

shcheklein (approved):

PR #826, opened 2025-01-17
Second iteration of cli command help

amritghimire:
Reference document: https://docs.google.com/document/d/1_QeMQ1NsguHSRSyJpF2n-1s57SHSuiyGzl10q9d1UaE/edit?usp=sharing

- **Rename studio to auth for cli command**
- **Drop aws_endpoint_url option**
- **Fix sources argument for cp**
- **Fix anonymous arg help**
- **Update cached list of files for the sources**
- **Reorder verbose and quiet**
- **Path to a directory or file to put data to**

### DIFF
--- a/src/datachain/cli/__init__.py
+++ b/src/datachain/cli/__init__.py
@@ -73,7 +73,7 @@ def main(argv: Optional[list[str]] = None) -> int:
 
 def handle_command(args, catalog, client_config) -> int:
     """Handle the different CLI commands."""
-    from datachain.studio import process_jobs_args, process_studio_cli_args
+    from datachain.studio import process_auth_cli_args, process_jobs_args
 
     command_handlers = {
         "cp": lambda: handle_cp_command(args, catalog),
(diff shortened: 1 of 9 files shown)


amritghimire:
Updated the document with the latest help messages.

src/datachain/cli/__init__.py
```diff
@@ -48,7 +48,6 @@ def main(argv: Optional[list[str]] = None) -> int:
     logger.setLevel(logging_level)
 
     client_config = {
-        "aws_endpoint_url": args.aws_endpoint_url,
```
shcheklein, line 51: can we remove `aws_endpoint_url` now completely (from the client)? (can be a separate small PR)

shcheklein (approved):
Approving to merge and move forward ... I assume the document is now up to date

src/datachain/cli/__init__.py
amritghimire: Why would we want that? client config is essentially how we pass the credentials config to datachain client for all other remotes as well from studio.

PR #828, opened 2025-01-18
Fix list of tuples. Closes #827

### LINKED ISSUE #827: List of tuples with multiple values is not supported:
### Description

Code:
```python
def get_action_intervals(file: File) -> list[tuple[float, float]]:
    return [(3, 12), (65, 83)]
```

Error:
```
TypeError: Cannot resolve type 'tuple[float, float]' for flattening features
```

the error in from `python_to_sql.py`

### Version Info

```Text
0.8.3
Python 3.10.8
```

dmpetrov:


src/datachain/lib/convert/python_to_sql.py
```diff
@@ -52,7 +52,7 @@ def python_to_sql(typ):  # noqa: PLR0911
 
     args = get_args(typ)
     if inspect.isclass(orig) and (issubclass(list, orig) or issubclass(tuple, orig)):
-        if args is None or len(args) != 1:
+        if args is None:
             raise TypeError(f"Cannot resolve type '{typ}' for flattening features")
 
         args0 = args[0]
```
shcheklein, line 58: what if we have different types in tuple? (e.g. tuple[float, str])?

shcheklein (approved):

PR #832, opened 2025-01-20
move functional tests out of unit test suite

### LINKED ISSUE #829: Unit tests for diff take 25 seconds
### Description

Our unit test suite becoming supper slow (`lib` unit-test time takes 58 seconds!) and useless as a result (now I cannot run it for every change I do - which works just fine before).

We need to reduce it's unit-tests run-time significantly (5-10 times). 

I see multiple anti-patterns but I'm sure there are more:

1. Too many `pytest.mark.parametrize`

The test below generates 50+ tests without a good reason. For example - there is not need to test `save()` in diff test - tests have to be "orthogonal". It needs to be replaced by 7-15 smaller tests.

```python
@pytest.mark.parametrize("added", (True, False))
@pytest.mark.parametrize("deleted", (True, False))
@pytest.mark.parametrize("modified", (True, False))
@pytest.mark.parametrize("same", (True, False))
@pytest.mark.parametrize("status_col", ("diff", None))
@pytest.mark.parametrize("save", (True, False))
def test_compare(test_session, added, deleted, modified, same, status_col, save):
    ...
```

2. Touching cloud resources

It looks like `test_resolve_file()` and such require access to clouds.

```
E               pytest_servers.exceptions.RemoteUnavailable: gs: Failed to setup mock remote: Error while fetching server API version: ('Connection aborted.', FileNotFoundError(2, 'No such file or directory'))

```

### Version Info

```Text

```

mattseddon:
closes #829

This PR moves functional tests from `tests/unit` underneath `tests/func`. 

When combined with the changes in #831 the total run time of the `tests/unit/lib` folder is < 15 seconds (on my machine) as opposed to ~40 seconds for `main`.


0x2b3bfa0 (approved):

tests/func/test_data_storage.py
```diff
@@ -0,0 +1,160 @@
...
+
+    assert to_compare == expected
+
+
+@pytest.mark.parametrize(
+    "cloud_type,version_aware",
+    [("s3", True)],
+    indirect=True,
+)
+def test_convert_type(cloud_test_catalog):
```
skshetry, line 101: This test looks like an unittest. Unrelated, but this probably does not `s3` fixture.

tests/func/test_datachain.py
```diff
@@ -613,6 +631,36 @@ def name_len(name):
     assert count == 7
 
 
+@pytest.mark.xdist_group(name="tmpfile")
+def test_udf_parallel_boostrap(test_session_tmpfile):
```
skshetry, line 635: Same question here: what was the reason to move this to `func` tests?

tests/func/test_datachain.py
```diff
@@ -555,6 +556,23 @@ def test_mutate_existing_column(test_session):
     assert list(ds.order_by("ids").collect()) == [(2,), (3,), (4,)]
 
 
+@pytest.mark.parametrize("processes", [False, 2, True])
+@pytest.mark.xdist_group(name="tmpfile")
+def test_parallel(processes, test_session_tmpfile):
```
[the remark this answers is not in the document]
mattseddon: Run time, the fact it is write to disk and the group it is in

PR #833, opened 2025-01-20
memoize usearch.sqlite_path()

skshetry:
If `usearch` fails to download the extension, it will keep retrying in the future. This adds significant cost - for example, in `tests/func/test_pytorch.py` run, it was invoked 111 times, taking ~30 seconds in total.

This PR caches the return value for the whole session.

Review with [`?w=1`](https://github.com/iterative/datachain/pull/833/files?w=1) for better diff. :)


mattseddon (approved):

skshetry:
For the record, the failing "Studio tests" are also failing in the `main` branch. Merging ...

PR #842, opened 2025-01-21
e2e tests: limit name_len_slow to 3, split e2e tests from others

skshetry:
[The `test_query_e2e` takes almost ~8mins to run][1] (whole CI job takes 11 mins). The `name_len_slow` script is the main culprit, since it sleeps for 1 sec in each udf function and that mapper is run in a single process parallel mode.

```
474.21s call     tests/test_query_e2e.py::test_query_e2e@tmpfile
```

This commit adds a limit of 3 files to the name_len_slow script, which is enough, since it's only running a single process.
(We immediately interrupt the running process after seeing "UDF Processing Started" gets printed).

This also split tests into two: one for the e2e tests and one for the rest, so that these things are more obvious in the future.

[1]: https://github.com/iterative/datachain/actions/runs/12879531971/job/35907168617#step:8:82

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -112,7 +112,11 @@ jobs:
         run: echo 'DISABLE_REMOTES_ARG=--disable-remotes=azure,gs' >> $env:GITHUB_ENV
 
       - name: Run tests
-        run: nox -s tests-${{ matrix.pyv }} -- $DISABLE_REMOTES_ARG
+        run: nox -s tests-${{ matrix.pyv }} -- -m "not e2e and not examples" $DISABLE_REMOTES_ARG
+        shell: bash
+
+      - name: Run E2E tests
+        run: nox -s tests-${{ matrix.pyv }} -- -m "e2e" $DISABLE_REMOTES_ARG
         shell: bash
 
       - name: Upload coverage report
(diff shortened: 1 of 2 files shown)


mattseddon (approved):

.github/workflows/tests.yml
```diff
@@ -112,7 +112,11 @@ jobs:
         run: echo 'DISABLE_REMOTES_ARG=--disable-remotes=azure,gs' >> $env:GITHUB_ENV
 
       - name: Run tests
-        run: nox -s tests-${{ matrix.pyv }} -- $DISABLE_REMOTES_ARG
+        run: nox -s tests-${{ matrix.pyv }} -- -m "not e2e and not examples" $DISABLE_REMOTES_ARG
+        shell: bash
+
+      - name: Run E2E tests
+        run: nox -s tests-${{ matrix.pyv }} -- -m "e2e" $DISABLE_REMOTES_ARG
```
mattseddon, line 119: [Q] Do we get some benefit from splitting this out?

mattseddon: Will the tests be distributed amongst different workers?

skshetry: > [Q] Do we get some benefit from splitting this out?

I hope that this will make these issues obvious by splitting them.


> Will the tests be distributed amongst different workers?

Yes, it runs in the same way as the above test run.  PTAL at https://github.com/iterative/datachain/actions/runs/12880225335/job/35909694647#step:9:17 where this workflow is actually run (we use `pull_request_target` which is running workflow from the default branch).

PR #667, opened 2024-12-06
Allow testing external contributions using secrets

0x2b3bfa0:
Delightful self-plagiarism from https://datachain.ai/blog/testing-external-contributions-using-github-actions-secrets

PR #843, opened 2025-01-21
ci: switch trigger from `pull_request_target` to `pull_request`

skshetry:
The current setup with `pull_request_target` is unnecessary because tests are skipped when tokens are unavailable.

This currently affects HuggingFace-related tests.
Maintainers can help when the contributor's change affects those areas. This situation is no different from other tests that require `ANTHROPIC_API_KEY` or `OPENAPI_KEY`, or Studio tests that don't run in fork.

Using `pull_request_target` also prevents changes to workflows from being tested, as it always runs workflows from the default branch. This makes it difficult to modify workflows.
Switching to `pull_request` resolves this issue.

Also, see https://github.com/iterative/datachain/pull/667#issuecomment-2527821664.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -3,7 +3,7 @@ name: Tests
 on:
   push:
     branches: [main]
-  pull_request_target:
+  pull_request:
   workflow_dispatch:
 
 env:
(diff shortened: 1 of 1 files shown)


0x2b3bfa0 (approved):
If external contributions needing secrets for testing are infrequent, maybe. 🤷🏼‍♂️

skshetry:
> If external contributions needing secrets for testing are infrequent, maybe. 🤷🏼‍♂️

We are doing all this for just one single example test. I don't find it worth it.

PR #848, opened 2025-01-21
catch HfHubHTTPError in hf-dataset-llm-eval example

mattseddon:
Patching the error shown here: https://github.com/iterative/datachain/actions/runs/12888313233/job/35932939032

Failed here too: https://github.com/iterative/datachain/actions/runs/12899335202/job/35967939172?pr=850

### DIFF
--- a/examples/llm_and_nlp/hf-dataset-llm-eval.py
+++ b/examples/llm_and_nlp/hf-dataset-llm-eval.py
@@ -1,4 +1,5 @@
 from huggingface_hub import InferenceClient
+from requests import HTTPError
 
 from datachain import C, DataChain, DataModel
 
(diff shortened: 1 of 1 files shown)


examples/llm_and_nlp/hf-dataset-llm-eval.py
```diff
@@ -20,15 +21,18 @@ def eval_dialog(
...
+            messages=[
+                {
+                    "role": "user",
+                    "content": f"{PROMPT}\n\nUser: {user_input}\nBot: {bot_response}",
+                },
+            ],
+            response_format={"type": "json", "value": DialogEval.model_json_schema()},
+        )
+    except HTTPError:
+        return DialogEval(result="Error", reason="Huggingface API returned an error.")
```
dreadatour, line 35: Strictly saying it can also be not only an error from Huggingface API, but also general HTTP error (404 Not Found, for example), we may want to raise something like "Error while interacting with the Hugging Face API".

dreadatour (approved):
👍

PR #849, opened 2025-01-21
move functional test

mattseddon:
The moved test calls the below code under the hood and took ~2 seconds to run 

https://github.com/iterative/datachain/blob/d0c22fc0fa28112bf0b6e51b8757e7602247c269/src/datachain/lib/listing.py#L106-L128


mattseddon:
#846 addresses the same issue

PR #850, opened 2025-01-22
add upload classmethod to File

### LINKED ISSUE #771: File upload from File class
### Description

As the next step from #758, we need a class method to simplify the upload:

```python
data = somefile.read() # bypes type
path = "s3://mybucket/dir/file.mp4"
...
f = File.upload(data, path)
```

User should not deal with `client` or any other internals.

mattseddon:
Closes #771

The original example https://github.com/iterative/datachain/pull/758 from becomes:

```python


import io

import imageio
import imageio.v3 as iio
from ultralytics import YOLO

from datachain.model.ultralytics import YoloPoses

base_path = "gs://redacted"


def pose_estimation(
    yolo: YOLO, file: File
) -> Iterator[tuple[File, int, float, YoloPoses]]:
    stem = file.get_file_stem()
    ext = file.get_file_ext()

    reader = imageio.get_reader(io.BytesIO(file.read()), format=ext)
    fps = reader.get_meta_data()["fps"]

    for frame, img in enumerate(reader):
        filename = f"{base_path}/frames/{stem}_{frame:06d}.jpg"
        img_encoded = iio.imwrite("<bytes>", img, extension=".jpeg")
        f = File.upload(filename, img_encoded)

        timestamp = frame / fps
        results = yolo(img)

        yield f, frame, timestamp, YoloPoses.from_results(results)


(
    DataChain.from_storage(f"{base_path}/videos/1e28")
    .limit(1)
    .setup(
        yolo=lambda: YOLO("yolo11n-pose.pt"),
    )
    .gen(pose_estimation, output=("file", "frame", "timestamp", "poses"))
    .save("videos-frames-poses")
)

DataChain.from_dataset("videos-frames-poses").show()

```

### DIFF
--- a/src/datachain/client/fsspec.py
+++ b/src/datachain/client/fsspec.py
@@ -390,6 +390,10 @@ class Client(ABC):
 
     def upload(self, path: str, data: bytes) -> "File":
         full_path = self.get_full_path(path)
+
+        parent = "/".join(full_path.split("/")[:-1])
+        self.fs.makedirs(parent, exist_ok=True)
+
         self.fs.pipe_file(full_path, data)
         file_info = self.fs.info(full_path)
         return self.info_to_file(file_info, path)
--- a/src/datachain/lib/file.py
+++ b/src/datachain/lib/file.py
@@ -191,6 +191,24 @@ class File(DataModel):
         self._caching_enabled: bool = False
 
     @classmethod
+    def upload(
+        cls, path: str, data: bytes, catalog: Optional["Catalog"] = None
+    ) -> "File":
+        if catalog is None:
+            from datachain.catalog.loader import get_catalog
+
+            catalog = get_catalog()
+
+        parts = path.split("/")
+        parent = "/".join(parts[:-1])
+        name = parts[-1]
+
+        client = catalog.get_client(parent)
+        file = client.upload(name, data)
+        file._set_stream(catalog)
+        return file
+
+    @classmethod
     def _from_row(cls, row: "RowDict") -> "Self":
         return cls(**{key: row[key] for key in cls._datachain_column_types})
 
--- a/tests/func/test_file.py
+++ b/tests/func/test_file.py
@@ -43,3 +43,23 @@ def test_resolve_file_no_exist(cloud_test_catalog):
     assert resolved_non_existent.size == 0
     assert resolved_non_existent.etag == ""
     assert resolved_non_existent.last_modified == TIME_ZERO
+
+
+def test_upload(cloud_test_catalog):
+    ctc = cloud_test_catalog
+
+    src_uri = ctc.src_uri
+    filename = "image_1.jpg"
+    source = f"{src_uri}/upload-test-images"
+    catalog = ctc.catalog
+
+    img_bytes = b"bytes"
+
+    f = File.upload(f"{source}/{filename}", img_bytes, catalog)
+
+    assert f.path == filename
+    assert f.source == source
+    assert f.read() == img_bytes
+
+    client = catalog.get_client(src_uri)
+    client.fs.rm(source, recursive=True)


src/datachain/lib/file.py
```diff
@@ -190,6 +190,21 @@ def __init__(self, **kwargs):
         self._catalog = None
         self._caching_enabled: bool = False
 
+    @classmethod
+    def upload(
+        cls, path: str, data: bytes, catalog: Optional["Catalog"] = None
+    ) -> "File":
+        if catalog is None:
```
mattseddon, line 197: [F] main problems here are:

- we have to instantiate a new catalog for each upload
- same but for client

skshetry: That's okay, I think. This is what we are doing everywhere (check the `file.open()` for example).

Besides, initializing `Client` should be cheap - fsspec's filesystem instances get cached automatically.

This is what bothers me the most at the moment though, would be nice if these things were done lazily and only once.

https://github.com/iterative/datachain/blob/d0c22fc0fa28112bf0b6e51b8757e7602247c269/src/datachain/client/s3.py#L50-L56

shcheklein (approved):

shcheklein (commented):
What happens if we run DataChain with some client config (e.g. anon=True) within a custom session for example? This information will be lost inside the UDF, right? And it will be instantiating some default client ...

I like the simplicity of this API though ... I'm just worried that it won't be enough

may be though it's fine to ask to pass catalog / client and do upload with a custom instnace?

src/datachain/client/fsspec.py
```diff
@@ -390,6 +390,12 @@ def open_object(
 
     def upload(self, path: str, data: bytes) -> "File":
         full_path = self.get_full_path(path)
+
+        is_local = self.protocol == "file"
+        if is_local:
+            self.fs.auto_mkdir = True
+            self.fs.touch(full_path)
```
skshetry, line 397:
```suggestion
		self.fs.makedirs(exist_ok=True)
```

src/datachain/lib/file.py
```diff
@@ -190,6 +190,23 @@ def __init__(self, **kwargs):
...
+            from datachain.catalog.loader import get_catalog
+
+            catalog = get_catalog()
+
+        parent = "/".join(path.split("/")[:-1])
+        name = PurePosixPath(path).name
+
```
skshetry, line 204: Is this necessary?

tests/func/test_file.py
```diff
@@ -43,3 +46,27 @@ def test_resolve_file_no_exist(cloud_test_catalog):
...
+    src_uri = ctc.src_uri
+    filename = "image_1.jpg"
+    source = f"{src_uri}/upload-test-images"
+    catalog = ctc.catalog
+
+    image = Image.new(mode="RGB", size=(64, 64))
+    img_bytes = io.BytesIO()
+    image.save(img_bytes, format="JPEG")
+    img_bytes = img_bytes.getvalue()
+    assert img_bytes
```
skshetry, line 63: This _setup_ is not necessary for the `upload` test. 

```suggestion
    img_bytes = b"bytes"
```

src/datachain/lib/file.py
skshetry: Consider using `client.split_url()` in that case.

src/datachain/lib/file.py
```diff
@@ -190,6 +190,23 @@ def __init__(self, **kwargs):
         self._catalog = None
         self._caching_enabled: bool = False
 
+    @classmethod
+    def upload(
+        cls, path: str, data: bytes, catalog: Optional["Catalog"] = None
```
skshetry, line 195: I think `upload` should also support file-like objects.

tests/func/test_file.py
mattseddon: The reason that I used the elaborate setup was for the test to demonstrate the expected usage of the API.

src/datachain/lib/file.py
mattseddon: `client.split_url` does not give what I need here as the client has already been set up with the parent path. Docs:
```
Splits the URL into two pieces:

bucket name without protocol (everything up until the first /)
path which is the rest of URL starting from bucket name
e.g s3://my-bucket/animals/dogs -> (my-bucket, animals/dogs)
```
For the test that I've written we end up writing files to `f"{src_uri}/upload-test-images/upload-test-images/image_1.jpg"`

mattseddon:
> What happens if we run DataChain with some client config (e.g. anon=True) within a custom session for example? This information will be lost inside the UDF, right? And it will be instantiating some default client ...
> 
> I like the simplicity of this API though ... I'm just worried that it won't be enough
> 
> may be though it's fine to ask to pass catalog / client and do upload with a custom instnace?

I put the catalog as an optional instance on the classmethod. I thought the point here was to hide those details for the simplest case.

src/datachain/lib/file.py
```diff
@@ -190,6 +190,24 @@ def __init__(self, **kwargs):
...
+    ) -> "File":
+        if catalog is None:
+            from datachain.catalog.loader import get_catalog
+
+            catalog = get_catalog()
+
+        parts = path.split("/")
+        parent = "/".join(parts[:-1])
+        name = parts[-1]
```
skshetry, line 204: Please use `posixpath.split()` instead of doing this by yourself.

src/datachain/client/fsspec.py
```diff
@@ -390,6 +390,10 @@ def open_object(
 
     def upload(self, path: str, data: bytes) -> "File":
         full_path = self.get_full_path(path)
+
+        parent = "/".join(full_path.split("/")[:-1])
```
skshetry, line 394: Can we use `posixpath.dirname()` here?

src/datachain/lib/file.py
mattseddon: There are examples of `.split("/")` and `"/".join` throughout the codebase.

shcheklein: (don't know the exact context here)

> There are examples of .split("/") and "/".join throughout the codebase.

in some case it is safe (cloud URLs), in some cases it's not (e.g. local file system URL - it can be `/` or `\`)

PR #852, opened 2025-01-23
Remove default value checks

ilongin:
Companion issue of https://github.com/iterative/studio/pull/11236
With making CH accept nullable columns, we no longer need special checks for default values in tests, as they will be `None` always.


src/datachain/data_storage/sqlite.py
```diff
@@ -276,8 +276,9 @@ def has_table(self, name: str) -> bool:
         )
         return bool(next(self.execute(query))[0])
 
-    def create_table(self, table: "Table", if_not_exists: bool = True) -> None:
+    def create_table(self, table: "Table", if_not_exists: bool = True) -> "Table":
```
ilongin, line 279: Changing this as it makes code better in Studio part where table columns are modified to allow nullable

shcheklein (approved):

dreadatour (approved):
👍

skshetry:
Can we introduce something like `strict` mode in schema, similar to `pydantic` to support non-nullable types?

https://docs.pydantic.dev/latest/concepts/strict_mode/

ilongin:
> Can we introduce something like `strict` mode in schema, similar to `pydantic` to support non-nullable types?
> 
> https://docs.pydantic.dev/latest/concepts/strict_mode/

The thing here is actually this is not null vs no null, but making sure

> Can we introduce something like `strict` mode in schema, similar to `pydantic` to support non-nullable types?
> 
> https://docs.pydantic.dev/latest/concepts/strict_mode/

shcheklein:
@ilongin what is the status of this?

ilongin:
> @ilongin what is the status of this?

It's ready for merge but I know we needed some more discussions / clarification in one of our meetings so I wrote this last comment (and put a link in slack). I would like to hear another opinion / confirmation before merging it.

skshetry (approved):

PR #855, opened 2025-01-24
use posixpath in upload methods

mattseddon:
from comments in https://github.com/iterative/datachain/pull/850

### DIFF
--- a/src/datachain/lib/file.py
+++ b/src/datachain/lib/file.py
@@ -199,9 +199,7 @@ class File(DataModel):
 
             catalog = get_catalog()
 
-        parts = path.split("/")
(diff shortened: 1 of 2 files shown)


src/datachain/lib/file.py
```diff
@@ -199,8 +199,8 @@ def upload(
 
             catalog = get_catalog()
 
-        parts = path.split("/")
-        parent = "/".join(parts[:-1])
+        parts = posixpath.split(path)
+        parent = posixpath.join(*parts[:-1])
         name = parts[-1]
```
skshetry, line 204:
```suggestion
        parent, name = posixpath.split(path)
```

skshetry (approved):

Issue #600, opened 2024-11-14
Error handling with bucket access issues


dmpetrov:
### Description

I got a massive stackstrace while a single line is enough like:

```
studio-prod-worker@studio-hor.iam.gserviceaccount.com does not have storage.objects.list access to the Google Cloud Storage bucket. Permission 'storage.objects.list' denied on resource (or it may not exist).
```

DC should carefully handle these types of errors.

The stack trace:
```
Using cached virtualenv
Listing gs://mpii-human-pose: 0 objects [00:00, ? objects/s]
Processed: 1 rows [00:00, 11.23 rows/s] [00:00, ? objects/s]
Traceback (most recent call last):
  File "<string>", line 3, in <module>
  File "/tmp/local/datachain_venv/python3.12/default/lib/python3.12/site-packages/datachain/lib/dc.py", line 474, in from_storage
    .save(list_ds_name, listing=True)
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/local/datachain_venv/python3.12/default/lib/python3.12/site-packages/datachain/lib/dc.py", line 764, in save
    query=self._query.save(
          ^^^^^^^^^^^^^^^^^
  File "/tmp/local/datachain_venv/python3.12/default/lib/python3.12/site-packages/datachain/query/dataset.py", line 1579, in save
    query = self.apply_steps()
            ^^^^^^^^^^^^^^^^^^
  File "/tmp/local/datachain_venv/python3.12/default/lib/python3.12/site-packages/datachain/query/dataset.py", line 1128, in apply_steps
    result = step.apply(
             ^^^^^^^^^^^
  File "/tmp/local/datachain_venv/python3.12/default/lib/python3.12/site-packages/datachain/query/dataset.py", line 572, in apply
    self.populate_udf_table(udf_table, query)
  File "/tmp/local/datachain_venv/python3.12/default/lib/python3.12/site-packages/datachain/query/dataset.py", line 492, in populate_udf_table
    process_udf_outputs(
  File "/tmp/local/datachain_venv/python3.12/default/lib/python3.12/site-packages/datachain/query/dataset.py", line 335, in process_udf_outputs
    for row in udf_output:
               ^^^^^^^^^^
  File "/tmp/local/datachain_venv/python3.12/default/lib/python3.12/site-packages/datachain/lib/udf.py", line 369, in <genexpr>
    output = (dict(zip(self.signal_names, row)) for row in udf_outputs)
[...]

PR #856, opened 2025-01-24
Handle permission error properly when checking for file

amritghimire:
# main > #856 (this) > #857

Currently, we had blanket catch for exception when trying to check the
file using _isfile. As a result, the exception stacktrace was repeated
and catching the exception in script was difficult as we had to capture
different exception. This convert the error to datachain native error
that can be captured safely and proceed accordingly.

This is first step toward handling #600


shcheklein:
> Currently, we had blanket catch for exception when trying to check the
file using _isfile. As a result, the exception stacktrace was repeated
and catching the exception in script was difficult as we had to capture
different exception

I'm not sure I understand this ... could you give an example please?

amritghimire:
> > Currently, we had blanket catch for exception when trying to check the
> > file using _isfile. As a result, the exception stacktrace was repeated
> > and catching the exception in script was difficult as we had to capture
> > different exception
> 
> I'm not sure I understand this ... could you give an example please?

https://github.com/iterative/datachain/pull/856/files#diff-ba026f8084c104850332b7e62ae0741b251f65b67d0271ce5870f4794a1346bfL102

shcheklein:
> https://github.com/iterative/datachain/pull/856/files#diff-ba026f8084c104850332b7e62ae0741b251f65b67d0271ce5870f4794a1346bfL102

is it an example?

amritghimire:
> > [#856 (files)](https://github.com/iterative/datachain/pull/856/files#diff-ba026f8084c104850332b7e62ae0741b251f65b67d0271ce5870f4794a1346bfL102)
> 
> is it an example?

Yes, as you see we had 

```
except: 
``` 
with supressed lint. That means no matter what the error we were returning false. So, if we did not have the permission, the script proceeded.

amritghimire:
https://github.com/iterative/datachain/pull/767/files#diff-ba026f8084c104850332b7e62ae0741b251f65b67d0271ce5870f4794a1346bfR102-R103

This code looks problematic but uncaught because we just ignored everything with except. The _isfile doesn't work in windows for glob as a result as seen in https://github.com/iterative/datachain/actions/runs/12948522550/job/36117365039?pr=856

src/datachain/lib/listing.py
```diff
@@ -99,8 +104,13 @@ def _isfile(client: "Client", path: str) -> bool:
...
 
         return info["type"] == "file"
-    except:  # noqa: E722
+    except FileNotFoundError:
         return False
+    except REMOTE_ERRORS as e:
+        raise ClientError(
+            message=str(e),
+            error_code=getattr(e, "code", None),
+        ) from e
```
dreadatour, line 113: Do we want to catch other exceptions here? 🤔

dreadatour (approved):

shcheklein:
> This code looks problematic but uncaught because we just ignored everything with except. The _isfile doesn't work in windows for glob as a result as seen in https://github.com/iterative/datachain/actions/runs/12948522550/job/36117365039?pr=856

yep, but that's why I guess we were just ignoring a wide range of exceptions and the rest of the code is built to handle that path as a non-file

should we just try to catch Remote exceptions that are relevant to credentials, etc. And then do a wide open `except:`? wdyt?

src/datachain/lib/listing.py
```diff
@@ -90,6 +91,10 @@ def _isfile(client: "Client", path: str) -> bool:
     Returns True if uri points to a file
```
skshetry, line 91: @shcheklein, was `_isfile()` added just for making it work with Google Cloud?

If so, I'd recommend extending `GCSFileSystem._is_file` (my preference), or add `GCSClient._is_file`.

Whitelisting exceptions on a fs-agnostic/generic code is not a good thing to do.

src/datachain/lib/listing.py
```diff
@@ -90,6 +91,10 @@ def _isfile(client: "Client", path: str) -> bool:
...
     """
     try:
+        if "://" in path:
```
shcheklein, line 94: why is this needed, btw?

src/datachain/lib/listing.py
shcheklein: no, I think it is a general code (there are special zero-byte files across all clouds I think) - we need to filter them

skshetry: Btw, we stopped doing these checks in dvc. We ignore files that end with slashes.

shcheklein: yes, so that's the point of this call - to always treat it as a directory AFAIR

shcheklein: I guess in some cases (not sure if this works the same across all clients) it returns a `file/` even if ask it to check `file` if zero-byte object exists

src/datachain/lib/listing.py
amritghimire: This is because when the s3 or other schema was in uppercase, the process would fail since it didn't match the specified regex. Its just that we swallowed the exception in test cases previously.

amritghimire:
> should we just try to catch Remote exceptions that are relevant to credentials, etc. And then do a wide open `except:`? wdyt?

A wide open except: would swallow a lot of issues IMO.

src/datachain/lib/listing.py
amritghimire: I think that should be the burden of outer call or calling function to capture exception that are totally unexpected.

src/datachain/lib/listing.py
amritghimire: I think this warrants separate discussion if we want to include _isfile or not. Merging the PR for now.

src/datachain/lib/listing.py
skshetry: Could you point me to those tests that use uppercase scheme?

src/datachain/lib/listing.py
skshetry: The discussion is not limited to only `_isfile`.
I am arguing that `_is_file` should not raise `ClientError`, or handle certain set of exceptions. `_is_file` should raise any exception that is not `FileNotFoundError`, and `get_listing` should convert it to `ClientError` without any filtering. There's no need for whitelisting.

skshetry: `_isfile` should be very close to `fs._is_file`.

src/datachain/lib/listing.py
amritghimire: https://github.com/iterative/datachain/blob/60aed77e41e88c44117568011d21aabdbf0176f7/tests/func/test_ls.py#L78

shcheklein: this still very weird place to do this kind of transformation ... how was it handled by ls downstream? Okay, it's swallowed here means we are losing some optimization - by `ls` still worked right? where do we handle this usually?

src/datachain/lib/listing.py
shcheklein: I guess in this case we can consider is_file more like `client.is_file` that already abstracts out `fs` - thus it is probably fine to convert exceptions if that is relevant

src/datachain/lib/listing.py
```diff
@@ -90,6 +92,15 @@ def _isfile(client: "Client", path: str) -> bool:
...
+            # This makes sure that the uppercase scheme is converted to lowercase
+            scheme, path = path.split("://", 1)
+            path = f"{scheme.lower()}://{path}"
+
+        if os.name == "nt" and "*" in path:
+            # On Windows, the glob pattern "*" is not supported
+            return False
```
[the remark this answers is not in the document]
amritghimire: `# On Windows, the glob pattern "*" is not supported`

When you try any listing with glob pattern, it used to fail If I remember correctly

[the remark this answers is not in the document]
skshetry: That seems odd. Let me check.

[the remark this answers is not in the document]
skshetry: I have created https://github.com/iterative/datachain/pull/964. Let's see if the test passes.

PR #858, opened 2025-01-24
query: interrupt query script on SIGTERM

skshetry:
This now supports gracefully terminating script on SIGTERM. Graceful shutdown is also done on Ctrl-C.

Also added logs.


shcheklein:
Is there a way to test this?

src/datachain/catalog/catalog.py
```diff
@@ -1506,13 +1550,50 @@ def query(
...
+                    send_interrupt,
+                    interrupt_timeout=interrupt_timeout,
+                    terminate_timeout=terminate_timeout,
+                )
+                if proc.returncode:
+                    raise QueryScriptCancelError(
+                        "Query script was canceled by user", return_code=proc.returncode
+                    ) from exc
+            finally:
+                if orig_handler is not None:
```
shcheklein, line 1591: it looks like we are triggering it twice?

skshetry: I am trying to reinstate old handler as soon as possible. It's okay if it gets called twice - there can only be one signal handler at a time for a given signal.

Anyway, I need to double-check all the logic here, and also think of what happens when someone resends SIGTERM/SIGINT.

skshetry:
> Is there a way to test this?

I don't see an easy way to test this unfortunately. I don't want to write some heavy tests, and simple mock tests are going to be useful.

src/datachain/cli/commands/query.py
```diff
@@ -33,6 +34,8 @@ def query(
         params=params,
     )
 
+    # ignore SIGINT in the main process
+    signal.signal(signal.SIGINT, signal.SIG_IGN)
```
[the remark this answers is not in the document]
skshetry: Decided not to complicate `CLI` for now.

src/datachain/catalog/catalog.py
```diff
@@ -1506,13 +1549,45 @@ def query(
...
+
         with subprocess.Popen(cmd, env=env, **popen_kwargs) as proc:  # noqa: S603
-            if capture_output:
-                args = (proc.stdout, output_hook)
-                thread = Thread(target=_process_stream, args=args, daemon=True)
-                thread.start()
-                thread.join()  # wait for the reader thread
+            orig_sigint_handler = signal.getsignal(signal.SIGINT)
+            # ignore SIGINT in the main process
+            signal.signal(signal.SIGINT, signal.SIG_IGN)
```
skshetry, line 1558: We need two different behaviors from this `query` API that this has become complicated.

CLI should ideally ignore `SIGINT` and forward `SIGTERM` to the child.
However, Celery tasks should not ignore `SIGINT`, but consider `SIGTERM` as a special signal for graceful termination.

Here, I have made a compromise: to ignore `SIGINT` and use `SIGTERM` as a special signal for graceful termination everywhere. 

So, `SIGINT` still works for CLI (where it matters most), and `SIGTERM` will work for canceling celery tasks.

src/datachain/catalog/catalog.py
skshetry: We need this. We need to rollback the signal handlers before we try to gracefully shutdown. Otherwise, if users send `SIGTERM` and `SIGINT`, we won't be able to forcefully exit.

src/datachain/catalog/catalog.py
```diff
@@ -1506,13 +1549,45 @@ def query(
...
+                shutdown_process(proc, interrupt_timeout, terminate_timeout)
+                if proc.returncode:
+                    raise QueryScriptCancelError(
+                        "Query script was canceled by user", return_code=proc.returncode
+                    ) from exc
+            finally:
+                signal.signal(signal.SIGTERM, orig_sigterm_handler)
+                signal.signal(signal.SIGINT, orig_sigint_handler)
+                if thread:
+                    thread.join()  # wait for the reader thread
```
mattseddon, line 1588: [Q] Will this still run into deadlocks while we are waiting for https://github.com/tqdm/tqdm/pull/1649?

mattseddon (approved):
Looks good

mattseddon:
> Is there a way to test this?

Unfortunately, we need an e2e test in Studio. I can help with this.

dreadatour (approved):
Looks good to me.

src/datachain/catalog/catalog.py
```diff
@@ -97,6 +98,46 @@ def noop(_: str):
...
+        return f"{self.__class__.__name__}({self.signal})"
+
+
+if sys.platform == "win32":
+    SIGINT = signal.CTRL_C_EVENT
+else:
+    SIGINT = signal.SIGINT
+
+
+def shutdown_process(
```
shcheklein, line 116: unit test this?

src/datachain/catalog/catalog.py
```diff
@@ -1506,13 +1549,45 @@ def query(
...
-            if capture_output:
-                args = (proc.stdout, output_hook)
-                thread = Thread(target=_process_stream, args=args, daemon=True)
-                thread.start()
-                thread.join()  # wait for the reader thread
+            orig_sigint_handler = signal.getsignal(signal.SIGINT)
+            # ignore SIGINT in the main process
+            signal.signal(signal.SIGINT, signal.SIG_IGN)
+
+            orig_sigterm_handler = signal.getsignal(signal.SIGTERM)
```
shcheklein, line 1560: this should be happening _inside_ try (to avoid getting the TerminationSignal raised in between)

src/datachain/catalog/catalog.py
```diff
@@ -1506,13 +1549,45 @@ def query(
...
+            thread: Optional[Thread] = None
+            try:
+                if capture_output:
+                    args = (proc.stdout, output_hook)
+                    thread = Thread(target=_process_stream, args=args, daemon=True)
+                    thread.start()
+
+                proc.wait()
+            except TerminationSignal as exc:
+                signal.signal(signal.SIGTERM, orig_sigterm_handler)
```
shcheklein, line 1573: why do we need it here?

src/datachain/catalog/catalog.py
skshetry: There's no point in unittesting this. We'll check `mock_popen.terminate.assert_called_once()` which is not that useful.

tests/unit/test_query.py
```diff
@@ -63,3 +66,15 @@ def test_non_zero_exitcode(catalog, mock_popen):
         catalog.query("pass")
     assert e.value.return_code == 1
     assert "Query script exited with error code 1" in str(e.value)
+
+
+def test_shutdown_process_on_sigterm(mocker, catalog, mock_popen):
```
shcheklein, line 71: not sure tbh this is a very meaningful test :)  seems we are testing mocks primarily

src/datachain/catalog/catalog.py
skshetry: `TerminationSignal` is raised only after we set a signal handler, which happens after this line.

src/datachain/catalog/catalog.py
shcheklein: so, I'm a bit lost - are we ignoring SIGINT in CLI or not?

src/datachain/catalog/catalog.py
skshetry: PTAL at the comment above: https://github.com/iterative/datachain/pull/858#discussion_r1930597833.

src/datachain/catalog/catalog.py
shcheklein: I mean function test this - it should not be about testing "mocks" or course

src/datachain/catalog/catalog.py
shcheklein: so it can be raised right before `try`?

src/datachain/catalog/catalog.py
skshetry: Deadlock was happening when `fp.write()` was raising exception. That won't happen now that we use `SIGTERM` to cancel and wait.

But I can imagine, when the conditions are favorable, the `tqdm` may still deadlock. 
But the chances are slim. Eg: when you press `Ctrl-C` right at the time `display` runs, it may deadlock if there are other `tqdm` processes running in other thread.

src/datachain/catalog/catalog.py
shcheklein: (brining it here)

> We need this. We need to rollback the signal handlers before we try to gracefully shutdown. Otherwise, if users send SIGTERM and SIGINT, we won't be able to forcefully exit.

could you clarify please? what is the exact sequence of events / signals are we talking about?

src/datachain/catalog/catalog.py
skshetry: We are ignoring `Ctrl-C` in the main process. If you are running in terminal, OS (`tty` driver) will send the signal to all processes in the foreground process group, so the script will also get the signal. If we forward signals, then the subprocess will receive the signal twice.

If you are running a background process (e.g.: via a script), you'd have to interrupt the script subprocess rather than the CLI process.

src/datachain/catalog/catalog.py
skshetry: Theoretically yes, it could.

src/datachain/catalog/catalog.py
skshetry: `SIGTERM` then (`SIGTERM` or `SIGINT`)

src/datachain/catalog/catalog.py
shcheklein: so, then we should move it inside and catch and make sure that we handle if for example of the original handlers is not yet initialized, etc

src/datachain/catalog/catalog.py
shcheklein: what are the default handlers? ignore both? how setting the originals back helps in this scenario (AFAIU it is about sending signals two times in a row?)

skshetry: > what are the default handlers? ignore both? how setting the originals back helps in this scenario (AFAIU it is about sending signals two times in a row?)

For `SIGINT`, Python adds a default handler which raises `KeyboardInterrupt()`.
There is no `SIGTERM` default handler in Python.

But this is not just about defaults. Celery tasks could have their own signal handlers' setup, so we just defer to that after we handle a `SIGTERM`.

src/datachain/catalog/catalog.py
shcheklein: > will send the signal to all processes in the foreground process group

okay, I see

I think it would be helpful to describe these details in the comment for the next generations, it's a dense, low level code

src/datachain/catalog/catalog.py
shcheklein: yes, but it means we won't likely get to the shutdown right? and won't be able to `be able to forcefully exit`. Unless I'm missing something.

shcheklein: would it be more meaningful to set both to ignore here (while we run shutdown)?

skshetry: > yes, but it means we won't likely get to the shutdown right? and won't be able to `be able to forcefully exit`. Unless I'm missing something.

Yes, it won't get to the shutdown if the two signals come in row.
I am interpreting two SIGTERM signals to trigger an abrupt termination without waiting (which is the default).

We could also choose to `kill` the subprocess if we get the response again, and exit cleanly as you are suggesting.


Yes, if two `SIGTERM` signals come in a row, it will prevent the process from reaching a clean shutdown phase. I’m interpreting the second `SIGTERM` as a signal to trigger an immediate termination, bypassing any cleanup (which is the default behavior for `SIGTERM`).

We could also opt to send a `kill` signal to the subprocess and exit cleanly, as you suggested.

I can see both perspectives - why would you send two `SIGTERM` signals if the goal isn’t an abrupt termination? Or, if you want to force an immediate exit, wouldn’t a `SIGKILL` be more appropriate than a second `SIGTERM`?

shcheklein: okay, SIGTERM I guess is a bit special (humans won't send it, right?) how about SIGINT?

skshetry: > okay, SIGTERM I guess is a bit special (humans won't send it, right?) how about SIGINT?

`SIGTERM` then `SIGINT` has the same behaviour. It won't cleanup. 

If you are talking about `SIGINT` only (or, `SIGINT` in a row), in CLI, the script is going to receive those signals. Otherwise, it will ignore them.

shcheklein: > or, SIGINT in a row), in CLI,

yep, that's what I mean

> the script is also going to receive those signals

yep, makes sense if that is the case ... does it always work this way on all platforms / terminals?

skshetry: > yep, makes sense if that is the case ... does it always work this way on all platforms / terminals?

I believe so. To reiterate, this will only work in interactive usecase. Sending `SIGINT` through other means won't work.

Also, I see other projects like [uv](https://redirect.github.com/astral-sh/uv/pull/5395/) doing the same.

shcheklein: and probably that's fine, we expect this to be only terminal CLI use case I guess

tests/unit/test_query.py
skshetry: The main goal of this test here is to assert that `shutdown_process` was called.

skshetry:
> Unfortunately, we need an e2e test in Studio. I can help with this.

I have added a functional test for `SIGTERM` shutdown. It looks very complicated, and I don't see any simpler way to test this. Open to suggestions.

tests/unit/test_query.py
shcheklein: yep, I understand. Doesn't see to be very useful though. Or let me put it in a different way - I think we rarely test things this way (to check if function is called w/o testing logic in some way). Still seems to be not very meaningful to my mind. While we can do in this case most likely a proper test

src/datachain/catalog/catalog.py
shcheklein: why is this resolved @skshetry ?

skshetry: I moved code around. So, we go to `try .. except` right after we set the signal handlers.

tests/unit/test_query.py
skshetry: Handling signals is a tricky thing, especially when the main process needs to receive it (where it happens to be a test process for tests).

>  I think we rarely test things this way (to check if function is called w/o testing logic in some way

We do that in plenty of places in DVC, especially in CLI tests.

> While we can do in this case most likely a proper test

If you have some example unittest that we can replace this with (without crossing boundaries into the working of `shutdown_process`), I'd be very interested.
Again, my goal with this test is to make sure that the `shutdown_process` gets called with expected values. If you see a way to do that, I'll be happy to write it, or review it.

I did add a functional test, but it's complex, and you have to tread very carefully not to get main process blocked/waiting forever.

src/datachain/catalog/catalog.py
shcheklein: so, what happens if signal comes right before try?

skshetry: I don't think that can happen. When `CPython` receives a signal, it sets a flag to execute handler at a later point.

shcheklein: so, that later point can be exactly before try and after handlers changed

skshetry: It could be, but AFAIK it only happens on certain byte codes, not all.
I don't think we need to be paranoid. We can shift it if that happens.
Don't see much point of this discussion tbh.

shcheklein: it's a classical race condition to my mind ... it doesn't happen until it does and it will be very hard to debug or find a cause (someone will be spending a month of work later) In this particular case it can be done properly it seems w/o too much effort at this point unless I'm missing something

> but AFAIK it only happens on certain byte codes, not all.

this seems a way way too technical detail that I would rely (if it is true) only if we do some super fancy performant lock less algorithm or something

skshetry: In 3.13, I don't see a bytecode for `try .. except`.

```py
>>> import dis

>>> dis.dis(
...:     """
...: try:
...:     print("hello world")
...: finally:
...:     print("hello world")
...: """
...: )
   0           RESUME                   0

   2           NOP

   3   L1:     LOAD_NAME                0 (print)
               PUSH_NULL
               LOAD_CONST               0 ('hello world')
               CALL                     1
               POP_TOP

   5   L2:     LOAD_NAME                0 (print)
               PUSH_NULL
               LOAD_CONST               0 ('hello world')
               CALL                     1
               POP_TOP
               RETURN_CONST             1 (None)

  --   L3:     PUSH_EXC_INFO

   5           LOAD_NAME                0 (print)
               PUSH_NULL
               LOAD_CONST               0 ('hello world')
               CALL                     1
               POP_TOP
               RERAISE                  0

  --   L4:     COPY                     3
               POP_EXCEPT
               RERAISE                  1
```

skshetry: > it's a classical race condition to my mind ... it doesn't happen until it does and it will be very hard to debug or find a cause (someone will be spending a month of work later) In this particular case it can be done properly it seems w/o too much effort at this point unless I'm missing something

There are bigger problems than this with signal handling with Python (sometimes it can swallow signals), etc.

shcheklein: @skshetry we are going way to far as I mentioned. This type of analysis is wrong to my mind in this kind of situations and are needed only if we do low level optimizations. Otherwise it's better (unless there are some strong reasons) just write thread safe code w/o race conditions.

shcheklein: > There are bigger problems than this with signal handling with Python (sometimes it can swallow signals), etc.

may be, but it seems a bit irrelevant to this thread unless I'm missing something, right?

skshetry: What you are asking is to write an overly defensive code without there being an evidence for being a problem. There are issues with that too.

shcheklein: this a few lines code change, no? 

also, doing multithreading / multiprocessing is always hard exactly because lot of problems manifest themselves rarely but bite hard in terms of time someone can then spend debugging them later ... so, yes, if there are obvious improvements while it is still possible w/o obvious downsides I would always err on the side of doing safer code

skshetry: The best thing to do here is move `sigterm` handler out of this `with` loop.

skshetry: > also, doing multithreading / multiprocessing is always hard exactly because lot of problems manifest themselves rarely but bite hard in terms of time someone can then spend debugging them later ... so, yes, if there are obvious improvements while it is still possible w/o obvious downsides I would always err on the side of doing safer code

There is no `multithreading` or `multiprocessing` involved here. The main thread always gets a signal.

shcheklein: > The best thing to do here is move sigterm handler out of this with loop.

probably, and in `except` clause and in `finally` check if proc is setup, if original handlers are set

> There is no multithreading or multiprocessing involved here. The main thread always gets a signal.

there is to my mind. An external process sends a signal to a another one.

skshetry: https://github.com/iterative/datachain/pull/873.

PR #861, opened 2025-01-27
Third iteration of cli help message

amritghimire:
- **datachain auth help remove extra line**
- **Use anon and update only where needed.**
- **Grammer**
- **Add available scopes options**


amritghimire:
Updated the google docs as well. cc. @shcheklein

ilongin (approved):

amritghimire:
@ilongin What is the difference between clone and cp now? Also, is --no-cp flag default  or anything changed there?

shcheklein (approved):
thanks! let's keep going ...

PR #863, opened 2025-01-27
Speed up unit tests

ilongin:
Removing not needed `.save()` on various unit tests to speed up tests run.


dmpetrov:
Great! Please share perf metrics.

ilongin:
> Great! Please share perf metrics.

It's not much, as we don't actually have that many `.save()` calls in unit tests. I got 26s for unit tests with `.save()` and 23s for unit tests with removed `.save()`. so it's small improvement. We can take a look at other places we can speed up tests

skshetry (approved):

skshetry:
> It's not much, as we don't actually have that many `.save()` calls in unit tests. I got 26s for unit tests with `.save()` and 23s for unit tests with removed `.save()`. so it's small improvement. We can take a look at other places we can speed up tests

I run the tests via the following most of the time:
```console
pytest -nlogical --disable-remotes all -m "not e2e and not examples" --durations 0
```

And it takes <50 seconds, usually ~40 seconds. So, I think we are in a good place.
IMO next step should be to organize remote tests better, and rely less on them.
(It still takes 2-3 minutes for me to run all the tests, and remote tests are to blame).

ilongin:
> > It's not much, as we don't actually have that many `.save()` calls in unit tests. I got 26s for unit tests with `.save()` and 23s for unit tests with removed `.save()`. so it's small improvement. We can take a look at other places we can speed up tests
> 
> I run the tests via the following most of the time:
> 
> ```
> pytest -nlogical --disable-remotes all -m "not e2e and not examples" --durations 0
> ```
> 
> And it takes <50 seconds, usually ~40 seconds. So, I think we are in a good place. IMO next step should be to organize remote tests better, and rely less on them. (It still takes 2-3 minutes for me to run all the tests, and remote tests are to blame).

Yea, I agree. On too many tests we check for all clouds which is probably not needed. I will  create a followup issue to deal with that.

PR #864, opened 2025-01-27
tests: set anon=True in GCS tests for performance

skshetry:


skshetry:
There are two tests failing, but looks like overall this reduces total tests' runtime by 30s.

shcheklein:
need to be careful since not every tests expects / built to have anon access. E.g. `test_signed_url` and `test_signed_url_versioned` first come to my mind. Not sure how this change affects it.

shcheklein:
@skshetry is it still needed?

skshetry:
Seems to save 30-45 seconds of runtime. But I am not sure if it is worth it.

PR #869, opened 2025-01-28
Added ability for constant literals in `DataChain.mutate(...)`

ilongin:
Adding ability to do something like: 
```python
dc = dc.mutate(val=0.5)
```
This will create column `val` with constant value `0.5` for all rows.

For now it's implement for basic primitives like `int`, `bool`, `str` and `float`. Later on we can extend it to things like lists etc.

### DIFF
--- a/src/datachain/lib/dc.py
+++ b/src/datachain/lib/dc.py
@@ -1134,8 +1134,7 @@ class DataChain:
 
         for col_name, expr in kwargs.items():
             if (
-                not isinstance(expr, primitives)
-                and not isinstance(expr, (Column, Func))
+                not isinstance(expr, primitives + (Column, Func))
                 and isinstance(expr.type, NullType)
             ):
                 raise DataChainColumnError(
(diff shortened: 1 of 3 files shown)


tests/unit/test_func.py
```diff
@@ -425,6 +425,20 @@ def test_lt_mutate(dc):
     assert list(res) == [0, 0, 0, 0, 0]
 
 
+def test_mutate_with_literal(dc):
```
dreadatour, line 428: It looks like we can parametrize this test.

src/datachain/lib/dc.py
```diff
@@ -1129,8 +1130,14 @@ def mutate(self, **kwargs) -> "Self":
         )
         ```
         """
+        primitives = (bool, str, int, float)
+
         for col_name, expr in kwargs.items():
-            if not isinstance(expr, (Column, Func)) and isinstance(expr.type, NullType):
+            if (
+                not isinstance(expr, primitives)
+                and not isinstance(expr, (Column, Func))
```
dreadatour, line 1138:
```suggestion
                not isinstance(expr, primitives + (Column, Func))
```
🤔

dreadatour (approved):
👍

tests/unit/test_func.py
ilongin: Done!

PR #870, opened 2025-01-28
rearrange input params for upload methods

mattseddon:
This PR changes the `upload` methods to have data (from) first and then path (to) second as input parameters.

### DIFF
--- a/src/datachain/remote/studio.py
+++ b/src/datachain/remote/studio.py
@@ -375,7 +375,7 @@ class StudioClient:
             method="GET",
         )
 
-    def upload_file(self, file_name: str, content: bytes) -> Response[FileUploadData]:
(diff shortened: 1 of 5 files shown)


src/datachain/remote/studio.py
```diff
@@ -375,7 +375,7 @@ def dataset_export_status(
             method="GET",
         )
 
-    def upload_file(self, file_name: str, content: bytes) -> Response[FileUploadData]:
+    def upload_file(self, content: bytes, file_name: str) -> Response[FileUploadData]:
```
dreadatour, line 378:
```suggestion
    def upload_file(self, data: bytes, path: str) -> Response[FileUploadData]:
```
Same as above? 🤔

dreadatour (approved):
🔥

PR #872, opened 2025-01-28
Added columns as case values

ilongin:
Up until now we only allowed python primitives as case values like str, int, float ...
With this PR we will allow columns as well (note that this also adds ability for columns as outcomes in `ifelse` as it uses `case` under the hood ), e.g
```python
# case example
res = dc.mutate(test=case((C("num") < 3, C("val")), else_="default"))

# ifelse example
res = dc.mutate(test=ifelse(C("num") < 3, C("val"), "default"))
```


dreadatour (approved):
❤️

PR #873, opened 2025-01-29
move signal handler for sigterm

skshetry:
Review with [`?diff=split&w=1`](https://github.com/iterative/datachain/pull/873/files?diff=split&w=1) for smaller diff.

### DIFF
--- a/src/datachain/catalog/catalog.py
+++ b/src/datachain/catalog/catalog.py
@@ -1554,42 +1554,46 @@ class Catalog:
             raise TerminationSignal(sig)
 
         thread: Optional[Thread] = None
-        with subprocess.Popen(cmd, env=env, **popen_kwargs) as proc:  # noqa: S603
-            logger.info("Starting process %s", proc.pid)
+        orig_sigterm_handler = signal.getsignal(signal.SIGTERM)
+        orig_sigint_handler = signal.getsignal(signal.SIGINT)
+        signal.signal(signal.SIGTERM, raise_termination_signal)
+        try:
+            with subprocess.Popen(cmd, env=env, **popen_kwargs) as proc:  # noqa: S603
+                try:
+                    # ignore SIGINT in the main process.
+                    # In the terminal, SIGINTs are received by all the processes in
+                    # the foreground process group, so the script will
+                    # receive the signal too.
+                    # (If we forward the signal to the child, it will receive it twice.)
+                    signal.signal(signal.SIGINT, signal.SIG_IGN)
+                    logger.info("Starting process %s", proc.pid)
+                    if capture_output:
+                        args = (proc.stdout, output_hook)
+                        thread = Thread(target=_process_stream, args=args, daemon=True)
+                        thread.start()
 
-            orig_sigint_handler = signal.getsignal(signal.SIGINT)
-            # ignore SIGINT in the main process.
-            # In the terminal, SIGINTs are received by all the processes in
-            # the foreground process group, so the script will receive the signal too.
-            # (If we forward the signal to the child, it will receive it twice.)
-            signal.signal(signal.SIGINT, signal.SIG_IGN)
+                    proc.wait()
+                except (TerminationSignal, KeyboardInterrupt) as exc:
+                    # SIGINT is ignored, but in case it ever happens, let's handle it
+                    signal.signal(signal.SIGTERM, orig_sigterm_handler)
+                    signal.signal(signal.SIGINT, orig_sigint_handler)
 
-            orig_sigterm_handler = signal.getsignal(signal.SIGTERM)
-            signal.signal(signal.SIGTERM, raise_termination_signal)
-            try:
-                if capture_output:
-                    args = (proc.stdout, output_hook)
-                    thread = Thread(target=_process_stream, args=args, daemon=True)
-                    thread.start()
-
-                proc.wait()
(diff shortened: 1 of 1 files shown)


src/datachain/catalog/catalog.py
```diff
@@ -1554,19 +1554,20 @@ def raise_termination_signal(sig: int, _: Any) -> NoReturn:
...
 
-            orig_sigint_handler = signal.getsignal(signal.SIGINT)
-            # ignore SIGINT in the main process.
-            # In the terminal, SIGINTs are received by all the processes in
-            # the foreground process group, so the script will receive the signal too.
-            # (If we forward the signal to the child, it will receive it twice.)
-            signal.signal(signal.SIGINT, signal.SIG_IGN)
+        orig_sigint_handler = signal.getsignal(signal.SIGINT)
+        orig_sigterm_handler = signal.getsignal(signal.SIGTERM)
+        signal.signal(signal.SIGTERM, raise_termination_signal)
```
skshetry, line 1560: We don't need to handle `TerminationSignal` to arrive in `subprocess.Popen`. If that happens before we `Popen` gets created, we'll bubble up the exception and there is no `popen`/`process` to shutdown.

shcheklein: what happens if it comes right after `with subprocess.Popen(cmd, env=env, **popen_kwargs) as proc` ?

skshetry: In that case, from what I can tell, it will happen when running the code inside the `try .. except` block. (At the next bytecode).

But, could it happen within `subprocess.Popen.__enter__`? Probably, yes, it could, and `__exit__` would not ever be called.

See https://github.com/python/cpython/issues/75569.

Again, this signal handling is on the best effort basis. (Also this code does not really belong here, but on Studio, so I'd rather not complicate any further).

skshetry: Also found this: https://peps.python.org/pep-0419/#interruption-inside-with-statement-expression

skshetry:
Closing, I think what we have in the default branch is good-enough, which uses the following pattern:

```
  old_handler = signal.getsignal(signal.SIGINT)
    signal.signal(signal.SIGINT, signal.SIG_IGN)
    try:
        ... # Cleanup code
    finally:
        signal.signal(signal.SIGINT, old_handler)
```

which is what is recommended in suggested in https://github.com/python/cpython/issues/75569.

src/datachain/catalog/catalog.py
```diff
@@ -1554,42 +1554,46 @@ def raise_termination_signal(sig: int, _: Any) -> NoReturn:
...
-            finally:
-                signal.signal(signal.SIGTERM, orig_sigterm_handler)
-                signal.signal(signal.SIGINT, orig_sigint_handler)
-                if thread:
-                    thread.join()  # wait for the reader thread
+        orig_sigterm_handler = signal.getsignal(signal.SIGTERM)
+        orig_sigint_handler = signal.getsignal(signal.SIGINT)
+        signal.signal(signal.SIGTERM, raise_termination_signal)
+        try:
+            with subprocess.Popen(cmd, env=env, **popen_kwargs) as proc:  # noqa: S603
```
skshetry, line 1561: I did not ignore `SIGTERM` before creating a child process.

If we ignore it, that will also set to ignored in the child process.
There is `preexec_fn` that I could use though (that is only available in POSIX though).

skshetry: That is the reason why `SIGTERM` was/is inside the `with ...` block.

src/datachain/catalog/catalog.py
```diff
@@ -1554,42 +1554,46 @@ def raise_termination_signal(sig: int, _: Any) -> NoReturn:
...
+                    logger.info("Starting process %s", proc.pid)
+                    if capture_output:
+                        args = (proc.stdout, output_hook)
+                        thread = Thread(target=_process_stream, args=args, daemon=True)
+                        thread.start()
+
+                    proc.wait()
+                except (TerminationSignal, KeyboardInterrupt) as exc:
+                    # SIGINT is ignored, but in case it ever happens, let's handle it
+                    signal.signal(signal.SIGTERM, orig_sigterm_handler)
```
shcheklein, line 1578: why do we need to restore them here?

skshetry: See https://github.com/iterative/datachain/pull/858#discussion_r1929552009 and https://github.com/iterative/datachain/pull/858#discussion_r1931425565.

shcheklein: Bringing it here

> We need this. We need to rollback the signal handlers before we try to gracefully shutdown. Otherwise, if users send SIGTERM and SIGINT, we won't be able to forcefully exit.

src/datachain/catalog/catalog.py
```diff
@@ -1554,42 +1554,46 @@ def raise_termination_signal(sig: int, _: Any) -> NoReturn:
...
+                    signal.signal(signal.SIGINT, orig_sigint_handler)
+
+                    logging.info("Shutting down process %s, received %r", proc.pid, exc)
```
skshetry, line 1581:
```suggestion
                    logger.info("Shutting down process %s, received %r", proc.pid, exc)
```

PR #875, opened 2025-01-31
do not call stats in ds pull (from Studio)

mattseddon:
Related to https://github.com/iterative/studio/issues/11283

We call Studio to get the stats for a particular dataset version but we've already retrieved the same information in an earlier call. This PR removes the dataset stats call from pull.

### DIFF
--- a/tests/func/test_pull.py
+++ b/tests/func/test_pull.py
@@ -155,14 +155,6 @@ def remote_dataset_info(requests_mock, remote_dataset):
 
 
 @pytest.fixture
-def remote_dataset_stats(requests_mock):
-    requests_mock.get(
-        f"{STUDIO_URL}/api/datachain/datasets/stats",
-        json={"num_objects": 5, "size": 1000},
-    )
-
-
-@pytest.fixture
 def dataset_export(requests_mock, remote_dataset_chunk_url):
     requests_mock.get(
         f"{STUDIO_URL}/api/datachain/datasets/export", json=[remote_dataset_chunk_url]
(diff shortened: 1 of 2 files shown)


tests/func/test_pull.py
```diff
@@ -322,33 +313,13 @@ def test_pull_dataset_not_found_in_remote(
...
-    requests_mock.get(
-        f"{STUDIO_URL}/api/datachain/datasets/stats",
-        status_code=400,
-        json={"message": "Internal error"},
-    )
-    catalog = cloud_test_catalog.catalog
-
-    with pytest.raises(DataChainError) as exc_info:
-        catalog.pull_dataset("ds://dogs@v1")
-    assert str(exc_info.value) == "Error from server: Internal error"
```
shcheklein, line 341: what was that?

tests/func/test_pull.py
```diff
@@ -154,14 +154,6 @@ def remote_dataset_info(requests_mock, remote_dataset):
     requests_mock.get(f"{STUDIO_URL}/api/datachain/datasets/info", json=remote_dataset)
 
 
-@pytest.fixture
-def remote_dataset_stats(requests_mock):
-    requests_mock.get(
-        f"{STUDIO_URL}/api/datachain/datasets/stats",
```
shcheklein, line 160: do we use stats for anything else outside?

mattseddon: only calls were in this file, I think we should remove the route from Studio

tests/func/test_pull.py
mattseddon: I assume this test was for `error_on_fetching_stats` because the fixture was missing. We are no longer calling stats so we can delete the test.

shcheklein (approved):

PR #881, opened 2025-02-01
Removing `SQLAlchemy` from `DataChain.compare()`

ilongin:
Removing SQLAlchemy constructs from `compare()` function.

TODO:

- [x] Remove `or_` when https://github.com/iterative/datachain/pull/894 is done


src/datachain/diff/__init__.py
```diff
@@ -44,66 +44,49 @@ def _compare(  # noqa: PLR0912, PLR0915, C901
     modified: bool = True,
     same: bool = True,
     status_col: Optional[str] = None,
-) -> "DataChain":
+):
```
dreadatour, line 47: 😥

dreadatour (approved):
Quite a big change. Code looks good to me 👍 didn’t check if it works though 👀

PR #886, opened 2025-02-03
drop expected output from common sql functions output

mattseddon:
follow up to #860

it seems weird that we would manually maintain a text copy of the expected output


dreadatour:
> it seems weird that we would manually maintain a text copy of the expected output

🤔 Not really agree tbh.

It was useful to include this output in the example so people can see what’s happening without running the code. This is common practice.

The same applies to Jupyter Notebooks — we have them "baked" in https://github.com/iterative/datachain-examples. In case of any backward-incompatible changes in the Datachain repo, we need to re-run the corresponding example to generate the new output.

mattseddon:
> > it seems weird that we would manually maintain a text copy of the expected output
> 
> 🤔 Not really agree tbh.
> 
> It was useful to include this output in the example so people can see what’s happening without running the code. This is common practice.
> 
> The same applies to Jupyter Notebooks — we have them "baked" in https://github.com/iterative/datachain-examples. In case of any backward-incompatible changes in the Datachain repo, we need to re-run the corresponding example to generate the new output.

This isn't an option with Jupyter Notebooks and the process for keeping them updated is more straightforward than remembering to run and then copy/pasting the output into the same file.

PR #890, opened 2025-02-03
Video models (take 2)

dreadatour:
Alternative approach to implement video models based on [this comment](https://github.com/iterative/datachain/pull/814#issuecomment-2626366436). Looks much cleaner IMO.

## New `VideoFile` model

```python
class VideoFile(File):
    """`DataModel` for reading video files."""

    def get_info(self) -> "Video":
        """Returns video file information."""

    def get_frame(self, frame: int) -> "VideoFrame":
        """
        Returns VideoFrame model for a video frame.

        Args:
            frame (int): Frame number to read.

        Returns:
            VideoFrame: Video frame model.
        """

    def get_frames(
        self,
        start: int = 0,
        end: Optional[int] = None,
        step: int = 1,
    ) -> "Iterator[VideoFrame]":
        """
        Returns VideoFrame models for a video frame.

        Args:
            start (int): Frame number to start reading from (default: 0).
            end (Optional[int]): Frame number to stop reading at, non-inclusive
                                 (default: None, read until the end).
            step (int): Step size for reading frames (default: 1).

        Returns:
            Iterator[VideoFrame]: List of video frame models.

        Note:
            If end is not specified, number of frames will be taken from the video file.
        """

    def get_fragment(self, start: float, end: float) -> "VideoFragment":
        """
        Returns VideoFragment model for a video interval.

        Args:
            start (float): Start time in seconds.
            end (float): End time in seconds.

        Returns:
            VideoFragment: Video fragment model.
        """

    def get_fragments(
        self,
        intervals: list[tuple[float, float]],
    ) -> "Iterator[VideoFragment]":
        """
        Returns VideoFragment models for video intervals.

        Args:
            intervals (list[tuple[float, float]]): List of start and end times
                                                   in seconds.

        Returns:
            Iterator[VideoFragment]: List of video fragment models.
        """
```

## New `VideoFrame` model

One can create `VideoFrame` without downloading video file, since it is "virtual" frame: original VideoFile + frame number.

If physical frame image is needed, call `save` method, which uploads frame image into storage and returns `ImageFile` new model.

API:

```python
class VideoFrame(VideoFile):
    """`DataModel` for reading video frames."""

    frame: int = Field(default=-1)

    def get_np(self) -> "ndarray":
        """
        Reads video frame from a video file and returns as numpy array.

        Returns:
            ndarray: Video frame.
        """

    def read_bytes(self, format: str = "jpg") -> bytes:
        """
        Reads video frame from a video file and returns as image bytes.

        Args:
            format (str): Image format (default: 'jpg').

        Returns:
            bytes: Video frame image as bytes.
        """

    def save(self, output: str, format: str = "jpg") -> "ImageFile":
        """
        Saves video frame as a new image file. If output is a remote path,
        the image file will be uploaded to the remote storage.

        Args:
            output (str): Output path, can be a local path or a remote path.
            format (str): Image format (default: 'jpg').

        Returns:
            ImageFile: Image file model.
        """
```

## New `VideoFragment` model

One can create `VideoFragment` without downloading video file, since it is "virtual" fragment: original video file + start/end timestamp.

If physical fragment video is needed, call `save` method, which uploads fragment video into storage and returns new `VideoFile` model.

API:

```python
class VideoFragment(VideoFile):
    """`DataModel` for reading video fragments."""

    start: float = Field(default=-1.0)
    end: float = Field(default=-1.0)

    def save(self, output: str, format: Optional[str] = None) -> "VideoFile":
        """
        Saves video interval as a new video file. If output is a remote path,
        the video file will be uploaded to the remote storage.

        Args:
            output (str): Output path, can be a local path or a remote path.
            format (Optional[str]): Output format (default: None). If not provided,
                                    the format will be inferred from the video fragment
                                    file extension.

        Returns:
            VideoFile: Video fragment model.
        """
```

## New `Video` model

Video file meta information.

```
class Video(DataModel):
    """`DataModel` for video file meta information."""

    width: int = Field(default=-1)
    height: int = Field(default=-1)
    fps: float = Field(default=-1.0)
    duration: float = Field(default=-1.0)
    frames: int = Field(default=-1)
    format: str = Field(default="")
```


src/datachain/lib/file.py
```diff
@@ -486,13 +489,205 @@ class ImageFile(File):
...
+class VideoFile(File):
+    """`DataModel` for reading video files."""
+
+    def get_info(self) -> "Video":
+        """Returns video file information."""
+        from .video import video_info
+
+        return video_info(self)
+
+    def get_frame(self, frame: int) -> "VideoFrame":
```
mattseddon, line 516: Minor, but should these be `to_` methods to match the DataChain class?

src/datachain/lib/video.py
```diff
@@ -0,0 +1,185 @@
...
+        Video: Video file information.
+    """
+    try:
+        probe = ffmpeg.probe(file.get_local_path())
+    except ffmpeg.Error as exc:
+        raise FileError(file, f"unable to probe video file: {exc.stderr}") from exc
+    except Exception as exc:
+        raise FileError(file, f"unable to probe video file: {exc}") from exc
+
+    if not probe:
```
mattseddon, line 45: [C] Could collapse these checks by giving probe a default (`or {}`) above

```python
probe = ffmpeg.probe(file.get_local_path()) or {}
...
all_streams = probe.get("streams")
video_format = probe.get("format")
if not all_streams or not video_format:
        raise FileError(file, "unable to probe video file")
```

mattseddon: and as the error is user-facing we might want to dumb it down a bit: "unable to extract metadata from video file" or something like that

src/datachain/lib/file.py
```diff
@@ -486,13 +489,205 @@ class ImageFile(File):
...
+        from .video import video_frame_np
+
+        return video_frame_np(self)
+
+    def read_bytes(self, format: str = "jpg") -> bytes:
+        """
+        Reads video frame from a video file and returns as image bytes.
+
+        Args:
+            format (str): Image format (default: 'jpg').
```
mattseddon, line 628: [Q] Should be optional str?

src/datachain/toolkit/ultralytics.py
```diff
@@ -0,0 +1,141 @@
+from typing import Union
```
mattseddon, line 1: [Q] How are these changes related to video models?

src/datachain/toolkit/video.py
```diff
@@ -0,0 +1,43 @@
...
+                .gen(frame=split_video_to_frames(step=5))
+                .save("frames")
+        )
+        ```
+    """
+
+    def _split_frames(file: VideoFile) -> Iterator[VideoFrame]:
+        yield from file.get_frames(start=start, end=end, step=step)
+
+    return _split_frames
```
mattseddon, line 43: would be good to have this under test

src/datachain/lib/video.py
```diff
@@ -0,0 +1,185 @@
...
+    all_streams = probe.get("streams")
+    video_format = probe.get("format")
+    if not all_streams or not video_format:
+        raise FileError(file, "unable to probe video file")
+
+    video_streams = [s for s in all_streams if s["codec_type"] == "video"]
+    if len(video_streams) == 0:
+        raise FileError(file, "no video streams found in video file")
+
+    video_stream = video_streams[0]
```
mattseddon, line 57: [Q] Why take the first one? Is there only ever one?

src/datachain/toolkit/ultralytics.py
mattseddon: I didn't review this file.

mattseddon (approved):

src/datachain/lib/video.py
dreadatour: Some video container formats (such as MPEG-4) support multiple video, audio, and subtitle streams. However, in practice, video files with multiple video streams are extremely rare. Most applications, including video players and streaming platforms, are not designed to handle multiple video streams because there is little practical use for them.

I have heard of a few specialized use cases, such as movies with different aspect ratios or sports event recordings featuring multiple camera angles. However, these are rare exceptions, and I have never encountered them in real-world scenarios. In most cases, it is much simpler to provide multiple separate video files and allow users to download or stream only the one they need.

src/datachain/toolkit/ultralytics.py
dreadatour: Moved to separate PR: https://github.com/iterative/datachain/pull/900

src/datachain/toolkit/video.py
dreadatour: Removed this.

PR #894, opened 2025-02-04
Adding explicit `or_` function

ilongin:
It's possible to use `|` for OR expressions, but  sometimes it's just more handy to have explicit `or_` function, just like sqlalchemy has.


dreadatour (approved):
❤️

PR #896, opened 2025-02-04
add upload xdist group to pytest suite

mattseddon:
Trying to mitigate errors like the following:

https://github.com/iterative/datachain/actions/runs/13128811251/job/36630011158#step:8:249


skshetry:
I think the error is happening due to https://github.com/apache/arrow/issues/31796.

PR #897, opened 2025-02-04
Improve datachain ds ls output

amritghimire:
This changes the way the datasets are listed in a CLI.

| Before | After |
|---|----|
|before|after|
|before|after|
|before|after|

New option:
-------

image

Closes https://github.com/iterative/studio/issues/11280

### DIFF
--- a/src/datachain/dataset.py
+++ b/src/datachain/dataset.py
@@ -628,6 +628,9 @@ class DatasetListRecord:
         self.versions.sort(key=lambda v: v.version)
         return self
 
+    def latest_version(self) -> DatasetListVersion:
+        return max(self.versions, key=lambda v: v.version)
+
     @property
     def is_bucket_listing(self) -> bool:
         """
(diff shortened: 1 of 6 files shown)


ilongin (approved):
LGTM

src/datachain/dataset.py
```diff
@@ -628,6 +628,9 @@ def merge_versions(self, other: "DatasetListRecord") -> "DatasetListRecord":
         self.versions.sort(key=lambda v: v.version)
         return self
 
+    def latest_version(self) -> DatasetListVersion:
+        return max(self.versions, key=lambda v: v.version)
```
dreadatour, line 632: Not sure if max version will always be the latest 🤔

dreadatour (approved):
Looks good to me! 👍
Thank you for such a beautiful description! ❤️

shcheklein:
@amritghimire does it accept a dataset name? e.g. "cat-and-dogs --versions"?

amritghimire:
> cat-and-dogs

Since it is generic ls, it doesnn't support dataset name. I can add an optional field for dataset name.

shcheklein:
> Since it is generic ls, it doesnn't support dataset name. I can add an optional field for dataset name.

yes, please. Can be a follow up.

PR #898, opened 2025-02-05
Add datasets filter for `datachain ds ls`

amritghimire:
With this change, a optional positional arg is added for datachain ds
ls.

Example output:
```
datachain ds ls temp
Name    Studio    Local
------  --------  -------
temp    v1        ✖
temp    v2        ✖
temp    v3        ✖
temp    v4        ✖
temp    v5        ✖
temp    v6        ✖
temp    v7        ✖
temp    v8        ✖
temp    v9        ✖
temp    v10       ✖
temp    v11       ✖
temp    v12       ✖
temp    v13       v13
```

### DIFF
--- a/src/datachain/cli/__init__.py
+++ b/src/datachain/cli/__init__.py
@@ -161,7 +161,7 @@ def handle_dataset_command(args, catalog):
             all=args.all,
             team=args.team,
             latest_only=not args.versions,
-            dataset_name=args.dataset_name,
+            name=args.name,
         ),
         "rm": lambda: rm_dataset(
             catalog,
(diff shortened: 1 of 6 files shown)


src/datachain/cli/__init__.py
```diff
@@ -161,6 +161,7 @@ def handle_dataset_command(args, catalog):
             all=args.all,
             team=args.team,
             latest_only=not args.versions,
+            dataset_name=args.dataset_name,
```
shcheklein, line 164: let's make it consistent with `rm`, `edit` (name vs dataset_name` (probably consider later updating pull?) - naming should be consistent and concise

shcheklein:
@amritghimire could you please add the exact command that you run?

amritghimire:
> @amritghimire could you please add the exact command that you run?

It is in PR description. 
```
datachain ds ls temp
```

where temp is the name of dataset.

PR #900, opened 2025-02-06
Visualize Ultralytics Yolo models

dreadatour:
Add function to visualize Yolo results.

Bounding boxes conversion will be added in separate PR to do not block video examples PR [here](https://github.com/iterative/datachain-examples/pull/28).

### TODO

- [ ] update docs (in or after #921)

### DIFF
--- a/examples/computer_vision/ultralytics-bbox.py
+++ b/examples/computer_vision/ultralytics-bbox.py
@@ -1,25 +1,37 @@
 import os
-
-os.environ["YOLO_VERBOSE"] = "false"
-
-
 from io import BytesIO
 
+from numpy import asarray
 from PIL import Image
 from ultralytics import YOLO
 
-from datachain import C, DataChain, File
+from datachain import DataChain, File
 from datachain.model.ultralytics import YoloBBoxes
+from datachain.toolkit.ultralytics import visualize_yolo
+
+OUTPUT_DIR = "output/bbox"
 
 
 def process_bboxes(yolo: YOLO, file: File) -> YoloBBoxes:
-    results = yolo(Image.open(BytesIO(file.read())))
-    return YoloBBoxes.from_results(results)
+    # read image
+    img = Image.open(BytesIO(file.read()))
 
+    # detect objects using YOLO model
+    results = yolo(img, verbose=False)
+    # convert results to YoloBBoxes signal
+    signal = YoloBBoxes.from_results(results)
+
+    # visualize results
+    img2 = visualize_yolo(asarray(img), signal)
+    img2.save(f"{OUTPUT_DIR}/{file.get_file_stem()}.jpg")
+
+    return signal
+
+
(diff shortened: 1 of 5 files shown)


src/datachain/toolkit/ultralytics.py
```diff
@@ -0,0 +1,141 @@
...
+from ultralytics.engine.results import Results
+
+from datachain.model.ultralytics.bbox import YoloBBox, YoloBBoxes
+from datachain.model.ultralytics.pose import YoloPose, YoloPoses
+from datachain.model.ultralytics.segment import YoloSegment, YoloSegments
+
+YoloSignal = Union[YoloBBox, YoloBBoxes, YoloPose, YoloPoses, YoloSegment, YoloSegments]
+
+
+def _signal_to_results(img: np.ndarray, signal: YoloSignal) -> Results:
```
ilongin, line 15: Maybe just add simple comment of what this function actually does, as it's hard to understand if you are not familiar with the domain

ilongin (approved):

shcheklein (approved):
we need some tests for this? I assume this is to make the tutorial simpler?

src/datachain/toolkit/ultralytics.py
dreadatour: Done 👍

dreadatour:
> we need some tests for this? I assume this is to make the tutorial simpler?

I am not sure how to test this 🤔 Still thinking.
I've updated examples to add ultralytics signals visualization: bboxes, poses and segments. We'll see if it fails since we run examples in CI.

examples/computer_vision/ultralytics-bbox.py
```diff
@@ -1,25 +1,37 @@
...
+from datachain.toolkit.ultralytics import visualize_yolo
+
+OUTPUT_DIR = "output/bbox"
 
 
 def process_bboxes(yolo: YOLO, file: File) -> YoloBBoxes:
-    results = yolo(Image.open(BytesIO(file.read())))
-    return YoloBBoxes.from_results(results)
+    # read image
+    img = Image.open(BytesIO(file.read()))
```
shcheklein, line 17: can we use ImageFile here?

dreadatour: Updated [here](https://github.com/iterative/datachain/pull/924/files#diff-31fc6b1b87fe919cc90d9c37eb5a2978edc44fbc3ed96e457dc51896dca240ca).

shcheklein: hmm, but can we use it here? :)

shcheklein: ah, I see

shcheklein (commented):
so, does this PR still need review?

dreadatour:
> so, does this PR still need review?

Not yet, I'll be back to this PR after https://github.com/iterative/datachain/pull/924

PR #901, opened 2025-02-06
Pull dataset from studio if not available locally

### LINKED ISSUE #874: Support Studio datasets in Python
### Description

Now AFAIU it is required to run `dc pull` (or clone? what is the difference @ilongin ?)  to be able then in code do `DataChain.from_dataset("name")`. 

Let's avoid this manual step and clone dataset automatically if token is set into the local DB (also clone the updated version). Consider remote datasets as an extension to local. 

That would simplify the workflow quite a lot.

Btw how do we handle version collisions atm @ilongin ?

amritghimire:
If the following case are met, this will pull dataset from Studio.
- User should be logged in to Studio.
- The dataset or version doesn't exist in local
- User has not pass fallback_to_remote=False to from_dataset.

In such case, this will pull the dataset from studio before continuing
further.

The test is added to check for such behavior.

Closes #874


src/datachain/query/dataset.py
```diff
@@ -1097,9 +1103,26 @@ def __init__(
...
+        try:
+            ds = self.catalog.get_dataset(name)
+            self.version = version or ds.latest_version
+            self.feature_schema = ds.get_version(self.version).feature_schema
+
+        except (DatasetNotFoundError, DatasetVersionNotFoundError):
+            if not studio:
+                raise
+
+            token = os.environ.get("DVC_STUDIO_TOKEN") or Config().read().get(
```
shcheklein, line 1115: can we move it somewhere / make a helper? it seems this should be a general logic that is used in many other places?

src/datachain/query/dataset.py
```diff
@@ -1112,6 +1135,21 @@ def __iter__(self):
     def __or__(self, other):
         return self.union(other)
 
+    def pull_dataset(self, name: str, version: Optional[int] = None) -> "DatasetRecord":
+        print("Dataset not found in local catalog, trying to get from studio")
```
shcheklein, line 1139: let's use logger here in debug mode? @skshetry what is your take?

amritghimire: I tried using logger with info here at beginning. But print seemed consistent with other similar messages. And also, we definitely want user to know we are trying to get it from Studio so that they can expect the delay in execution of the code.

src/datachain/query/dataset.py
```diff
@@ -1097,9 +1103,23 @@ def __init__(
...
+        try:
+            ds = self.catalog.get_dataset(name)
+            self.version = version or ds.latest_version
+            self.feature_schema = ds.get_version(self.version).feature_schema
+
+        except (DatasetNotFoundError, DatasetVersionNotFoundError):
+            if not studio:
+                raise
+
+            if not is_token_set():
```
ilongin, line 1115: Can we merge these 2 checks for studio and token in one line

src/datachain/query/dataset.py
```diff
@@ -1097,9 +1103,23 @@ def __init__(
         self.column_types: Optional[dict[str, Any]] = None
 
         self.name = name
-        ds = self.catalog.get_dataset(name)
-        self.version = version or ds.latest_version
-        self.feature_schema = ds.get_version(self.version).feature_schema
+        try:
```
ilongin, line 1106: I think this logic should be in `Catalog` instead. Here we are putting a lot of custom code related to how and when we pull datasets into `DatasetQuery` which should not be it's domain IMO.
We currently have method `Catalog.get_dataset_with_version_uuid(...)` so I would create similar method with signature:
```python
def get_dataset_with_version(name: str, version: int, studio=False) -> DatasetRecord
    ...
```
This method would ensure to pull dataset from studio if the one with specific name + version doesn't exist locally and `studio` flag is set to `True`. Otherwise it would raise an Exception.
Then we wouldn't even need this special `pull_dataset()` method here, we would just call `ds = self.catalog.get_dataset_with_version(...)` and that's it.

src/datachain/lib/dc.py
```diff
@@ -481,6 +481,7 @@ def from_dataset(
         version: Optional[int] = None,
         session: Optional[Session] = None,
         settings: Optional[dict] = None,
+        studio: bool = True,
```
ilongin, line 484: This flag is little bit ambiguous to me. If I'm user I wouldn't know if this means we fetch from studio first or locally? In any case, it's not clear that fetching from studio is just in case it doesn't exist locally.

ilongin (changes requested):

ilongin (approved):

PR #903, opened 2025-02-06
assume monkeypatch.chdir does not work in multiprocess spawn

mattseddon:


skshetry:
That test only started failing since yesterday, so maybe some dependency got updated?

skshetry:
It's also interesting that the test only fails on Python 3.12.

skshetry:
I am able to reproduce this with:
```console
nox -s tests-3.12 -- tests/func/test_query.py::test_shutdown_on_sigterm
```

skshetry:
`datamodel-code-generator` had a new release (`0.27.0`) that is now bringing more dependencies, including `anyio` (which comes from `httpx`). `anyio` is what is causing issues, but idk what.
I'll try looking into this on Monday.

We can pin to `datamodel-code-generator<0.27` temporarily.

https://github.com/koxudaxi/datamodel-code-generator/pull/2288#issuecomment-2642098256

shcheklein:
@skshetry 

> We can pin to datamodel-code-generator<0.27 temporarily.

sounds good, thanks

skshetry:
That has been fixed upstream, so we don't need to pin it. But I do need to look why it hangs with `anyio`.

PR #905, opened 2025-02-08
fix(cli): remove extra Error from the message

shcheklein:
Instead of:

```
(.venv) √ Projects/cattle-care-train % datachain ds pull ds://something
Error: Error from server: Dataset something not found
```

we now do:

```
(.venv) √ Projects/cattle-care-train % datachain ds pull ds://something
Error: Dataset something not found
```

I think this should be enough

### DIFF
--- a/tests/func/test_pull.py
+++ b/tests/func/test_pull.py
@@ -332,7 +332,7 @@ def test_pull_dataset_exporting_dataset_failed_in_remote(
 
     with pytest.raises(DataChainError) as exc_info:
         catalog.pull_dataset("ds://dogs@v1")
-    assert str(exc_info.value) == (f"Dataset export {export_status} in Studio")
+    assert str(exc_info.value) == f"Dataset export {export_status} in Studio"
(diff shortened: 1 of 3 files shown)


0x2b3bfa0 (approved):

tests/func/test_pull.py
```diff
@@ -332,9 +332,7 @@ def test_pull_dataset_exporting_dataset_failed_in_remote(
 
     with pytest.raises(DataChainError) as exc_info:
         catalog.pull_dataset("ds://dogs@v1")
-    assert str(exc_info.value) == (
-        f"Error from server: Dataset export {export_status} in Studio"
-    )
+    assert str(exc_info.value) == (f"Dataset export {export_status} in Studio")
```
skshetry, line 335:
```suggestion
    assert str(exc_info.value) == f"Dataset export {export_status} in Studio"
```

skshetry (approved):

PR #907, opened 2025-02-08
fix(cli, studio): don't retry on non request/http errors

shcheklein:
W/o this fix, if team is not set via env variable `DVC_STUDIO_TEAM` and / or DataChain config, `self.team` in Studio remote client raises a `DataChainError` (which is fine). What is not fine is that `_send_request` keeps retying with an exponential backoff on those.

Pretty much means that first time someone is trying to run something like `datachain ds pull ...` there are big chances to hit a long hang w/o any messages at all. Then if you are like to wait you can see that it just expect team to be set.

Also improves the messaging around `get_team` errors a bit.

## TODO:

- [ ] Add tests
- [x] Add logger to at least see what is going on with retries in debug mode.

### DIFF
--- a/src/datachain/remote/studio.py
+++ b/src/datachain/remote/studio.py
@@ -14,6 +14,7 @@ from typing import (
 from urllib.parse import urlparse, urlunparse
 
 import websockets
+from requests.exceptions import HTTPError, Timeout
 
 from datachain.config import Config
 from datachain.error import DataChainError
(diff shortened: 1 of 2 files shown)


0x2b3bfa0 (approved):

src/datachain/remote/studio.py
```diff
@@ -158,15 +159,14 @@ def _send_request_msgpack(
...
+    @retry_with_backoff(retries=3, errors=(HTTPError, Timeout))
     def _send_request(
         self, route: str, data: dict[str, Any], method: Optional[str] = "POST"
     ) -> Response[Any]:
         """
         Function that communicate Studio API.
         It will raise an exception, and try to retry, if 5xx status code is
-        returned, or if ConnectionError or Timeout exceptions are thrown from
-        requests lib
+        returned, or if Timeout exceptions is thrown from the requests lib
```
shcheklein, line 169: [C]: removed `ConnectionError` from the one we retry on. It is usually Proxy, or something SSL - don't think we can recover from those

PR #913, opened 2025-02-10
Inline script metadata

ilongin:
Adding class called `ScriptMeta` that knows how to parse script inline metadata (https://packaging.python.org/en/latest/specifications/inline-script-metadata/#inline-script-metadata) .

Data that we collect from it are:
* python version
* dependencies
* number of workers
* files
* params

With this, we can even avoid having special run script settings in Studio, as user can do everything in the script codebase and can seamlessly c/p script between local and Studio.

Example of script with inline metadata:
```python
# /// script
# requires-python = ">=3.12"
#
# dependencies = [
#   "pandas < 2.1.0",
#   "numpy == 1.26.4"
# ]
#
# [tools.datachain.workers]
# num_workers = 3
#
# [tools.datachain.files]
# image1 = "s3://ldb-public/image1.jpg"
# file1 = "s3://ldb-public/file.pdf"
#
# [tools.datachain.params]
# min_length_sec = 1
# cache = false
#
# ///
import sys
import pandas as pd
print(f"Python version: {sys.version_info}")
print(f"Pandas version: {pd.__version__}")
```

Example of parsing:
```python
meta = ScriptMeta.parse(script)
print(meta.python_version)
print(meta.get_param("cache"))
```


src/datachain/script_meta.py
```diff
@@ -0,0 +1,115 @@
...
+        return self.params.get(name)
+
+    def get_file(self, name: str) -> Any:
+        return self.files.get(name)
+
+    @staticmethod
+    def read_inline_meta(script: str) -> Optional[dict]:
+        """Converts inline script metadata to dict with all found data"""
+        regex = (
+            r"(?m)^# /// (?P<type>[a-zA-Z0-9-]+)$\s(?P<content>(^#(| .*)$\s)+)^# ///$"
```
dreadatour, line 79: I know this regex was taken from https://packaging.python.org/en/latest/specifications/inline-script-metadata/, but this is just an example and I would like to suggest to update this regexp:

```suggestion
            r"(?m)^# \/\/\/ (?P<type>[a-zA-Z0-9-]+)[ \t]*$\s(?P<content>(^#(| .*)$\s)+)^# \/\/\/[ \t]*$"
```

1. Escape the `/` symbol (it is ok in this case, but in general in regex it should be escaped)
2. Allow trailing spaces (and tabs) in first and last strings (where `///` goes)

Also I am not really sure about using `\s` as a newline matcher, I know `$\s` combination will only catches newline, but not space or tab, anyway, `[\r\n|\r|\n]` looks more strict to me:

```suggestion
            r"(?m)^# \/\/\/ (?P<type>[a-zA-Z0-9-]+)[ \t]*$[\r\n|\r|\n](?P<content>(^#(| .*)$[\r\n|\r|\n])+)^# \/\/\/[ \t]*$"
```

Also it is a good idea to use non-capturing groups if no capturing is needed, as it will make regex much faster. For example, in this case I would suggest to use `(?:| .*)` instead of `(| .*)`, one more group can be converted to non-capturing and the final regex will looks like this:

```suggestion
            r"(?m)^# \/\/\/ (?P<type>[a-zA-Z0-9-]+)[ \t]*$[\r\n|\r|\n](?P<content>(?:^#(?:| .*)$[\r\n|\r|\n])+)^# \/\/\/[ \t]*$"
```

This regex is more strict in terms of regex specifications, a bit more loose in terms of trailing newlines and faster in terms of performance.

Also we may want to use verbose mode (`re.VERBOSE`) to make it more readable, let me know if I can help with this.

dreadatour (approved):
Looks good to me! Only one comment above about regex 👀

dreadatour:
Oh, and tests fails because of:
```
ModuleNotFoundError: No module named 'tomllib'
```

src/datachain/script_meta.py
shcheklein: is there a regular parser (not regexp based) that we could use? complex regexs can be a can worms (not saying that this is the case here, but in general it can be tricky and sometimes leads to very bad performance)

dreadatour: > is there a regular parser (not regexp based) that we could use? complex regexs can be a can worms (not saying that this is the case here, but in general it can be tricky and sometimes leads to very bad performance)

I don't think so, in [docs](https://packaging.python.org/en/latest/specifications/inline-script-metadata/) they are suggesting to use regex, but we need to take a look 👀

ilongin: @dreadatour thanks for the suggestions, I've updated the regex as you mentioned.
@shcheklein I also haven't seen anything else, but I think it shouldn't be a problem as this is just to get the `toml` content itself from the comment and then parsing is done with toml library afterwards so regex will not change when we add more things to toml.

src/datachain/script_meta.py
```diff
@@ -0,0 +1,116 @@
...
+import tomli
+
+
+class ScriptMetaParsingError(Exception):
+    def __init__(self, message):
+        super().__init__(message)
+
+
+@dataclass
+class ScriptMeta:
```
shcheklein, line 14: @ilongin what is the long term plan for this class? 

can we put this information into an existing Query class / table?

ilongin: Long term plan is to use it for parsing env data from script itself with which we could get rid of that settings UI panel and user could c/p code easier between local and Studio.
I'm not sure what did you mean by putting information into existing Query class. Did you mean `DatasetQuery` ? I would keep it  like this and use it in `Catalog` when we run the scripts.

shcheklein:
> Yea, tomllib is added in python 3.11 version so this is failing in older versions. I'm now using tomli instead which has the same API and works the same.

can we avoid new deps at least in Python 3.11+ ? every dependency counts

src/datachain/script_meta.py
shcheklein: I see. Sorry, disregard the "put somewhere" part - it's not relevant for now. So, we won't use it atm, right? Should we postpone merging this until we have a clear feature that requires it?

> with which we could get rid of that settings UI panel and user could c/p code easier between local and Studio.

so, is the next step - using this info in Studio and / or CLI (is this code even needed to run it in CLI?) ... I'm just not sure if we should be merging code that we won't use

ilongin: This will be needed when / if we decide to replace Studio query settings with it, or make it as optional if it makes sense to have two ways to define settings in the first place. 
I think it needs to be in datachain codebase as currently we have method in `Catalog` to run query script which accepts all those env / settings data - now those arguments will be removed and parsed and read from script itself with this class.

We can postpone merge until we have final decision

src/datachain/script_meta.py
```diff
@@ -0,0 +1,121 @@
...
+    # we need tomli
+    import tomli as tomllib  # type: ignore[no-redef]
+
+
+class ScriptMetaParsingError(Exception):
+    def __init__(self, message):
+        super().__init__(message)
+
+
+@dataclass
```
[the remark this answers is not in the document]
ilongin: The thing is that I wanted to be as similar as official python naming for this which is Inline script metadata https://packaging.python.org/en/latest/specifications/inline-script-metadata/#inline-script-metadata  ... with this in mind `Metadata` sounds better than `Config` ... but I'm not 100% sure.

src/datachain/script_meta.py
```diff
@@ -0,0 +1,121 @@
...
+        self.num_workers = num_workers
+
+    def get_param(self, name: str) -> Any:
+        return self.params.get(name)
+
+    def get_file(self, name: str) -> Any:
+        return self.files.get(name)
+
+    @staticmethod
+    def read_inline_meta(script: str) -> Optional[dict]:
```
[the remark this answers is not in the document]
ilongin: Changed it to only `def read(...)`

src/datachain/script_meta.py
```diff
@@ -0,0 +1,121 @@
...
+        params: Optional[dict[str, Any]] = None,
+        num_workers: Optional[int] = None,
+    ):
+        self.python_version = python_version
+        self.dependencies = dependencies or []
+        self.files = files or {}
+        self.params = params or {}
+        self.num_workers = num_workers
+
+    def get_param(self, name: str) -> Any:
```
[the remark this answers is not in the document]
ilongin: I'm not sure I understood 100% the need for always returning `str`. Isn't it beneficial if someone can set `false` as a parameter and get boolean `False` in the script?. Python lib for parsing TOML does automatic conversion for basic types: https://docs.python.org/3/library/tomllib.html#conversion-table

src/datachain/script_meta.py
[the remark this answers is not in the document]
ilongin: Renamed to `ScriptConfig` for now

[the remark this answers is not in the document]
dmpetrov: We overuse term metadata because our domain, I'll try my best to avoid it here.

src/datachain/script_meta.py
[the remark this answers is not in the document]
dmpetrov: Because we cannot define type from ui and cli.

PR #918, opened 2025-02-11
Add documentation for new from_dataset behavior

amritghimire:
This adds missing parameter and behavior to the documentation.

Before:
image

After:
image

### DIFF
--- a/src/datachain/lib/dc.py
+++ b/src/datachain/lib/dc.py
@@ -481,7 +481,7 @@ class DataChain:
         version: Optional[int] = None,
         session: Optional[Session] = None,
         settings: Optional[dict] = None,
-        fallback_to_remote: bool = True,
+        fallback_to_studio: bool = True,
     ) -> "Self":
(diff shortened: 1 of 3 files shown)


src/datachain/lib/dc.py
```diff
@@ -484,10 +484,15 @@ def from_dataset(
         fallback_to_remote: bool = True,
```
skshetry, line 484: I thought we were trying to avoid "remote" terminology.

dreadatour (commented):
Should we also update "Example" section to add more examples? Like to add examples with different settings. Also it may be a good idea to describe some settings here, since it is a raw dictionary, it is not really clear, what kind of settings can I use here.

src/datachain/lib/dc.py
shcheklein: yes, agree, probably better to use `studio` ?

dreadatour (approved):
Thank you for updating the docs with examples ❤️

PR #919, opened 2025-02-11
Update models

dreadatour:
Add validation and tests for the following models:

- Bounding box
- Pose
- Segments

Add conversion to/from different bounding box formats, including normalized versions, for example:

```python
In [1]: from datachain.model.bbox import BBox

In [2]: BBox.from_coco([10, 20, 80, 60])
Out[2]: BBox(title='', coords=[10, 20, 90, 80])

In [3]: BBox.from_coco([0.1, 0.2, 0.8, 0.6], normalized_to=(320, 240))
Out[3]: BBox(title='', coords=[32, 48, 288, 192])

In [4]: BBox.from_list([10, 20, 90, 80]).to_yolo()
Out[4]: [50, 50, 80, 60]

In [5]: BBox.from_list([10, 20, 90, 80]).to_yolo_normalized((100, 100))
Out[5]: [0.5, 0.5, 0.8, 0.6000000000000001]
```

### TODO

- [x] update docs: #921


src/datachain/model/pose.py
```diff
@@ -19,22 +24,76 @@ class Pose(DataModel):
...
+
+        If the input coordinates are normalized (i.e., floats between 0 and 1),
+        they will be converted to absolute pixel values based on the provided
+        image size. The image size should be given as a tuple (width, height)
+        via the `normalized_to` argument.
+
+        Args:
+            points (Sequence[Sequence[float]]): The x and y coordinates
+                of the keypoints. List of 2 lists: x and y coordinates.
+            normalized_to (Sequence[int], optional): The reference image size
```
shcheklein, line 42: let's just to image size - all tools usually accept size, not `normalized_to` AFAIK

src/datachain/model/pose.py
```diff
@@ -19,22 +24,76 @@ class Pose(DataModel):
...
         assert all(
             isinstance(value, (int, float)) for value in [*points_x, *points_y]
         ), "Pose coordinates must be floats or integers."
+
+        if normalized_to is not None:
+            assert all(0 <= coord <= 1 for coord in [*points_x, *points_y]), (
+                "Normalized coordinates must be floats between 0 and 1."
+            )
+            width, height = validate_img_size(normalized_to)
+            points_x = [coord * width for coord in points_x]
```
shcheklein, line 67: can it be a vectorized operation with numpy?

src/datachain/model/pose.py
```diff
@@ -19,22 +24,76 @@ class Pose(DataModel):
...
+            points (Sequence[Sequence[float]]): The x and y coordinates
+                of the keypoints. List of 2 lists: x and y coordinates.
+            normalized_to (Sequence[int], optional): The reference image size
+                (width, height) for denormalizing the bounding box. If None (default),
+                the coordinates are assumed to be absolute pixel values.
+
+        Returns:
+            Pose: A Pose object.
+        """
+        assert isinstance(points, (tuple, list)), "Pose must be a list of 2 lists."
```
shcheklein, line 49: probably we should be raising proper ValueErrors here, not do asserts

src/datachain/model/pose.py
dreadatour: I was thinking about using `img_size` as the name of the parameter here, but then it might not be quite clear if we need to define `img_size` or not and what is the difference here. If `normalized_to` is not set, then `points` are absolute pixel coordinates, if `normalized_to` is set, then points are relative floats within `normalized_to` image size.

Other possible option might be to separate those methods, such as `from_list` takes points with absolute coordinates and no image size is required, and `from_list_normalized` takes relative positions and image size is required.

What do you think in terms of API?

shcheklein: okay, let me even step back - why do we need a such complicated method? that behaves differently depending on the input and the arguments passed? even from this discussion it seems it's too complicated wdyt?

src/datachain/model/pose.py
```diff
@@ -19,22 +24,76 @@ class Pose(DataModel):
...
+        if normalized_to is not None:
+            assert all(0 <= coord <= 1 for coord in [*points_x, *points_y]), (
+                "Normalized coordinates must be floats between 0 and 1."
+            )
+            width, height = validate_img_size(normalized_to)
+            points_x = [coord * width for coord in points_x]
+            points_y = [coord * height for coord in points_y]
+
         return Pose(
             x=[round(coord) for coord in points_x],
```
shcheklein, line 71: same, all these could be vectorized operations most likely

can we outsource it to some exiting lib btw? (I'm afraid that this seemingly simple code has a lot of complexity potentially ... also, AFAIR yolo has some utils and since we depend on it already should we use it?

src/datachain/model/pose.py
dreadatour: Oh, rhetorical question! 😅 Any function will behaves differently depending on the input and the arguments passed 🙃 Let me describe the way I was thinking.

In this case we have two options:

1. Split the function into two: `from_list` + `from_list_normalized` -> this means we should also have third function to follow the DRY principle, or drop all the validations.
2. Use one function `from_list` with additional argument (`normalized_to=...`)

In terms of API I do not really see the difference, but in terms of "complexity" there is a difference. It is absolutely the same in both cases, except for additional conversion (removing all checks for simplicity here):

```python
# validation

points_x, points_y = points

if normalized_to is not None:
    width, height = normalized_to
    points_x = [coord * width for coord in points_x]
    points_y = [coord * height for coord in points_y]

# processing
```

It is kind of the same but same time it is complicated and kind of subjective, I would say. I was considering to have two separate functions, but in the end I have decided to make a single one. I still have no strong opinion on this, and if you feel we need to separate these functions, I will be happy to update this PR ❤️

I also have this described in docstrings [here](https://github.com/iterative/datachain/pull/919/files/db9edbe9d26627b8eb0a00fa48beefbc2f09c17d#diff-0707deb137d9a1b72a157b25738ad783ba4a06386019f6a7e52c8c24f0929e30R34-R37). Which also might signalize this was a bad idea, but same time this exactly comment I think we should keep in separate function if we decided to split these.

src/datachain/model/pose.py
dreadatour: I will take a look at vectorized operations, thank you for the suggestion 🙏

> also, AFAIR yolo has some utils and since we depend on it already should we use it?

I don't want to import `ultralytics` here because user might not have it installed, and I am not sure we should keep it as an optional dependency. Here I am trying to keep these models dead simple and do not rely on additional optional libraries. These base models (`BBox, `Pose`, `Segment`) in general should works with any library, not only Yolo, that's why we do have all these conversions here.

**Note**

> These base models (`BBox, `Pose`, `Segment`) in general should works with any library, not only Yolo, that's why we do have all these conversions here.

Strictly saying `Pose` validation [requires exactly 17 points](https://github.com/iterative/datachain/pull/919/files/db9edbe9d26627b8eb0a00fa48beefbc2f09c17d#diff-0707deb137d9a1b72a157b25738ad783ba4a06386019f6a7e52c8c24f0929e30R55-R57), it is OK for Yolo but may not be OK for other pose models. I think I should remove this validation from here.

src/datachain/model/pose.py
shcheklein: > Any function will behaves differently depending on the input and the arguments passed 

in this context specifically - not every. All `from_` and `to_` are stable to my mind and have precise semantics of an output

> Split the function into two: from_list + from_list_normalized -> this means we should also have third function to follow the DRY principle, or drop all the validations.

okay, why do we need `from_list` at all?

it should be rather from_coco, from_yolo, from_albumentations ... 

or it can be from(array, type: Literal['coco', 'yolo', ....])

src/datachain/model/bbox.py
```diff
@@ -20,28 +30,258 @@ class BBox(DataModel):
...
+        If the input coordinates are normalized (i.e., floats between 0 and 1),
+        they will be converted to absolute pixel values based on the provided
+        image size. The image size should be given as a tuple (width, height)
+        via the `normalized_to` argument.
+
+        Args:
+            coords (Sequence[float]): The bounding box coordinates.
+            title (str, optional): The title or label for the bounding box.
+                Defaults to "".
+            normalized_to (Sequence[int], optional): The reference image size
```
shcheklein, line 55: so, does it mean we have the same model that can be normalize and not?

src/datachain/model/bbox.py
```diff
@@ -20,28 +30,258 @@ class BBox(DataModel):
...
+        Create a bounding box from coordinates in PASCAL VOC format.
+
+        PASCAL VOC format represents bounding boxes as [x_min, y_min, x_max, y_max],
+        where:
+            - (x_min, y_min) are the coordinates of the top-left corner.
+            - (x_max, y_max) are the coordinates of the bottom-right corner.
+
+        If the input coordinates are normalized (i.e., floats between 0 and 1),
+        they will be converted to absolute pixel values based on the provided
+        image size. The image size should be given as a tuple (width, height)
```
shcheklein, line 48: what is the image size here?

src/datachain/model/bbox.py
shcheklein: also, even here it is weird ... `normalized_to ... the reference image size`

src/datachain/model/bbox.py
```diff
@@ -20,28 +30,258 @@ class BBox(DataModel):
...
+                round(x_center - width / 2),
+                round(y_center - height / 2),
+                round(x_center + width / 2),
+                round(y_center + height / 2),
+            ],
+        )
+
+    def to_yolo(self) -> list[int]:
+        """
+        Convert the bounding box to YOLO format.
```
shcheklein, line 233: yolo is normalized by default, no? (at when you train something it asks it to be normalized) - let's refer to utils in albumentations lib to be sure about formats, naming, etc

src/datachain/model/utils.py
```diff
@@ -0,0 +1,76 @@
...
+def validate_bbox_normalized(
+    coords: Sequence[float], img_size: Sequence[int]
+) -> Sequence[float]:
+    """Validate the bounding box coordinates and normalize them to the image size."""
+    assert isinstance(coords, (tuple, list)), "Bounding box must be a tuple or list."
+    assert len(coords) == 4, "Bounding box must be a tuple or list of 4 coordinates."
+    assert all(isinstance(value, float) for value in coords), (
+        "Bounding box normalized coordinates must be floats."
+    )
+    assert all(0 <= value <= 1 for value in coords), (
```
shcheklein, line 37: how do bounding boxes behave on the edges (the same with poses) - is it true that it should be always <1 (since coodrinate starts with 0 and is always < image size)?

src/datachain/model/utils.py
```diff
@@ -0,0 +1,76 @@
...
+    assert isinstance(coords, (tuple, list)), "Coords must be a tuple or list."
+    assert len(coords) == 4, "Coords must be a tuple or list of 4 coordinates."
+    assert all(isinstance(value, int) for value in coords), (
+        "Coords must be a tuple or list of 4 ints."
+    )
+
+    width, height = validate_img_size(img_size)
+
+    assert (
+        0 <= coords[0] <= width
```
shcheklein, line 65: same question here - how do they behave on the edges?

dreadatour:
So many great questions 🤔
It looks like this approach failed. Please, take a look at the new one here #924 ?

src/datachain/model/pose.py
dreadatour: Just wonder if it will be faster for the list of 17 items 🤔

src/datachain/model/bbox.py
dreadatour: > yolo is normalized by default, no? 

Yes and no. It depends.

By default internally Yolo [stores result data](https://github.com/ultralytics/ultralytics/blob/main/ultralytics/engine/results.py#L1004) [not](https://github.com/ultralytics/ultralytics/blob/main/ultralytics/engine/results.py#L1062) [normalized](https://github.com/ultralytics/ultralytics/blob/main/ultralytics/engine/results.py#L1072) (in `[x1, y1, x2, y2]` format for bounding boxes, btw), also Yolo outputs is not normalized (see [here](https://github.com/ultralytics/ultralytics/blob/main/ultralytics/engine/results.py#L769)).

But model training requires bounding boxes to be in yolo format (`x_center, y_center, wigth, height]`, normalized).

So "yolo format is normalized" is for model training.

PR #922, opened 2025-02-12
Rename export_files to to_storage with cloud support

### LINKED ISSUE #859: Rename export_files() to to_storage()
### Description

`export_files()` needs to be renamed in order to get a consistent naming with `from_storages()`, `to_parque()`, `to_json()` etc. It's also need too support all params acceptable in `from_storage()` like cloud paths `s3://mybkt/dir1/`

- [ ] Rename to to_storage()
- [ ] Support clouds

Open question: should we do a bit opposite - rename `from_storage()` --> `from_files()`, `export_files()` --> `to_files()` in order to get a consistency with `from_parquet()`, `from_json()`. It seems like a cleaner way.

@iterative/datachain WDYT folks?

```python
def predict_dog(File: file) -> float:
   model = tf.keras.applications.ResNet50(weights='imagenet')
   img = tf.keras.preprocessing.image.load_img(io.BytesIO(file.read()), target_size=(224, 224))
   pred = model.predict(tf.expand_dims(tf.keras.preprocessing.image.img_to_array(img), axis=0))
   return sum(pred[0][151:269])

(
    DataChain
    .from_storage("s3://mybucket/data/*.jpg'")
    .map(dog_score=predict_dog)
    .filter(C("dog_score") > 0.96)
    .to_storage("s3://mybucket/out/dogs")
)
```

PS:
- it's ok to break compatibility with export_file since it always was a workaround

amritghimire:
This renames the export files to to storage adding the support for the
cloud destinations as well.

With this change, the following code will work:
```py

ds = DataChain.from_storage("az://amrit-test-az/image.png")
ds.save("az")

ds.to_storage("gs://amrit-datachain-test/destination", placement="filename")

```

Closes #859

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -68,7 +68,7 @@ high confidence scores.
 
     likely_cats = annotated.filter((Column("meta.inference.confidence") > 0.93) \
                                    & (Column("meta.inference.class_") == "cat"))
-    likely_cats.export_files("high-confidence-cats/", signal="file")
+    likely_cats.to_storage("high-confidence-cats/", signal="file")
 
 
 Example: LLM based text-file evaluation
@@ -109,7 +109,7 @@ Python code:
     )
 
     successful_chain = chain.filter(Column("is_success") == True)
-    successful_chain.export_files("./output_mistral")
+    successful_chain.to_storage("./output_mistral")
 
     print(f"{successful_chain.count()} files were exported")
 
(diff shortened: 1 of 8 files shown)


dreadatour (approved):
Looks good to me! ❤️

Just wonder if it will be a good idea to add example [to the docs](https://amrit-rename-export-files.datachain-documentation.pages.dev/references/datachain/#datachain.lib.dc.DataChain.to_storage)?

PR #923, opened 2025-02-13
Added `num_threads` to export files in parallel

ilongin:
Adding `num_threads` argument to `DataChain.to_storage(...)` function to speedup exporting which was done one by one up until now.

Performance changes (tested on **200** images from `s3://ldb-public/remote/data-lakes/dogs-and-cats/` and no cache):

Num Threads | 1 | 2 | 5 (default) | 10 | 15
--- | --- | --- | --- | --- | --- 
Time | 99s | 52s | 23s | 13s | 10.5s


ilongin:
> I like the simplicity of the implementation, @ilongin! However, prefetching uses a temporary cache when `cache=False`, so this solution might not be ideal in that case.

@skshetry can you explain why it's not ideal? I'm not sure I 100% understand the implication

skshetry:
> > I like the simplicity of the implementation, @ilongin! However, prefetching uses a temporary cache when `cache=False`, so this solution might not be ideal in that case.
> 
> @skshetry can you explain why it's not ideal? I'm not sure I 100% understand the implication

When `prefetch>0` and `cache=False`, prefetching uses a temporary cache.

https://github.com/iterative/datachain/blob/de171c6b0e33de3b40b1cf29c8543f11b1c02a86/src/datachain/lib/udf.py#L348-L349

So, all the files are prefetched to a temporary location in the background. So what this PR does in that case is that it saves files to a temporary location first before "exporting" it back out from that cache. We set `caching_enabled=True` so all the file operations will also try to use the cache. This is not ideal for performance.

https://github.com/iterative/datachain/blob/de171c6b0e33de3b40b1cf29c8543f11b1c02a86/src/datachain/lib/file.py#L325-L327

We also remove prefetched items after we run the mapper function too. And we remove the cache directory later. This change, as a result, also breaks the export of `symlink` link type.

https://github.com/iterative/datachain/blob/de171c6b0e33de3b40b1cf29c8543f11b1c02a86/src/datachain/lib/udf.py#L310-L313

Also, by "all the files" above, I mean we prefetch all the `File` signals, which might also be suboptimal.

https://github.com/iterative/datachain/blob/de171c6b0e33de3b40b1cf29c8543f11b1c02a86/src/datachain/lib/udf.py#L299-L301

ilongin:
@skshetry so if I understood correctly, there is only a problem when `prefetch=0`? ... if that's the true, in that case I can just return the old code (looping through the results of `.collect()` and exporting individual file), and for `prefetch > 0` it can stay new code... it won't be pretty as now but still better than introducing threads and all that boilerplate. wdyt?

dreadatour (commented):
Code looks good to me, couple comments below about naming/docstrings.
Not approving _yet_ since @skshetry have a reasonable comment above.

src/datachain/lib/dc.py
```diff
@@ -2451,6 +2451,7 @@ def export_files(
         placement: FileExportPlacement = "fullpath",
         use_cache: bool = True,
         link_type: Literal["copy", "symlink"] = "copy",
+        prefetch: Optional[int] = None,
```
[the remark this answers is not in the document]
ilongin: Changed to `num_workers` as it's the most correct name I think

src/datachain/lib/dc.py
```diff
@@ -2462,15 +2463,22 @@ def export_files(
             use_cache: If `True`, cache the files before exporting.
             link_type: Method to use for exporting files.
                 Falls back to `'copy'` if symlinking fails.
+            prefetch: number of workers to use for downloading files in advance.
+                      This is enabled by default and uses 2 workers.
+                      To disable prefetching, set it to 0.
```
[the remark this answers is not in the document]
ilongin: Fixed

ilongin:
@shcheklein @skshetry I've tried to use `AsyncMapper` instead of thread pool but it wasn't really working as it's needed to re-implement all of our methods in `File` to be async (simple wrapping with `to_thread()` doesn't work) and also there were some issues with SQLite as different threads are trying to use same connection.
Eventually I decided to implement cancelation to our thread pool wrapper class, which is basically the only thing that was missing from it comparing to `AsyncMapper`. IMO this is ok solution for now. Follow up task is to rename that `NodesThreadPool` to something more generic as we are using it in multiple places already, not only for listing / nodes.

skshetry:
> also there were some issues with SQLite as different threads tried to use the same connection.

The `AsynMapper` runs the producer in a different thread, so you'd need to clone the sqlite3 connection.

https://github.com/iterative/datachain/blob/df0ec95e6b3dcf4a3e82cc253f7cc96238345af9/src/datachain/query/dataset.py#L1265-L1268

> @shcheklein @skshetry I've tried to use `AsyncMapper` instead of thread pool but it wasn't really working as for it we need to re-implement all of our methods in `File` to be async (simple wrapping with `to_thread()` doesn't work)

That is true. That is the downside of using Async. Note, using `to_thread()` is not the right thing to do here, because we are doing this for performance and that will add significant overhead. Also, threads are not cancellable in Python.

TBH, I am not sure if we even need an `AsyncMapper` here (which runs multiple workers).
My intuition is that `await fs._get([lpaths], [rpaths])` on a single worker will be good-enough. (The only issue is that `fsspec` is not atomic, but our current implementation is not either).


> 
> Eventually I decided to implement cancelation to our thread pool wrapper class (`NodesThreadPool`), which is basically the only thing that was missing from it comparing to `AsyncMapper`. IMO this is ok solution for now. Follow up task is to rename that `NodesThreadPool` to something more generic as we are using it in multiple places already, not only for listing / nodes.

Okay, I don't have a strong opinion here. I'll let you and @shcheklein decide.

shcheklein:
@ilongin I'm fine if @skshetry is fine and you feel that it's a good option for now :).

src/datachain/lib/dc.py
```diff
@@ -2525,8 +2537,22 @@ def to_storage(
...
             raise ValueError("Files with the same name found")
 
-        for file in self.collect(signal):
-            file.export(output, placement, use_cache, link_type=link_type)  # type: ignore[union-attr]
+        progress_bar = tqdm(
+            desc=f"Exporting files to {output}: ",
+            unit=" files",
+            unit_scale=True,
+            unit_divisor=10,
+            total=self.count(),  # type: ignore [union-attr]
```
shcheklein, line 2545: why do we need this ignore?

shcheklein:
@ilongin can we please cleanup the name and description (if needed for the PR), thanks!

shcheklein:
@ilongin does it affect commands like cp / clone (I still don't know what is the difference tbh)?

src/datachain/nodes_thread_pool.py
```diff
@@ -57,44 +57,65 @@ def __init__(self, max_threads):
...
+    def cancel_all(self):
+        self.cancel = True
+        # Canceling tasks just in case any of them is scheduled to run.
+        # Note that running tasks cannot be canceled, instead we will wait for
+        # them to finish when shutting down thread loop executor by calling
+        # shutdown() method.
+        for task in self.tasks:
+            task.cancel()
+        if self.th_pool:
+            self.th_pool.shutdown()  # this will wait for running tasks to finish
```
shcheklein, line 117: so, what does it mean for the cancellation? (potentially taking long time?)

@skshetry how does cancellation for prefetch work? is it more or less immediate / with some finite timeout?

src/datachain/nodes_thread_pool.py
```diff
@@ -57,44 +57,65 @@ def __init__(self, max_threads):
         self._max_threads = max_threads
         self._thread_counter = 0
         self._thread_lock = threading.Lock()
```
shcheklein, line 59: separate: probably can be done w/o a lock if this only about `+1`. locks imposes some performance loss. There should be something like atomicint or something

src/datachain/lib/dc.py
ilongin: Not needed, c/p leftover. Removed

ilongin:
> Does this affect only to_storage? How about a regular case when we do `file.upload` in the mapper for example?

This only affect `to_storage()`.  btw I assumed you meant on `file.export()`, not `file.upload()`

ilongin:
> @ilongin does it affect commands like cp / clone (I still don't know what is the difference tbh)?

No, those commands still use "old" codebase which haven't been refactored to `DataChain`. We kind of deprioritized it for now but eventually they should use this API I think, which would make it much more simpler than current.

shcheklein:
> This only affect to_storage(). btw I assumed you meant on file.export(), not file.upload()

No, I mean `File.upload()` call that we are using now in some UDFs. But this is probably out of scope for this issue / PR.

Q: why can't we re-implement export_files (to_storage) with a UDF that already has prefetch. Is it because of symlinks?

ilongin:
> > This only affect to_storage(). btw I assumed you meant on file.export(), not file.upload()
> 
> No, I mean `File.upload()` call that we are using now in some UDFs. But this is probably out of scope for this issue / PR.
> 
> Q: why can't we re-implement export_files (to_storage) with a UDF that already has prefetch. Is it because of symlinks?

Symlinks and other things mentioned by @skshetry here https://github.com/iterative/datachain/pull/923#issuecomment-2656871328

src/datachain/nodes_thread_pool.py
skshetry: > @skshetry how does cancellation for prefetch work? is it more or less immediate / with some finite timeout?

We schedule a (worker) task to run in a separate thread (not the main thread).
When we receive `KeyboardInterrupt` or any exception on the main thread, we cancel the task which raises `asyncio.CancelledError` inside the worker task. This gets raised "at the next opportunity", i.e on some `await` point.

The cancellation should be nearly immediate for this discussion. However, technically, it depends on what is currently running in the asyncio event loop, such as a synchronous function (in which it will block unfortunately until completion).

ilongin:
@shcheklein I wanted to merge this now as I feel like most of the conversation is resolved, do you agree?

shcheklein:
@ilongin @skshetry 

> When prefetch>0 and cache=False, prefetching uses a temporary cache.

so, can we run just this particular UDF in the cache=True mode?

ilongin:
> @ilongin @skshetry
> 
> > When prefetch>0 and cache=False, prefetching uses a temporary cache.
> 
> so, can we run just this particular UDF in the cache=True mode?

I think there is a performance issue with that as then we will first save files to cache and then c/p to actual destination and maybe user doesn't want that (if he sets cache = false)

shcheklein:
> to cache and then c/p to actual destination

how would we do this operation? can we do move instead, right in UDF?

src/datachain/lib/dc.py
```diff
@@ -2500,8 +2508,10 @@ def to_storage(
         placement: FileExportPlacement = "fullpath",
         use_cache: bool = True,
         link_type: Literal["copy", "symlink"] = "copy",
+        num_workers: Optional[int] = EXPORT_FILES_MAX_THREADS,
```
dmpetrov, line 2511: Please change the name to something else since we use term `worker` for distributed compute. `threads` looks like a better fit. 

`EXPORT_FILES_MAX_THREADS` also suggests using threads, not workers.

dmpetrov (approved):
Looks good!

Added a comment about naming - it's important! Also, I'd appreciate some use-cases examples (ideally in the docs).

shcheklein:
@ilongin 

> if not use_cache:

where does it come from, could you elaborate please?

ilongin:
> @ilongin
> 
> > if not use_cache:
> 
> where does it come from, could you elaborate please?

It comes from `to_storage()` method

shcheklein:
@ilongin 

> It comes from to_storage() method

oh, I see. I wonder if we actually need it (why not use DataChain settings for this?) 🤔 

> About moving instead, you meant something like this?

yes, I would love @skshetry to review it as well. It feels we can further simplify all this caching logic tbh. I don't like that we are exposing it a bit. Try to think what need to be done for this UDF implementation to be clean and use as less internal stuff as possible. Even if it involves changes to the caching outside of export.

ilongin:
> @ilongin
> 
> > It comes from to_storage() method
> 
> oh, I see. I wonder if we actually need it (why not use DataChain settings for this?) 🤔
> 

You are right,  I've removed it.

> yes, I would love @skshetry to review it as well. It feels we can further simplify all this caching logic tbh. I don't like that we are exposing it a bit. Try to think what need to be done for this UDF implementation to be clean and use as less internal stuff as possible. Even if it involves changes to the caching outside of export.

I realized `symlink` is not an actual issue as if cache is not used or source of the file is not on local FS, we do normal copy instead. 

So the only issue that remains is that of performance as we discussed above.
Honestly, I'm not sure what is the ideal solution to this ... to me it seems like there is nothing wrong neither with prefetch nor cache itself, it just the use case of exporting file that maybe doesn't fit prefetch completely from performance standpoint. Prefetch always uses cache (temp or real) to store files somewhere and in that way makes sure file is available to the mapper and I don't see anything wrong with that. Cache is also simple part of the system and I don't see what can be changed there as well.
There are 3 solutions we need to choose between:
1. Don't use prefetch, but custom solution with threads / async to export files -> IMO the best option
2. Use prefetch and ignore performance overhead comparing to 1) -> Even though this looks best from code standpoint, we should make sure we have the best performance as well.
3. Use `mv` or similar 'hacking" as in my last example -> I wouldn't do this as it's not correct and could lead to bugs, e.g file won't be available in mapper after we do explicit move.

@skshetry I'm also wondering about your opinion.

skshetry:
> 1. Don't use prefetch, but custom solution with threads / async to export files -> IMO the best option
> 2. Use prefetch and ignore performance overhead comparing to 1) -> Even though this looks best from code standpoint, we should make sure we have the best performance as well.
> 3. Use `mv` or similar 'hacking" as in my last example -> I wouldn't do this as it's not correct and could lead to bugs, e.g file won't be available in mapper after we do explicit move.

3rd option is just an optimization that we could do. It does not have to be `mv`, we can symlink, reflink or hardlink files if needed. It's part of 2).

So, that leaves two options: either we reimplement `to_storage` using threading or async, or we use prefetching.

Regarding the prefetching, the issue is that we'll be making the same number of system calls at the very least, as there are files to download (let's ignore the syscalls required to read the file).
I think if we are not too worried about performance, this might be a better option. But this does give me a feeling of relying on side effects from _prefetching_ - it feels like a black magic from the pov of `export`. We can ignore the `mv`/link options for this PR, so I'd avoid optimizing here.

Now, on the first option, my preference would be using async, because of the issues the thread have (eg: cancellation) and the concurrent file transfers you can do on async that is not possible with threads. This is more important for small file transfers.

But as I mentioned above, you'd need to reimplement most of what `export` to achieve this, and we'd need to compare performance of this with `AsyncMapper`.

So, I am okay with the PR if you prefer to continue with the current approach. We can come back later if we need to improve.

PR #924, opened 2025-02-16
Models update 2

dreadatour:
Another approach to make better Yolo models along with bounding boxes conversions.

Instead of trying to solve unsolvable problem with having one universal model for Bounding Boxes (same for Poses and Segments), it looks like it is better to have dedicated models for every common format, e.g. Yolo. Plus to have really basic and trivial universal models and let users to override them and decide what to do next.

See also [updated docs page](https://models-update-2.datachain-documentation.pages.dev/references/data-types/yolo/).

Checklist:
- [x] update docs
- [x] update examples
- [x] tests full coverage


src/datachain/model/yolo.py
```diff
@@ -0,0 +1,500 @@
...
+    @property
+    def img_size(self) -> tuple[int, int]:
+        """Get the image size (width, height) from the original shape."""
+        return (
+            (self.orig_shape[1], self.orig_shape[0])
+            if len(self.orig_shape) == 2
+            else (0, 0)
+        )
+
+    def to_albumentations(self) -> Iterator[list[float]]:
```
dreadatour, line 102: `YoloBox` have bounding box type conversion (also below)

src/datachain/model/yolo.py
```diff
@@ -0,0 +1,500 @@
...
+    right_wrist = 10
+    left_hip = 11
+    right_hip = 12
+    left_knee = 13
+    right_knee = 14
+    left_ankle = 15
+    right_ankle = 16
+
+
+class YoloPose(YoloBox):
```
dreadatour, line 350: Yolo pose result have everything yolo bbox have + poses.

src/datachain/model/yolo.py
```diff
@@ -0,0 +1,500 @@
...
+    def img_size(self) -> tuple[int, int]:
+        """Get the image size (width, height) from the original shape."""
+        return (
+            (self.orig_shape[1], self.orig_shape[0])
+            if len(self.orig_shape) == 2
+            else (0, 0)
+        )
+
+
+class YoloSeg(YoloBox):
```
dreadatour, line 257: Yolo segment result have everything yolo bbox have + segments.

src/datachain/model/yolo.py
```diff
@@ -0,0 +1,500 @@
...
+            cls=cls,
+            name=name,
+            confidence=confidence,
+            box=box,
+            keypoints=keypoints,
+            orig_shape=list(results[0].orig_shape),
+        )
+
+
+class YoloCls(DataModel):
```
dreadatour, line 426: Not sure if anyone will use it but Yolo have models so I've added this model here too.

src/datachain/model/bbox.py
```diff
@@ -1,102 +1,84 @@
...
-    def from_list(coords: list[float], title: str = "") -> "BBox":
-        assert len(coords) == 4, "Bounding box must be a list of 4 coordinates."
-        assert all(isinstance(value, (int, float)) for value in coords), (
-            "Bounding box coordinates must be floats or integers."
-        )
-        return BBox(
-            title=title,
-            coords=[round(c) for c in coords],
-        )
+    def from_list(coords: Sequence[float], title: str = "") -> "BBox":
```
dreadatour, line 26: Leave this method for now since it is used in many places, but mark it as `deprecated`. No need in separate method if one can build this model with `__init__` method directly.

tests/unit/model/test_yolo.py
```diff
@@ -0,0 +1,190 @@
...
+    assert voc_boxes[0] == [10, 20, 190, 80]
+    assert voc_boxes[1] == [160, 80, 200, 100]
+
+    yolo_boxes = list(model.to_yolo())
+    assert len(yolo_boxes) == 2
+    assert_array_almost_equal(yolo_boxes[0], [0.5, 0.5, 0.9, 0.6], decimal=3)
+    assert_array_almost_equal(yolo_boxes[1], [0.9, 0.9, 0.2, 0.2], decimal=3)
+
+
+def test_yolo_obb():
```
shcheklein, line 59: Q: what to we test with this kind of tests?

dreadatour: We are testing if models were not changed, since other code may rely on them.

tests/unit/model/test_pose.py
```diff
@@ -0,0 +1,26 @@
+from datachain.model.pose import Pose, Pose3D
+
+
+def test_pose():
+    pose = Pose(x=[10, 20, 30], y=[40, 50, 60])
+    assert pose.model_dump() == {
+        "x": [10.0, 20.0, 30.0],
```
shcheklein, line 7: should these be ints?

src/datachain/model/bbox.py
```diff
@@ -1,102 +1,84 @@
...
 
 from datachain.lib.data_model import DataModel
+from datachain.model.utils import BBoxType, convert_bbox
 
 
 class BBox(DataModel):
     """
     A data model for representing bounding box.
 
+    Use `datachain.model.Yolo` or for YOLO-specific bounding boxes.
```
shcheklein, line 14: looks like there is a typo

tests/unit/model/test_pose.py
dreadatour: Since `Pose` is now a generic bounding box model I believe we should have `floats` here. See also my comment [here](https://github.com/iterative/datachain/pull/924/files/fdc7f1b033dd6845caf380a8884949a5509cf372#r1957254868) and model declaration [here](https://github.com/iterative/datachain/pull/924/files/fdc7f1b033dd6845caf380a8884949a5509cf372#diff-86234d5135b709fdb026741ca44df7a33d0dbdc127593b1d50b87a94c3c59217R10-R23).

src/datachain/model/bbox.py
```diff
@@ -1,102 +1,84 @@
...
-            [coords["x1"], coords["y1"], coords["x2"], coords["y2"]],
-            title=title,
+        Args:
+            coords (list[float]): The bounding box coordinates.
+            title (str): The title of the bounding box.
+
+        Returns:
+            BBox: The bounding box instance.
+        """
+        warn(
```
shcheklein, line 37: we don't need to deprecate anything - we can break it for now

src/datachain/model/bbox.py
shcheklein: internally?

src/datachain/model/bbox.py
```diff
@@ -1,102 +1,60 @@
...
+    This model is intended for general bounding box representations or other formats.
+
     Attributes:
         title (str): The title of the bounding box.
-        coords (list[int]): The coordinates of the bounding box.
-
-    The bounding box is defined by two points:
-        - (x1, y1): The top-left corner of the box.
-        - (x2, y2): The bottom-right corner of the box.
+        coords (list[float]): The coordinates of the bounding box.
```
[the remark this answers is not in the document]
shcheklein: why? could you give more details?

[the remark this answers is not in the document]
dreadatour: Sure! Since it is a generic model, user can decide to store bounding boxes of any format here: COCO, YOLO, Albumentation, etc. Normalized or not. It is just a bare model.

To be honest I don't think we even need these bare models for BBox, Pose, etc, we can leave only models for YOLO for now and add other dedicated models later.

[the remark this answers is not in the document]
shcheklein: it can break some downstream stuff in data loader / training though? - if it expects ints

src/datachain/model/bbox.py
dreadatour: I mean instead of using:

```
BBox.from_list([10, 20, 30, 40])
```

customers can use:

```
BBox(coords=[10, 20, 30, 40])
```

We don't use this model anymore in the datachain source code for now (except for the examples).

src/datachain/model/bbox.py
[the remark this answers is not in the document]
shcheklein: also, should we include type into the class? to know is it yolo or something else?

Issue #883, opened 2025-02-02
`group_by` is broken on nested columns


shcheklein:
I have a dataset with these model:

```
class Meta(DataModel):
    width: int
    height: int
    num1: int
    date: str
    location: str
    hour: str
    timestamp: int
    num2: int
```

I run a few queries w/o loading this model (it was used only to create the dataset).

Operation like:

```python
DataChain.from_dataset("detector").group_by(
    cnt=func.count(),
    partition_by=("meta.location"),
).select("meta.location", "cnt").show(100)
```

raises: 

```
 cannot resolve signal name 'meta.location': is not found
```

while pure `select("meta.location")` works fine.

Also:

```
DataChain.from_dataset("detector").group_by(
    cnt=func.count(),
    partition_by=("meta__location"),
).select("meta__location", "cnt").show(100)
```

raises `KeyError: 'sys__id'`

(not sure that we should be allowing using such direct select at all though)

dreadatour:
This is a tricky part, because after applying `group_by`:

```
DataChain.from_dataset("detector").group_by(
    cnt=func.count(),
    partition_by=("meta.location"),
)
```

chain does not contains `meta` signal with `Meta` model type anymore. After this `group_by` schema will be dead simple:

```
meta__location: str
cnt: int
```

`meta.location` is `meta__location` because we do not store dots in the column names in DB, we are replacing them with double underscore symbols `__`.

During `select` in case columns with dots (`meta.location`) are selected, we are checking schema first to find nested signals. In case of `meta.location` there are no nested signals (`meta` -> `location`) in DB anymore, it is raw `meta__location` and that's why this call failed with error:

```
cannot resolve signal name 'meta.location': is not found
```

Possible solution here is to do not check for schema and always replace dot with double underscore and to check if column exists in the DB, sounds like a possible option, but I am not sure if this is what we want.

shcheklein:
Do we serialize / store schema as a separate object in DB or do we restore it when we read dataset? 

E.g. what happens if I:

```
DataChain.from_dataset("detector").group_by(
    cnt=func.count(),
    partition_by=("meta.location"),
).save("something")
```

and read `something` back? will it consider `meta__location` as a nested field if there is no schema in memory to refer to?


> here are no nested signals (meta -> location) in DB anymore

in DB or in datachain schema at that point? coming back to the previous question - do we even store schema (nested fields vs flat fields)?

>

dreadatour:
[...]

PR #927, opened 2025-02-18
Fix `group_by` for nested columns

dreadatour:
Fix for #883:

1. Keep partial model for nested columns as discussed [here](https://github.com/iterative/datachain/issues/883#issuecomment-2655358308).
2. Fix for `KeyError: 'sys__id'` error: it looks like sys signals are not always exists in dataset (see also #81)

TODO:

- [ ] more tests for `to_partial` method required


src/datachain/lib/signal_schema.py
```diff
@@ -749,3 +749,120 @@ def _build_tree_for_model(
...
+                else:
+                    # If the parent type is not in the schema, add it
+                    # to the custom types with the signal.
+                    schema["_custom_types"][parent_type_partial] = {
+                        "schema_version": 2,
+                        "name": parent_type_partial,
+                        "fields": {signal: partial_type},
+                        "bases": [
+                            (get_type_name(partial_type), "__main__", partial_type),
+                            ("DataModel", "datachain.lib.data_model", "DataModel@v1"),
```
dreadatour, line 842: This part I don't like 😢 Is there a better way to do this? 🤔

tests/unit/lib/test_datachain.py
```diff
@@ -770,6 +770,26 @@ def test_collect_nested_feature(test_session):
         assert nested == features_nested[n]
 
 
+def test_select_no_sys_id(test_session):
```
shcheklein, line 773: not clear from the name what we are testing, can we rephrase it?

src/datachain/lib/signal_schema.py
```diff
@@ -749,3 +749,120 @@ def _build_tree_for_model(
...
+
+def type_name_to_partial(type_: str) -> str:
+    """Convert a type name to a partial type name with the same version."""
+    if "@" in type_:
+        type_name, version = type_.split("@")
+        type_ = f"{type_name}Partial@{version}"
+    return type_
+
+
+def get_type_name(type_: str) -> str:
```
shcheklein, line 864: can we use `parse_name_version`?

src/datachain/lib/signal_schema.py
```diff
@@ -749,3 +749,120 @@ def _build_tree_for_model(
...
+        if not schema["_custom_types"]:
+            del schema["_custom_types"]
+
+        return SignalSchema.deserialize(schema)
+
+
+def type_name_to_partial(type_: str) -> str:
+    """Convert a type name to a partial type name with the same version."""
+    if "@" in type_:
+        type_name, version = type_.split("@")
```
shcheklein, line 859: can in some cases version be empty (just looking into `parse_name_version`)

src/datachain/lib/signal_schema.py
```diff
@@ -749,3 +749,120 @@ def _build_tree_for_model(
...
+
+                parent_type_origin, parent_type_partial = signal_type, partial_type
+
+        if not schema["_custom_types"]:
+            del schema["_custom_types"]
+
+        return SignalSchema.deserialize(schema)
+
+
+def type_name_to_partial(type_: str) -> str:
```
shcheklein, line 856: if these are helpers - let's make them protected

src/datachain/lib/signal_schema.py
```diff
@@ -749,3 +749,120 @@ def _build_tree_for_model(
...
+        """
+        serialized = self.serialize()
+        custom_types = serialized.get("_custom_types", {})
+
+        schema: dict[str, Any] = {"_custom_types": {}}
+
+        for column in columns:
+            parent_type_origin, parent_type_partial = "", ""
+            for i, signal in enumerate(column.split(".")):
+                # We are expecting for the first part of the column name (signal name)
```
shcheklein, line 792: this amount or comments is a bit out of style? is it because method is complicated? can we split it / simplify it then to make a bit more self- descriptive?

src/datachain/lib/signal_schema.py
```diff
@@ -749,3 +749,120 @@ def _build_tree_for_model(
...
+                    # If the parent type is already in the schema, just add the signal
+                    # to the custom types.
+                    schema["_custom_types"][parent_type_partial]["fields"][signal] = (
+                        partial_type
+                    )
+                else:
+                    # If the parent type is not in the schema, add it
+                    # to the custom types with the signal.
+                    schema["_custom_types"][parent_type_partial] = {
+                        "schema_version": 2,
```
shcheklein, line 837: it's better to use `CustomType` model and serialize it and we do in signal schema itself ... so that we don't have hard coded dict outside w/i control (we are breaking isolation and leaking internals this way)

src/datachain/lib/signal_schema.py
shcheklein: yep, that's quite ugly :) why not just take this code:

```
for type_ in fr.__mro__:
            model_store_name = (
                ModelStore.get_name(type_) if issubclass(type_, DataModel) else None
            )
            bases.append((type_.__name__, type_.__module__, model_store_name))
```

make it a helper, run it agains general `DataModel` (+ add a tuple with partial type)

less hardcoded at least

src/datachain/lib/signal_schema.py
```diff
@@ -749,3 +749,120 @@ def _build_tree_for_model(
...
+                        "name": parent_type_partial,
+                        "fields": {signal: partial_type},
+                        "bases": [
+                            (get_type_name(partial_type), "__main__", partial_type),
```
shcheklein, line 841: can we double check if it belongs to `__main__` usually

src/datachain/lib/signal_schema.py
```diff
@@ -749,3 +749,120 @@ def _build_tree_for_model(
...
+        custom_types = serialized.get("_custom_types", {})
+
+        schema: dict[str, Any] = {"_custom_types": {}}
+
+        for column in columns:
+            parent_type_origin, parent_type_partial = "", ""
+            for i, signal in enumerate(column.split(".")):
+                # We are expecting for the first part of the column name (signal name)
+                # to be in the schema, other parts should be in the custom types.
+                if i == 0:
```
shcheklein, line 794: can be a column with a custom type w/o any additional parts `bbox`

in this case it seems we won't copy `_custom_type`, right?

src/datachain/lib/signal_schema.py
dreadatour: 🙏 Fixed 👍

tests/unit/lib/test_datachain.py
dreadatour: Done

src/datachain/lib/signal_schema.py
dreadatour: Done

src/datachain/lib/signal_schema.py
dreadatour: I can not think of any cases not can find one. If you have something in mind, please, share 🙏

src/datachain/lib/signal_schema.py
dreadatour: Done

src/datachain/lib/signal_schema.py
dreadatour: Removed all the comments, simplify code a bit.

src/datachain/lib/signal_schema.py
dreadatour: Done, thank you 🙏

src/datachain/lib/signal_schema.py
dreadatour: Checked, again, I can not think of any cases, let me know if you have any

src/datachain/lib/signal_schema.py
dreadatour: Not sure 🤔
Added one more test for our custom model (`BBox`), but still not sure if this is what you have meant.

src/datachain/lib/dc.py
```diff
@@ -1128,9 +1130,13 @@ def group_by(
             signal_columns.append(column)
             schema_fields[col_name] = func.get_result_type(self.signals_schema)
 
+        signal_schema = SignalSchema(schema_fields)
```
shcheklein, line 1133: Q: (trying to understand this a bit better)

Why don't we apply `to_partial` to `schema_fields`? can one of the `schema_fields` be similar in a sense that we'll need to preserve a partial model?

src/datachain/lib/signal_schema.py
shcheklein: what are the possible inputs to this function? Can I pass `bbox` (and expect all nested fields included), for example? Or is always the most granular level nested fields (and reconstruct parents from them)?

src/datachain/lib/signal_schema.py
```diff
@@ -749,3 +754,108 @@ def _build_tree_for_model(
...
+
+        return SignalSchema.deserialize(schema)
+
+
+def _type_name_to_partial(type_name: str) -> str:
+    """Convert a type name to a partial type name with the same version."""
+    if "@" not in type_name:
+        return type_name
+    model_name, version = ModelStore.parse_name_version(type_name)
+    return f"{model_name}Partial@v{version}"
```
shcheklein, line 861: can we hit to different Partial constructed from the same model with a bit different set of fields? (this function will generate the same name for both which might be wrong)

src/datachain/lib/signal_schema.py
```diff
@@ -749,3 +754,108 @@ def _build_tree_for_model(
...
+
+                parent_type_origin, parent_type_partial = signal_type, partial_type
+
+        if schema_custom_types:
+            schema["_custom_types"] = {
+                type_name: ct.model_dump()
+                for type_name, ct in schema_custom_types.items()
+            }
+
+        return SignalSchema.deserialize(schema)
```
shcheklein, line 853: can we construct the object itself instead of deserializing it?

src/datachain/lib/signal_schema.py
```diff
@@ -749,3 +754,108 @@ def _build_tree_for_model(
...
+        (e.g. `FooPartial@v1` instead of `Foo@v1`) to avoid conflicts
+        with the original schema.
+
+        Args:
+            *columns (str): The columns to include in the partial schema.
+
+        Returns:
+            SignalSchema: The new partial schema.
+        """
+        serialized = self.serialize()
```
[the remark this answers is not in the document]
dreadatour: I am serializing it also to get list of fields (so whole serialize result is used, not only custom types), same time this can be done in a different way. I don't really like serialization-deserialization, but I can not find more elegant way to do it.

src/datachain/lib/signal_schema.py
dreadatour: Now I see. Good question. It is supposed to always take the most granular level nested fields, but since this function is exposed, anyone can pass, let's say, `bbox`. Let me check and add new test for this.

src/datachain/lib/signal_schema.py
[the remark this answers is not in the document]
shcheklein: what was the blocker from using just `values`, `tree` fields or something like that?

src/datachain/lib/signal_schema.py
shcheklein: I think that's fine, can be an assert I guess - if we don't need a general helper let's keep it scoped, I would clarify the expected inputs though

src/datachain/lib/dc.py
dreadatour: This is because `schema_fields` here is:

```
schema_fields: dict[str, DataType]
```

In case of partial signals (let's say, `file.path`) it will be: `schema_fields = {"file.path": str}`. We can use `{"file": File}`, for example, but we still need to know partial signal names to preserve them.

That's why we keep simple signals in `schema_fields` and store partial signals in `keep_columns` and creating new schema from existing one using `to_partial`.

src/datachain/lib/signal_schema.py
dreadatour: I think it is possible in theory, but I can not think of any valid real-world example. Partial schema is created from existing dataset and if dataset's schema is correct, partial schema will also be correct.

Same is possible in theory, for example, if we will use `merge` from different datasets or `map` from different Python files with different custom models with the same name.

Not sure if we need to address this issue in this PR, but this is a good question for sure and we should think about it.

src/datachain/lib/dc.py
shcheklein: `{"file.path": str}`

does it mean that partial File won't be created and preserved, though?

I still don't quite understand why for some columns (defined as names) we preserve and construct partials, for other (defined as Func) we don't 🤔

src/datachain/lib/signal_schema.py
shcheklein: Two files (File model). Different name in the DB schema:

`file1`: File
`file2`: File

we use columns `file1.path` and `file2.source`.

will it lead to a collision here potentially?

src/datachain/lib/signal_schema.py
dreadatour: Sure we can, but I think it will be almost identical to `deserialize` 👀 IMO it is better to reuse existing function.

src/datachain/lib/signal_schema.py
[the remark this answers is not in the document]
shcheklein: my concern here is that as we change the format now we'll also have to consider this helper (which is not an easy on at all) and carefully include new fields, etc

src/datachain/lib/signal_schema.py
dreadatour: One partial model will be created:

```
FilePartial@v1
    path: str
    source: str
```

shcheklein: kk, is it correct though? does it mean we'll keep some extra fields?

src/datachain/lib/signal_schema.py
```diff
@@ -179,6 +179,16 @@ def from_column_types(col_types: dict[str, Any]) -> "SignalSchema":
                 )
         return SignalSchema(signals)
 
+    @staticmethod
+    def _get_bases(fr: type) -> list[tuple[str, str, Optional[str]]]:
+        bases: list[tuple[str, str, Optional[str]]] = []
+        for base in fr.__mro__:
```
amritghimire, line 185: How will it fare if the type is somehow int or some basic types? 
image

Should we use getattr instead?

src/datachain/lib/signal_schema.py
```diff
@@ -749,3 +754,120 @@ def _build_tree_for_model(
...
+                        fields={signal: partial_type},
+                        bases=[
+                            (partial_type_name, "__main__", partial_type),
+                            *data_model_bases,
+                        ],
+                    )
+
+                parent_type, parent_type_partial = signal_type, partial_type
+
+        if schema_custom_types:
```
amritghimire, line 867: Can you check if the hidden fields are also serialized? I recently merged #931

src/datachain/lib/signal_schema.py
amritghimire: Sorry for such a silly mistake 🤦

PR #928, opened 2025-02-19
Incremental (delta) update

ilongin:
Adding the ability to do incremental, or delta updates with dataset. Idea is to not re-build the whole dataset from source once source has some changes (new or modified files in s3 bucket for example), but to create diff chain between the last version of dataset and source with all modifications added (chain methods like mappers, filters etc. what has been added to create original chain from source), and then merge diff with last version of dataset. This way we will have much better performance.

The way user will run delta updates is to just re-run the whole script where it creates dataset, with one small modification - adding `delta=True` on `DataChain.save()` method.

Example:
```python
from datachain import DataChain, Column, C

def my_embedding(file: File) -> list[float]:
    return [0.5, 0.5]

(
      DataChain.from_storage(
          "s3://ldb-public/remote/data-lakes/dogs-and-cats/", type="image", anon=True
      )
      .filter(C.name.glob("*cat*"))
      .filter(C.name.glob("*.jpg"))
      .map(emb=my_embedding)
      .mutate(dist=func.cosine_distance("emb", (0.1, 0.2)))
      .save("cats", delta=True)
)
```

Comparison between current approach and alternative one:

Comparison | 1. Current approach | 2. Alternative approach (similar to current) | 3. Apply diff on every starting step
--- | --- | --- | ---
Description | We calculate diff between last version of starting dataset (listing or any other "normal" dataset) and current version of resulting dataset. We then apply chain functions on that diff and merge results with current version of resulting dataset to create new version of it | From dataset dependencies of resulting dataset we get starting dataset name and version (listing or any other "normal" dataset) and calculate diff between that and last version of starting dataset. We then apply chain functions on that diff and merge results with current version of resulting dataset to create new version of it | Idea is to run whole chain in special "delta" mode which would ensure that on each starting step (reading from listing or normal dataset) we don't read whole data but calculate diff and run on that. We would then union that "delta" chain with current last version of dataset. The problem is that this is not possible to implement because of structure of our codebase - diff is implemented in `DataChain` or "upper" level and it would need to be used in `DatasetQuery` or "lower" level ... we need to refactor / remove `DatasetQuery` first.
Someone removes starting dataset and all of it's versions (e.g listing) |  Everything works as expected (listing will be re-created and delta update will work fine) | We wouldn't have exact listing version for delta -> we would need to re-create the whole dataset as it was ran the first time (without delta performance gains) | We wouldn't have exact listing version for delta -> we would need to re-create the whole dataset as it was ran the first time (without delta performance gains)
File object is removed in resulting dataset (e.g using one of the chain functions) | Delta update doesn't work | Delta update doesn't work | Delta update doesn't work | Delta update doesn't work
How it will look in UI | The same as in local (just adding `delta=True` flag in `.save()`) | The same as in local (just adding `delta=True` flag in `.save()`) | The same as in local (just adding `delta=True` flag in `.save()`)
Pros | If starting dataset (e.g listing) is removed we still can use delta update and have performance gain | - |
Cons | If file signal is removed from resulting dataset, delta update doesn't work | 1. If someone removes listings, we need to re-create whole dataset from beginning, without delta.  2. If file signal is removed from resulting dataset, delta update doesn't work |
Not supported `DataChain` methods | `union()`, `agg()`, `distinct()`, `group_by()`, `merge(inner=False)` |`union()`, `agg()`, `distinct()`, `group_by()`, `merge(inner=False)` | `agg()`, `distinct()`, `group_by()`, `merge(inner=False)`

Q&A

**Q: Difference between approaches regarding supporting queries**
A: They should support the same type of queries

Q: **How will different approaches look in code?**
A: The same as normal queries, just without union or merge and by adding `delta` flag to `.save()`

Q: **In the second approach the idea was to use delta on the source side (per each source). This would allow to have a single file source for example. Both approaches don't allow that atm AFAIU. Are there cases like this?**
A: In both approaches when we speak about "source" we actually mean on source or starting dataset from which resulting dataset is created. It can be listing dataset if `.from_storage()` is used or just a regular dataset if `.from_dataset()` is used for example. In the question, if I understood correctly, suggestion is to get distinct values from `source` column and then do `diff` for each of those and apply chain to each diff and then union with current dataset. This is not possible because in order to get multiple sources into resulting dataset some kind of union / merge must have happened in the chain process and even if union is supported we don't know the chain of dataset which have been added with union. For example:
```python
(
    .filter(C("file.path").glob("*cat*"))
    .union(DataChain.from_dataset("one_file"))
    .save("cats", delta=True)
)
```
In this chain we are adding dataset `one_file` which we don't know how is created as it's chain is not in this script.  Maybe it's chain was created in some other script and looks something like:
```python
(
    DataChain.from_storage("gs://datachain-demo")
    .limit(1)
    .save("one_file")
)
```
but we don't know this and we won't be able to apply `limit(1)` to it after doing `diff` for source `gs://datachain-demo` and would end up with whole `gs://datachain-demo` bucket instead of one file in resulting dataset.  Maybe this could be done in future if we will have ability to go through the tree of datasets and extracting it's chains from other scripts but that is way out of the scope of this task and I'm not sure if it's even possible to do atm. Another issue is that we cannot just apply chain to diff for every source. In above example, chain is the instance that starts with `DataChain.from_storage("s3://ldb-public/remote/data-lakes/dogs-and-cats/", anon=True)` and ends up with `.save()` and if we apply it to diff of source  `gs://datachain-demo`  code won't work as for example filter `.filter(C("file.path").glob("*cat*"))` is not meant to be applied to it ... in general, I don't see any easy solution of how in any case (not just edge cases) we could make this kind of logic to work. I'm open to suggestions as maybe I'm missing something but ATM I just don't see it work.

Regarding the one file, yes that one is tricky. With current implementation it doesn't break but every time dataset is calculated from start (delta doesn't work as there is no dependency in created dataset as there is no listing created since we just extract data from file and create dataset rows) .. example:
```python
    DF_DATA = {
        "first_name": ["Alice", "Bob", "Charlie", "David", "Eva"],
        "age": [25, 30, 35, 40, 45],
        "city": ["New York", "Los Angeles", "Chicago", "Houston", "Phoenix"],
    }
    pd.DataFrame(DF_DATA).to_parquet(Path(os.path.join(os.path.abspath(os.getcwd()), "test.parquet")))
    DataChain.from_storage(path.as_uri()).parse_tabular().save("tabular", delta=True)
```
**Q: We need actual examples of code (e.g. union / merge where it breaks)**
A: Code doesn't brake on union / merge, instead it just creates duplicates. Example:
```python
(
    DataChain.from_storage("gs://datachain-demo/datacomp-small/metadata")
    .limit(1)
    .save("one_file")
)

(
    .filter(C("file.path").glob("*cat*"))
    .union(DataChain.from_dataset("one_file"))
    .save("cats", delta=True)
)
```
Here, on a second run of script if there is any diff between old and new version of `s3://ldb-public/remote/data-lakes/dogs-and-cats/` we would end up with duplicate file from `gs://datachain-demo/datacomp-small/metadata` as that one would be added with applying chain (which has union in it) to diff and then merging (union) with current latest version of resulting dataset which already has that file. Currently I don't see the way to fix this and even if there is some smart way to avoid this, I would leave it for follow up.

Q: **how about group by and other stuff - will those work?**
A: Group by cannot work as well. I'm not sure it can ever work with delta update even in theory as the whole point of delta is to apply chain (which is where group by is) to just a subset of data that has been added / changed in order for group by to be correct it must be applied to whole dataset. 
Other stuff should work IMO.


dmpetrov:
@ilongin it would be great to extract all logic outside of the fat file `dc.py` to `increment.py` or `dc_incremental.py`

Also, should we call it incremental or delta? :) Delta seems better but I don't like it do to a conflict with Delta Lake. Any ideas? :)

dmpetrov:
> Currently `DataChain.diff()` returns only added and changed records by default...

Let's use the same default for the incremental update.

> delta makes more sense

Then let's use Delta 🙂

tests/unit/lib/test_signal_schema.py
```diff
@@ -1005,3 +1005,8 @@ def test_column_types(column_type, signal_type):
 
     assert len(signals) == 1
     assert signals["val"] is signal_type
+
+
+def test_get_file_signal():
+    assert SignalSchema({"name": str, "f": File}).get_file_signal() == "f"
```
dreadatour, line 1011: Should we also test for nested file signal or is it out of the scope of this method?

E.g.:

```python
class CustomModel(DataModel):
    file: File
    foo: str
    bar: float

assert SignalSchema({"name": str, "custom": CustomModel}).get_file_signal() == "custom.file"
```

dreadatour (approved):
Looks good to me 👍🔥

tests/unit/lib/test_signal_schema.py
ilongin: For now it only works for top level file objects. In future we can add nested as well if needed

src/datachain/lib/dc.py
```diff
@@ -333,6 +334,15 @@ def clone(self) -> "Self":
         """Make a copy of the chain in a new table."""
         return self._evolve(query=self._query.clone(new_table=True))
 
+    def append_steps(self, chain: "DataChain") -> "Self":
```
shcheklein, line 337: Q: do we need this a as user facing API? can it be hidden?

src/datachain/lib/dc.py
```diff
@@ -760,16 +770,33 @@ def listings(
...
+        **kwargs,
     ) -> "Self":
         """Save to a Dataset. It returns the chain itself.
 
         Parameters:
             name : dataset name. Empty name saves to a temporary dataset that will be
                 removed after process ends. Temp dataset are useful for optimization.
             version : version of a dataset. Default - the last version that exist.
+            delta : If True, we optimize on creation of the new dataset versions
+                by calculating diff between source and the last version and applying
```
shcheklein, line 786: could you describe a bit - how do we do this exactly? do we calculate diff for each "source" (e.g. bucket)?

src/datachain/lib/dc.py
ilongin: Well, I'm using this outside of `DataChain` class so I wanted to make it public.

src/datachain/lib/dc.py
```diff
@@ -1621,6 +1648,7 @@ def compare(
         modified: bool = True,
         same: bool = False,
         status_col: Optional[str] = None,
+        sys: Optional[bool] = False,
```
skshetry, line 1651:
```suggestion
        sys: bool = False,
```

src/datachain/lib/dc.py
shcheklein: what if there are two sources (merge of two datasets)?

src/datachain/lib/dc.py
```diff
@@ -760,16 +761,42 @@ def listings(
...
+                Then we merge modified diff with the last version of dataset to
+                create new version. This way we avoid applying modifications to all
+                records from source every time since that can be expensive operation.
+                Source can be cloud storage or other dataset which has File object
+                in schema.
+                Diff is calculated using `DataChain.diff()` method which looks into
+                File `source` and `path` for matching, and File `version` and `etag`
+                for checking if the record is changed.
+                Note that this takes in account only added and changed records in
+                source while deleted recordsare not removed in the new dataset version.
```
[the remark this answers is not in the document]
shcheklein: > by calculating diff between source and the last version of dataset

I'm not sure I understand this. I assume source == bucket here (or dataset) and dataset - is it an actual named dataset from the `save("name", delta=True)` operation? How can we compare source with some datasets, I'm not sure ...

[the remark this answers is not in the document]
shcheklein:
```
Diff is calculated using `DataChain.diff()` method which looks into
                File `source` and `path` for matching, and File `version` and `etag`
                for checking if the record is changed.
```

I see. 

This won't work in so many cases though, no? File got renamed, dropped. There are > 1 sources, etc, etc.

Why can't we compare source with itself to get diff, apply chan to that diff (or multiple diffs if there are multiple sources) and the end merge result with a previous version of the saved dataset. I think that semantics would be cleaner. WDYT?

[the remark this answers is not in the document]
ilongin: I started with premise that source = some other dataset with `File` objects,  to make it more generic as every `DataChain` construct starts with some dataset (whether it's listing or other "normal" one). 
You are right about this not working if `File` is removed and your suggestion does sound better (I will try to re-implement that way). 

The only thing I'm not sure is about  what you said "multiple diffs if there are multiple sources". From this I assumed you mean: 1. get distinct sources from dataset (e.g distinct bucket names) 2. get their listings 3. for all of them calculate diffs with self 3. apply chain to each and merge. Disadvantage of this approach is that then we cannot do delta update with other dataset as starting point and it's more complex to implement as that "get their listings" part is not so easy to find since we have partial listings (in `source` we have just bucket name but listing from which it's created could be something more specific depending of what is defined in `DataChain.from_storage()` method ). Also if someone does filtering using path and maybe glob in the `.from_storage()` itself, e.g `DataChain.from_storage('s3://ldb-public/remote/data-lakes/dogs-and-cats/cat*') this will be tricky to implement. Also, applying chain to every of this little diffs (for each source) is not possible as some source can come from union / merge with other chain and those methods (e.g filters, mappers) that were in the chain before union/merge cannot be applied to it as the code will break.
Because of this I suggest to keep this "dataset" as a original source notation and with your suggestion from above it would be something like:
1. Get original starting dataset from which chain is being created (listing dataset, other dataset etc.)
2. Calculate diff between latest version and version before of that starting dataset
3. Apply chain to that diff and union with current version

To refer to your question below as well, diff doesn't work with union / merge in a way that will catch changes from that other chain which was "added" to the original one, but the code will not break i.e user can still use union / merge - it will  recalculate whole other chain that was added with union / merge so there will be no performance gains.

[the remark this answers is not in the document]
ilongin: Now I realized that "compare source with itself to get a diff" is also little bit tricky. What if someone created dataset from some listing that was then in latest version 5 and then there were multiple versions (re-listings) added and now it's in version 10. We should compare version 10 to version 5 but we don't save that info that it was created from version 5 anywhere ... in general it's questionable what comparing with self even means.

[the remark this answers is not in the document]
shcheklein: 1. We don't save with which version of source listing we created dataset so we don't know which version of source to diff. This is explained in more details in my last post here as well. 

I think we can / and probably should be saving this (regardless of this epic). Are there any issues with that?

3. If someone removes listings we will not be able to do diff with previous versions and we won't be able to do incremental update. 

I guess it will be the same for any implementation to a certain degree. If someone removes the previous dataset we also won't be able to do an incremental update, right?

are there any other fundamental problems with this? what is the difference this approach vs alternatives in regards to different types of queries they would support? How will different approaches look in code? Will they affect how this can be done in UI?

==================

Can we please create a table with different approaches, brief description, brief examples, pros/cons and more examples how they would looks like and work (or not) for different types of queries.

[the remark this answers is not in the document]
ilongin: > I think we can / and probably should be saving this (regardless of this epic). Are there any issues with that?

You are right actually, I totally forgot about dataset dependencies which can be used for this.

> I guess it will be the same for any implementation to a certain degree. If someone removes the previous dataset we also won't be able to do an incremental update, right?

When all listing is removed then in current implementation delta still works fine while in alternative approach whole dataset needs to be re-created as if there was no delta (it will still work but without delta performance enchacement)

> are there any other fundamental problems with this? what is the difference this approach vs alternatives in regards to different types of queries they would support? How will different approaches look in code? Will they affect how this can be done in UI?
> 
> ==================
> 
> Can we please create a table with different approaches, brief description, brief examples, pros/cons and more examples how they would looks like and work (or not) for different types of queries.

I will write table in description. It's hard to say if there are some fundamental problems with alternative approach until I implement it but with current approach the only issue is that removal of file objects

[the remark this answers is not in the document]
ilongin: Added new function called `delta_update_alternative()` which implements alternative approach (what you suggested) so we can compare them and also added table of comparison in description between approaches.
TL;DR; Alternative approach does look slightly better now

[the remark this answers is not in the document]
shcheklein: thanks for the table!

- we need actual examples of code (e.g. union / merge where it breaks)
- how about group by and other stuff - will those work?
- in the second approach the idea was to use `delta` on the source side (per each source). This would allow to have a single file source for example. Both approaches don't allow that atm AFAIU. Are there cases like this?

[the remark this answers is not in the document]
ilongin: Added some answers to those in PR description. I will add some more code examples if needed.
I found one major flaw of alternative approach (or any approach that allows `File` objects to disappear in final dataset) - it is impossible to filter out old records that are modified in new version of source dataset.
My approach was:
```python
diff = source_dc_latest.diff(source_dc)  # diff source by itself to get `added` and `modified` rows
res = DataChain.from_dataset(name, latest_version).diff(diff, added=True, modified=False)  # diff current dataset with diff to filter out those rows that are modified
res = res.union(diff)  # final result
```
Step 2 is problematic since we cannot diff current dataset with anything if we assume `File` objects might not be there.
I will try to think of about some other approach but ATM I don't see any viable solution. 
If we allow final dataset to have old and new modified rows then this is not an issue, but this issue requirements were to include only newest versions of a files in final result.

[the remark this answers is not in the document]
shcheklein: Also, I'm still not sure if there are cases when we read multiple sources (e.g. one is incremental - second is static - e.g. single file). And we run the whole query in an incremental mode we do want to get delta for the first one, but we don't want to do it for the second (it will an empty result)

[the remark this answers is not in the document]
ilongin: > Also, I'm still not sure if there are cases when we read multiple sources (e.g. one is incremental - second is static - e.g. single file). And we run the whole query in an incremental mode we do want to get delta for the first one, but we don't want to do it for the second (it will an empty result)

I'm not sure I understand. Can you give me an example? i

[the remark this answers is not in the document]
shcheklein: > From my understanding I don't see how group by

so, what else doesn't work / requires clarification of semantics? merge, group by ... how about window, partioned, etc? can we come up with a table after all? 

> I'm not sure I understand. Can you give me an example? 

https://colab.research.google.com/github/iterative/datachain-examples/blob/main/formats/json-metadata-tutorial.ipynb

or

```python
DataChain.from_csv("s3://ainventory-store/prices_datasets/annotation.csv", object_name="annotation").save("annotations") # <--- this file is static, and has only static information in it
DataChain.from_storage("s3://ainventory-store/prices_datasets/images").save("images")

annotations = DataChain.from_dataset("annotations")
images = DataChain.from_dataset("images")


annotations.select("annotation").merge(images, on..., right_on,  inner=True).save("catalog", incremental=True)
```

In those we have a file with annotations (static or not). We keep updating it. It would be enough to either just read a new version of that file. `merge` should be working in such cases to my mind.

-----------------------

I feel we really need more examples and scenarios. Can we check our examples, tests, etc? Try to see what happens in those if we do incremental mode?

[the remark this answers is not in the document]
ilongin: @shcheklein  I think it's important to brainstorm around the implementation idea you mentioned on 1-1 again about finding all the distinct sources in resulting dataset and doing delta update for each of those separately.  IMO this is just impossible to implement and I've put already some explanation in PR description (Q&A section,  third question / answer). Please take a look and let me know if you think I'm missing something.

src/datachain/dataset.py
```diff
@@ -105,24 +105,21 @@ def parse(
...
         return cls(
             id,
-            dependency_type,
-            dependency_name,
+            (
+                DatasetDependencyType.STORAGE
+                if is_listing_dataset(dataset_name)
+                else DatasetDependencyType.DATASET
+            ),
+            dataset_name,
```
shcheklein, line 122: just to double check -does it mean that names will change for all datasets going forward? is it to preserve the full listing name?

src/datachain/delta.py
```diff
@@ -0,0 +1,85 @@
...
+        latest_version = catalog.get_dataset(name).latest_version
+    except DatasetNotFoundError:
+        # first creation of delta update dataset
+        return None, True
+
+    dependencies = catalog.get_dataset_dependencies(
+        name, latest_version, indirect=False
+    )
+    if len(dependencies) > 1:
+        raise Exception(
```
shcheklein, line 47: it seems we need some tests

also,  let's not use Exception - let's use DataChainError or what we have for these cases

src/datachain/delta.py
```diff
@@ -0,0 +1,85 @@
...
+    but just the diff part which is very important for performance.
+
+    Note that currently delta update works only if there is only one direct dependency.
+    """
+    catalog = dc.session.catalog
+    dc._query.apply_listing_pre_step()
+
+    chain_file_signal = dc.signals_schema.get_file_signal()
+    if not chain_file_signal:
+        raise ValueError("Chain doesn't produce file signal, cannot do delta update")
```
shcheklein, line 35: what does it mean? how users can understand this?

src/datachain/delta.py
```diff
@@ -0,0 +1,85 @@
...
+
+    dep = dependencies[0]
+    if not dep:
+        # starting dataset (e.g listing) was removed so we are backing off to normal
+        # dataset creation, as it was created first time
+        return None, True
+
+    source_ds_name = dep.name
+    source_ds_version = int(dep.version)
+    source_ds_latest_version = catalog.get_dataset(source_ds_name).latest_version
```
shcheklein, line 59: will this logic work with studio flags ?

src/datachain/delta.py
```diff
@@ -0,0 +1,85 @@
...
+
+    source_dc = datachain.read_dataset(source_ds_name, source_ds_version)
+    source_dc_latest = datachain.read_dataset(source_ds_name, source_ds_latest_version)
+    source_file_signal = source_dc.signals_schema.get_file_signal()
+    if not source_file_signal:
+        raise ValueError("Source dataset doesn't have file signals")
+
+    diff = source_dc_latest.diff(source_dc, on=source_file_signal)
```
shcheklein, line 67: I think we should not be expecting file signal to exist

shcheklein, line 67: should there be `persist` at the end? to avoid calculating the diff a few times?

src/datachain/delta.py
```diff
@@ -0,0 +1,85 @@
...
+    # We append all the steps from the original chain to diff, e.g filters, mappers.
+    diff = _append_steps(diff, dc)
+
+    if diff.is_empty():
```
shcheklein, line 71: actually it might calculate the whole computation twice, no?

src/datachain/diff/__init__.py
```diff
@@ -155,7 +155,11 @@ def _to_list(obj: Optional[Union[str, Sequence[str]]]) -> Optional[list[str]]:
     if status_col:
         cols_select.append(diff_col)
 
-    dc_diff = dc_diff.select(*cols_select)
+    if not dc_diff._sys:
+        # TODO workaround when sys signal is not available in diff
```
shcheklein, line 159: what does it mean?

src/datachain/lib/dc/datachain.py
```diff
@@ -66,11 +68,34 @@
...
 
 
+T = TypeVar("T", bound="DataChain")
+
+
+def delta_disabled(
+    method: "Callable[Concatenate[T, P], T]",
+) -> "Callable[Concatenate[T, P], T]":
+    """
+    Decorator for disabling DataChain methods (e.g `.agg()` or `.union()`) to
```
shcheklein, line 83: does it mean we'll throw an exception only when we start executing it? (might be very long time and expensive for users). Can we do this before we execute it?

src/datachain/lib/dc/datachain.py
```diff
@@ -176,6 +202,16 @@ def __repr__(self) -> str:
         self.print_schema(file=file)
         return file.getvalue()
 
+    def as_delta(self, delta: bool = False) -> "Self":
```
shcheklein, line 205: does it become a public method? do users need this?

src/datachain/lib/dc/datachain.py
```diff
@@ -470,8 +506,41 @@ def save(  # type: ignore[override]
             description : description of a dataset.
             attrs : attributes of a dataset. They can be without value, e.g "NLP",
                 or with a value, e.g "location=US".
+            delta : If True, we optimize on creation of the new dataset versions
```
shcheklein, line 509: so, am I wrong or `delta` was not moved back to the source?

shcheklein: or why do we need it in save?

src/datachain/lib/dc/storage.py
```diff
@@ -151,7 +167,7 @@ def lst_fn(ds_name, lst_uri):
 
         chain = ls(dc, list_path, recursive=recursive, column=column)
 
-        storage_chain = storage_chain.union(chain) if storage_chain else chain
+        storage_chain = storage_chain.union(chain) if storage_chain else chain  # type: ignore[attr-defined]
```
shcheklein, line 170: can we fix it please?

src/datachain/lib/dc/storage.py
```diff
@@ -47,6 +48,21 @@ def read_storage(
         update : force storage reindexing. Default is False.
         anon : If True, we will treat cloud bucket as public one
         client_config : Optional client configuration for the storage client.
+        delta : If True, we optimize on creation of the new dataset versions
```
shcheklein, line 51: run through chatGTP please. Things like `optimize on creation` look wrong a bit ...

src/datachain/delta.py
```diff
@@ -0,0 +1,85 @@
...
+    from datachain.lib.dc import DataChain
+
+
+def _append_steps(dc: "DataChain", other: "DataChain"):
+    """Returns cloned chain with appended steps from other chain.
+    Steps are all those modification methods applied like filters, mappers etc.
+    """
+    dc = dc.clone()
+    dc._query.steps += other._query.steps.copy()
+    dc.signals_schema = dc.signals_schema.append(other.signals_schema)
```
shcheklein, line 16: why do we need this, not sure I understand ...

tests/func/test_delta.py
```diff
@@ -0,0 +1,286 @@
...
+from PIL import Image
+
+import datachain as dc
+from datachain import func
+from datachain.error import DatasetVersionNotFoundError
+from datachain.lib.dc import C
+from datachain.lib.file import File, ImageFile
+
+
+def test_delta_update_from_dataset(test_session, tmp_dir, tmp_path):
```
shcheklein, line 14: we should have a test with map and we should count number of maps applied in a regular mode and in delta (to make sure we don't do things twice)

tests/unit/lib/test_datachain.py
```diff
@@ -274,6 +274,10 @@ def test_read_record_empty_chain_without_schema(test_session):
     )
 
 
+def test_is_empty(test_session):
```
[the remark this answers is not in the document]
ilongin: No, I've added additional simple method in `DataChain` called `.is_empty()` .. I think it can come handy in general.

src/datachain/delta.py
ilongin: `File` signals is still needed in final dataset because of `changed / modified` rows. Delta update is looking for new and changed / modified rows in sources and those modified needs to be replaced in dataset i.e it's not enough just append new versions of rows and leave old versions in final dataset (this was task requirement from @dmpetrov ). In regards to this, file signals needs to be present in final dataset as well to be able to compare with `diff` and remove old versions of rows that are going to be replaced with new versions which are modified in source. 
Please let me know if you can think of way to avoid having file signals as from what I can think of they must be present.

src/datachain/delta.py
ilongin: Good question. I assumed all datasets are present locally and I think this will always be the case since the only case where one of the dependencies is not present locally, but in Studio, is when we pull dataset from it (we don't pull it's dependencies). But when using delta update, user is creating / saving dataset so this is not about pulling existing one.

src/datachain/delta.py
ilongin: We cannot do `diff` of source with itself (current vs old version from dependencvies) if file signals are not present, as `diff` is relying on it to be present. Can you explain how it can be avoided?

ilongin: Good point. This is not only good for avoid re-calculating but also to avoid extremely complex query at the end which is in risk of SQLite query size limit for example. I was avoiding persist as it slows things down a little bit but here it makes sense. BTW final query would be much simpler if there's not that requirement about changed / modified rows mentioned in above comment https://github.com/iterative/datachain/pull/928#discussion_r2063345238

Adding `.persist()`

src/datachain/lib/dc/datachain.py
ilongin: Sorry, leftover. Removing it.

src/datachain/delta.py
ilongin: Based on above comments, I've added `.persist()` and this comes after it so there will be no re-calculation. Without persist, yes there would be as `.is_empty()` uses `.count()` underneath which applies all steps.

src/datachain/lib/dc/datachain.py
ilongin: No, exception is thrown when method is called. I've put some tests for it and notice I didn't put `.save()` in them but they still work (throw exception)

src/datachain/lib/dc/datachain.py
ilongin: Yes, no need to be public, changing to private

src/datachain/lib/dc/storage.py
ilongin: Done

src/datachain/delta.py
ilongin: Actually this is not even reachable as the only way to get multiple dependencies is using one of those disabled methods like `.union()` so `NotImplementedError` is thrown before. Removing it.

src/datachain/lib/dc/storage.py
ilongin: Fixed by refactoring that part of the codebase a little bit

src/datachain/delta.py
ilongin: This is tricky part of the code. We want to run the chain on "diff" and currently it works in a way to add all the steps from original chain but also we need to append all the signals schema changes as some steps like `.map()` are adding new signals to the signals schema ( [link](https://github.com/iterative/datachain/blob/main/src/datachain/lib/dc/datachain.py#L558) ) .. btw I noticed that here instead of append we need to do `dc.signals_schema = other.signals_schema` as some steps might rename signals.

Also, when we refactor codebase to be able to run for all the sources this will not be needed as we will just run the whole chain in "special" mode and then all those `DatasetQuery.starting_step`s will just return "diff" instead of full data so we won't need tricky logics like this so this is only temporary.

src/datachain/lib/dc/datachain.py
shcheklein: so, when do we call the method? before we execute the whole datachain?

src/datachain/dataset.py
shcheklein: okay, I see ... should we be saving actual UUID , some id of the dataset? name since unreliable anyways, even if it is now the full name

src/datachain/delta.py
shcheklein: so, will I be able to use something like `from_dataset(studio=True, delta=True)`? (and it automatically pulls the versions of the dataset that I need .. wdyt?

src/datachain/delta.py
shcheklein: yeah, but the we should have a way to specify which field to use ...

also, I wonder if we should by default just don't use file, just don't update existing rows ... instead of this File magic - it seems like an edge case that we are making a general case

how can we introduce a setting on which we will be comparing?

src/datachain/delta.py
shcheklein: so, is this one done?

src/datachain/dataset.py
ilongin: That seems like a good idea, yes, but I would do it separately in other issue / PR as it's related to dataset dependencies more than to delta updates

src/datachain/delta.py
ilongin: Yes.

src/datachain/lib/dc/datachain.py
ilongin: Execution of datachain happens when applying steps (on `.save()` or `.collect()` etc) and this check is happening before when calling methods like `.union()`

src/datachain/delta.py
ilongin: Do you mean we should specify which `File` field to use, or we should specify which specific fields in general to use for matching / comparing (might not be one of `File` fields but something else)?
If later is the case: one of the requirements of the issue, when I was speaking with Dmitry, was to specifically use `File` signals and it's related method [`DataChain.diff(...)`](https://github.com/iterative/datachain/blob/main/src/datachain/lib/dc/datachain.py#L1401) to calculate diff (that one works only with files and btw should/will be renamed to `file_diff()` in future) in delta logic. 
If we are going to loose-up this requirement to allow user to specify arbitrary fields to do matching and comparing on to, we can do that as well by adding `on`, `right_on`, `compare`, `right_compare` arguments, and then to calculate diff we would need to use [`DataChain.compare()`](https://github.com/iterative/datachain/blob/main/src/datachain/lib/dc/datachain.py#L1325) which accepts those args as well (btw we should probably rename this method to general `diff()` but it's another topic).

This is more of a product decision (@dmpetrov adding you to the loop as well) but my suggestion is to use `File` fields to do matching / comparing on to for first version. If this is not enough, this can be changed in future to allow arbitrary fields. We will be changing interface a little bit but it should be backward compatible as well. Now we are using first found `File` field in datasets for diff and this will be problematic only if there are more than one of those present in dataset which is IMO pretty rare and not expected.

src/datachain/dataset.py
shcheklein: yep, let's create an issue for this ... I'm a bit surprised that we are using names to identify datasets (considering that we allow renaming)

src/datachain/delta.py
shcheklein: > If we are going to loose-up this requirement to allow user to specify arbitrary fields to do matching and comparing on to, we can do that as well by adding on, right_on, compare, right_compare arguments, and then to calculate diff we would need to use [DataChain.compare()](https://github.com/iterative/datachain/blob/main/src/datachain/lib/dc/datachain.py#L1325) which accepts those args as well (btw we should probably rename this method to general diff() but it's another topic).

yes, I mean this. I'm not sure why it all should be limited to files

src/datachain/delta.py
shcheklein: what happens if someone else is running this code and there are already a few versions of the dataset on Studio?

src/datachain/diff/__init__.py
shcheklein: is it fixed? can we cleanup this?

src/datachain/lib/dc/datachain.py
shcheklein: okay, good, just wanted to confirm. thanks!

src/datachain/lib/dc/datachain.py
```diff
@@ -66,11 +68,34 @@
...
-    from typing_extensions import ParamSpec, Self
+    from typing_extensions import Concatenate, ParamSpec, Self
 
     P = ParamSpec("P")
 
 
+T = TypeVar("T", bound="DataChain")
+
+
+def delta_disabled(
```
shcheklein, line 79: should it be protected (otherwise it can becomes visible in docs, no?)

shcheklein: btw, can we move it to delta.py ? to keep this file cleaner ...

src/datachain/lib/dc/datachain.py
```diff
@@ -176,6 +202,16 @@ def __repr__(self) -> str:
         self.print_schema(file=file)
         return file.getvalue()
 
+    def _as_delta(self, delta: bool = False) -> "Self":
```
shcheklein, line 205: can we make it a regular ctr parameter (like we do with all other attrs) ... why do we need a special way of setting this up?

src/datachain/lib/dc/datachain.py
```diff
@@ -176,6 +202,16 @@ def __repr__(self) -> str:
...
+        """Marks this chain as delta, which means special delta process will be
+        called on saving dataset for optimization"""
+        self._delta = delta
+        return self
+
+    @property
+    def delta(self) -> bool:
```
shcheklein, line 212: is it public? do we want it to be public?

src/datachain/lib/dc/datachain.py
```diff
@@ -472,6 +508,25 @@ def save(  # type: ignore[override]
                 or with a value, e.g "location=US".
         """
         schema = self.signals_schema.clone_without_sys_signals().serialize()
+        if self.delta and name:
+            delta_ds, has_changes = delta_update(self, name)
```
shcheklein, line 512: just to make sure - does it mean that the whole chain is evaluated at this point? (it's probably fine for save, i'm just trying to understand how execution was triggered before in save - at what point, where and if there are any implications of this change (e.g. some downstream code affected)

src/datachain/lib/dc/datachain.py
```diff
@@ -2145,6 +2205,10 @@ def count(self) -> int:
         """Return the number of rows in the chain."""
         return self._query.count()
 
+    def is_empty(self) -> bool:
```
shcheklein, line 2208: if we use `is_` notation, we should probably use it consistently - please check in this PR (e..g is_delta) and if we have some other bool methods

shcheklein (commented):
Can we test it on the cattle care scenario? Video files, we extract and save frames for them / or extract and save subvideos ... (it is similar to one of the tutorias). Should it be working for such scenario? (let me know if you need access to the code)

ilongin:
> Can we test it on the cattle care scenario? Video files, we extract and save frames for them / or extract and save subvideos ... (it is similar to one of the tutorias). Should it be working for such scenario? (let me know if you need access to the code)

Yes, please give me access to the code as I dont think I have it

shcheklein:
> Yes, please give me access to the code as I dont think I have it

DMed you the link!

src/datachain/diff/__init__.py
ilongin: No, I will create separate issue to deal with this later on. I don't think it's high priority as this workaround works fine for now. It's tricky to fix though so that's why I don't want to spent too much time on it now

src/datachain/lib/dc/datachain.py
ilongin: Moved to `delta.py`

src/datachain/lib/dc/datachain.py
ilongin: User should never set this parameter so I wanted to "hide" it a little bit by making it a private attribute and special private method to set it to `True`

src/datachain/lib/dc/datachain.py
ilongin: In `delta_update()` function only calculating diff is evaluated and saved in temp table (we call `.persist()`). The rest of the chain (using that persisted diff, union it with current dataset etc.) is evaluated when calling `delta_ds._query.save(...)`.

src/datachain/lib/dc/datachain.py
ilongin: Since it's true we don't have `is_` notation in  `DataChain` I've moved to simple property `empty` method

src/datachain/dataset.py
ilongin: Created issue https://github.com/iterative/datachain/issues/1087

src/datachain/lib/dc/datachain.py
shcheklein: no, that's fine .. but if we keep it public we need proper docs for it then ... and an example if you have something in mind

src/datachain/lib/dc/datachain.py
shcheklein: I think users never actually use DataChain ctr anyway - it is kinda private already (and probably it has some attributes already that are technical, not use facing)

src/datachain/diff/__init__.py
shcheklein: when does it happen? if we have an issue with context - let's put a link in this todo please ... it should be clear as much as possible from the context what is happening here (and why)

what is `sys=True` - it's not documented it seems (why?)

src/datachain/delta.py
shcheklein: so, should we actually check / pull two versions from Studio in this case (initial run) - to be able to get the diff?

src/datachain/delta.py
ilongin: I've pushed the version that doesn't uses `File` fields by default, but user should set list of fields for matching and comparing. Let me know what do you think.

src/datachain/lib/dc/datasets.py
```diff
@@ -37,6 +43,32 @@ def read_dataset(
...
+            fields to determine if a record has changed. Note that this process only
+            considers added and modified records in storage; deleted records are not
+            removed from the new dataset version.
+            This calculation is based on the difference between the current version
+            of the source and the version used to create the dataset.
+        delta_on: A list of fields that uniquely identify rows in the source.
+            If two rows have the same values, they are considered the same (e.g., they
+            could be different versions of the same row in a versioned source).
+            This is used in the delta update to calculate the diff.
+        delta_right_on: A list of fields in the final dataset that correspond to the
```
shcheklein, line 65: what is the final dataset? is it the result itself of the whole chain?

should we then specify this in `save`? 🤔

ilongin: Yes, that's it. I don't think we should specify this in `.save()` as these fields can be specific to source. For example when we will be able to handle multiple sources in delta update each source could have it's own  match fields that correspond to different fields in final dataset. I was thinking to rename this field but haven't find better name...

shcheklein: yep, `right_on` (why right, why not left?) is not clear a all 

> we will be able to handle multiple sources in delta update each source could have it's own match fields that correspond to different fields in final dataset

are we going to do subtract per source? is it pretty much about changed / removed objects only?

ilongin: > yep, `right_on` (why right, why not left?) is not clear a all

I've updated it to `delta_result_on` and added better explanation in docs. Let me know if this is better.

> are we going to do subtract per source? is it pretty much about changed / removed objects only?

Currently delta works only for one (first) "starting" point e.g `from_storage(...)` or `.from_dataset(...)` but in future it will work for any starting point in chain e.g here we have 2 "starting points":
```python
(
    chain.from_storage("s3://first-bucket", delta=True, delta_on=["id_first"], delta_result_on=["id"])
    .map(...)
    .union(dc.from_storage("s3://second-bucket", delta=True, delta_on=["id_second"], delta_result_on=["id"]))
    .mutate(id=ifelse(isnone(chain.C("id_first") ), chain.C("id_second"), chain.C("id_first)))
    .select_except("id_first", "id_second")
)
```
In this example `id_first` and `id_second` are all normalized to just `id` in final dataset and then we need to use that to match "diff" datasets with final dataset. Matching "diff" datasets with final datasets is important to keep only latest modified files i.e. remove old ones from final dataset.

src/datachain/lib/dc/storage.py
```diff
@@ -48,6 +54,36 @@ def read_storage(
...
+            removed from the new dataset version.
+            This calculation is based on the difference between the current version
+            of the source and the version used to create the dataset.
+        delta_on: A list of fields that uniquely identify rows in the source.
+            If two rows have the same values, they are considered the same (e.g., they
+            could be different versions of the same row in a versioned source).
+            This is used in the delta update to calculate the diff.
+        delta_result_on: A list of fields in the resulting dataset that correspond
+            to the `delta_on` fields from the source.
+            This is needed to identify rows that have changed in the source but are
```
shcheklein, line 78: do we / can we detect deletions btw?

shcheklein (approved):
If we tested at least on some real scenarios, let's merge it.

src/datachain/lib/dc/datachain.py
shcheklein: one minor things that is left here @ilongin  ... let's please take care of it

ilongin:
> If we tested at least on some real scenarios, let's merge it.

Testing today on cattle-care for some scripts...  found some issue with dataset dependencies, pushed a fix today. I want to test a little bit more today evening and then merge

src/datachain/lib/dc/storage.py
ilongin: at beginning we decided to ignore deletions for now. I do think we should re-consider this

src/datachain/delta.py
shcheklein: so, this one is also not resolved yet ... can be a followup, but it seems like an issue for me

PR #931, opened 2025-02-20
Add option to specify the hidden field

### LINKED ISSUE #830: Hidden fields
### Description

Users usually only need to see basic fields like `path`, `size` and `etag` when viewing File. We should hide other fields by default to reduce UI clutter, while keeping them accessible when needed.

We need a way to mark rarely used fields as "internal" in our data model.

Example:

```python
class File(DataModel):
    """`DataModel` for reading binary files."""

    source: str = Field(default="", hidden=True)
    path: str
    size: int = Field(default=0)
    version: str = Field(default="", hidden=True)
    etag: str = Field(default="")
```

Note, it should be general approach. User should be able to have hidden fields in their classes.

Open question: 
- Should `size` be hidden? Likely yes.
- `etag` should be visible (to emphasise the consistency and see changes). It looks like it needs to be moved next to path (before `size` and `version`).

amritghimire:
For any hidden fields, a `_hidden_fields` class variable will hide the
fields from the values when showing using `dc.show`.

Example on how to hide fields for custom model:
```python

from datachain import DataModel, DataChain


class InnerClass(DataModel):
    inner_value: float
    hide_inner: float

    _hidden_fields = ["hide_inner"]


class OuterClass(DataModel):
    outer_value: float
    hide_outer: float

    inner_object: InnerClass
    _hidden_fields = ["hide_outer"]


inner = InnerClass(inner_value=1.1, hide_inner=1.2)
outer = OuterClass(outer_value=1.3, hide_outer=1.4, inner_object=inner)
DataChain.from_values(
        outer=[outer],
        nums=[1]
).show()
```

This also takes care when we display the datachain using following command:
```
datachain show az
```

Similarly, https://github.com/iterative/studio/pull/11357 takes care of the UX in Studio. 

Test is also added.

Closes #830

### DIFF
--- a/src/datachain/cli/commands/show.py
+++ b/src/datachain/cli/commands/show.py
@@ -16,6 +16,7 @@ def show(
     columns: Sequence[str] = (),
     no_collapse: bool = False,
     schema: bool = False,
+    include_hidden: bool = False,
 ) -> None:
     from datachain import Session
     from datachain.lib.dc import DataChain
@@ -24,9 +25,13 @@ def show(
 
     dataset = catalog.get_dataset(name)
     dataset_version = dataset.get_version(version or dataset.latest_version)
-    hidden_fields = SignalSchema.get_flatten_hidden_fields(
-        dataset_version.feature_schema
-    )
+
+    if include_hidden:
+        hidden_fields = []
+    else:
+        hidden_fields = SignalSchema.get_flatten_hidden_fields(
+            dataset_version.feature_schema
+        )
 
     query = (
         DatasetQuery(name=name, version=version, catalog=catalog)
--- a/src/datachain/cli/commands/show.py
+++ b/src/datachain/cli/commands/show.py
@@ -1,6 +1,8 @@
 from collections.abc import Sequence
 from typing import TYPE_CHECKING, Optional
 
+from datachain.lib.signal_schema import SignalSchema
+
 if TYPE_CHECKING:
     from datachain.catalog import Catalog
 
@@ -22,6 +24,7 @@ def show(
 
     dataset = catalog.get_dataset(name)
     dataset_version = dataset.get_version(version or dataset.latest_version)
+    hidden_fields = SignalSchema.get_hidden_fields(dataset_version.feature_schema)
 
     query = (
         DatasetQuery(name=name, version=version, catalog=catalog)
@@ -30,7 +33,8 @@ def show(
         .offset(offset)
     )
     records = query.to_db_records()
-    show_records(records, collapse_columns=not no_collapse)
+    show_records(records, collapse_columns=not no_collapse, hidden_fields=hidden_fields)
+
     if schema and dataset_version.feature_schema:
         print("\nSchema:")
         session = Session.get(catalog=catalog)
--- a/src/datachain/cli/commands/show.py
+++ b/src/datachain/cli/commands/show.py
@@ -24,7 +24,9 @@ def show(
 
     dataset = catalog.get_dataset(name)
     dataset_version = dataset.get_version(version or dataset.latest_version)
-    hidden_fields = SignalSchema.get_hidden_fields(dataset_version.feature_schema)
+    hidden_fields = SignalSchema.get_flatten_hidden_fields(
+        dataset_version.feature_schema
+    )
 
     query = (
         DatasetQuery(name=name, version=version, catalog=catalog)
--- a/src/datachain/lib/dc.py
+++ b/src/datachain/lib/dc.py
@@ -1254,7 +1254,7 @@ class DataChain:
         )
         with self._query.ordered_select(*db_signals).as_iterable() as rows:
             if row_factory:
-                rows = (row_factory(db_signals, r) for r in rows)  # type: ignore[assignment]
+                rows = (row_factory(db_signals, r) for r in rows)
             yield from rows
 
     def to_columnar_data_with_names(
--- a/src/datachain/lib/dc.py
+++ b/src/datachain/lib/dc.py
@@ -65,7 +65,6 @@ _T = TypeVar("_T")
 D = TypeVar("D", bound="DataChain")
 UDFObjT = TypeVar("UDFObjT", bound=UDFBase)
 
-
 DEFAULT_PARQUET_CHUNK_SIZE = 100_000
 
 
@@ -1226,22 +1225,36 @@ class DataChain:
     def collect_flatten(self) -> Iterator[tuple[Any, ...]]: ...
 
     @overload
+    def collect_flatten(self, *, exclude_hidden: bool) -> Iterator[tuple[Any, ...]]: ...
+
+    @overload
     def collect_flatten(
         self, *, row_factory: Callable[[list[str], tuple[Any, ...]], _T]
     ) -> Iterator[_T]: ...
 
-    def collect_flatten(self, *, row_factory=None):
+    @overload
+    def collect_flatten(
+        self,
+        *,
+        row_factory: Callable[[list[str], tuple[Any, ...]], _T],
+        exclude_hidden: bool,
+    ) -> Iterator[_T]: ...
+
+    def collect_flatten(self, *, row_factory=None, exclude_hidden: bool = False):
         """Yields flattened rows of values as a tuple.
 
         Args:
             row_factory : A callable to convert row to a custom format.
                           It should accept two arguments: a list of column names and
                           a tuple of row values.
+            exclude_hidden: Whether to exclude hidden signals from the schema.
         """
-        db_signals = self._effective_signals_schema.db_signals()
+        db_signals = self._effective_signals_schema.db_signals(
+            exclude_hidden=exclude_hidden
+        )
         with self._query.ordered_select(*db_signals).as_iterable() as rows:
             if row_factory:
-                rows = (row_factory(db_signals, r) for r in rows)
+                rows = (row_factory(db_signals, r) for r in rows)  # type: ignore[assignment]
             yield from rows
 
     def to_columnar_data_with_names(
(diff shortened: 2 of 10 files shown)


shcheklein (commented):
we already have `sys_id` and may be some other fields - how do we manage them? (how and where we filter them)?

amritghimire:
> we already have `sys_id` and may be some other fields - how do we manage them? (how and where we filter them)?

We manually hide them for eg https://github.com/iterative/studio/blob/main/frontend/src/composite/DatasetsEditor/PythonQueryTab/DatasetObjectsTable/columns.tsx

src/datachain/lib/signal_schema.py
```diff
@@ -384,6 +392,37 @@ def deserialize(schema: dict[str, Any]) -> "SignalSchema":
 
         return SignalSchema(signals)
 
+    @staticmethod
```
shcheklein, line 395: Q: why is it a static method?

Q: should we rename it? get_flatten_hidden_fields or something like that ...

src/datachain/lib/dc.py
```diff
@@ -1225,23 +1224,37 @@ def _effective_signals_schema(self) -> "SignalSchema":
...
+            exclude_hidden: Whether to exclude hidden signals from the schema.
         """
-        db_signals = self._effective_signals_schema.db_signals()
+        db_signals = self._effective_signals_schema.db_signals(
+            exclude_hidden=exclude_hidden
+        )
         with self._query.ordered_select(*db_signals).as_iterable() as rows:
             if row_factory:
-                rows = (row_factory(db_signals, r) for r in rows)
+                rows = (row_factory(db_signals, r) for r in rows)  # type: ignore[assignment]
```
shcheklein, line 1257: why do we need this type: ignore? can we fix it?

src/datachain/cli/commands/show.py
```diff
@@ -22,6 +24,7 @@ def show(
 
     dataset = catalog.get_dataset(name)
     dataset_version = dataset.get_version(version or dataset.latest_version)
+    hidden_fields = SignalSchema.get_hidden_fields(dataset_version.feature_schema)
```
shcheklein, line 27: is there a way to avoid introducing the `get_hidden_fields` and use get_flat_tree that does filtering already?

it feels we are implementing it in a few places in two different ways, and this one is only for show to work 

can we use `get_flat_tree` to get the list of fields?

amritghimire: `get_hidden_fields` was initially for studio implementation. We do need a list of the hidden fields in Studio to mark them as hidden in UX.

src/datachain/lib/signal_schema.py
amritghimire: It is static method because it doesn't really use any values from the class and makes it easy to call from Studio.

src/datachain/lib/dc.py
amritghimire: Not sure, it was throwing CI error even if the line was not touched. We could come back to fix this but seems like out of scope of this PR.

src/datachain/lib/signal_schema.py
amritghimire: Renamed to get_flatten_hidden_fields.

src/datachain/lib/dc.py
shcheklein: hmm, is CI broken atm? or does it happen only for this PR?

src/datachain/lib/signal_schema.py
shcheklein: I mean, should be just schema.get_flatten_hidden_fields() instead (instance method). It's a bit strange to see that we class method and pass pretty much an instance of the same class

src/datachain/cli/commands/show.py
shcheklein: same question - can we use get_flat_tree? or implement `get_hidden_fields` using `get_flat_tree`?

amritghimire: I don't understand what benefit will that achieve by complicating `get_flat_tree`? `get_flat_tree` is to get the list of the fields in a schema. It will not get into the `_hidden_fields` class variable. It just removes the hidden field that it finds in the current schema. We need the list of hidden field explicitly here.

src/datachain/lib/signal_schema.py
amritghimire: I see the confusion, we are not passing the schema object itself, but we are passing the serialized part of feature schema we have at the moment inside the dataset version. We don't need to form a signalschema object just for this IMO.

src/datachain/cli/commands/show.py
amritghimire: I believe these two should be separate just for simplicity and for the principle of separation of concern. These two functions serves entirely different purposes.

src/datachain/lib/dc.py
amritghimire: It was pre commit hook that was causing the issue which runs only on changed file

shcheklein: so, if we run the full mypy on the full code base in main it fails, correct?

src/datachain/lib/signal_schema.py
shcheklein: Ok, I see. Parameter should be renamed then? So, it means we kinda expose some internals (operate on the serialized format). What if that format changes? It means we'll have to have some backward compatibility? Is it compatible with version 1? (before things like bases were introduced)? are we testing this per format?

src/datachain/cli/commands/show.py
shcheklein: okay, may be I'm missing something ... what does `get_flat_tree` return?

amritghimire: It returns the information about the fields in flattened structure like `([field1], [field2, inside_field2])` and so on.

src/datachain/lib/signal_schema.py
amritghimire: It is from dataset versions we save in database.  It does have backward compatibility because we access the field like below if you check the PR changes:
```
schema.get("hidden_fields", [])
```
and so on.

src/datachain/cli/commands/show.py
shcheklein: and `inside_field2` can also be an array, or it is more like? is it a single level of arrays?

src/datachain/lib/dc.py
amritghimire: Not sure, let me try to remove it and push it and see how it behaves in CI.

src/datachain/cli/commands/show.py
shcheklein: so, can we do `_`.join() and get the list of fields in the same format as that second method returns?

amritghimire: Yes, we can get the list of fields in same format. But the issue is not with the format but with what is returned. The method `SignalSchema.get_hidden_fields` now `SignalSchema.get_flatten_hidden_fields` returns only the hidden field. We will not need the list of all the fields we have. 

This `get_flat_tree` returns all the fields that we have. The flag `include_hidden` introduced with the PR handles whether we want to include the hidden field among it or not, but not only hidden field. 

If you have strong opinion on reusing the get_flat_tree, the only way I can see is reusing it two times, one with `include_hidden` to true to get all the fields, next time with `include_hidden` to false to get the list without hidden fields and then calculate the difference. Not sure it is the way to go.

As I have mentioned, the `get_flat_tree` and `get_flatten_hidden_fields` serves two completely different purposes.

shcheklein: can `show` use this and pass `include_hidden` (instead of getting hidden fields and filtering)? That's the only place where `get_flatten_hidden_fields` is needed, right?

amritghimire: `get_flatten_hidden_fields` is needed in Studio as well to get the list of hidden field to show in columns filtering.

src/datachain/lib/signal_schema.py
shcheklein: but it means we'll have to always now keep in mind in one extra place this compatibility aspect? can we avoid using serialized format? (I think @dreadatour is kinda looking into a similar issue btw)

amritghimire: I don't think so, I mean we already are using the feature schema in other places directly in studio and probably in datachain. I think having to create signal schema object just to parse the certain fields within sounds expensive to me specially when listing the versions.

amritghimire: We maintain the format in CustomType object in itself in the same file for this.

shcheklein: > and probably in datachain

hopefully not (?)

> in other places directly in studio

means it should be improved if we see more place coming (like in this case)?

>  I think having to create signal schema object just to parse the certain fields within sounds expensive to me specially when listing the versions.

it's a single `deserialize` call on a smallish dict unless I'm missing something? what does deserialize do, does it create models and such?

amritghimire: > it's a single deserialize call on a smallish dict unless I'm missing something? what does deserialize do, does it create models and such?

Deserialization creates a signal based on the field. The issue is the deserialized signal doesn't have the type definition with it so, we did not have the hidden signal with us. I tried implementing it in  e6767be997ed09822e5851a99b996002b075b1f9 but it did not work in studio.  We need to look into the feature schema as we are already doing in Studio already.

amritghimire: I honestly do not see the benefit in calling a class method by initializing a class instead of a static class.  TBH

src/datachain/lib/dc.py
amritghimire: It did not work for some reason. We can come back as a followup but I don't know how to fix this currently honestly

shcheklein (approved):

PR #935, opened 2025-02-27
Fix the to_storage for relative path

amritghimire:
This fixes the case where the path is passed to to_storage

### DIFF
--- a/src/datachain/lib/file.py
+++ b/src/datachain/lib/file.py
@@ -275,7 +275,7 @@ class File(DataModel):
         client: Client = self._catalog.get_client(destination)
 
         if client.PREFIX == "file://" and not destination.startswith(client.PREFIX):
-            destination = client.PREFIX + os.path.abspath(destination)
+            destination = Path(destination).absolute().as_uri()
 
         client.upload(self.read(), destination)
 
--- a/src/datachain/lib/file.py
+++ b/src/datachain/lib/file.py
@@ -272,8 +272,12 @@ class File(DataModel):
     def save(self, destination: str):
         """Writes it's content to destination"""
         destination = stringify_path(destination)
-        client: Client = self._catalog.get_client(str(destination))
-        client.upload(self.read(), str(destination))
+        client: Client = self._catalog.get_client(destination)
+
+        if client.PREFIX == "file://" and not destination.startswith(client.PREFIX):
+            destination = f"{client.PREFIX}{destination}"
+
+        client.upload(self.read(), destination)
 
     def _symlink_to(self, destination: str):
         if self.location:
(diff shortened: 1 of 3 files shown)


skshetry:
Where does this relative path come from? Why don't we pass the full path to the `upload()`?

amritghimire:
> Where does this relative path come from? Why don't we pass the full path to the `upload()`?

https://iterativeai.slack.com/archives/C04A9RWEZBN/p1740628663390679

skshetry:
Not sure that I understand, could you please elaborate? I have read that thread.

amritghimire:
> Not sure that I understand, could you please elaborate? I have read that thread.

Updated the code. Basically the relative path is provided when running the script to to_storage. Moved the logic to to_storage instead.

skshetry:
This should be fixed in `export()` and/or `save()` as you could pass the relative paths there too, not in `to_storage()`.

Also, a few other things that I noticed:

1. Now that `export` supports exporting to any filesystems, it should not using `os.path`.


https://github.com/iterative/datachain/blob/88cee30f7ea495676a395b97bbcf69785f1be775/src/datachain/lib/file.py#L304


2. Should `export()` be renamed to be in symmetry with `to_storage()`?

src/datachain/lib/file.py
```diff
@@ -272,8 +272,12 @@ def read_text(self):
     def save(self, destination: str):
         """Writes it's content to destination"""
         destination = stringify_path(destination)
-        client: Client = self._catalog.get_client(str(destination))
-        client.upload(self.read(), str(destination))
+        client: Client = self._catalog.get_client(destination)
+
+        if client.PREFIX == "file://" and not destination.startswith(client.PREFIX):
+            destination = f"{client.PREFIX}{destination}"
```
skshetry, line 278: There is no such thing as relative `file://` URI. The path needs to be converted to an absolute path, and then to an URI.

src/datachain/lib/file.py
```diff
@@ -272,8 +272,12 @@ def read_text(self):
     def save(self, destination: str):
         """Writes it's content to destination"""
         destination = stringify_path(destination)
-        client: Client = self._catalog.get_client(str(destination))
-        client.upload(self.read(), str(destination))
+        client: Client = self._catalog.get_client(destination)
+
+        if client.PREFIX == "file://" and not destination.startswith(client.PREFIX):
+            destination = client.PREFIX + os.path.abspath(destination)
```
skshetry, line 278: This is still not correct. `file://` URI uses forward-slash as a separator, but by using `os.path`, the separator could be a backward-slash (`\\`) here.

You can use either of the options:
```python
from pathlib import Path
Path(destination).absolute().as_uri()
```

or, `urllib.request.pathname2url` or, `make_path_posix` from fsspec.

dreadatour (approved):
🙏 looks good to me!

PR #941, opened 2025-03-04
support Python 3.13

skshetry:


0x2b3bfa0 (approved):
As long as it passes tests...

noxfile.py
```diff
@@ -8,6 +11,9 @@
 nox.options.default_venv_backend = "uv|virtualenv"
 nox.options.reuse_existing_virtualenvs = True
 nox.options.sessions = "lint", "tests"
+
+project = nox.project.load_toml()
+python_versions = nox.project.python_versions(project)
```
skshetry, line 16: This uses a new API: [nox.project.python_versions](https://nox.thea.codes/en/stable/config.html#nox.project.python_versions) that reads from `classifiers` from `pyproject.toml`.

noxfile.py
```diff
@@ -93,3 +99,7 @@ def examples(session: nox.Session) -> None:
         "examples",
         *session.posargs,
     )
+
+
+if __name__ == "__main__":
+    nox.main()
```
skshetry, line 105: Added this and above inline metadata so that `uv run noxfile.py ...` works.

skshetry:
Note: I have kept the studio tests to run with 3.12 so that it is in sync with what we run in Studio.

https://github.com/iterative/datachain/blob/ddfb2f6394e26ef56d9f651c8657c84c75ce31c5/.github/workflows/tests-studio.yml#L23

amritghimire (approved):

PR #942, opened 2025-03-04
Fix ffmpeg setup in GitHub CI

dreadatour:
FFmpeg setup is broken for now because of rate limiting.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -80,7 +80,7 @@ jobs:
 
       - name: Set up FFmpeg
         if: runner.os != 'macOS' || runner.arch != 'ARM64'
-        uses: FedericoCarboni/setup-ffmpeg@v3
+        uses: FedericoCarboni/setup-ffmpeg@v3.1
         with:
           ffmpeg-version: release
           github-token: ${{ github.server_url == 'https://github.com' && github.token || '' }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -79,14 +79,15 @@ jobs:
           ref: ${{ github.event.pull_request.head.sha || github.ref }}
 
       - name: Set up FFmpeg
-        if: runner.os != 'macOS' || runner.arch != 'ARM64'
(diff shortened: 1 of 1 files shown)


skshetry:
This github action does not seem to support macOS arm64.

.github/workflows/tests.yml
```diff
@@ -79,14 +79,15 @@ jobs:
...
-          ffmpeg-version: release
-          github-token: ${{ github.server_url == 'https://github.com' && github.token || '' }}
-      - name: Set up FFmpeg
-        if: runner.os == 'macOS' && runner.arch == 'ARM64'
+        uses: AnimMouse/setup-ffmpeg@v1
+        id: ffmpeg-install
+        continue-on-error: true
+
+      # https://github.com/AnimMouse/setup-ffmpeg/issues/5
+      - if: steps.ffmpeg-install.conclusion == 'failure' && runner.os == 'macOS'
```
skshetry, line 87:
```suggestion
      - if: steps.ffmpeg-install.outcome == 'failure' && runner.os == 'macOS'
```

skshetry: > The result of a completed step after [continue-on-error](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) is applied. Possible values are success, failure, cancelled, or skipped. When a continue-on-error step fails, the outcome is failure, but the final conclusion is success.

From https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#steps-context.

I made a mistake on my PR.

skshetry (approved):

.github/workflows/tests.yml
```diff
@@ -80,6 +80,14 @@ jobs:
 
       - name: Set up FFmpeg
         uses: AnimMouse/setup-ffmpeg@v1
+        id: ffmpeg-install
+        continue-on-error: true
```
skshetry, line 84:
```suggestion
        continue-on-error: runner.os == 'macOS'
```

PR #956, opened 2025-03-10
Add studio support in DataChain.datasets()

### LINKED ISSUE #952: `DataChain.datasets()` support Studio
Need to have a flag to being able to see all datasets (similar to how we already do ls I think) in Studio.

Let's also make sure that the flag name alight with from_dataset. Probably we should rename it to just basic `studio=True/False` cc @amritghimire 

We need this to be able to calculate, print statistics across datasets in time.

amritghimire:
This adds two arguments named `studio: bool` and `team: str|None` for
the datasets class method.

Closes #952

### DIFF
--- a/src/datachain/catalog/catalog.py
+++ b/src/datachain/catalog/catalog.py
@@ -1167,6 +1167,8 @@ class Catalog:
     def ls_datasets(
         self, include_listing: bool = False, studio: bool = False
     ) -> Iterator[DatasetListRecord]:
+        from datachain.remote.studio import StudioClient
+
         if studio:
             client = StudioClient()
             response = client.ls_datasets()
--- a/src/datachain/catalog/catalog.py
+++ b/src/datachain/catalog/catalog.py
@@ -1753,17 +1753,16 @@ class Catalog:
         )
 
     def studio_dataset_versions(
-        self, team: Optional[str] = None, include_listing=False
+        self, include_listing=False
     ) -> Iterator[tuple[DatasetListRecord, "DatasetListVersion", Optional["Job"]]]:
         """Get Studio versions from Studio.
 
         Parameters:
-            team: Team name.
             include_listing: Whether to include listing datasets.
         """
         from datachain.remote.studio import StudioClient
 
-        client = StudioClient(team=team)
+        client = StudioClient()
 
         response = client.ls_datasets()
 
--- a/src/datachain/catalog/catalog.py
+++ b/src/datachain/catalog/catalog.py
@@ -1751,3 +1751,41 @@ class Catalog:
             client_config=client_config or self.client_config,
             only_index=True,
         )
+
+    def studio_dataset_versions(
+        self, team: Optional[str] = None, include_listing=False
+    ) -> Iterator[tuple[DatasetListRecord, "DatasetListVersion", Optional["Job"]]]:
+        """Get Studio versions from Studio.
+
+        Parameters:
+            team: Team name.
+            include_listing: Whether to include listing datasets.
+        """
+        from datachain.remote.studio import StudioClient
+
+        client = StudioClient(team=team)
+
+        response = client.ls_datasets()
+
+        if not response.ok:
+            raise DataChainError(response.message)
+
+        if not response.data:
+            return
+
+        datasets = [DatasetListRecord.from_dict(d) for d in response.data]
+
+        # preselect dataset versions jobs from db to avoid multiple queries
+        jobs_ids: set[str] = {
+            v.job_id for ds in datasets for v in ds.versions if v.job_id
+        }
+        jobs: dict[str, Job] = {}
+        if jobs_ids:
+            jobs = {j.id: j for j in self.metastore.list_jobs_by_ids(list(jobs_ids))}
+
+        for d in datasets:
+            yield from (
+                (d, v, jobs.get(str(v.job_id)) if v.job_id else None)
+                for v in d.versions
+                if not d.is_bucket_listing or include_listing
+            )
--- a/src/datachain/catalog/catalog.py
+++ b/src/datachain/catalog/catalog.py
@@ -1164,8 +1164,25 @@ class Catalog:
 
         return direct_dependencies
 
-    def ls_datasets(self, include_listing: bool = False) -> Iterator[DatasetListRecord]:
-        datasets = self.metastore.list_datasets()
+    def ls_datasets(
+        self, include_listing: bool = False, studio: bool = False
+    ) -> Iterator[DatasetListRecord]:
+        if studio:
+            client = StudioClient()
+            response = client.ls_datasets()
+            if not response.ok:
+                raise DataChainError(response.message)
+            if not response.data:
+                return
+
+            datasets: Iterator[DatasetListRecord] = (
+                DatasetListRecord.from_dict(d)
+                for d in response.data
+                if not d.get("name", "").startswith(QUERY_DATASET_PREFIX)
+            )
+        else:
+            datasets = self.metastore.list_datasets()
+
         for d in datasets:
             if not d.is_bucket_listing or include_listing:
                 yield d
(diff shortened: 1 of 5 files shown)


src/datachain/lib/dc.py
```diff
@@ -731,6 +745,44 @@ def datasets(
...
+        response = client.ls_datasets()
+
+        if not response.ok:
+            raise DataChainError(response.message)
+
+        if not response.data:
+            return
+
+        for d in response.data:
+            dataset = DatasetListRecord(
```
amritghimire, line 774: Is there a better option to form dataset list record from the response dictionary? cc. @dreadatour

src/datachain/lib/dc.py
```diff
@@ -731,6 +745,44 @@ def datasets(
...
+        cls,
+        team: Optional[str] = None,
+    ) -> Iterator[DatasetListRecord]:
+        """Get datasets from Studio.
+
+        Parameters:
+            team: Team name.
+
+        Returns:
+            Iterator[DatasetInfo]: Iterator of DatasetInfo objects.
```
amritghimire, line 759:
```suggestion
            Iterator[DatasetListRecord]: Iterator of DatasetInfo objects.
```

dreadatour (commented):
Code look good to me, but I am sure we can improve it a bit 👀
Couple comments below, please, take a look? 🙏 Will be happy to discuss it further ❤️

src/datachain/lib/dc.py
```diff
@@ -723,6 +728,15 @@ def datasets(
             )
         ]
 
+        if studio:
+            # Add datasets from Studio
+            token = Config().read().get("studio", {}).get("token")
+            if token:
+                for d in cls.studio_datasets(team=team):
+                    datasets.extend(
+                        [DatasetInfo.from_models(d, v, None) for v in d.versions]
```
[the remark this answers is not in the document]
ilongin: I would vote against adding these flags into internal classes as much as we can, otherwise we will couple our code with Studio too much. If they can only be in public method, that would be ideal.

ilongin (changes requested):

src/datachain/dataset.py
```diff
@@ -302,6 +302,7 @@ def parse(
         size: Optional[int],
         query_script: str = "",
         job_id: Optional[str] = None,
+        **kwargs,
```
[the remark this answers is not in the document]
amritghimire: Some field like created_by_id and so on are not relevant to datachain.

src/datachain/lib/dc.py
amritghimire: Thanks, created a from_dict, moved the logic to catalog to return the proper types as well.

src/datachain/lib/dc.py
```diff
@@ -731,6 +745,44 @@ def datasets(
             **{object_name: datasets},  # type: ignore[arg-type]
         )
 
+    @classmethod
+    def studio_datasets(
```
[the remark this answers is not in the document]
amritghimire: Moved.

src/datachain/catalog/catalog.py
```diff
@@ -1751,3 +1751,41 @@ def index(
             client_config=client_config or self.client_config,
             only_index=True,
         )
+
+    def studio_dataset_versions(
```
shcheklein, line 1755: can it be abstracted somehow via metastore client or something? 

if not, should we move implementation into some studio related package? 

It feels studio stuff should not be part of the API (even a bit hidden one like catalog)

src/datachain/lib/dc.py
```diff
@@ -723,6 +726,16 @@ def datasets(
             )
         ]
 
+        if studio and Config().read().get("studio", {}).get("token"):
```
shcheklein, line 729: I think we should be throwing an exception if `studio` is specified and there is no token and team set.

src/datachain/lib/dc.py
```diff
@@ -701,6 +702,8 @@ def datasets(
         in_memory: bool = False,
         object_name: str = "dataset",
         include_listing: bool = False,
+        studio: bool = True,
+        team: Optional[str] = None,
```
shcheklein, line 706: let's not complicate it for now? keep it just studio

src/datachain/lib/dc.py
```diff
@@ -723,6 +726,16 @@ def datasets(
             )
         ]
 
+        if studio and Config().read().get("studio", {}).get("token"):
+            datasets.extend(
```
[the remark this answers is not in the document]
amritghimire: No, we are forming same object.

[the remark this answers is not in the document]
shcheklein: I think we need a way to distinguish them, or we need to make it exclusive - return _only_ studio datasets

src/datachain/lib/dc.py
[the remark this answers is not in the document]
dreadatour: > I would vote against adding these flags into internal classes as much as we can, otherwise we will couple our code with Studio too much. If they can only be in public method, that would be ideal.

Hm, I agree. But still, we need to be able to distinguish these datasets. What are the better options? 🤔

src/datachain/catalog/catalog.py
amritghimire: The part to fetch the datasets from studio still lies in Studio Client. This is to parse the dataset and get the dataset versions or process it accordingly just similar to how we do the list for the `list_datasets_versions` . Since this serves the same purpose as the same. I think this should be here .

amritghimire: Also, there is some logic to fetch the remote dataset and other.

amritghimire:
@shcheklein @ilongin @dreadatour Updated the code to make it exclusive, so we will either get only studio datasets or local only.

src/datachain/catalog/catalog.py
shcheklein: I still think we should not have `studio_...` named methods in API. How about `list_datasets_versions` accepting `studio=True` for example?

src/datachain/lib/dc.py
```diff
@@ -701,9 +701,22 @@ def datasets(
...
         """Generate chain with list of registered datasets.
 
+        Args:
+            session: Optional session instance. If not provided, uses default session.
+            settings: Optional dictionary of settings to configure the chain.
+            in_memory: If True, creates an in-memory session. Defaults to False.
+            object_name: Name of the output object in the chain. Defaults to "dataset".
+            include_listing: If True, includes listing datasets. Defaults to False.
+            studio: If True, returns datasets from Studio only,
+                otherwise returns all datasets locally.
```
shcheklein, line 715: `all local datasets. Defaults to False`

src/datachain/catalog/catalog.py
amritghimire: That works for me.

src/datachain/lib/dc.py
amritghimire: It does now.

src/datachain/dataset.py
```diff
@@ -648,6 +649,20 @@ def is_bucket_listing(self) -> bool:
     def has_version_with_uuid(self, uuid: str) -> bool:
         return any(v.uuid == uuid for v in self.versions)
 
+    @classmethod
+    def from_dict(cls, d: dict[str, Any]) -> "DatasetListRecord":
+        versions = [DatasetListVersion.parse(**v) for v in d.get("versions", [])]
+        return cls(
+            id=d.get("id"),  # type: ignore[arg-type]
```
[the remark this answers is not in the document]
amritghimire: Because id and name has required type and so on. But updated the logic without mypy suppression.

src/datachain/catalog/catalog.py
amritghimire: Updated accordingly.

shcheklein (approved):

ilongin (approved):

PR #964, opened 2025-03-11
refactor(listing): pull glob check out of `_isfile`

skshetry:
Related: https://github.com/iterative/datachain/pull/856#discussion_r1989389883

### DIFF
--- a/src/datachain/lib/listing.py
+++ b/src/datachain/lib/listing.py
@@ -1,3 +1,4 @@
+import glob
 import logging
 import os
 import posixpath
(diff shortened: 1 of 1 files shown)


src/datachain/lib/listing.py
```diff
@@ -173,7 +170,7 @@ def get_listing(
         uri = str(uri)
 
     # we don't want to use cached dataset (e.g. for a single file listing)
-    if not uri.endswith("/") and _isfile(client, uri):
+    if not glob.has_magic(uri) and not uri.endswith("/") and _isfile(client, uri):
```
skshetry, line 173: The `fs.info` does not support a glob. So, it should never go down that path, regardless of the platform.

amritghimire (approved):

Issue #947, opened 2025-03-05
reduce import time


skshetry:
### Description

```
$ ipython
```                                                                                                                                      
```py
In [1]: %time import datachain
CPU times: user 1.03 s, sys: 297 ms, total: 1.32 s
Wall time: 2.73 s
```

<img width="1119" alt="Image" src="https://github.com/user-attachments/assets/8ddce29e-a3f7-42e3-8a79-3d66d4e005b7" />

skshetry:
<img width="1119" alt="Image" src="https://github.com/user-attachments/assets/333044df-a717-4a36-8fe3-6ce0c0a3a3dc" />


Now the import time has dropped to <450ms. `sqlalchemy` and `pydantic` are major contributors now taking 60% of the importtime. The rest is a mix of `fsspec`, `attrs`, and `lib.File` (which also points to `pydantic`).

sqlalchemy is all over the place. To really fix it, we need better organization of modules and lot of discipline. I don't see a way around `pydantic`.

I'll add a benchmark so that we don't regress, but I am closing this for now.

PR #965, opened 2025-03-12
refactor(listing): reraise any exception as ClientError on `isdir` check

skshetry:
- Reraises any exception as `ClientError` in `datachain.lib.listing.get_listing` when checking if a path is a directory. This helps us remove the imports of two libraries, `gcsfs` and `botocore`  which reduces import time: #947.
- Move `isfile` to `datachain.fs.utils`.
- Remove unused utilities from `datachain.utils`

### DIFF
--- a/src/datachain/error.py
+++ b/src/datachain/error.py
@@ -1,15 +1,3 @@
-import botocore.errorfactory
-import botocore.exceptions
-import gcsfs.retry
-
-REMOTE_ERRORS = (
-    gcsfs.retry.HttpError,  # GCS
-    OSError,  # GCS
-    botocore.exceptions.BotoCoreError,  # S3
-    ValueError,  # Azure
-)
-
-
 class DataChainError(RuntimeError):
     pass
 
--- a/src/datachain/fs/utils.py
+++ b/src/datachain/fs/utils.py
@@ -1,0 +1,30 @@
+from typing import TYPE_CHECKING
+
+from fsspec.implementations.local import LocalFileSystem
+
+if TYPE_CHECKING:
+    from fsspec import AbstractFileSystem
+
+
+def _isdir(fs: "AbstractFileSystem", path: str) -> bool:
+    info = fs.info(path)
+    return info["type"] == "directory" or (
+        info["size"] == 0 and info["type"] == "file" and info["name"].endswith("/")
+    )
+
+
+def isfile(fs: "AbstractFileSystem", path: str) -> bool:
+    """
+    Returns True if uri points to a file.
+
+    Supports special directories on object storages, e.g.:
+    Google creates a zero byte file with the same name as the directory with a trailing
+    slash at the end.
+    """
+    if isinstance(fs, LocalFileSystem):
+        return fs.isfile(path)
+
+    try:
+        return not _isdir(fs, path)
+    except FileNotFoundError:
+        return False
--- a/src/datachain/lib/listing.py
+++ b/src/datachain/lib/listing.py
@@ -3,14 +3,16 @@ import logging
 import os
 import posixpath
 from collections.abc import Iterator
+from contextlib import contextmanager
 from typing import TYPE_CHECKING, Callable, Optional, TypeVar, Union
 
 from fsspec.asyn import get_loop
 from sqlalchemy.sql.expression import true
 
+import datachain.fs.utils as fsutils
 from datachain.asyn import iter_over_async
 from datachain.client import Client
-from datachain.error import REMOTE_ERRORS, ClientError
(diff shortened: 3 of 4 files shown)


src/datachain/utils.py
```diff
@@ -193,14 +192,6 @@ def suffix_to_number(num_str: str) -> int:
         raise ValueError(f"Invalid number/suffix for: {num_str}") from None
 
 
-def force_create_dir(name):
```
skshetry, line 196: This is unused.

src/datachain/utils.py
```diff
@@ -210,13 +201,6 @@ def datachain_paths_join(source_path: str, file_paths: Iterable[str]) -> Iterabl
     return (f"{source_stripped}/{path.lstrip('/')}" for path in file_paths)
 
 
-# From: https://docs.python.org/3/library/shutil.html#rmtree-example
-def remove_readonly(func, path, _):
```
skshetry, line 214: This is also unused.

src/datachain/error.py
```diff
@@ -1,15 +1,3 @@
-import botocore.errorfactory
```
skshetry, line 1: This increases import time, so I am removing this and reraising any unhandled error as `ClientError`.

src/datachain/fs/utils.py
```diff
@@ -0,0 +1,27 @@
+from typing import TYPE_CHECKING
+
+from fsspec.implementations.local import LocalFileSystem
+
+if TYPE_CHECKING:
+    from fsspec import AbstractFileSystem
+
+
+def isdir(fs: "AbstractFileSystem", path: str) -> bool:
```
skshetry, line 9: Renamed `isfile` to `isdir`. I think we are more interested to see if something is a directory or not than files.

src/datachain/fs/utils.py
```diff
@@ -0,0 +1,27 @@
...
+        return fs.isdir(path)
+
+    try:
+        info = fs.info(path)
+    except FileNotFoundError:
+        return False
+    else:
+        return info["type"] == "directory" or (
+            info["size"] == 0 and info["type"] == "file" and info["name"].endswith("/")
+        )
```
skshetry, line 27: Improving the check for zero-byte file with a trailing slash.

dreadatour: 🤔 Looks reasonable, but same time looks a bit suspicious 👀
I know this was done by a reason and was checked many times, so I am ok with this 🙃

src/datachain/lib/listing.py
```diff
@@ -167,12 +155,20 @@ def get_listing(
...
-        storage_uri, path = Client.parse_url(uri)
-        return None, f"{storage_uri}/{path.lstrip('/')}", path, False
+    isfile = _reraise_as_client_error()(fsutils.isfile)
+    if not glob.has_magic(uri) and isfile(client.fs, uri):
+        if uri.endswith(client.fs.sep):
+            # should be fixed upstream in fsspec
+            # boils down to https://github.com/fsspec/filesystem_spec/pull/1567#issuecomment-2563160414
+            # fsspec removes the trailing slash and returns a file, that's why we are
+            # are not getting an error here
+            raise NotADirectoryError(errno.ENOTDIR, os.strerror(errno.ENOTDIR), uri)
```
skshetry, line 168: cc @shcheklein. I changed this to raise `NotADirectoryError` when the given path is a path to a file and has a trailing slash.

```console
$ datachain ls pyproject.toml/
Error: [Errno 20] Not a directory: 'pyproject.toml/'
```

I think it's better to fail eagerly here.

src/datachain/fs/utils.py
```diff
@@ -0,0 +1,30 @@
...
+
+if TYPE_CHECKING:
+    from fsspec import AbstractFileSystem
+
+
+def _isdir(fs: "AbstractFileSystem", path: str) -> bool:
+    info = fs.info(path)
+    return info["type"] == "directory" or (
+        info["size"] == 0 and info["type"] == "file" and info["name"].endswith("/")
+    )
```
skshetry, line 13: I extract this out so that if we need to implement `isdir`, it would not be too difficult. 

We cannot implement `isdir` to be a negative of `isfile` as it misses non-existing file.

src/datachain/lib/listing.py
skshetry: I think we could also move this logic inside `isfile`, but that won't allow us to raise `NotADirectoryError`.

skshetry: It looks like `azure` and `gs` raise `FileNotFoundError` in this case.

PR #966, opened 2025-03-12
Improve models and file methods

dreadatour:
- BBox model now have methods to convert to/from different bbox formats: Albumentations, COCO, PASCAL VOC and YOLO
- Add more validations and tests for BBox, Pose and Segment models
- Add tests for Ultralytics models
- Add File conversion methos: to TextFile, ImageFile and VideoFile (see Ultralytcs examples for usage)
- Add 'image_info' method to get meta from ImageFile (same as for video file)


src/datachain/lib/file.py
```diff
@@ -242,6 +242,30 @@ def __init__(self, **kwargs):
         self._catalog = None
         self._caching_enabled: bool = False
 
+    def as_text_file(self) -> "TextFile":
+        """Convert the file to a `TextFile` object."""
+        if isinstance(self, TextFile):
+            return self
+        file = TextFile(**self.model_dump())
```
amritghimire, line 249: Should we add check for the file content for example if it is possible to dump a file as text or not? I don't think we benefit from dumping certain binary files as text file IMO.

src/datachain/lib/file.py
```diff
@@ -242,6 +242,30 @@ def __init__(self, **kwargs):
...
+        file._set_stream(self._catalog, caching_enabled=self._caching_enabled)
+        return file
+
+    def as_image_file(self) -> "ImageFile":
+        """Convert the file to a `ImageFile` object."""
+        if isinstance(self, ImageFile):
+            return self
+        file = ImageFile(**self.model_dump())
```
amritghimire, line 257: Same here, we should probably look into proper handling of error if the file cannot be converted to image file or if it is an invalid image file.

src/datachain/lib/file.py
```diff
@@ -242,6 +242,30 @@ def __init__(self, **kwargs):
...
+        file._set_stream(self._catalog, caching_enabled=self._caching_enabled)
+        return file
+
+    def as_video_file(self) -> "VideoFile":
+        """Convert the file to a `VideoFile` object."""
+        if isinstance(self, VideoFile):
+            return self
+        file = VideoFile(**self.model_dump())
```
amritghimire, line 265: ⬆️

src/datachain/lib/image.py
```diff
@@ -1,17 +1,44 @@
...
+
+    Returns:
+        Image: Image file information.
+    """
+    file = file.as_image_file()
+
+    try:
+        with file.open() as stream:
+            img = PILImage.open(stream)
+    except Exception as exc:
```
amritghimire, line 24: Sounds like a broad catch. I am not sure about it, even though we reraise or wrap the exception. So, should be fine. Just thinking through.

src/datachain/model/bbox.py
```diff
@@ -1,47 +1,216 @@
...
+
+    def pose_inside(self, pose: Union["Pose", "Pose3D"]) -> bool:
+        """Return True if the pose is inside the bounding box."""
+        return all(
+            self.point_inside(x, y) for x, y in zip(pose.x, pose.y) if x > 0 or y > 0
         )
 
+    @staticmethod
+    def from_list(coords: Sequence[float], title: str = "") -> "BBox":
+        return BBox.from_voc(coords, title=title)
```
amritghimire, line 206: Not familiar with the logic TBH, but we should probably add a test covering this.

src/datachain/lib/file.py
dreadatour: Hm... For this we need to download file first. This helper is needed for user to be able to use methods from `TextFile` or `VideoFile` models, for example, to split video into frames. I suppose users know what they are doing, otherwise they'll get an error.

src/datachain/lib/file.py
dreadatour: User will get an error in case they try to use methods from `ImageFile` class on non-image file. Here is just a matter of model type, not real file.

src/datachain/lib/image.py
dreadatour: Good catch, let me think 👍

src/datachain/lib/file.py
```diff
@@ -242,6 +242,30 @@ def __init__(self, **kwargs):
...
 
+    def as_text_file(self) -> "TextFile":
+        """Convert the file to a `TextFile` object."""
+        if isinstance(self, TextFile):
+            return self
+        file = TextFile(**self.model_dump())
+        file._set_stream(self._catalog, caching_enabled=self._caching_enabled)
+        return file
+
+    def as_image_file(self) -> "ImageFile":
```
shcheklein, line 253: would it better to actually transform it on the UDF preparation side? 

let's say we see that we have a model File and UDF accepts ImageFile - we would try to instantiate the object as ImageFile

wdyt?

otherwise, why can't users initially create DC with proper object types?

src/datachain/lib/file.py
```diff
@@ -547,6 +571,17 @@ def save(self, destination: str, client_config: Optional[dict] = None):
 class ImageFile(File):
     """`DataModel` for reading image files."""
 
+    def get_info(self) -> "Image":
```
shcheklein, line 574: let's rename it after all? I mean Image to ImageMeta or something? it is confusing tbh otherwise

src/datachain/lib/image.py
```diff
@@ -1,17 +1,44 @@
...
+    Args:
+        file (ImageFile): Image file object.
+
+    Returns:
+        Image: Image file information.
+    """
+    file = file.as_image_file()
+
+    try:
+        with file.open() as stream:
```
shcheklein, line 22: just to double check - is there guarantee that it is a stream (that we don't download it)

e.g. does it mean that users in reality will have to disable cahce and prefetch for such operations?

shcheklein (approved):
A few questions, otherwise amazing work, let's move forward with it

src/datachain/lib/file.py
dreadatour: > would it better to actually transform it on the UDF preparation side?
>
> let's say we see that we have a model File and UDF accepts ImageFile - we would try to instantiate the object as ImageFile

Sounds like an awesome idea! This should works in 99% cases. Let me create a follow-up issue.

> otherwise, why can't users initially create DC with proper object types?

It is better for sure, and one should use it if possible, but let's say we have a bucket with mixed types of files, for example, images and annotations (e.g. JSON or plain text). Then you can not specify type of file in DC and you need to be able to convert `File` to, let's say, `ImageFile`.

As an brief example:

```python
def parse_image_meta(file: File) -> ImageMeta:
    img_file = file.as_image_file()
    ...

def parse_annotations(file: File) -> Annotation:
    txt_file = file.as_text_file()
    ...

source = DataChain.from_storage("s3://bucket/")
frames = (
    source
    .filter(C("file.path").glob('*.jpg'))
    .map(image_meta=parse_image_meta)
)
annotations = (
    source
    .filter(C("file.path").glob('*.txt'))
    .map(annotations=parse_annotations)
)
dc = (
    frames.merge(
        annotations,
        on=func.path.file_stem("file.path"),
        right_on=func.path.file_stem("file.path"),
    )
)
```

Btw, sidenote, I think we should also add `JSONFile` with helpers for reading and parsing JSON's 👀 Same for `ParquetFile`, may be?

src/datachain/lib/file.py
dreadatour: This is how it was initially, see [this comment](https://github.com/iterative/datachain/pull/814#pullrequestreview-2547435306). Me personally, me too prefer `ImageMeta` instead of `Image`, but we do have this conversation before and decided to do it this way.

cc @dmpetrov

src/datachain/model/bbox.py
dreadatour: Added 🙏

src/datachain/lib/image.py
dreadatour: Hm, I think it is OK here to catch general exception since we are going to show it to user.

src/datachain/lib/image.py
dreadatour: We are using [File.open](https://github.com/iterative/datachain/blob/main/src/datachain/lib/file.py#L279-L292) here, if cache is disabled, no caching will be used and file will be downloaded using [our regular mechanics](https://github.com/iterative/datachain/blob/main/src/datachain/client/fsspec.py#L377-L386).

shcheklein: right, but is cache enabled by default? is it possible to disable it specifically for this operation? so, that simple query that is fetching metadata doesn't trigger the full download by default? Disabling explicitly won't be an obvious thing to do to end users ...

src/datachain/lib/file.py
shcheklein: let's rename it please. Anything is better then just `Image` to my mind. It is the most confusing name for it.

src/datachain/lib/file.py
shcheklein: it's not a very good example, in this example you can "easily" do two `from_storage` with different types. And arguably it will be easier to manage and read and probably more efficient (?). Anyways, it's fine to keep those helpers I think, it's not a big deal.

PR #967, opened 2025-03-12
lazily import from `datachain.data_storage.sqlite`

skshetry:
`datachain.sql.sqlite.setup()` runs on import of this module. That takes more than 750ms on my machine.

https://github.com/iterative/datachain/blob/4d4d4298434f0ba7836a0d9b52f68e1955a531cd/src/datachain/data_storage/sqlite.py#L57

After this PR, the total importtime drops to <600ms.

Related: #947.


dreadatour (approved):
❤️

PR #970, opened 2025-03-13
client(fsspec): remove botocore import

skshetry:


src/datachain/client/fsspec.py
```diff
@@ -287,11 +285,6 @@ async def process(queue) -> None:
                     worker.cancel()
             if excs:
                 raise excs[0]
-        except ClientError as exc:
-            raise DataChainClientError(
-                exc.response.get("Error", {}).get("Message") or exc,
-                exc.response.get("Error", {}).get("Code"),
-            ) from exc
```
skshetry, line 294: I might be regressing the error message a bit by doing this, but this is not the right place to handle this, and I haven't seen a case where this happens (it is probably in exceptional cases).

I could not find any examples in https://github.com/iterative/dvcx/pull/233 or in https://github.com/iterative/dvcx/issues/50.

PR #971, opened 2025-03-13
tests(unit/video): skip if ffmepg is not installed

skshetry:


dreadatour (approved):
🙏 Thank you for this improvement ❤️

PR #972, opened 2025-03-13
tests: add an unit test for catching import time regressions in the future

skshetry:
Part of https://github.com/iterative/datachain/issues/947.

Taken an inspiration from https://github.com/pola-rs/polars/blob/63acc86ef657f1b404355ee8b6d4ed9786a8ed1a/py-polars/tests/unit/test_polars_import.py.

### DIFF
--- a/tests/test_import_time.py
+++ b/tests/test_import_time.py
@@ -1,0 +1,48 @@
+import logging
+import subprocess
+import sys
+from pathlib import Path
+
+from datachain import C, DataChain
+
+logger = logging.getLogger(__name__)
+
+MAX_ATTEMPTS = 3
+MAX_IMPORT_TIME_MS = 700
+
+
+def _import_time_chain(test_session):
+    proc = subprocess.run(  # noqa: S603
+        [sys.executable, "-X", "importtime", "-c", "import datachain"],
+        stderr=subprocess.PIPE,
+        check=True,
+    )
+    out = proc.stderr.replace(b"import time:", b"").strip()
+    Path("import_time.csv").write_bytes(out)
+
+    dc = DataChain.from_csv("import_time.csv", session=test_session, delimiter="|")
+    dc = dc.save("import_time")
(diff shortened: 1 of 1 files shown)


0x2b3bfa0 (approved):

tests/test_import_time.py
```diff
@@ -0,0 +1,48 @@
...
+    # columns are: cumulative_ms, self_ms, import
+    # TODO: use `mutate` instead of `map`
+    return (
+        dc.map(cumulative_ms=lambda cumulative: cumulative // 1000, output=int)
+        .map(self_ms=lambda self_us: self_us // 1000, output=int)
+        .map(**{"import": lambda imported_package: imported_package.lstrip()})
+    )
+
+
+def test_import_time(catalog, test_session):
```
shcheklein, line 34: Amazing! Can we put some brief comment how would people approach fixing / profiling things in case it breaks. For a lot of folks it will be something new to learn and it would help if we have some steps / best practices here.

skshetry: Done. I have added a small comment on how to use the datachain to find the issue.
And also added a docstring with a how to analyze it yourself outside of the test.

tests/test_import_time.py
```diff
@@ -0,0 +1,81 @@
...
+        dc = _import_time_chain(test_session)
+        (import_time_ms,) = dc.filter(
+            C("import") == "datachain",
+        ).collect("cumulative_ms")
+        import_timings.append((dc, import_time_ms))
+        # pass `--log-cli-level=info` to see these logs live
+        logger.info("attempt %d, import time: %dms", attempt + 1, import_time_ms)
+
+    dc, min_import_time = min(import_timings, key=lambda x: x[1])
+    # If there is a regression, uncomment the following to check the culprit:
```
skshetry, line 68:
```suggestion
    # If there is a regression, uncomment the following to find the culprit:
```

PR #973, opened 2025-03-14
Check UDF params types in SignalSchema

dreadatour:
Follow-up for the https://github.com/iterative/datachain/pull/966

See this thread for more details: https://github.com/iterative/datachain/pull/966#discussion_r1994190977

We are now checking types of UDF arguments if they are defined. Also there is a special case for `File` signal: it will be automatically converted to `TextFile`, `ImageFile` or `VideoFile` if one of these types are used in UDF. See also my comments about implementation and examples below 🙏

## Very basic example

Let's say we have Dataset `my-ds` with two signals: `name: str` and `value: int`.

We can now define UDF to process these signals, e.g.:

```python
def process(name: str, value: int) -> float:
    return len(name) / value

DataChain.from_dataset("my-ds").map(signal=process)
```

In this PR we are adding type checking for UDF param types based on dataset, such as the following UDF:

```python
def process(name: int, value: int) -> float:
    return len(name) / value

DataChain.from_dataset("my-ds").map(signal=process)
```

will return an error:

```
SignalResolvingError: cannot resolve signal name 'name': types mismatch: <class 'int'> != <class 'str'>
```

because in dataset `my-ds` signal `name` type is `str`, not `int`.

## Another real-life example

In this example we are creating dataset from Google Storage bucket with signal `file` type `File`. Next we are going to process these files as images to get image meta from them. Since general `File` type have no methods to work with images, we need type conversion (`File` -> `ImageFile`), which can be done by using `file.as_image_file()`, but it is also possible to do it on-the-fly based on UDF signature types, which is the subject of this PR.

#### Query

```python
from datachain import DataChain, Image, ImageFile


def parse_meta(file: ImageFile) -> Image:
    return file.get_info()


(
    DataChain.from_storage("gs://datachain-test-vlad/human-pose-estimation")
    .limit(100)
    .map(meta=parse_meta)
    .show()
)
```

#### Before

```
$ python query.py
  File "/Users/vlad/work/iterative/datachain/src/datachain/lib/udf.py", line 174, in process
    return self._func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/vlad/work/iterative/playground/query.py", line 5, in parse_meta
    return file.get_info()
           ^^^^^^^^^^^^^
  File "/Users/vlad/.virtualenvs/datachain/lib/python3.12/site-packages/pydantic/main.py", line 892, in __getattr__
    raise AttributeError(f'{type(self).__name__!r} object has no attribute {item!r}')
AttributeError: 'File' object has no attribute 'get_info'
==========================================================
  File "/Users/vlad/work/iterative/playground/query.py", line 12, in <module>
    .show()
     ^^^^^^
  File "/Users/vlad/work/iterative/datachain/src/datachain/lib/dc.py", line 1887, in show
    df = dc.to_pandas(flatten, include_hidden=include_hidden)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/vlad/work/iterative/datachain/src/datachain/lib/dc.py", line 1864, in to_pandas
    results = self.results(include_hidden=include_hidden)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/vlad/work/iterative/datachain/src/datachain/lib/dc.py", line 1334, in results
    return list(self.collect_flatten(include_hidden=include_hidden))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/vlad/work/iterative/datachain/src/datachain/lib/dc.py", line 1285, in collect_flatten
    with self._query.ordered_select(*db_signals).as_iterable() as rows:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.12/3.12.9/Frameworks/Python.framework/Versions/3.12/lib/python3.12/contextlib.py", line 137, in __enter__
    return next(self.gen)
           ^^^^^^^^^^^^^^
  File "/Users/vlad/work/iterative/datachain/src/datachain/query/dataset.py", line 1264, in as_iterable
    query = self.apply_steps().select()
            ^^^^^^^^^^^^^^^^^^
  File "/Users/vlad/work/iterative/datachain/src/datachain/query/dataset.py", line 1210, in apply_steps
    result = step.apply(
             ^^^^^^^^^^^
  File "/Users/vlad/work/iterative/datachain/src/datachain/query/dataset.py", line 602, in apply
    self.populate_udf_table(udf_table, query)
  File "/Users/vlad/work/iterative/datachain/src/datachain/query/dataset.py", line 520, in populate_udf_table
    process_udf_outputs(
  File "/Users/vlad/work/iterative/datachain/src/datachain/query/dataset.py", line 340, in process_udf_outputs
    for udf_output in udf_results:
                      ^^^^^^^^^^^
  File "/Users/vlad/work/iterative/datachain/src/datachain/lib/udf.py", line 100, in run
    yield from self.inner.run(
  File "/Users/vlad/work/iterative/datachain/src/datachain/lib/udf.py", line 386, in run
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/vlad/work/iterative/datachain/src/datachain/lib/udf.py", line 284, in process_safe
    raise DataChainError(
datachain.lib.utils.DataChainError: Error in user code in class 'Mapper': 'File' object has no attribute 'get_info'
$
```

#### After
```
$ python query.py
                            file     file              file      file                             file     file  meta   meta   meta
                            path     size              etag is_latest                    last_modified location width height format
0    human-pose-estimation/0.jpg   730027  CKfQgNGSyIkDEAE=         1 2024-11-06 16:39:41.795000+00:00     None  3580   5370   JPEG
1    human-pose-estimation/1.jpg  4744618  CPbpqtGSyIkDEAE=         1 2024-11-06 16:39:42.510000+00:00     None  3217   4825   JPEG
2   human-pose-estimation/10.jpg  1726248  CO2DmtGSyIkDEAE=         1 2024-11-06 16:39:42.236000+00:00     None  2792   3634   JPEG
3   human-pose-estimation/11.jpg  2257008  CPy6ndGSyIkDEAE=         1 2024-11-06 16:39:42.291000+00:00     None  3573   5242   JPEG
4   human-pose-estimation/12.jpg   792329  CNH9g9GSyIkDEAE=         1 2024-11-06 16:39:41.865000+00:00     None  5247   3498   JPEG
5   human-pose-estimation/13.jpg  2676811  CITOpdGSyIkDEAE=         1 2024-11-06 16:39:42.423000+00:00     None  2997   4496   JPEG
6   human-pose-estimation/14.jpg  1274475  CJPemdGSyIkDEAE=         1 2024-11-06 16:39:42.224000+00:00     None  5472   3648   JPEG
7   human-pose-estimation/15.jpg   993245  CPyXj9GSyIkDEAE=         1 2024-11-06 16:39:42.047000+00:00     None  2969   4453   JPEG
8   human-pose-estimation/16.jpg  2368068  CIGYotGSyIkDEAE=         1 2024-11-06 16:39:42.356000+00:00     None  2968   4016   JPEG
9   human-pose-estimation/17.jpg  1965589  CI3bmdGSyIkDEAE=         1 2024-11-06 16:39:42.225000+00:00     None  3265   4898   JPEG
10  human-pose-estimation/18.jpg  3172189  CP6g39GSyIkDEAE=         1 2024-11-06 16:39:43.337000+00:00     None  3744   5616   JPEG
11  human-pose-estimation/19.jpg  1044142  CISzkdGSyIkDEAE=         1 2024-11-06 16:39:42.072000+00:00     None  4698   7050   JPEG
12   human-pose-estimation/2.JPG  2809438  CPLPoNGSyIkDEAE=         1 2024-11-06 16:39:42.323000+00:00     None  3648   5472   JPEG
13  human-pose-estimation/20.jpg  1257389  CJiulNGSyIkDEAE=         1 2024-11-06 16:39:42.140000+00:00     None  4896   3264   JPEG
14  human-pose-estimation/21.jpg  2062542  CIKZntGSyIkDEAE=         1 2024-11-06 16:39:42.304000+00:00     None  4851   3234   JPEG
15  human-pose-estimation/22.jpg  3014260  CMPSqNGSyIkDEAE=         1 2024-11-06 16:39:42.464000+00:00     None  5461   8192   JPEG
16  human-pose-estimation/23.jpg  1778821  CJL/ltGSyIkDEAE=         1 2024-11-06 16:39:42.179000+00:00     None  3705   5557   JPEG
17  human-pose-estimation/24.jpg  3823660  CPPu49GSyIkDEAE=         1 2024-11-06 16:39:43.436000+00:00     None  3648   5472   JPEG
18  human-pose-estimation/25.jpg   779455  CP2TidGSyIkDEAE=         1 2024-11-06 16:39:41.927000+00:00     None  2829   3149   JPEG
19  human-pose-estimation/26.jpg   750505  CMDrwdGSyIkDEAE=         1 2024-11-06 16:39:42.884000+00:00     None  3456   5184   JPEG

[Limited by 20 rows]
$
```

## TODO

- [x] tests

### DIFF
--- a/src/datachain/lib/signal_schema.py
+++ b/src/datachain/lib/signal_schema.py
@@ -472,16 +472,16 @@ class SignalSchema:
         """
         Returns new schema that combines current schema and setup signals.
         """
-        setup = setup or {}
-        setup_no_types = dict.fromkeys(setup.keys(), str)
-        union = SignalSchema(self.values | setup_no_types)
-
-        # Slice combined schema by keys
+        setup_params = setup.keys() if setup else []
         schema: dict[str, DataType] = {}
 
         for param, param_type in params.items():
+            if param in setup_params:
+                schema[param] = str
+                continue
+
             try:
-                schema_type = union._find_in_tree(param.split("."))
+                schema_type = self._find_in_tree(param.split("."))
             except SignalResolvingError:
                 continue
 
--- a/src/datachain/lib/signal_schema.py
+++ b/src/datachain/lib/signal_schema.py
@@ -465,7 +465,7 @@ class SignalSchema:
 
     def slice(
         self,
-        params: Union[Sequence[str], dict[str, Optional[type]]],
+        params: Union[Sequence[str], dict[str, Optional[DataType]]],
         setup: Optional[dict[str, Callable]] = None,
     ) -> "SignalSchema":
         """
@@ -474,22 +474,19 @@ class SignalSchema:
         setup = setup or {}
         setup_no_types = dict.fromkeys(setup.keys(), str)
         union = SignalSchema(self.values | setup_no_types)
-        # Slice combined schema by keys
-        schema = {}
 
-        if isinstance(params, dict):
-            for param, param_type in params.items():
-                schema[param] = (
-                    param_type
-                    if param_type is not None
-                    else union._find_in_tree(param.split("."))
-                )
-        else:
-            for param in params:
-                try:
-                    schema[param] = union._find_in_tree(param.split("."))
-                except SignalResolvingError:
-                    pass
+        # Slice combined schema by keys
+        schema: dict[str, DataType] = {}
+
+        for param in params:
+            if isinstance(params, dict) and params[param] is not None:
+                schema[param] = params[param]  # type: ignore[assignment]
+                continue
+            try:
+                schema[param] = union._find_in_tree(param.split("."))
+            except SignalResolvingError:
+                pass
+
         return SignalSchema(schema, setup)
 
     def row_to_features(
--- a/src/datachain/lib/signal_schema.py
+++ b/src/datachain/lib/signal_schema.py
@@ -154,9 +154,9 @@ class SignalSchema:
             if not callable(func):
                 raise SetupError(key, "value must be function or callable class")
 
-    def _init_setup_values(self):
+    def _init_setup_values(self) -> None:
         if self.setup_values is not None:
-            return self.setup_values
+            return
 
         res = {}
         for key, func in self.setup_func.items():
@@ -398,7 +398,7 @@ class SignalSchema:
         return SignalSchema(signals)
 
     @staticmethod
-    def get_flatten_hidden_fields(schema):
+    def get_flatten_hidden_fields(schema: dict):
         custom_types = schema.get("_custom_types", {})
         if not custom_types:
             return []
@@ -476,14 +476,12 @@ class SignalSchema:
         schema: dict[str, DataType] = {}
 
         for param, param_type in params.items():
+            # This is special case for setup params, they are always treated as strings
             if param in setup_params:
                 schema[param] = str
                 continue
 
-            try:
-                schema_type = self._find_in_tree(param.split("."))
-            except SignalResolvingError:
-                continue
+            schema_type = self._find_in_tree(param.split("."))
 
             if param_type is Any:
                 schema[param] = schema_type
--- a/src/datachain/lib/signal_schema.py
+++ b/src/datachain/lib/signal_schema.py
@@ -464,19 +464,32 @@ class SignalSchema:
         return False
 
     def slice(
-        self, keys: Sequence[str], setup: Optional[dict[str, Callable]] = None
+        self,
+        params: Union[Sequence[str], dict[str, Optional[type]]],
+        setup: Optional[dict[str, Callable]] = None,
     ) -> "SignalSchema":
-        # Make new schema that combines current schema and setup signals
+        """
+        Returns new schema that combines current schema and setup signals.
+        """
         setup = setup or {}
         setup_no_types = dict.fromkeys(setup.keys(), str)
         union = SignalSchema(self.values | setup_no_types)
         # Slice combined schema by keys
         schema = {}
-        for k in keys:
-            try:
-                schema[k] = union._find_in_tree(k.split("."))
-            except SignalResolvingError:
-                pass
+
+        if isinstance(params, dict):
+            for param, param_type in params.items():
+                schema[param] = (
+                    param_type
+                    if param_type is not None
+                    else union._find_in_tree(param.split("."))
+                )
+        else:
+            for param in params:
+                try:
+                    schema[param] = union._find_in_tree(param.split("."))
+                except SignalResolvingError:
+                    pass
         return SignalSchema(schema, setup)
 
     def row_to_features(
--- a/src/datachain/lib/signal_schema.py
+++ b/src/datachain/lib/signal_schema.py
@@ -489,19 +489,23 @@ class SignalSchema:
                 schema[param] = schema_type
                 continue
 
+            schema_origin = get_origin(schema_type)
+            param_origin = get_origin(param_type)
+
+            if schema_origin is Union and type(None) in get_args(schema_type):
+                schema_type = get_args(schema_type)[0]
+                if param_origin is Union and type(None) in get_args(param_type):
+                    param_type = get_args(param_type)[0]
+
             if is_batch:
                 if param_type is list:
                     schema[param] = schema_type
                     continue
 
-                if get_origin(param_type) is not list:
+                if param_origin is not list:
                     raise SignalResolvingError(param.split("."), "is not a list")
 
                 param_type = get_args(param_type)[0]
-
-            if param_type == schema_type:
-                schema[param] = schema_type
-                continue
 
             if param_type == schema_type or (
                 isclass(param_type)
--- a/src/datachain/lib/signal_schema.py
+++ b/src/datachain/lib/signal_schema.py
@@ -465,8 +465,9 @@ class SignalSchema:
 
     def slice(
         self,
-        params: Union[Sequence[str], dict[str, Optional[DataType]]],
+        params: dict[str, Union[DataType, Any]],
         setup: Optional[dict[str, Callable]] = None,
+        is_batch: bool = False,
     ) -> "SignalSchema":
         """
         Returns new schema that combines current schema and setup signals.
@@ -478,14 +479,43 @@ class SignalSchema:
         # Slice combined schema by keys
         schema: dict[str, DataType] = {}
 
-        for param in params:
-            if isinstance(params, dict) and params[param] is not None:
-                schema[param] = params[param]  # type: ignore[assignment]
-                continue
+        for param, param_type in params.items():
             try:
-                schema[param] = union._find_in_tree(param.split("."))
+                schema_type = union._find_in_tree(param.split("."))
             except SignalResolvingError:
-                pass
+                continue
+
+            if param_type is Any:
+                schema[param] = schema_type
+                continue
+
+            if is_batch:
+                if param_type is list:
+                    schema[param] = schema_type
+                    continue
+
+                if get_origin(param_type) is not list:
+                    raise SignalResolvingError(param.split("."), "is not a list")
+
+                param_type = get_args(param_type)[0]
+
+            if param_type == schema_type:
+                schema[param] = schema_type
+                continue
+
+            if param_type == schema_type or (
+                isclass(param_type)
+                and isclass(schema_type)
+                and issubclass(param_type, File)
+                and issubclass(schema_type, File)
+            ):
+                schema[param] = schema_type
+                continue
+
+            raise SignalResolvingError(
+                param.split("."),
+                f"types mismatch: {param_type} != {schema_type}",
+            )
 
         return SignalSchema(schema, setup)
 
--- a/src/datachain/lib/udf.py
+++ b/src/datachain/lib/udf.py
@@ -159,6 +159,7 @@ class UDFBase(AbstractUDF):
         ```
     """
 
+    is_input_batched = False
     is_output_batched = False
     prefetch: int = 0
 
@@ -395,6 +396,7 @@ class Mapper(UDFBase):
 class BatchMapper(UDFBase):
     """Inherit from this class to pass to `DataChain.batch_map()`."""
 
+    is_input_batched = True
     is_output_batched = True
 
     def run(
@@ -481,6 +483,7 @@ class Generator(UDFBase):
 class Aggregator(UDFBase):
     """Inherit from this class to pass to `DataChain.agg()`."""
 
+    is_input_batched = True
     is_output_batched = True
 
     def run(
--- a/src/datachain/lib/udf_signature.py
+++ b/src/datachain/lib/udf_signature.py
@@ -1,7 +1,7 @@
 import inspect
 from collections.abc import Generator, Iterator, Sequence
 from dataclasses import dataclass
-from typing import Callable, Optional, Union, get_args, get_origin
+from typing import Any, Callable, Union, get_args, get_origin
 
 from datachain.lib.data_model import DataType, DataTypeNames, is_chain_type
 from datachain.lib.signal_schema import SignalSchema
@@ -18,7 +18,7 @@ class UdfSignatureError(DataChainParamsError):
 @dataclass
 class UdfSignature:
     func: Union[Callable, UDFBase]
-    params: dict[str, Optional[type]]
+    params: dict[str, Union[DataType, Any]]
     output_schema: SignalSchema
 
     DEFAULT_RETURN_TYPE = str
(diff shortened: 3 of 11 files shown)


src/datachain/lib/signal_schema.py
```diff
@@ -464,19 +464,63 @@ def contains_file(self) -> bool:
         return False
 
     def slice(
-        self, keys: Sequence[str], setup: Optional[dict[str, Callable]] = None
+        self,
+        params: dict[str, Union[DataType, Any]],
+        setup: Optional[dict[str, Callable]] = None,
+        is_batch: bool = False,
     ) -> "SignalSchema":
```
dreadatour, line 471: Main changes of this PR goes into this function.

`params` is now a dictionary with param names as keys and types as values instead of just a list of input params names.

`is_batch` is needed to be able to check batching UDF input params types.

Let's say we do have a signal `foo` with type `int`, then regular UDF might looks like this:

```python
def process(foo: int) -> int:
    ...
```

and batching UDF looks like this:

```python
def process_batch(foo: list[int]) -> int:
    ...
```

To be able to check types here we need to know if UDF is batching or not.

src/datachain/lib/signal_schema.py
```diff
@@ -464,19 +464,63 @@ def contains_file(self) -> bool:
...
-        # Make new schema that combines current schema and setup signals
-        setup = setup or {}
-        setup_no_types = dict.fromkeys(setup.keys(), str)
-        union = SignalSchema(self.values | setup_no_types)
```
dreadatour, line 472: No need to create new schema, since we are only wanted to know if param with name `foo` is came from setup or not. We are going to just check if param `foo` in setup params or not.

src/datachain/lib/signal_schema.py
```diff
@@ -464,19 +464,63 @@ def contains_file(self) -> bool:
...
+        """
+        Returns new schema that combines current schema and setup signals.
+        """
+        setup_params = setup.keys() if setup else []
+        schema: dict[str, DataType] = {}
+
+        for param, param_type in params.items():
+            if param in setup_params:
+                schema[param] = str
+                continue
```
dreadatour, line 481: Just check if input param is case from setup values (using `str` here because it was like this before, I suppose it is hard or not possible to get output type of lambda functions).

src/datachain/lib/signal_schema.py
```diff
@@ -464,19 +464,63 @@ def contains_file(self) -> bool:
...
             try:
-                schema[k] = union._find_in_tree(k.split("."))
+                schema_type = self._find_in_tree(param.split("."))
             except SignalResolvingError:
-                pass
+                continue
+
+            if param_type is Any:
+                schema[param] = schema_type
+                continue
```
dreadatour, line 490: This is happens if UDF function param have no type, e.g.

```
def process(foo) -> int:
    ...
```

src/datachain/lib/signal_schema.py
```diff
@@ -464,19 +464,63 @@ def contains_file(self) -> bool:
...
+
+            schema_origin = get_origin(schema_type)
+            param_origin = get_origin(param_type)
```
dreadatour, line 493: We will use it couple times below, so just cache the result of `get_origin` here.

src/datachain/lib/signal_schema.py
```diff
@@ -464,19 +464,63 @@ def contains_file(self) -> bool:
...
+
+            if schema_origin is Union and type(None) in get_args(schema_type):
```
dreadatour, line 495: This happens if DataSet schema signal is `Optional`. In this case just for simplicity we might want to allow users to do not specify `Optional` type in the UDF, so these two declarations will be identical:

```python
def process(foo: Optional[int]) -> int:
    ...
```

and

```python
def process(foo: int) -> int:
    ...
```

**Note**: the most common use case is `DataChain.from_csv("path_to_csv_file")`, all signals will be optional in this case 🤔 We might want to revisit `from_csv` method later.

src/datachain/lib/signal_schema.py
```diff
@@ -464,19 +464,63 @@ def contains_file(self) -> bool:
...
+                schema_type = get_args(schema_type)[0]
+                if param_origin is Union and type(None) in get_args(param_type):
+                    param_type = get_args(param_type)[0]
+
+            if is_batch:
```
dreadatour, line 500: In case of batch UDFs these both declarations will be the same:

```python
def process(foo: list[int]) -> int:
    ...
```

and

```python
def process(foo: list) -> int:
    ...
```

src/datachain/lib/signal_schema.py
```diff
@@ -464,19 +464,63 @@ def contains_file(self) -> bool:
...
+                param_type = get_args(param_type)[0]
+
+            if param_type == schema_type or (
+                isclass(param_type)
+                and isclass(schema_type)
+                and issubclass(param_type, File)
+                and issubclass(schema_type, File)
+            ):
+                schema[param] = schema_type
+                continue
```
dreadatour, line 517: UDF input param type should be the same as DataSet signal type.

Special case here: we are converting `File` signal to `TextFile`, `ImageFile` or `VideoFile` based on UDF input param type signature, and vise-versa, `TextFile` DataSet signal can be used as regular `File` type if needed.

Let's say we do have DataSet with `File` type:

```python
DataChain.from_storage("s3://bucket/")
```

if we know these files are image files, we can use them in UDF like this:

```python
def process(file: ImageFile) -> ImageMeta:
    return file.get_info()
```

This might be useful, since `File` class have no `get_info` method.

JFYI, other possible way to do this is:

```python
def process(file: File) -> ImageMeta:
    return file.as_image_file().get_info()
```

Same forks in the opposite way. Let's say, we do have DataSet with `ImageFile` files:

```python
DataChain.from_storage("s3://bucket/", type="image")
```

we can use regular `File` in UDF:

```python
def process(file: File) -> int:
    ...
```

type will be converted automatically in both ways.

src/datachain/lib/udf_signature.py
```diff
@@ -18,7 +18,7 @@ def __init__(self, chain: str, msg):
 @dataclass
 class UdfSignature:
     func: Union[Callable, UDFBase]
-    params: Sequence[str]
+    params: dict[str, Union[DataType, Any]]
```
dreadatour, line 21: UDF signature params is now a dictionary with param names as keys and param types as values, instead of just a list of param names.

src/datachain/lib/udf_signature.py
```diff
@@ -58,15 +58,23 @@ def parse(
         if not isinstance(udf_func, UDFBase) and not callable(udf_func):
             raise UdfSignatureError(chain, f"UDF '{udf_func}' is not callable")
 
-        func_params_map_sign, func_outs_sign, is_iterator = (
-            UdfSignature._func_signature(chain, udf_func)
+        func_params_map_sign, func_outs_sign, is_iterator = cls._func_signature(
+            chain, udf_func
         )
```
dreadatour, line 63: Not sure why not to use `self` here 👀

src/datachain/lib/udf_signature.py
```diff
@@ -58,15 +58,23 @@ def parse(
...
+        udf_params: dict[str, Union[DataType, Any]] = {}
         if params:
-            udf_params = [params] if isinstance(params, str) else params
-        elif not func_params_map_sign:
-            udf_params = []
-        else:
-            udf_params = list(func_params_map_sign.keys())
+            udf_params = (
+                {params: Any} if isinstance(params, str) else dict.fromkeys(params, Any)
+            )
```
dreadatour, line 69: Use `Any` type here for params with unknown type, since we don't have an `Unknown` type in Python 👀

src/datachain/lib/udf.py
```diff
@@ -159,6 +159,7 @@ def process(self, file) -> list[float]:
         ```
     """
 
+    is_input_batched = False
```
shcheklein, line 162: why not use the same `is_output_batch`? (may be rename it to `is_batched)? it seems it always has the same value

src/datachain/lib/signal_schema.py
shcheklein: > o get output type of lambda functions

output or input?

src/datachain/lib/udf_signature.py
shcheklein: can it be None? what is parameter type is type annotation is missing ... e.g. how does mypy treat it?

src/datachain/lib/signal_schema.py
```diff
@@ -464,19 +464,63 @@ def contains_file(self) -> bool:
...
+                if param_type is list:
+                    schema[param] = schema_type
+                    continue
+
+                if param_origin is not list:
+                    raise SignalResolvingError(param.split("."), "is not a list")
+
+                param_type = get_args(param_type)[0]
+
+            if param_type == schema_type or (
```
shcheklein, line 510: do we do this only for File? does it make sense to make it general?

src/datachain/lib/udf.py
dreadatour: For generators input will not be batched (single row), but output is batched, thus we need to have separate params for these.

src/datachain/lib/signal_schema.py
dreadatour: Good question. For Files we are sure no additional fields are exists in different File type models (File, ImageFile, VideoFile, etc). But in general other models inherited from each other may have additional fields and simple type conversion may fails. We can also check this, but it will complicate things and I am not sure I can see the value here.

We can do it later if needed.

src/datachain/lib/udf_signature.py
dreadatour: This is the case when `params` is set in `map` chain method. Here we are not checking UDF signature param types, e.g.:

```python
dc.map(lambda file: file.parse_info(), params="file")
```

or

```python
def process_signals(signal1, signal2) -> int:
    return signal1 + signal2

dc.map(process_signals, params=["signal1", "signal2"])
```

---

Answering your question, if parameter type is missing, `inspect` will return it as `inspect.Parameter.empty` (see [below](https://github.com/iterative/datachain/pull/973/files#diff-5d3be55eec360ab345279685e953f961a442346fd180ddfd1df214af2a96eccaR71-R76)). In this case we are returning it's type as `Any` (since where is no `Unknown` type in Python).

ilongin:
Haven't looked into the PR yet but am I the only one little bit concerned with these implicit `File` conversions? ... What worries me is that we are not consistent, i.e `int` -> `str` is not converted automatically but `File` -> `ImageFile` is.  Is the only reason for this to avoid that `as_image_file()` method call by user in UDF itself or there's more to it?

dreadatour:
> `int` -> `str` is not converted automatically but `File` -> `ImageFile` is

This is because `int` and `str` are completely different types, while `ImageFile` is inherited from the `File` model and is, basically, the same type, just with some additional helper methods defined.

I prefer to think about this the same way as `int` and `float` types behave:

```
$ cat sum.py
def sum(x: float, y: float) -> float:
    return x + y

print(sum(3, 4))
$ python sum.py
7
$ mypy sum.py
Success: no issues found in 1 source file
$
```

Different types, but converted seamless without any warning. Looks logical, intuitive and natural to me 😇

ilongin:
> > `int` -> `str` is not converted automatically but `File` -> `ImageFile` is
> 
> This is because `int` and `str` are completely different types, while `ImageFile` is inherited from the `File` model and is, basically, the same type, just with some additional helper methods defined.
> 
> I prefer to think about this the same way as `int` and `float` types behave:
> 
> ```
> $ cat sum.py
> def sum(x: float, y: float) -> float:
>     return x + y
> 
> print(sum(3, 4))
> $ python sum.py
> 7
> $ mypy sum.py
> Success: no issues found in 1 source file
> $
> ```
> 
> Different types, but converted seamless without any warning. Looks logical, intuitive and natural to me 😇

Yes, there are some implicit casting happening in some languages, but what I was worried more is to be inconsistent and to only do implicit casting for `File` objects. If we are always implicitly casting subclasses for any kind of custom user classes then I'm ok with this.

dreadatour:
> If we are always implicitly casting subclasses for any kind of custom user classes then I'm ok with this.

Then IMO we need to make sure no additional fields were added in subclasses. See [my comment above](https://github.com/iterative/datachain/pull/973#discussion_r1997790885).

shcheklein (approved):

PR #974, opened 2025-03-16
Test if 'test_import_time' test is reliable in GitHub CI

dreadatour:
Too many failures of `test_import_time` test in my PR https://github.com/iterative/datachain/pull/973.
I just wanted to be sure if this is not relevant to my changes and only depends on GitHub CI Actions.

### DIFF
--- a/tests/test_import_time.py
+++ b/tests/test_import_time.py
@@ -10,7 +10,7 @@ from datachain import C, DataChain
 logger = logging.getLogger(__name__)
 
 MAX_ATTEMPTS = 3
-MAX_IMPORT_TIME_MS = 700
+MAX_IMPORT_TIME_MS = 701
 
 lazy_modules = [
     "adlfs",


skshetry:
It was taking ~450ms on `ubuntu-latest-8-cores` last I checked, so I added some slack and set it to 700ms. It should never reach that number, much less exceed it

I'll investigate the issue.

dreadatour:
> It was taking ~450ms on `ubuntu-latest-8-cores` last I checked, so I added some slack and set it to 700ms. It should never reach that number, much less exceed it
> 
> I'll investigate the issue.

I know, I have run it on my localhost and it wasn't even close to 700ms, but in my PR it failed every second time 😥 I think it is because of parallel tests running, may be because of the shared GitHub CI workers.

dreadatour:
Fixed in https://github.com/iterative/datachain/pull/975

PR #975, opened 2025-03-17
fix(tests): disable coverage in test_import_time

skshetry:
This commit disables coverage for the test_import_time test to avoid unpredictable results. The import usually takes  ~400ms but with coverage enabled, it was frequently going past 600ms on my machine.

Which I believe is also happening on the CI.

- Also added a log message to display the full contents of the output from `-Ximporttime` in case of a parse error in `from_csv`.

Thanks to @dreadatour for noticing this issue. Fixes #974.

### DIFF
--- a/tests/test_import_time.py
+++ b/tests/test_import_time.py
@@ -52,7 +52,7 @@ def _import_time_chain(test_session):
     )
 
 
-# disable coverage for this test to avoid unpredictable results
+# disable coverage for this test to minimize import time overhead
 @pytest.mark.no_cover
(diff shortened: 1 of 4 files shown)


tests/test_import_time.py
```diff
@@ -47,6 +52,8 @@ def _import_time_chain(test_session):
     )
 
 
+# disable coverage for this test to avoid unpredictable results
```
skshetry, line 55:
```suggestion
# disable coverage for this test to minimize import time overhead
```

dreadatour (approved):
Thank you! 🙏❤️

ilongin (approved):

Issue #317, opened 2024-08-19
Make `.get_storage()` caching listing lazy


ilongin:
This is follow up for PR https://github.com/iterative/datachain/pull/294 where listing is saved (cached) inside `.from_storage()` method itself which means it's not lazy unlike all the other `DataChain` steps. 
We should think how to postpone caching / saving listing when all chain steps are applied.

Ideas:
1. Make some generic step that accepts function (closure) in `DatasetQuery`
2. Create `.terminate()` on `DataChain` class
3. ...

PR #976, opened 2025-03-17
Making listing lazy in `DatasetQuery`

ilongin:
Before we had listing process happening in `.from_storage()` method itself which meant it wasn't lazy. 
Idea was to move it to `DatasetQuery.apply_steps()` instead.

Fixes: #317


skshetry:
@ilongin, do we have some caching here? I believe the chain would rerun when executed?
Eg: 

```python
dc = DataChain.from_storage(...)
dc.exec() # runs once
dc.exec() # will it rerun the generator function again?
```

src/datachain/lib/dc.py
```diff
@@ -475,25 +475,29 @@ def from_storage(
...
+                    cls.from_records(
+                        DataChain.DEFAULT_FILE_RECORD,
+                        session=session,
+                        settings=settings,
+                        in_memory=in_memory,
+                    )
+                    .settings(prefetch=0)
+                    .gen(
+                        list_bucket(list_uri, cache, client_config=client_config),
+                        output={f"{object_name}": File},
```
dreadatour, line 495: If I am not mistaken:

```suggestion
                        output={f"{object_name}": file_type},
```

src/datachain/query/dataset.py
```diff
@@ -1097,26 +1091,43 @@ def __init__(
...
+        self.list_ds_name: Optional[str] = None
 
-        if fallback_to_studio and is_token_set():
-            ds = self.catalog.get_dataset_with_remote_fallback(name, version)
+        self.name = name
+        self.dialect = self.catalog.warehouse.db.dialect
+        if version:
+            self.version = version
+
+        if is_listing_dataset(self.name):
```
dreadatour, line 1109: Even if they are the same (`self.name` and `name`), we are using `name` below in this branch, so IMO it should be either `name` or `self.name` everywhere, just to be consistent.

```suggestion
        if is_listing_dataset(name):
```

src/datachain/query/dataset.py
```diff
@@ -1097,26 +1091,43 @@ def __init__(
...
+            # not setting query step yet as listing dataset might not exist at
+            # this point
+            self.list_ds_name = name
+        elif fallback_to_studio and is_token_set():
```
dreadatour, line 1113: Not related to this PR, but `is_token_set` here looks odd and raises questions.

We may want to import it as:

```
from datachain.remote.studio import is_token_set as is_studio_token_set
```

above, for example, just for the better readability of the code here.

src/datachain/query/dataset.py
```diff
@@ -1097,26 +1091,43 @@ def __init__(
         self.temp_table_names: list[str] = []
         self.dependencies: set[DatasetDependencyType] = set()
         self.table = self.get_table()
-        self.starting_step: StartingStep
+        self.query_step: Optional[QueryStep] = None
```
dreadatour, line 1094: I am wonder if we can think of better name for `query_step` attribute. I have a lot of questions lower in the code and keep confusing about this name. Should we, may be, keep `starting_step` name as it is more verbose? 🤔

src/datachain/query/dataset.py
```diff
@@ -1180,11 +1191,24 @@ def c(self, column: Union[C, str]) -> "ColumnClause[Any]":
...
+
     def apply_steps(self) -> QueryGenerator:
         """
         Apply the steps in the query and return the resulting
         sqlalchemy.SelectBase.
         """
+        if self.list_fn:
+            self.list_fn()
+
+        if self.list_ds_name:
```
dreadatour, line 1209: I am not sure why do we need to pass `list_ds_name`? Can `list_fn` returns the listed dataset name?

dreadatour (commented):
Looks good to me overall, thank you, this is a great improvement!

Couple comments below and I also haven't seen tests for the case described in the PR (making listing lazy). If I missed one, sorry, if not, should we consider to add one?

src/datachain/query/dataset.py
ilongin: `list_fn` is optional so if it's not defined, we still need `list_ds_name` to create that starting step which we couldn't create in constructor as there listing maybe hasn't happened yet so listing dataset could not exist. We also cannot use `self.name` as that one is present only for "attached" chains which are those that are pointing to whole dataset  and on any modification method call (e.g `.filter()`) it is removed as chain becomes "unattached"

ilongin:
> @ilongin, do we have some caching here? I believe the chain would rerun when executed? Eg:
> 
> ```python
> dc = DataChain.from_storage(...)
> dc.exec() # runs once
> dc.exec() # will it rerun the generator function again?
> ```

Yes, chain will rerun in this example, together with listing if it is needed (listing doesn't exist yet or `update` flag is provided) so I would say no caching.

src/datachain/query/dataset.py
dreadatour: Thank you! I really think it will be more verbose 🙏

src/datachain/query/dataset.py
dreadatour: Got it, thank you so much for explanation 🙏

dreadatour (approved):
Looks good to me 👍❤️

src/datachain/query/dataset.py
shcheklein: yes, agreed cc @amritghimire  ... it is still not a good idea to have Studio exposed this way

ideally it should be just `get_dataset`, inside it it should be deciding on fallback

ilongin: Listing is already done with `gen` but we cannot just append the rest of the chain to that part as we want to cache listing at some point, i.e call `save()` on it and if we call it the middle of the chain it's not lazy any more. It needs to happen in the `save()` of the dataset when we apply other steps. 
So we could do
```python
def from_ storage():
    return (
        cls.from_records(DEFAULT_FILE_RECORDS)
           .gen(list_bucket(,,,))
           .save(list_ds_name, listing=True) 
    )

ds = DataChain.from_storage("s3://ldb-public").filter(...).map(...).save("my_dataset")
```
This is similar as it was before this PR but it's not lazy and to make it lazy we need to add some step in `DatasetQuery` as there we start to apply steps. 
Ideal solution would be to move all those steps and `apply_step` function from `DatasetQuery` to `DataChain` as there is no point for main logic to be there IMO and maybe even remove `DatasetQuery` alltogether but that's whole another topic.

src/datachain/query/dataset.py
```diff
@@ -1180,11 +1191,24 @@ def c(self, column: Union[C, str]) -> "ColumnClause[Any]":
         col.table = self.table
         return col
 
+    def set_listing_pre_step(self, list_fn: Callable) -> None:
```
shcheklein, line 1194: Q: does it have to have a "listing" in its name? or is it a general mechanism that can be applied even in the future for other steps?

essentially, we can make it less listing specific here and do `before_apply_callback` or something

ilongin: It doesn't need to be. I thought of making this generic but didn't see any current need for it atm so decided to make it explicit. Also, this is all in internal `DatasetQuery`. If we would need something like `before_apply_callback` we would need to add public method in `DataChain` as well. This is all easy to add if needed in future.

shcheklein (approved):

src/datachain/query/dataset.py
shcheklein: >  I thought of making this generic but didn't see any current need for it atm so decided to make it explicit.

it might confuse people in the future (it looks specific, and I would be looking why it is specific and if I can reuse it for something else).

shcheklein: or it might even lead for someone duplicating it

src/datachain/query/dataset.py
```diff
@@ -1180,11 +1190,23 @@ def c(self, column: Union[C, str]) -> "ColumnClause[Any]":
...
+        """
+        self.before_steps.append(fn)
+
     def apply_steps(self) -> QueryGenerator:
         """
         Apply the steps in the query and return the resulting
         sqlalchemy.SelectBase.
         """
+        for fn in self.before_steps:
+            fn()
```
amritghimire, line 1205: With this I saw a caveat, that fn seems to be called every time a step is performed since we don't clear the before steps at any time. So, whenever I try to use the collect or chain, I am getting the query to refetch the table instead.

src/datachain/lib/dc/storage.py
```diff
@@ -95,24 +94,28 @@ def from_storage(
...
+            (
+                from_records(
+                    DataChain.DEFAULT_FILE_RECORD,
+                    session=session,
+                    settings=settings,
+                    in_memory=in_memory,
+                )
+                .settings(prefetch=0)
+                .gen(
+                    list_bucket(list_uri, cache, client_config=client_config),
```
amritghimire, line 113: This seems to be called everytime I use the datachain to apply steps. Should'nt this be applied only once?

src/datachain/lib/dc/storage.py
```diff
@@ -95,24 +94,28 @@ def from_storage(
         dc.signals_schema = dc.signals_schema.mutate({f"{object_name}": file_type})
         return dc
 
+    dc = from_dataset(list_ds_name, session=session, settings=settings)
```
amritghimire, line 97: Calling from_dataset when list_ds_exists is false also doesn't seem right

src/datachain/query/dataset.py
ilongin: This is expected and it's how it was before when listing was lazy (before we refactored it using `DataChain` higher level functions).

src/datachain/lib/dc/storage.py
ilongin: No, it should be called every time you apply steps. The whole idea is for user to apply steps only once anyway as it's very expensive operation.

src/datachain/lib/dc/storage.py
ilongin: Lower level code (`DatasetQuery`) is aware of listing being lazy so this is ok. We will start chain with listing dataset and the fact it doesn't exists yet is just the nature of it's "laziness"

src/datachain/query/dataset.py
amritghimire: Yes, but it seems to run every time I run `chain.collect()` or `chain.count()` . As you can see in the test `test_from_storage_multiple_uris_cache` in https://github.com/iterative/datachain/pull/994 , it is called every time for chains.

src/datachain/lib/dc/storage.py
amritghimire: I mean we could get dataset not found error when the ist_ds_name doesn't exist

src/datachain/query/dataset.py
ilongin: `chain.collect()` applies steps every time it's called

amritghimire: Yes, my question was should we rerun listing every time collect is called when update is passed? Or once should suffice?

PR #980, opened 2025-03-19
feat(dc): add repr to show schema for DataChain instances

skshetry:
This is a quick and dirty implementation to display `__repr__` for `DataChain` instances. The implementation uses `DataChain.print_schema()` method to capture the output and return it.

This makes working in the REPL and notebooks easier. PTAL at the tests for the output.

Also note that I changed the default indentation to `2` from `4` before. I think less indentation makes it easier on REPL to work with.

And also fixed a small bug in `SignalSchema.print_tree` that was always indenting the schema by at least a single space.

```py
>>> from datachain import DataChain
>>> dc = DataChain.from_values(col1=["a", "b", "c"])
>>> dc
col1: str
>>> dc.select_except("col1")
Empty DataChain
>>> dc.map(col2=lambda col1: col1 * 2)
col1: str
col2: str
```

### DIFF
--- a/src/datachain/lib/dc.py
+++ b/src/datachain/lib/dc.py
@@ -6,6 +6,7 @@ import sys
 from collections.abc import Iterator, Sequence
 from functools import wraps
 from typing import (
+    IO,
     TYPE_CHECKING,
     Any,
     BinaryIO,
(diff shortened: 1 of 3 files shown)


amritghimire (approved):

skshetry:
We could alternatively represent it like a tree in a table. Perhaps like the following?
```
>>> dc
Columns (8 total):
Column                 Type       Full path
------                 -----      ---------
sign1                  MyNested   sign1
├── label              str        sign1.label
├── fr                 MyFr       sign1.fr
│    ├── nnn           str        sign1.fr.nnn
│    └── count         int        sign1.fr.count
├── col1               str        col1
└── col2               str        col2

```

PR #981, opened 2025-03-19
Enhance Dataset Management: Add description and labels support

amritghimire:
- Updated Catalog, Metastore, and DataChain classes to include optional description and labels parameters for dataset creation.
- Modified show command to display dataset description and labels.
- Added a unit test to verify the new functionality for saving datasets with description and labels.


dreadatour (approved):
Looks good to me 👍 Only two comments:

- We might want to add validation, for labels, at least.
- More tests needed 👀

amritghimire:
> Looks good to me 👍 Only two comments:
> 
> * We might want to add validation, for labels, at least.

Couldn't find the proper validation as it doesn't seem to exist in studio side or here so far.

shcheklein (commented):
How about CLI / local tool?  I think we need to have this metadata for all datasets cc @dmpetrov

amritghimire:
> How about CLI / local tool? I think we need to have this metadata for all datasets cc @dmpetrov

WDYM by CLI/local tool?

We can edit the labels and descriptions from UI:

```
datachain ds edit --help
usage: datachain dataset edit [-h] [-v] [-q] [--new-name NEW_NAME] [--description DESCRIPTION] [--labels LABELS [LABELS ...]] [--studio] [-L] [-a] [--team TEAM] name

Edit dataset metadata.

positional arguments:
  name                  Dataset name

options:
  --new-name NEW_NAME   Dataset new name
  --description DESCRIPTION
                        Dataset description
  --labels LABELS [LABELS ...]
                        Dataset labels
  --studio              Edit dataset from Studio
  -L, --local           Edit local dataset only
  -a, --all             Edit both datasets from studio and local
  --team TEAM           The team to edit a dataset. By default, it will use team from config
  -h, --help            show this help message and exit
  -v, --verbose         Be verbose
  -q, --quiet           Be quiet
```

We can view it using show:
```
datachain show new_name --schema --hidden
Name:  new_name
Description:  new description
Labels:  new_label,old_label


   sys__id            sys__rand key
0        1   765796559596144837   a
1        2  8876953003906727361   b
2        3  3380766838906907097   c

Schema:
key: str
```

shcheklein:
@amritghimire I mean a scenario like - pull dataset from Studio, see the list of datasets / list them with DataChain - we should be able to see the tags assigned in Studio. Other way around - if we push a dataset to Studio back (can we do this at all?) and we have some local tags and such - they should be propagated to the backend.

PR #985, opened 2025-03-20
POC: Add an inexpensive test to set anon if no creds are found in gcs

### LINKED ISSUE #740: GCP missing creds slows down significantly
### Description

See quick start.

This command take 20 seconds (sometimes 30) without any progressbar if I don't have any GCP creds setup. If I put `anon=True` - it works instantly like it suppose to.

```python
meta = DataChain.from_json("gs://datachain-demo/dogs-and-cats/*json", object_name="meta")
```

This affects all users since we use GCP in most of our examples and, I'm sure, majority of user do not have GCP creds set up.

What's needed: we need to avoid the delay when GCP creds are not set up. Ideally, without specifying `anon=True` which adds more code in examples and mental overhead for users.

ToDo:
 - [ ] Fix the issue
 - [ ] Remove `anon=True` from Readme and Get Started

amritghimire:
Regarding the discussion on #740, this is a possible approach to specify
anon token in gcs if no creds is found.
This is the similar to the approach we already have s3.

The approach is to try different method of connection before defaulting
back to anon as gcs filesystem does. But we are skipping the metadata
check for the case of google cloud engine.

Testing the improvement with the code:
```py
import time

start = time.time()
images = DataChain.from_storage("gs://datachain-demo/dogs-and-cats/*jpg")
images.show()
end = time.time()
print(f"Time taken: {end - start} seconds")
```

The improvement was from:
27.73566699028015 seconds

to:
Time taken: 6.101557016372681 seconds

Closes #740

> Note: this is just a proof of concept PR, if we want to go this route or not. Needs refinement if we want to go this way. Created a PR this way so it is easier to view the difference.

### DIFF
--- a/src/datachain/client/gcs.py
+++ b/src/datachain/client/gcs.py
@@ -16,6 +16,7 @@ from .fsspec import DELIMITER, Client, ResultQueue
 # Patch gcsfs for consistency with s3fs
 GCSFileSystem.set_session = GCSFileSystem._set_session
 PageQueue = asyncio.Queue[Optional[Iterable[dict[str, Any]]]]
+ANONYMOUS_TOKEN = "anon"  # noqa: S105
 
 
 class GCSClient(Client):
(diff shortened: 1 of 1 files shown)


src/datachain/client/gcs.py
```diff
@@ -24,12 +25,85 @@ class GCSClient(Client):
...
+        # Try various authentication methods
+        gcreds = GoogleCredentials(
+            token=ANONYMOUS_TOKEN,
+            project=DEFAULT_PROJECT,
+            access="full_control",
+        )
+
+        # Define authentication methods to try
+        auth_methods = ["google_default", "cache"]
+        if is_on_gce(request_callback):
```
shcheklein, line 87: @amritghimire what is the logic behind the request_callback? will always raise or only in certain cases?

src/datachain/client/gcs.py
```diff
@@ -24,12 +25,85 @@ class GCSClient(Client):
...
+            return kwargs
+
+        # If anonymous access requested, force anonymous access
+        if kwargs.get("anon"):
+            kwargs["token"] = ANONYMOUS_TOKEN
+            return kwargs
+
+        # Try various authentication methods
+        gcreds = GoogleCredentials(
+            token=ANONYMOUS_TOKEN,
```
shcheklein, line 80: why do we pass anon token here?

src/datachain/client/gcs.py
```diff
@@ -24,12 +25,85 @@ class GCSClient(Client):
...
-            kwargs["token"] = "anon"  # noqa: S105
+            return kwargs
+
+        # If token is provided, use it
+        if "token" in kwargs:
+            return kwargs
+
+        # If anonymous access requested, force anonymous access
+        if kwargs.get("anon"):
+            kwargs["token"] = ANONYMOUS_TOKEN
```
shcheklein, line 75: is it a way fsspec internally distinguishes it?

src/datachain/client/gcs.py
amritghimire: It is to disable the connection to metadata server. It raises exception and the is_on_gce will check for existence of a file to determine if certain file exist. You can look into the code of is_on_gce for more context.

src/datachain/client/gcs.py
amritghimire: Yes,

src/datachain/client/gcs.py
shcheklein: why can't it be disabled on the fsspec side? does it mean that metadata server can be ignored, or does it mean that it won't be retrying anymore?

amritghimire: It won't be retrying it anymore. 
Also, this is just a proof of concept PR, if we want to go this route or not. Needs refinement if we want to go this way. Created a PR this way so it is easier to view the difference.

amritghimire: It is not in fsspec. It is in `from google.auth.compute_engine._metadata import is_on_gce` as you can see from the import above.

shcheklein: `fsspec` can use it the same way we do here, quickly check in the cloud auth path (probably that's what Martin was suggesting in the corresponding fsspec discussion, but I don't remember the details by now)

shcheklein: btw, why is it considered safe? was there a reason for this retries or it's just a mistake on the Google lib side?

amritghimire: Before we go that route, let's see if this works for us or not first internally IMO.

This is a hacky fix for now to be sure.

 @0x2b3bfa0 may have more context on this.  If we can skip the metadata check or not. 

I was wondering is it safe for datachain to skip the metadata check altogether or not? 


So, users on google compute engine could use token to cloud to force this way of communication.

0x2b3bfa0: I got sidetracked by a lot of things and I lack enough context to respond. Please ping me in a few day if you still need my input. Sorry! 🙈

shcheklein: > I was wondering is it safe for datachain to skip the metadata check altogether or not?

answer should be the same as for fsspec to my mind, or google lib

> So, users on google compute engine could use token to cloud to force this way of communication.

they could, but probably won't expect the discrepancy

amritghimire: As I mentioned, this is a hacky fix. We can look for proper way to fix this or look for proper way either to suggest it to fsspec or google lib or use a request which doesn't retry for 5 times when it fails for gce check.

We first need to figure out or discuss if we want to implement something like this. @shcheklein cc. @iterative/datachain

amritghimire: If we don't want some hacky fix like this here, I don't think we should spend more time to see how to skip the metadata check here.

PR #994, opened 2025-03-25
Enhance `from_storage` function to support multiple URIs

amritghimire:
- Updated the `from_storage` function to accept a list of URIs, allowing for batch processing of storage locations.
- Introduced a new internal function `_extract_chain` to streamline the extraction of data from individual URIs.
- Enhanced documentation to clarify usage and added examples for multiple URIs.
- Improved caching logic to avoid redundant updates for URIs pointing to the same storage location.

This change increases the flexibility and efficiency of data retrieval
in the DataChain library.

### DIFF
--- a/src/datachain/client/fsspec.py
+++ b/src/datachain/client/fsspec.py
@@ -89,9 +89,9 @@ class Client(ABC):
         from .local import FileClient
         from .s3 import ClientS3
 
-        protocol = urlparse(str(url)).scheme
+        protocol = urlparse(os.fspath(url)).scheme
 
-        if not protocol or _is_win_local_path(str(url)):
+        if not protocol or _is_win_local_path(os.fspath(url)):
             return FileClient
         if protocol == ClientS3.protocol:
             return ClientS3
@@ -122,7 +122,7 @@ class Client(ABC):
         source: Union[str, os.PathLike[str]], cache: Cache, **kwargs
     ) -> "Client":
         cls = Client.get_implementation(source)
-        storage_url, _ = cls.split_url(str(source))
+        storage_url, _ = cls.split_url(os.fspath(source))
         if os.name == "nt":
             storage_url = storage_url.removeprefix("/")
 
--- a/src/datachain/lib/dc/storage.py
+++ b/src/datachain/lib/dc/storage.py
@@ -1,4 +1,5 @@
 import os.path
+from collections.abc import Iterable
 from typing import (
     TYPE_CHECKING,
     Optional,
@@ -9,7 +10,12 @@ from datachain.lib.file import (
     FileType,
     get_file_type,
 )
-from datachain.lib.listing import get_file_info, get_listing, list_bucket, ls
+from datachain.lib.listing import (
+    get_file_info,
+    get_listing,
+    list_bucket,
+    ls,
+)
 from datachain.query import Session
 
 if TYPE_CHECKING:
@@ -17,7 +23,7 @@ if TYPE_CHECKING:
 
 
 def from_storage(
-    uri: Union[str, os.PathLike[str]],
+    uri: Union[str, os.PathLike[str], Iterable[str], Iterable[os.PathLike[str]]],
     *,
     type: FileType = "binary",
     session: Optional[Session] = None,
@@ -29,11 +35,12 @@ def from_storage(
     anon: bool = False,
     client_config: Optional[dict] = None,
 ) -> "DataChain":
-    """Get data from a storage as a list of file with all file attributes.
+    """Get data from storage(s) as a list of file with all file attributes.
     It returns the chain itself as usual.
 
     Parameters:
-        uri : storage URI with directory. URI must start with storage prefix such
+        uri : storage URI with directory or list of URIs.
+            URIs must start with storage prefix such
             as `s3://`, `gs://`, `az://` or "file:///"
         type : read file as "binary", "text", or "image" data. Default is "binary".
         recursive : search recursively for the given path.
@@ -42,16 +49,26 @@ def from_storage(
         anon : If True, we will treat cloud bucket as public one
         client_config : Optional client configuration for the storage client.
 
-    Example:
-        Simple call from s3
-        ```py
+    Returns:
+        DataChain: A DataChain object containing the file information.
+
+    Examples:
+        Simple call from s3:
+        ```python
         import datachain as dc
         chain = dc.from_storage("s3://my-bucket/my-dir")
         ```
 
-        With AWS S3-compatible storage
-        ```py
-        import datachain as dc
+        Multiple URIs:
+        ```python
+        chain = dc.from_storage([
+            "s3://bucket1/dir1",
+            "s3://bucket2/dir2"
+        ])
+        ```
+
+        With AWS S3-compatible storage:
+        ```python
         chain = dc.from_storage(
             "s3://my-bucket/my-dir",
             client_config = {"aws_endpoint_url": "<minio-endpoint-url>"}
@@ -61,9 +78,15 @@ def from_storage(
         Pass existing session
         ```py
         session = Session.get()
-        import datachain as dc
-        chain = dc.from_storage("s3://my-bucket/my-dir", session=session)
+        chain = dc.from_storage([
+            "path/to/dir1",
+            "path/to/dir2"
+        ], session=session, recursive=True)
         ```
+
+    Note:
+        When using multiple URIs with `update=True`, the function optimizes by
+        avoiding redundant updates for URIs pointing to the same storage location.
     """
     from .datachain import DataChain
     from .datasets import from_dataset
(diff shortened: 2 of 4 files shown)


src/datachain/lib/dc/storage.py
```diff
@@ -79,40 +154,35 @@ def from_storage(
...
-            )
-            .save(list_ds_name, listing=True)
-        )
-
-    dc = from_dataset(list_ds_name, session=session, settings=settings)
-    dc.signals_schema = dc.signals_schema.mutate({f"{object_name}": file_type})
+        dc = dc.union(chain) if dc else chain
 
-    return ls(dc, list_path, recursive=recursive, object_name=object_name)
+    return dc  # type: ignore[return-value]
```
shcheklein, line 188: why do we ingore types here?

src/datachain/client/fsspec.py
```diff
@@ -89,9 +89,9 @@ def get_implementation(url: Union[str, os.PathLike[str]]) -> type["Client"]:
         from .local import FileClient
         from .s3 import ClientS3
 
-        protocol = urlparse(str(url)).scheme
+        protocol = urlparse(os.fspath(url)).scheme
```
shcheklein, line 92: what is the difference, why do we need this?

src/datachain/lib/dc/storage.py
```diff
@@ -79,40 +154,35 @@ def from_storage(
...
-            .gen(
-                list_bucket(list_uri, cache, client_config=client_config),
-                output={f"{object_name}": File},
-            )
-            .save(list_ds_name, listing=True)
-        )
-
-    dc = from_dataset(list_ds_name, session=session, settings=settings)
-    dc.signals_schema = dc.signals_schema.mutate({f"{object_name}": file_type})
+        dc = dc.union(chain) if dc else chain
```
shcheklein, line 186: can we avoid doing union for files? it can be quite expensive creating a table per each file in the list and then doing all these unions

it is indeed (most likely) needed for non-file sources (need a listing and then union them), unfortunately

src/datachain/client/fsspec.py
amritghimire: https://peps.python.org/pep-0519/

https://python-forum.io/thread-32670.html

src/datachain/lib/dc/storage.py
amritghimire: Because mypy assumes that the value can be None, but when the reality is it cannot be none as we have enough safeguards above.

src/datachain/lib/dc/storage.py
amritghimire: We have special case for files where we use dataset from values in case of file 

https://github.com/iterative/datachain/blob/main/src/datachain/lib/dc/storage.py#L84-L98

src/datachain/lib/dc/storage.py
shcheklein: let's put assert a line above then please

`assert dc`

src/datachain/client/fsspec.py
shcheklein: could you please summarize? is it a cosmetic change? linter complaining? any particular edge case it it solving?

src/datachain/lib/dc/storage.py
shcheklein: will it be a single table per each file + union at the end?

src/datachain/client/fsspec.py
amritghimire: > This lack of support required users of pathlib to manually convert path objects to strings by calling str(path) which many found error-prone.

One issue in converting path objects to strings comes from the fact that the only generic way to get a string representation of the path was to pass the object to str().
This can pose a problem when done blindly as nearly all Python objects have some string representation whether they are a path or not,
e.g. str(None) will give a result that builtins.open() [5] will happily use to create a new file.

src/datachain/lib/dc/storage.py
amritghimire: I am not sure what assert will achieve or improve here tbh, is it only to correct mypy type detection?

If for providing feedback to user, you can see we raise value error if uri is empty, the only case when it can be None.

src/datachain/lib/dc/storage.py
amritghimire: > will it be a single table per each file + union at the end?

Sorry, I don't understand, WDYM?

shcheklein: does it call a single `from_records` per each single file?

src/datachain/client/fsspec.py
shcheklein: okay, I see, thanks! Minor note: let's try to do this in a separate PR if possible.

src/datachain/lib/dc/storage.py
amritghimire: As you can see from the code linked above, it doesn't even call from records but call from_values for a single file

shcheklein: okay, but it does it for every single file, right? essentially creating as many table in DB as the number of files you passed?

amritghimire: Yeah, I think so. I will look into the possible way of creating a single table for each files passed,

amritghimire: Updated PTAL

skshetry:
> * Updated the `from_storage` function to accept a list of URIs, allowing for batch processing of storage locations.

Could you please expand more on this? Is this required for Studio? If so, can you point me to a particular issue?

amritghimire:
> > * Updated the `from_storage` function to accept a list of URIs, allowing for batch processing of storage locations.
> 
> 
> 
> Could you please expand more on this? Is this required for Studio? If so, can you point me to a particular issue?

https://github.com/iterative/studio/issues/11423#issuecomment-2734037698

src/datachain/lib/dc/storage.py
```diff
@@ -79,40 +142,67 @@ def from_storage(
...
+            _, lst_uri, _ = parse_listing_uri(os.fsdecode(single_uri), client_config)
+            if lst_uri in listed_uris:
+                listed_uris.add(lst_uri)
+                update_for_uri = False
+
+        list_ds_name, list_uri, list_path, list_ds_exists = get_listing(
+            single_uri, session, update=update
+        )
+
+        # ds_name is None if object is a file, we don't want to use cache
```
ilongin, line 168:
```suggestion
        # list_ds_name is None if object is a file, we don't want to use cache
```

src/datachain/lib/dc/storage.py
```diff
@@ -79,40 +142,67 @@ def from_storage(
...
+        storage_chain = storage_chain.union(chain) if storage_chain else chain
+
+    if file_values:
+        file_chain = from_values(
+            session=session,
+            settings=settings,
+            in_memory=in_memory,
+            file=file_values,
         )
+        file_chain.signals_schema = file_chain.signals_schema.mutate(
```
ilongin, line 201: Not specifically related to this PR as I've seen we are doing this in multiple places, but I'm wondering if this can be done better, i.e. we shouldn't need to manually change `signals_schema` at all - it should be changed based on public methods that are applied to the chain, e.g renaming file etc.

ilongin (changes requested):

src/datachain/lib/dc/storage.py
```diff
@@ -10,15 +10,21 @@
...
 from datachain.query import Session
 
 if TYPE_CHECKING:
     from .datachain import DataChain
 
 
 def from_storage(
-    uri: Union[str, os.PathLike[str]],
+    uri: Union[str, os.PathLike[str], list[str], list[os.PathLike[str]]],
     *,
```
dreadatour, line 28: Should we use `Iterable` or `Sequence` instead of `list` type to also allow `tuple`, for example?

Also just wonder if we can use `*args` for this.

```python
dc.from_storage("s3://bucket1/dir1")
```

and

```python
dc.from_storage("s3://bucket1/dir1", "s3://bucket1/dir2")
```

instead of using lists:

```python
dc.from_storage(["s3://bucket1/dir1", "s3://bucket1/dir2"])
```

src/datachain/lib/dc/storage.py
```diff
@@ -43,16 +50,26 @@ def from_storage(
         anon : If True, we will treat cloud bucket as public one
         client_config : Optional client configuration for the storage client.
 
-    Example:
-        Simple call from s3
-        ```py
+    Returns:
+        DataChain: A DataChain object containing the file information.
```
dreadatour, line 54: ❤️

src/datachain/lib/dc/storage.py
```diff
@@ -62,9 +79,15 @@ def from_storage(
...
-        chain = dc.from_storage("s3://my-bucket/my-dir", session=session)
+        chain = dc.from_storage([
+            "path/to/dir1",
+            "path/to/dir2"
+        ], session=session, recursive=True)
         ```
+
+    Note:
+        When using multiple URIs with `update=True`, the function optimizes by
+        avoiding redundant updates for URIs pointing to the same storage location.
```
dreadatour, line 90: ❤️

src/datachain/lib/dc/storage.py
```diff
@@ -79,40 +102,73 @@ def from_storage(
...
+                from_records(
+                    DataChain.DEFAULT_FILE_RECORD,
+                    session=session,
+                    settings=settings,
+                    in_memory=in_memory,
+                )
+                .settings(prefetch=0)
+                .gen(
+                    list_bucket(list_uri, cache, client_config=client_config),
+                    output={f"{object_name}": File},
```
dreadatour, line 148: It may be obvious, but I can not see the answer. Why do we need

```python
        file_chain.signals_schema.mutate(
            {f"{object_name}": file_type}
        )
```

below? Why don't we just set

```python
output={f"{object_name}": file_type}
```

here?

src/datachain/lib/dc/storage.py
```diff
@@ -79,40 +102,73 @@ def from_storage(
...
+            # disable prefetch for listing, as it pre-downloads all files
+            (
+                from_records(
+                    DataChain.DEFAULT_FILE_RECORD,
+                    session=session,
+                    settings=settings,
+                    in_memory=in_memory,
+                )
+                .settings(prefetch=0)
+                .gen(
```
dreadatour, line 146: May be naive question, but still. Is it possible to use single generator to generate listings for all the URIs here? 🤔

dreadatour (commented):
Looks good to me overall, couple comments below.

Also we need more tests for this. List paths and files in different combinatons, with `update` set to `True` and `False`, etc.

src/datachain/lib/dc/storage.py
amritghimire: This is literal copy paste of existing implementation for list. Any improvement for the logic can be a separate PR if needed IMO

ilongin (approved):

src/datachain/lib/dc/storage.py
```diff
@@ -79,40 +102,73 @@ def from_storage(
...
+    file_values = []
+
+    for single_uri in uris:
+        update_for_uri = update
+        if update:
+            # If update is True, we need to check if the URI has already been listed
+            # If it has, we don't need to update the listing
+            _, lst_uri, _ = parse_listing_uri(os.fsdecode(single_uri), client_config)
+            if lst_uri in listed_uris:
+                listed_uris.add(lst_uri)
```
shcheklein, line 121: would it be easier to just do `set(listed_uris)` at the end?

src/datachain/lib/dc/storage.py
```diff
@@ -79,40 +102,73 @@ def from_storage(
...
+    listed_uris = set()
+    file_values = []
+
+    for single_uri in uris:
+        update_for_uri = update
+        if update:
+            # If update is True, we need to check if the URI has already been listed
+            # If it has, we don't need to update the listing
+            _, lst_uri, _ = parse_listing_uri(os.fsdecode(single_uri), client_config)
+            if lst_uri in listed_uris:
```
shcheklein, line 120: should it be `not in` at the very least? 

do we need a test for this then?

src/datachain/lib/dc/storage.py
```diff
@@ -79,40 +102,73 @@ def from_storage(
...
+                listed_uris.add(lst_uri)
+                update_for_uri = False
+
+        list_ds_name, list_uri, list_path, list_ds_exists = get_listing(
```
shcheklein, line 124: should we be collecting these `list_ds_name` to deduplicate / instead of the `lst_uri` above?

src/datachain/lib/dc/storage.py
shcheklein: I think we need a test for deduplication logic ...

amritghimire: Added a test

amritghimire:
Merging this. Will work on followup PR for the optimization of setting update to false for following uris. It is currently not able to identify if a dataset will be created in current chain due to recent change in from_storage at https://github.com/iterative/datachain/pull/976/files#diff-eaca8e963dcff8dd88a958ca9d5bab9b4b153b1f76aba2b97c755fba2dde0a0c. 


Posting Comment from @ilongin  here for visibility:
> I don't think this can be checked here any more since listing is now lazy and happens only on applying steps. I know it was easier before when listing was happening on  .from_storage()  call itself and we could assure it was present, but that was wrong approach and we need to change our way of thinking about it from now on.

src/datachain/lib/dc/storage.py
```diff
@@ -23,7 +22,7 @@
 
 
 def from_storage(
-    uri: Union[str, os.PathLike[str], Iterable[str], Iterable[os.PathLike[str]]],
+    uri: Union[str, os.PathLike[str], list[str], list[os.PathLike[str]]],
```
skshetry, line 25: This should allow passing a mixed-type collections/iterable of `str` and `os.PathLike[str]` at the same time.

tests/func/test_datachain.py
```diff
@@ -369,6 +370,85 @@ def test_export_images_files(test_session, tmp_dir, tmp_path, use_cache):
...
+            "cat1",
+            "cat2",
+            "dog1",
+            "dog2",
+            "dog3",
+            "dog4",
+        }
+
+        # Verify from_records was called exactly twice
+        assert mock_get_listing.call_count == 4  # TODO FIX THIS
```
shcheklein, line 449: what is this TODO about?

PR #1001, opened 2025-03-28
Rename from_x to read_x for dc

### LINKED ISSUE #984: introduce a functional datachain API
### Description

Many popular dataframe libraries such as `pandas`, `pyarrow`, `polars`, and `daft` offer a functional API to create dataframes or tables. For instance, functions like [`read_parquet`](https://pandas.pydata.org/docs/reference/api/pandas.read_parquet.html#pandas.read_parquet) allow users to easily read data from Parquet files and create a dataframe.

However, in the case of `datachain`, users are required to import the `DataChain` class and use a classmethod to achieve similar functionality.

To improve the usability and accessibility of `datachain`, it would be beneficial to introduce a functional API. This would allow users to perform common tasks, such as reading data from various sources, in a more straightforward and familiar manner. For example, instead of:

```python

chain = DataChain.read_parquet('file.parquet')
```

Users could write:

```python
import datachain as dc

chain = dc.read_parquet('file.parquet')
```

This change would align `datachain` with the conventions of other dataframe libraries, making it easier for new users. Additionally, a functional API can lead to cleaner and more readable code.

Another thing to note, the _input_ functions are only going to be functional, the `write` function is an instance method and will remain so. For example, in our case, `from_storage` will be a functional API, but `to_storage` will remain an instance-method.

> [!NOTE]
> `pandas` does not seem very consistent. There are some methods like `DataFrame.from_dict()` and `DataFrame.from_records()`. `polars` does not provide such APIs.

### Version Info

```Text

```

amritghimire:
- **Rename from_storage to read_storage**
- **Rename from_csv to read_csv**
- **Rename from_hf to read_hf**
- **Rename from_dataset to read_dataset**
- **Rename from json to read json**
- **Rename from_pandas to read_pandas**
- **Rename from_parquet to read_parquet**
- **Rename from_records to read_records**
- **Rename from_values to read_values**

I decided not to update to_x to write_x because write_storage doesn't makes sense in our context. We are writing to storage but not writing storage itself and so on. 
It is also in line with other tools like  https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_parquet.html#pandas-dataframe-to-parquet

cc. @dmpetrov 

Closes #984

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -61,7 +61,7 @@ high confidence scores.
     import datachain as dc
 
     meta = dc.from_json("gs://datachain-demo/dogs-and-cats/*json", object_name="meta", anon=True)
-    images = dc.from_storage("gs://datachain-demo/dogs-and-cats/*jpg", anon=True)
+    images = dc.read_storage("gs://datachain-demo/dogs-and-cats/*jpg", anon=True)
 
     images_id = images.map(id=lambda file: file.path.split('.')[-2])
     annotated = images_id.merge(meta, on="id", right_on="meta.id")
@@ -102,7 +102,7 @@ Python code:
          return result.lower().startswith("success")
 
     chain = (
-       dc.from_storage("gs://datachain-demo/chatbot-KiT/", object_name="file", anon=True)
+       dc.read_storage("gs://datachain-demo/chatbot-KiT/", object_name="file", anon=True)
        .settings(parallel=4, cache=True)
        .map(is_success=eval_dialogue)
        .save("mistral_files")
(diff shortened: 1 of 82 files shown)


dmpetrov:
> I decided not to update to_x to write_x

looks good!

> * **Rename from_storage to read_storage**

I always feel that `storage` sounds a bit strange. Should we rename it to `read_files` instead?

dreadatour (approved):
Yay! Love it 😍

And thank you for fixing all the small code styles things along the way 🙏

PR #1012, opened 2025-03-31
Refactoring `dc.datasets()` arguments

ilongin:
Changes to `dc.datasets()` :

- [x] renamed `object_name` to `column`
- [x] removed `column` default value of `dataset` and set default to `None` which means no top level object will be created if not specified explicitly

### DIFF
--- a/src/datachain/lib/dc/datasets.py
+++ b/src/datachain/lib/dc/datasets.py
@@ -110,8 +110,8 @@ def datasets(
         session: Optional session instance. If not provided, uses default session.
         settings: Optional dictionary of settings to configure the chain.
         in_memory: If True, creates an in-memory session. Defaults to False.
-        column: Name of the output object in the chain. Defaults to None which
-            means no top level object will becreated.
+        column: Name of the output column in the chain. Defaults to None which
+            means no top level column will be created.
         include_listing: If True, includes listing datasets. Defaults to False.
         studio: If True, returns datasets from Studio only,
             otherwise returns all local datasets. Defaults to False.
--- a/src/datachain/lib/dc/datasets.py
+++ b/src/datachain/lib/dc/datasets.py
@@ -1,7 +1,4 @@
-from typing import (
-    TYPE_CHECKING,
-    Optional,
-)
+from typing import TYPE_CHECKING, Optional, get_type_hints
 
 from datachain.lib.dataset_info import DatasetInfo
 from datachain.lib.file import (
@@ -12,6 +9,7 @@ from datachain.lib.signal_schema import SignalSchema
 from datachain.query import Session
 from datachain.query.dataset import DatasetQuery
 
+from .records import read_records
 from .utils import Sys
 from .values import read_values
 
@@ -102,7 +100,7 @@ def datasets(
     session: Optional[Session] = None,
     settings: Optional[dict] = None,
     in_memory: bool = False,
-    object_name: str = "dataset",
+    column: Optional[str] = None,
     include_listing: bool = False,
     studio: bool = False,
 ) -> "DataChain":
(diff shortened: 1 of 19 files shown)


src/datachain/lib/dc/datasets.py
```diff
@@ -102,7 +100,7 @@ def datasets(
     session: Optional[Session] = None,
     settings: Optional[dict] = None,
     in_memory: bool = False,
-    object_name: str = "dataset",
```
shcheklein, line 105: do we use object_name in some other places? should we review and change that? (for consistency)

src/datachain/lib/dc/datasets.py
```diff
@@ -112,7 +110,8 @@ def datasets(
         session: Optional session instance. If not provided, uses default session.
         settings: Optional dictionary of settings to configure the chain.
         in_memory: If True, creates an in-memory session. Defaults to False.
-        object_name: Name of the output object in the chain. Defaults to "dataset".
+        column: Name of the output object in the chain. Defaults to None which
+            means no top level object will becreated.
```
shcheklein, line 114: typo here (please use some local linter?)

src/datachain/lib/dc/datasets.py
```diff
@@ -112,7 +110,8 @@ def datasets(
         session: Optional session instance. If not provided, uses default session.
         settings: Optional dictionary of settings to configure the chain.
         in_memory: If True, creates an in-memory session. Defaults to False.
-        object_name: Name of the output object in the chain. Defaults to "dataset".
+        column: Name of the output object in the chain. Defaults to None which
```
shcheklein, line 113: this description (column is an object) looks weird ...  if we are renaming this, we should also change the description I think

src/datachain/lib/dc/datasets.py
ilongin: Yes, we use it on multiple places. I will create separate issue for that to be clean with issues and related PRs

ilongin: https://github.com/iterative/datachain/issues/1016

dreadatour:
We have this discussion during planning today. We now have three different options:
- `object_name`
- `column`
- `signal`

We should choose one and stick to it. Me personally I prefer `signal` rather than `column`, since we may have complex nested signals, e.g. `file: File` _signal_ which have it own _signals_ inside, like `file.name`. In case of _columns_ it will be `file__name`, but we don't really want to expose these to users. This is what I am keen to, will be happy to hear your thoughts, just wanted to discuss this here.

shcheklein (approved):

ilongin:
> We have this discussion during planning today. We now have three different options:
> 
> * `object_name`
> * `column`
> * `signal`
> 
> We should choose one and stick to it. Me personally I prefer `signal` rather than `column`, since we may have complex nested signals, e.g. `file: File` _signal_ which have it own _signals_ inside, like `file.name`. In case of _columns_ it will be `file__name`, but we don't really want to expose these to users. This is what I am keen to, will be happy to hear your thoughts, just wanted to discuss this here.
> 
> **Update**: e.g. from the nearest [PR](https://github.com/iterative/datachain/pull/1020) we can see we have `SignalSchema`, `select_except_signals`, "cannot resolve signal name 'fr1.label'" error etc.

I agree this must be consistent all across the codebase. I've created related issue to do so in other parts after this PR is merged.
Regarding the exact name, I'm not 100% sure what's better. It makes sense what you said, but on the other hand `signal` is kind of a strange name to me .... not sure if its widely used in the industry for things like this? `column` is maybe more clear to the users, even though it's not mapped 1-1 to actual column in DB. 
I would like to hear other opinions as well @shcheklein @dmpetrov

dreadatour (approved):
❤️

dmpetrov:
> `signal` is kind of a strange name to me .... not sure if its widely used in the industry for things like this? `column` is maybe more clear to the users

Agree. We should use common term Column instead of Signal and other terms.

amritghimire (approved):
LGTM. Also. would be great if we could add a test where the column is None, i.e default arg.

amritghimire:
> > LGTM. Also. would be great if we could add a test where the column is None, i.e default arg.
> 
> @amritghimire the test is already there [#1012 (files)](https://github.com/iterative/datachain/pull/1012/files#diff-7fa3fd7accc76fc7bbc0c5e9b696be6e54cdcc5126b8c1f682b0f75cba1f95baR304-R308) Did you mean on something different or?

Yes, sorry did not notice that

PR #1017, opened 2025-04-01
Add callbacks for 'downloaded' and 'generated' rows while processing distributed UDF

dreadatour:
While processing distributed UDF we need not only `processed` rows callback, but also callbacks for `downloaded` and `generated` rows. In this PR we make it possible to use this callback in SaaS UDF distributor.

See related Studio PR: https://github.com/iterative/studio/pull/11467

### DIFF
--- a/src/datachain/lib/udf.py
+++ b/src/datachain/lib/udf.py
@@ -16,7 +16,6 @@ from datachain.lib.convert.flatten import flatten
 from datachain.lib.data_model import DataValue
 from datachain.lib.file import File
 from datachain.lib.utils import AbstractUDF, DataChainError, DataChainParamsError
-from datachain.progress import CombinedDownloadCallback
 from datachain.query.batch import (
     Batch,
     BatchingStrategy,
@@ -327,8 +326,9 @@ def _prefetch_inputs(
 
     if after_prefetch is None:
         after_prefetch = noop
-        if isinstance(download_cb, CombinedDownloadCallback):
-            after_prefetch = download_cb.increment_file_count
+        if download_cb and hasattr(download_cb, "increment_file_count"):
+            increment_file_count: Callable[[], None] = download_cb.increment_file_count
+            after_prefetch = increment_file_count
 
     f = partial(_prefetch_input, download_cb=download_cb, after_prefetch=after_prefetch)
     mapper = AsyncMapper(f, prepared_inputs, workers=prefetch)
--- a/src/datachain/query/dispatch.py
+++ b/src/datachain/query/dispatch.py
@@ -4,9 +4,8 @@ from itertools import chain
 from multiprocessing import cpu_count
 from sys import stdin
 from threading import Timer
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING, Literal, Optional
 
-import attrs
 import multiprocess
 from cloudpickle import load, loads
 from fsspec.callbacks import DEFAULT_CALLBACK, Callback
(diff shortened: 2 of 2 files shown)


src/datachain/lib/udf.py
```diff
@@ -327,8 +326,9 @@ def _prefetch_inputs(
 
     if after_prefetch is None:
         after_prefetch = noop
-        if isinstance(download_cb, CombinedDownloadCallback):
-            after_prefetch = download_cb.increment_file_count
+        if download_cb and hasattr(download_cb, "increment_file_count"):
```
dreadatour, line 329: Not really like to use `hasattr`, another option is to use common abstract class and to check for instance.

src/datachain/query/dispatch.py
```diff
@@ -152,16 +154,14 @@ def send_stop_signal_to_workers(task_queue, n_workers: Optional[int] = None):
         for _ in range(n_workers):
             put_into_queue(task_queue, STOP_SIGNAL)
 
-    def create_input_queue(self):
```
dreadatour, line 155: Leftover, not used anywhere.

src/datachain/query/dispatch.py
```diff
@@ -308,29 +325,22 @@ def run(self) -> None:
...
                     self.udf,
-                    cb=generated_cb,
+                    cb=self.generated_cb,
                 )
+        put_into_queue(self.done_queue, {"status": FINISHED_STATUS})
 
-        put_into_queue(
-            self.done_queue,
-            {"status": FINISHED_STATUS, "processed": processed_cb.processed_rows},
-        )
```
dreadatour, line 326: Already notified, no need to notify again.

src/datachain/query/dispatch.py
```diff
@@ -308,29 +325,22 @@ def run(self) -> None:
...
-
-    def notify_and_process(self, udf_results, processed_cb):
+    def notify_and_process(self, udf_results):
         for row in udf_results:
-            put_into_queue(
-                self.done_queue,
-                {"status": OK_STATUS, "processed": processed_cb.processed_rows},
```
dreadatour, line 332: No need to notify `processed_rows` here, already notified.

src/datachain/query/dispatch.py
```diff
@@ -260,46 +262,61 @@ def run_udf_parallel(  # noqa: C901, PLR0912
...
+        self.name = name
+        self.queue = queue
         super().__init__()
 
     def relative_update(self, inc: int = 1) -> None:
-        self.processed_rows = inc
+        put_into_queue(self.queue, {"status": NOTIFY_STATUS, self.name: inc})
 
 
-@attrs.define
```
dreadatour, line 281: Move callbacks init to `__init__` method, thus `attrs.define` not needed.

src/datachain/query/dispatch.py
```diff
@@ -260,46 +262,61 @@ def run_udf_parallel(  # noqa: C901, PLR0912
...
-    def __init__(self, queue: "multiprocess.Queue"):
+class DownloadCallback(Callback):
+    def __init__(self, queue: "multiprocess.Queue") -> None:
         self.queue = queue
         super().__init__()
 
     def relative_update(self, inc: int = 1) -> None:
+        # This callback is used to notify the size of the downloaded files
+        pass
+
```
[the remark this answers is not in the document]
dreadatour: Wrong 👀

shcheklein (approved):

PR #1018, opened 2025-04-01
Increase MAX_IMPORT_TIME_MS in tests up to 800ms

dreadatour:
This test fails too often :(

Latest example: https://github.com/iterative/datachain/actions/runs/14176251298/job/39711770663

Will be happy to have better option and also will be happy to discuss this.

### DIFF
--- a/tests/test_import_time.py
+++ b/tests/test_import_time.py
@@ -10,7 +10,7 @@ import datachain as dc
 logger = logging.getLogger(__name__)
 
 MAX_ATTEMPTS = 3
-MAX_IMPORT_TIME_MS = 700
+MAX_IMPORT_TIME_MS = 800
 
 lazy_modules = [
     "adlfs",


shcheklein:
@dreadatour why is it happening? is there a chance that we actually making it worse gradually and need to do a research again?

ilongin:
What is the max import time we are aiming for? ... Anyway, I would rather skip the test for now and create an issue to fix it, then increasing import time but I don't have strong opinion.
It makes sense that it's because of Github workers and running tests in parallel as you mentioned in related issue.

dreadatour:
> What is the max import time we are aiming for?

from @skshetry ([here](https://github.com/iterative/datachain/pull/974#issuecomment-2728232512)):

> It was taking ~450ms on ubuntu-latest-8-cores last I checked, so I added some slack and set it to 700ms. It should never reach that number, much less exceed it

skshetry (approved):

skshetry:
> What is the max import time we are aiming for? 

I'd like it to be <100ms, but we are pretty far away from that.

dreadatour:
> > What is the max import time we are aiming for?
> 
> I'd like it to be <100ms, but we are pretty far away from that.

I know 😥 We'll be there, one day 🤗

PR #1020, opened 2025-04-01
Better error hadling in `.select_except()`

ilongin:
Splitting error messages when signal is not found and when trying to remove nested signal which is not supported.
Also making error message little bit better.

### DIFF
--- a/src/datachain/lib/signal_schema.py
+++ b/src/datachain/lib/signal_schema.py
@@ -87,6 +87,12 @@ class SignalResolvingTypeError(SignalResolvingError):
         )
 
 
+class SignalRemoveError(SignalSchemaError):
+    def __init__(self, path: Optional[list[str]], msg: str):
+        name = " '" + ".".join(path) + "'" if path else ""
+        super().__init__(f"cannot remove signal name{name}: {msg}")
+
+
 class CustomType(BaseModel):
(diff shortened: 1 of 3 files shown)


src/datachain/lib/signal_schema.py
```diff
@@ -620,18 +620,28 @@ def _find_in_tree(self, path: list[str]) -> DataType:
         return curr_type
 
     def select_except_signals(self, *args: str) -> "SignalSchema":
-        schema = copy.deepcopy(self.values)
-        for field in args:
-            if not isinstance(field, str):
-                raise SignalResolvingTypeError("select_except()", field)
+        def has_signal(signal: str):
+            signal = signal.replace(".", DEFAULT_DELIMITER)
+            return any(signal == s for s in self.db_signals())
```
dreadatour, line 625: Just wonder, why not to move it on the parent class level 👀

dreadatour (approved):
❤️

dmpetrov (approved):
thank you!

src/datachain/lib/signal_schema.py
ilongin: I haven't seen any need as it's only used here and I wanted to keep public API as clean as possible (easier to maintain and understand)

PR #1029, opened 2025-04-07
Added `DataChain.persist()`

ilongin:
Added new method to create temp datasets `.persist()`. Old way to create temp datasets was to call `.save()` without name but now name is mandatory in `.save()`.

### DIFF
--- a/src/datachain/lib/dc/datachain.py
+++ b/src/datachain/lib/dc/datachain.py
@@ -444,8 +444,11 @@ class DataChain:
         return listings(*args, **kwargs)
 
     def persist(self) -> "Self":
-        """Save to temporary dataset that will be removed after the process ends.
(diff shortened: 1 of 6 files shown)


src/datachain/lib/dc/datachain.py
```diff
@@ -443,9 +443,17 @@ def listings(
         )
         return listings(*args, **kwargs)
 
+    def persist(self) -> "Self":
+        """Save to temporary dataset that will be removed after the process ends.
```
shcheklein, line 447: Can we improve the docs a bit?

- Don't mention dataset here (it's not important)
- Make sure that it's clear what is the use case here (what is the optimization here)

shcheklein (commented):
Please improve docs for this

PR #1034, opened 2025-04-09
dc: add read_database API

### LINKED ISSUE #1023: Add `read_database` method
We need a way to read data from a SQL database into DataChain.

It should be accepting a SQL, runs it on a selected connection (how do we specify the connection in local Python script (not Studio)?) and return a DataChain with data.

As an inspiration we can look into DVC `import db`.

skshetry:
Closes #1023.


tests/func/test_read_database.py
```diff
@@ -0,0 +1,172 @@
...
+    assert chain.schema == {"id": int, "value": Optional[str]}
+    assert sorted(chain.to_records(), key=lambda r: r["id"]) == [
+        {"id": i, "value": None if i < 50 else str(i)} for i in range(1, 1000)
+    ]
+
+
+# FIXME: `read_records`/`SignalSchema.db_signals` strips `Optional` types from the
+# SignalSchema, and creates the column as `T` instead.
+# This works on sqlite3, but fails on other databases.
+@skip_if_not_sqlite
```
skshetry, line 107: I tried to support nullable types in `read_database` but could not figure out a simpler way. (We use `ColumnClause` which does not provide a way to set `nullable`).

skshetry:
@shcheklein, I looked into possibility of determining types from a given query result.

According to PEP 249, the `cursor.description` attribute must include `type_code`. The `type_code` is intended to be compared against standard "Type Objects" (like `STRING`, `NUMBER`, `DATETIME`) defined within the PEP ([see Type Objects section](https://peps.python.org/pep-0249/#type-objects-and-constructors)).

However, in practice, it is inconsistent across different database drivers, and requires knowledge of specific drivers/databases and their internal types:

 1. `sqlite3`: Due to SQLite's dynamic typing system, the standard Python sqlite3 module does not set `type_code` value. https://github.com/python/cpython/issues/55900

 2. Snowflake (`snowflake-connector-python`): The official Snowflake connector returns  integer constants for `type_code` instead of the specified Type Objects.

 3. DuckDB (duckdb): The DuckDB Python driver appears to return string representations of database types (e.g., "VARCHAR", "INTEGER") in the `type_code`, rather than the required PEP 249 Type Objects.

4. Postgres `psycopg`: Similar to `snowflake`, postgres returns integer code for their internal types.

In any case, the `type_code` is not very reliable, and requires mapping to the internal types of the databases/drivers.


Polars does implement it, but I find it very complex for us to implement. https://github.com/pola-rs/polars/blob/79177fc14ee8cfb072e4975677b7c210236595a0/py-polars/polars/io/database/_inference.py#L201

`duckdb` returns stringified types, which polars parses [here](https://github.com/pola-rs/polars/blob/79177fc14ee8cfb072e4975677b7c210236595a0/py-polars/polars/io/database/_inference.py#L41), and some may return TypeObjects, which is handled [here](https://github.com/pola-rs/polars/blob/79177fc14ee8cfb072e4975677b7c210236595a0/py-polars/polars/datatypes/_parse.py#L72).

For pandas, there is an open issue: https://github.com/pandas-dev/pandas/issues/13049 for this.

src/datachain/query/schema.py
```diff
@@ -40,12 +40,15 @@ def __getattr__(cls, name: str):
...
-    def __init__(self, text, type_=None, is_literal=False, _selectable=None):
+    def __init__(
+        self, text, type_=None, is_literal=False, nullable=None, _selectable=None
+    ):
         """Dataset column."""
         self.name = ColumnMeta.to_db_name(text)
         super().__init__(
             self.name, type_=type_, is_literal=is_literal, _selectable=_selectable
         )
+        self.nullable = nullable
```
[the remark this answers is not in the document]
skshetry: Unfortunately, it does not work on Clickhouse. I was able to make `save()` work by wrapping `clickhouse_sqlalchemy.types.Nullable`, but haven't figured out how to make `read` work as well (null values get changes to default values).

src/datachain/lib/dc/records.py
```diff
@@ -83,8 +82,7 @@ def read_records(
...
     warehouse = catalog.warehouse
     dr = warehouse.dataset_rows(dsr)
-    db = warehouse.db
-    insert_q = dr.get_table().insert()
-    for record in to_insert:
-        db.execute(insert_q.values(**record))
+    table = dr.get_table()
+    warehouse.insert_rows(table, to_insert)
+    warehouse.insert_rows_done(table)
     return read_dataset(name=dsr.name, session=session, settings=settings)
```
skshetry, line 88: Migrated to use `insert_rows()` instead of saving each row individually. Improves performance by a lot.

tests/unit/lib/test_feature_utils.py
```diff
@@ -64,11 +64,6 @@ def test_not_array_value_error():
         dc.read_values(value=True)
 
 
-def test_empty_value_list_error():
-    with pytest.raises(ValuesToTupleError):
-        dc.read_values(value=[])
```
skshetry, line 69: Dropped the test. Now the column type will default to `Optional[str]` in case of empty values.

skshetry:
> @skshetry thanks for the research! are we ready to land this? will it work with RDS / PG btw?

I have tested it with postgres.

```python
import datachain as dc

dc.read_database(
   "SELECT * FROM users",
   "postgresql+psycopg://myuser:mypassword@localhost:5432/mydb",
)
```

This needs `psycopg` or `psycopg-binary` driver to be installed. (Every db requires a corresponding driver to be installed).

If RDS provides a db connection uri, then this should work without any issues. If the db uses more advanced authentication mechanisms (eg: `Oauth`/`kerberos`), then the user needs to instantiate a `sqlalchemy.Engine`/`sqlalchemy.Connection` and pass that to `read_database`.

Issue #1031, opened 2025-04-07
Error when calling `read_pandas`


luczeng:
### Description

Hello,

I get `AttributeError: 'tuple' object has no attribute 'lower'` with the following script:

```
df = dc.to_pandas()
new_dc = read_pandas(df)
```
A little surprising that I can't recreate the datachain from a dataframe that came from a datachain.

`df` is a basic datachain obtained like so: `dc = read_storage("image_folder", type="image", anon=True, update=True).filter(C("file.path").glob("*.png")).save('some_name')`

### Version Info

```Text
0.14.4
Python 3.13.2
```

amritghimire:
The issue seems to be when we have a datachain with multi level index. For the datachain read from the storage, we usually have indexes as MultiIndex unless flattened. 

Looking into the issue

luczeng:
Fyi, for now I am using this dirty fix. Does the job for now.

```
new_remote_location = "remoteB"

df = dc.to_pandas()

sources = df[("file", "source")]
paths = df[("file", "path")]
sizes = df[("file", "size")]
versions = df[("file", "version")]
etags = df[("file", "etag")]
is_latests = df[("file", "is_latest")]
# last_modifieds = df[("file", "last_modified")]  #not working?
locations = df[("file", "location")]

sources = [new_location for _ in range(len(sources))]

files = [
    File(
        path=path,
        size=size,
        source=source,
        version=version,
        etag=etag,
        is_latest=is_latest,
        location=location,
    )
    for path, size, source, version, etag, is_latest, location in zip(
        paths,
        sizes,
        sources,
        versions,
        etags,
        is_latests,
        locations,
    )
]

dc_from_remoteB = read_values(file=files)
dc_from_remoteB.to_storage("tmp")

```

PR #1035, opened 2025-04-09
Solves #1031

kshitiz305:
solves #1031 

```
AttributeError: 'tuple' object has no attribute 'lower' with the following script:

df = dc.to_pandas()
new_dc = read_pandas(df)
A little surprising that I can't recreate the datachain from a dataframe that came from a datachain.

df is a basic datachain obtained like so: dc = read_storage("image_folder", type="image", anon=True, update=True).filter(C("file.path").glob("*.png")).save('some_name')
```


dreadatour (approved):
Looks good to me 👍 Thank you for the improvement 🙏

luczeng:
btw does this allow the "identity" mapping? Ie `to_pandas` followed by `read_pandas`

kshitiz305:
Closes #1031

PR #1037, opened 2025-04-10
Added `DataChain.storage_switch()` method

ilongin:
Added `DataChain.storage_switch()` method to update `source` column with new storage value (e.g different s3 bucket).
Optionally `old_storage` argument can be passed so only those rows with `old_storage` values are updated which can be convenient when dataset has multiple sources and we only want to update one of those.

TODO:

- [ ] Reset `version`, `etag`, `last_modified` and `size`
- [ ] Create new method called `bucket_update()` to update `version`, `etag`, `last_modified`


shcheklein:
Closing for now ...

PR #1038, opened 2025-04-10
dc.from_records() doesn't work with dicts in Clickhouse

### LINKED ISSUE #1024: `dc.from_records()` doesn't work with dicts in Clickhouse
### Description

Clickhouse fails to insert dict as string correctly.
Example test that works for SQLite but fails for Clickhouse:

```python
def test_read_records_with_dict(test_session):
    schema = {"d": dict, "my_col": int}
    chain = dc.read_records(
        [{"d": {"a": 1}, "my_col": 1}], schema=schema, session=test_session
    )
```

### Version Info

```Text
0.6.10.dev278+g8805401.d20250403
Python 3.12.4
```

kshitiz305:
Closes #1024

### DIFF
--- a/src/datachain/sql/types.py
+++ b/src/datachain/sql/types.py
@@ -335,6 +335,14 @@ class Array(SQLType):
 class JSON(SQLType):
     impl = types.JSON
 
+    def process_bind_param(self, value: Any, dialect) -> Any:
+        """Convert Python dict to JSON string for storage."""
+        if value is not None:
+            # Use orjson for efficient serialization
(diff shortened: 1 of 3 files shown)


ilongin:
@kshitiz305 Thanks for your effort but this PR doesn't fix the real issue (test in original issue still fails). Note that to fix this you probably need access to some proprietary codebase (Clickhouse adapter) which is not open sourced so I suggest working on some other issues if you are interested. I'm closing the PR and removing the issue from this repo to avoid further confusion.

PR #1043, opened 2025-04-15
Update the callback after the row has been processed by the UDF, not before

dreadatour:
Another tiny update useful for distributed processing.

We use the "Downloaded", "Processed", and "Generated" callbacks to track rows processed by UDFs. In this PR, I’m updating these counters **after** the row(s) have been processed — not before. This might seem like a minor change, but it makes more sense when generators are involved. It’s odd to see the "Processed" count go up when nothing has happened yet.


shcheklein (approved):

src/datachain/lib/udf.py
```diff
@@ -387,8 +387,8 @@ def _prepare_rows(udf_inputs) -> "abc.Generator[Sequence[Any], None, None]":
                 result_objs = self.process_safe(udf_args)
                 udf_output = self._flatten_row(result_objs)
                 output = [{"sys__id": id_} | dict(zip(self.signal_names, udf_output))]
-                processed_cb.relative_update(1)
                 yield output
+                processed_cb.relative_update(1)
```
skshetry, line 391: We process before incrementing the callback, no? `process_safe` is what processes the udf.

dreadatour: You're right, it is only related to generator. Updated the PR.

skshetry (approved):

PR #1044, opened 2025-04-17
Renamed dataset `labels` to `attrs` and added attributes filters in `dc.datasets()`

ilongin:
Renaming `labels` to `attrs` in dataset model and related methods / functions.

Also added ability to filter datasets in `dc.datasets()` by attributes.
Attributes can be just normal name e.g `NLP` or combination of name and value e.g `location=US`. Examples of attribute filters are:
* `NLP` -> search all datasets that have `NLP` attribute value
* `location=US` -> search all datasets that have `location=US` attribute (combination of name and value)
* `location=*` -> search all datasets that have `location` name and any value .. e.g `location=US`, `location=UK` etc.

Note that for now filtering logic is implemented in application level. Ideal solution would be to implement it in DB level and for the users to use `.filter()` methods with functions, but that will be done as follow-up later on.

### DIFF
--- a/src/datachain/catalog/catalog.py
+++ b/src/datachain/catalog/catalog.py
@@ -776,7 +776,7 @@ class Catalog:
         listing: Optional[bool] = False,
         uuid: Optional[str] = None,
         description: Optional[str] = None,
-        labels: Optional[list[str]] = None,
+        attrs: Optional[list[str]] = None,
     ) -> "DatasetRecord":
         """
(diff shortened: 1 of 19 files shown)


amritghimire (approved):
Looks good to me. But I think having a option to filter the datasets by attrs in cli could also be helpful.

shcheklein (approved):

shcheklein:
@ilongin let's ship it? :)

PR #1046, opened 2025-04-18
fix(docs): file name to file path

shcheklein:
Fixes docs a bit to reflect the current naming of the file object attributes.


0x2b3bfa0 (approved):
🤖 📖

PR #1054, opened 2025-04-22
read_records: insert records in batches

skshetry:
insert_rows does not do chunking, so we go OOM with very large no. of records.

### DIFF
--- a/src/datachain/lib/dc/records.py
+++ b/src/datachain/lib/dc/records.py
@@ -38,8 +38,9 @@ def read_records(
         single_record = dc.read_records(dc.DEFAULT_FILE_RECORD)
         ```
     """
-    from datachain.query.dataset import adjust_outputs, get_col_types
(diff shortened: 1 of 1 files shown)


src/datachain/lib/dc/records.py
```diff
@@ -38,8 +38,9 @@ def read_records(
         single_record = dc.read_records(dc.DEFAULT_FILE_RECORD)
         ```
     """
-    from datachain.query.dataset import adjust_outputs, get_col_types
+    from datachain.query.dataset import INSERT_BATCH_SIZE, adjust_outputs, get_col_types
```
skshetry, line 41: https://github.com/iterative/datachain/blob/45fa2f7a6e4e675d50d3165ce981b0e060e4aa17/src/datachain/query/dataset.py#L86

^ this is the default.

shcheklein (approved):

PR #1068, opened 2025-04-29
Moving `attrs` from dataset to dataset version

ilongin:
Moving `attrs` from dataaset to dataset version to be able to have different attributes for specific version.


ilongin:
Closing as related issue is being closed for now

PR #1070, opened 2025-04-30
fix(setup): allow non callable values

### LINKED ISSUE #1069: `setup()` support any value, not only callable / function
Now we are getting an error that looks like:

`datachain.lib.signal_schema.SetupError: cannot setup value 'client': value must be function or callable class`

This is confusing and requires a non obvious workaround, e.g.:

`client=lambda: anthropic.Anthropic(api_key=API_KEY)`

(Note: the `lambda` above)

We should be able to just do:

`client=anthropic.Anthropic(api_key=API_KEY)` as well.

- [ ] Don't forget to update docs

shcheklein:
Fixes #1069

Allow regular values in `setup()`:

`client=anthropic.Anthropic(api_key=API_KEY)` instead of using `lambda`: `client=lambda: anthropic.Anthropic(api_key=API_KEY)`

Callable is still supported and cached.

TODO:

- [ ] Needs some testing with distributed mode (can we serialize non callable properly)
- [x] Update docs

### DIFF
--- a/src/datachain/lib/dc/datachain.py
+++ b/src/datachain/lib/dc/datachain.py
@@ -1993,7 +1993,8 @@ class DataChain:
         """Setup variables to pass to UDF functions.
 
         Use before running map/gen/agg/batch_map to save an object and pass it as an
-        argument to the UDF.
+        argument to the UDF. If callable / function value is passed it will be called
+        and the result value is cached and used.
 
         Example:
             ```py
@@ -2004,7 +2005,7 @@ class DataChain:
             (
                 dc.read_storage(DATA, type="text")
                 .settings(parallel=4, cache=True)
-                .setup(client=lambda: anthropic.Anthropic(api_key=API_KEY))
+                .setup(client=anthropic.Anthropic(api_key=API_KEY))
                 .map(
                     claude=lambda client, file: client.messages.create(
                         model=MODEL,
--- a/src/datachain/lib/signal_schema.py
+++ b/src/datachain/lib/signal_schema.py
@@ -144,31 +144,28 @@ def create_feature_model(
 class SignalSchema:
     values: dict[str, DataType]
     tree: dict[str, Any]
-    setup_func: dict[str, Callable]
+    setup_args: dict[str, Any]
     setup_values: Optional[dict[str, Any]]
 
     def __init__(
         self,
         values: dict[str, DataType],
-        setup: Optional[dict[str, Callable]] = None,
+        setup: Optional[dict[str, Any]] = None,
     ):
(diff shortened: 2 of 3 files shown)


shcheklein:
@dreadatour do you foresee any potential issues in the distributed mode (where we serialize values). As far as I understand now we can serialize callable (do we do this?), while after this change it will be an object. Not sure if there are any issues with this.

shcheklein:
@dreadatour when you have a minute, could please take a look :) should be a small PR ... my only question is about serialize / deserialize logic - if it is impacted or not

amritghimire (approved):

shcheklein:
@dreadatour friendly ping, can it cause any issues with distributed?

shcheklein:
@dreadatour PTAL

dreadatour:
Sorry 😢🙏 Will handle this today.

dreadatour:
Sorry again for the delay 🙏

> @dreadatour do you foresee any potential issues in the distributed mode (where we serialize values). As far as I understand now we can serialize callable (do we do this?), while after this change it will be an object. Not sure if there are any issues with this.

> @dreadatour when you have a minute, could please take a look :) should be a small PR ... my only question is about serialize / deserialize logic - if it is impacted or not

Yes, UDF serialization logic is impacted.

For example for YOLO model (`YOLO("yolo11n.pt")`):
- without lambda (`.setup(yolo=YOLO("yolo11n.pt"))`) serialized UDF size is ~11Mb, because it put whole `yolo11n.pt` model (5.35Mb) into the serialized object.
- with lambda (`.setup(yolo=lambda: YOLO("yolo11n.pt"))`) serialized UDF size is only 18kb.

We also have limit for serialized UDF size (10Mb for now), so in case of YOLO without lambda UDF will fail:

```
UDF 413fa700-3e72-4aa5-8468-eda15a1ff766 failed: UDF function/class data is too large!
```

shcheklein:
Thanks @dreadatour , it makes sense.

Tbh it seems critical enough to probably force lambda  (just improve docs and error handling?) :(

Otherwise I have no idea how to handle this so it is also not confusing ...

dreadatour:
> Tbh it seems critical enough to probably force lambda (just improve docs and error handling?) :(

We can allow lambdas and basic types (e.g. stings, numbers, bool values, etc) with docs and error handling improved.

> Otherwise I have no idea how to handle this so it is also not confusing ...

I agree, this can cause more issues than it solves.

shcheklein:
Let's close this to not confuse users. I'll update the docs for now. Thanks @dreadatour for the review.

PR #1071, opened 2025-04-30
Fix for reading exported parquet

ilongin:
Fixing the issue when we read parquet file that is being created with datachain itself and already has source fields inside it. What was happening is that we were adding duplicated source fields on read and this PR avoids that even if `source=True` flag is set.


shcheklein:
@ilongin tests are broken?

@ilongin should we actually rewrite source if source is True? 🤔 since the existing source is kinda wrong by now ...

ilongin:
> @ilongin should we actually rewrite source if source is True? 🤔 since the existing source is kinda wrong by now ...

Make sense, I've overwritten it now with new source.

src/datachain/lib/arrow.py
```diff
@@ -126,7 +126,16 @@ def _process_record(
             if isinstance(kwargs.get("format"), CsvFileFormat):
                 kwargs["format"] = "csv"
             arrow_file = ArrowRow(file=file, index=index, kwargs=kwargs)
+
+            if self.output_schema and hasattr(vals[0], "source"):
+                # if we are reading parquet file written by datachain it might have
+                # source inside of it already, so we should not duplicate it, instead
+                # we are re-creating it
```
shcheklein, line 133: cleanup the comment a bit ... (no extra line is needed)

shcheklein: let's also add to docs (source=True) - that if it enabled and the file already has it - it will be rewritten (e.g. when file was generated by datachain before)

src/datachain/lib/arrow.py
```diff
@@ -126,7 +126,16 @@ def _process_record(
...
+                # of the self.source flag
+                vals[0].source = arrow_file  # type: ignore[attr-defined]
```
dreadatour, line 135:
```suggestion
                vals[0].source = arrow_file  # type: ignore[attr-defined]
                
```

dreadatour (approved):
Looks good to me 👍

shcheklein:
@ilongin what is the status here?

PR #1082, opened 2025-05-02
Changing dataset versions from int to semver

ilongin:
TODO

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -167,7 +167,7 @@ dialog-rating@v2
 By default, when a saved dataset is loaded, the latest version is fetched but another version can be requested:
 
 ```python
-ds = dc.read_dataset("dialog-rating", version=1)
+ds = dc.read_dataset("dialog-rating", version="1.0.0")
 ```
 
 ### Chain execution, optimization and parallelism
--- a/src/datachain/lib/dc/datachain.py
+++ b/src/datachain/lib/dc/datachain.py
@@ -214,7 +214,7 @@ class DataChain:
         return self._query.name
 
     @property
-    def version(self) -> Optional[int]:
+    def version(self) -> Optional[str]:
         """Version of the underlying dataset, if there is one."""
         return self._query.version
 
@@ -457,7 +457,7 @@ class DataChain:
     def save(  # type: ignore[override]
         self,
         name: str,
-        version: Optional[int] = None,
+        version: Optional[str] = None,
         description: Optional[str] = None,
         attrs: Optional[list[str]] = None,
         **kwargs,
--- a/src/datachain/catalog/catalog.py
+++ b/src/datachain/catalog/catalog.py
@@ -33,6 +33,7 @@ from datachain.cache import Cache
 from datachain.client import Client
 from datachain.dataset import (
     DATASET_PREFIX,
+    DEFAULT_DATASET_VERSION,
     QUERY_DATASET_PREFIX,
     DatasetDependency,
     DatasetListRecord,
@@ -153,9 +154,9 @@ class DatasetRowsFetcher(NodesThreadPool):
         metastore: "AbstractMetastore",
         warehouse: "AbstractWarehouse",
         remote_ds_name: str,
-        remote_ds_version: int,
+        remote_ds_version: str,
         local_ds_name: str,
-        local_ds_version: int,
+        local_ds_version: str,
         schema: dict[str, Union[SQLType, type[SQLType]]],
         max_threads: int = PULL_DATASET_MAX_THREADS,
         progress_bar=None,
@@ -285,7 +286,7 @@ class NodeGroup:
     # (not including the bucket name or s3:// prefix)
     source_path: str = ""
     dataset_name: Optional[str] = None
-    dataset_version: Optional[int] = None
+    dataset_version: Optional[str] = None
     instantiated_nodes: Optional[list[NodeWithPath]] = None
 
     @property
@@ -606,7 +607,7 @@ class Catalog:
         return lst, client, list_path
 
     def _remove_dataset_rows_and_warehouse_info(
-        self, dataset: DatasetRecord, version: int, **kwargs
+        self, dataset: DatasetRecord, version: str, **kwargs
     ):
         self.warehouse.drop_dataset_rows_table(dataset, version)
         self.update_dataset_version_with_warehouse_info(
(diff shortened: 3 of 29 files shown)


docs/examples.md
```diff
@@ -167,7 +167,7 @@ dialog-rating@v2
 By default, when a saved dataset is loaded, the latest version is fetched but another version can be requested:
 
 ```python
-ds = dc.read_dataset("dialog-rating", version=1)
+ds = dc.read_dataset("dialog-rating", version="1.0.0")
```
shcheklein, line 170: can be a followup - can we still accept int in these methods + also accept 1 (meaning - give me the latest 1.0.* ?

we'll need a way to specify wildcards I think ...

shcheklein (approved):

amritghimire (approved):
Looks good to me. One question, how  does it work with existing versions in use with int ?

ilongin:
> Looks good to me. One question, how does it work with existing versions in use with int ?

In Studio I've added DB migration to migrate everything to semver (major versions), example: `1` -> `1.0.0`, `2` -> `2.0.0` etc. 
This way we won't work wiht int anymore

shcheklein:
> This way we won't work wiht int anymore

I think we can / and probably should allow int on the API level to keep it backward compatible, and also just for simplicity for users.

shcheklein:
Btw, just to double check - do we validate the input string? that it is an actual semver ...

docs/examples.md
shcheklein: let's do wildcard filtering / syntax as a followup? we can discuss it there ... wdyt?

src/datachain/lib/dc/datachain.py
```diff
@@ -457,9 +457,10 @@ def persist(self) -> "Self":
     def save(  # type: ignore[override]
         self,
         name: str,
-        version: Optional[int] = None,
+        version: Optional[str] = None,
         description: Optional[str] = None,
         attrs: Optional[list[str]] = None,
+        update_version: Optional[str] = "patch",
```
dmpetrov, line 463: please get rid of this. it should be as simple as possible for users.

dmpetrov (approved):
It looks good! 

I'm approving to not to block you, but please don't overcomplicate this with different updating strategies - it needs to be removed.

src/datachain/lib/dc/datachain.py
shcheklein: how would user decide which version to update, though? (or am I missing an alternative here? 🤔 )

(I agree that we should keep it simple, but I though we at least need a way to say which version to update, no?)

shcheklein: > We just need to choose the defaults:

I'm fine with this.

> Let's not implement anything beyond this until we have enough user requests and stories about this.

but we need at least _some_ way to update those versions? Otherwise it will be updating only patch (or only major). Or am I still missing something? 🤔

dmpetrov: User updates versions manually, right? or I'm missing something? 🙂 

User `ds.save("mycats")`, then decides to increase the version in another run `ds.save("mycats", "2.0.0")`, then it runs 10 auto-updates up to `2.0.10`. then user modifies code and run it one more time when it increments minor to `2.1.0`, later - another 20 auto-updates up to `2.1.20`. 

Is that how you see it too?

shcheklein: I see. It breaks (makes it non-trivial) any automation (around major / minor updates). But we can do a followup, I'm fine with that.

dmpetrov: could you please elaborate this?

shcheklein: I mean, I imagine cases where users won't be ready to do manual code change every time they run a job with an update. E.g. something triggered by CD / tag, or job runs on a schedule and updates a major version, etc. I hope it is more or less clear what I mean :)

dmpetrov: Why does this need a major version bump?

shcheklein: It's an outside decision (e.g. a tag on CI that means a new version of code and triggers a full pipeline run).

dmpetrov: How can tags be used without explicitly passing them in code like `ds.save("mycats", os.getenv("CI_CATS_VER"))`?

shcheklein: I mean Git tags. And we don't have to read its value (which is also possible though, eg using env var as you mentioned). It is enough that it triggers a job that does a major bump automatically.

dmpetrov: yeah, but that seems a pretty regular workflow with packaging systems and it's all done with explicit versioning - no need in auto increments

shcheklein: I've seen (even within our org) different examples, we have / had workflows that bump versions automatically. Again, I don't think it's a blocker / critical - I'm totally fine to keep it simpler for now.

docs/examples.md
ilongin: Sure, I will create a followup issue after this PR is merged!

ilongin:
> Btw, just to double check - do we validate the input string? that it is an actual semver ...

There was a validation and valid error message, but I still improved it now a little bit to be more explicit i.e. to have explicit `validate(...)` method in `semver.py` and use it explicitly in `DataChain.save()`.

src/datachain/lib/dc/datachain.py
ilongin: Ok, I will then implement just default patch increments in `.save()` and that's it. For everything else user will need to explicitly specify version. Listing datasets will be increasing automatically major version though, but this is all internal thing and not relevant to the user.

ilongin: It's done (removing `update_version` param)

ilongin:
> > This way we won't work wiht int anymore
> 
> I think we can / and probably should allow int on the API level to keep it backward compatible, and also just for simplicity for users.

Added backward compatibility to accept `int` in `dc.read_dataset()`. It tries to find latest version of the major value of that input, e.g if `dc.read_dataset("my-ds", version=3)` it will try to find latest `3.*` version, e.g something like `3.4.3`

PR #1088, opened 2025-05-07
Add cli command to get the list of the jobs

amritghimire:
This provides a way to get the list of jobs from the studio.

Currently, I selected the following field as table headers. We can modify it as we want.
image

QQ:Should we show certain number of jobs here?


dreadatour (approved):
Looks good to me 👍

shcheklein (approved):
We need to update docs / also add some tests. Thanks for taking care of this @amritghimire .

PR #1093, opened 2025-05-11
Use underscores for semantic versions in table names

0x2b3bfa0:
Continuation of https://github.com/iterative/datachain/pull/1082 & https://github.com/iterative/studio/pull/11552.

### DIFF
--- a/src/datachain/data_storage/warehouse.py
+++ b/src/datachain/data_storage/warehouse.py
@@ -253,12 +253,16 @@ class AbstractWarehouse(ABC, Serializable):
         name = parsed.path if parsed.scheme == "file" else parsed.netloc
         return parsed.scheme, name
 
-    def dataset_table_name(self, dataset_name: str, version: str) -> str:
+    def dataset_table_name(self, dataset_name: str, version: Union[str, int]) -> str:
         prefix = self.DATASET_TABLE_PREFIX
         if Client.is_data_source_uri(dataset_name):
             # for datasets that are created for bucket listing we use different prefix
             prefix = self.DATASET_SOURCE_TABLE_PREFIX
-        return f"{prefix}{dataset_name}_{version.replace('.', '_')}"
+        if isinstance(version, int):
+            normalized_version = f"{version}_0_0"
+        else:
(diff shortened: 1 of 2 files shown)


src/datachain/data_storage/warehouse.py
```diff
@@ -258,7 +258,7 @@ def dataset_table_name(self, dataset_name: str, version: str) -> str:
         if Client.is_data_source_uri(dataset_name):
             # for datasets that are created for bucket listing we use different prefix
             prefix = self.DATASET_SOURCE_TABLE_PREFIX
-        return f"{prefix}{dataset_name}_{version}"
+        return f"{prefix}{dataset_name}_{version.replace(".", "_")}"
```
0x2b3bfa0: I wonder if we should be performing a heavier normalization as well for `dataset_name` — like `normalize_col_names` from https://github.com/iterative/datachain/pull/542

shcheklein (approved):

src/datachain/data_storage/warehouse.py
dmpetrov: Yes, we need stricter rules for the names. Eventually, names should be exposable in urls. So, lowercase only, no spaces and special characters, limit to length (100?), etc

tests/func/test_warehouse.py
```diff
@@ -1,6 +1,6 @@
 def test_dataset_stats_no_table(cloud_test_catalog, dogs_dataset):
     catalog = cloud_test_catalog.catalog
-    catalog.warehouse.drop_dataset_rows_table(dogs_dataset, 1)
-    num_objects, size = catalog.warehouse.dataset_stats(dogs_dataset, 1)
+    catalog.warehouse.drop_dataset_rows_table(dogs_dataset, version="1.0.0")
+    num_objects, size = catalog.warehouse.dataset_stats(dogs_dataset, version="1.0.0")
```
0x2b3bfa0, line 4: Leftover `int` versions from https://github.com/iterative/datachain/pull/1082; using keyword arguments so they appear the next time someone has to search/replace _en masse_.

PR #1101, opened 2025-05-16
add incremental processing example

shcheklein:
Adds an example of delta (incremental) processing.

## TODO

- [x] fix tests / make it run as a test

### DIFF
--- a/examples/incremental_processing/delta.py
+++ b/examples/incremental_processing/delta.py
@@ -34,13 +34,12 @@ def process_files_with_delta():
     This demonstrates incremental processing - only new files are processed.
     """
     chain = (
-        dc
-            .read_storage("test/", update=True, delta=True, delta_on="file.path")
-            .filter(C("file.path").glob("*.txt"))
-            .map(file_number=extract_file_number)
-            .map(content=lambda file: file.read_text())
-            .map(processed_at=lambda: time.strftime("%Y-%m-%d %H:%M:%S"))
-            .save(name="test_files")
+        dc.read_storage("test/", update=True, delta=True, delta_on="file.path")
+        .filter(C("file.path").glob("*.txt"))
+        .map(file_number=extract_file_number)
+        .map(content=lambda file: file.read_text())
+        .map(processed_at=lambda: time.strftime("%Y-%m-%d %H:%M:%S"))
+        .save(name="test_files")
     )
 
     # Show information about the dataset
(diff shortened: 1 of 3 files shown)


examples/incremental_processing/delta.py
```diff
@@ -0,0 +1,64 @@
...
+    return -1
+
+
+def process_files_with_delta():
+    """
+    Process files in the test directory using DataChain with delta mode.
+    This demonstrates incremental processing - only new files are processed.
+    """
+    chain = (
+        dc.read_storage("test/", update=True, delta=True, delta_on="file.path")
```
ilongin, line 37: Note that without explicit `delta_compare` it will look for all fields in schema except `file.path` (since it's in `delta_on` already) to say if file is changed or not. This means that two rows need to be identical (all fields the same) in order to not count them as "modified / changed". If it count's them as changed there is no performance gain in delta update. You usually want to set `delta_compare=["file.version", "file.etag"]. 

There is the case though with non-versioned sources where `file.version` and `file.etag` are randomly set every time on re-index which causes the same thing to happen regardless as it will catch everything as modified. In this cases, and in every other case where user doesn't even want or can track changed rows, workaround is to put `delta_update` to be the same as `delta_on` but we need a better way.
Options are:
* make default `delta_compare=None` to disable tracking changed rows instead to look into all fields. If we go with this path then `DataChain.compare()` and `DataChain.diff()` needs to be changed as well to be consistent.
* Add additional flag for this, e.g `delta_ignore_changed`.

I'm leaning more on first option, although then user then needs to explicitly set all fields in some cases (we loose "shortcut" of default being all fields). I don't have strong opinion though.

ilongin (changes requested):

examples/incremental_processing/delta.py
shcheklein: thanks, I think we are fine in this particular example (?)

> There is the case though with non-versioned sources where file.version and file.etag are randomly set every time on re-index

where did you experience this?

shcheklein: @ilongin please let me know ^^

ilongin: I thought I saw it in one of our `gs` buckets but now I checked and it seems like it's ok. `version` and `etag` should be set to empty string if they don't exist.

Regarding your example, yea you don't need to put anything as non of the columns will be changed since you only append. If you would re-create files every time when calling `generate_next_file` then it would be a problem as for local files `etag` we put `mtime` which would mean that delta would find all files being modified.

PR #1107, opened 2025-05-19
Add 'slice' and 'join' array functions

dreadatour:
See also related SaaS PR.

### DIFF
--- a/src/datachain/func/array.py
+++ b/src/datachain/func/array.py
@@ -219,7 +219,8 @@ def slice(
             try:
                 return list[element_type(el[0])]
             except IndexError:
-                return str
+                # if the array is empty, return list[str] as default type
+                return list[str]
         return type(el)
 
     def type_from_args(arr, *_):
@@ -227,7 +228,8 @@ def slice(
             try:
                 return list[element_type(arr[0])]
             except IndexError:
-                return str  # if the array is empty, return str as default type
+                # if the array is empty, return list[str] as default type
+                return list[str]
         return None
 
     if isinstance(arr, (str, Func)):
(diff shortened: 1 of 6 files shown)


src/datachain/func/array.py
```diff
@@ -178,6 +178,124 @@ def inner(arg):
...
+
+    Returns:
+        Func: A Func object that represents the slice function. Result of the
+            function will be a slice of the array starting from the offset
+            and with the given length.
+
+    Example:
+        ```py
+        dc.mutate(
+            contains1=func.array.slice("signal.values", 3),
```
shcheklein, line 206: does it accept `C` syntax (just to double check)

shcheklein (approved):

src/datachain/func/array.py
```diff
@@ -178,6 +178,124 @@ def inner(arg):
...
+    Returns:
+        Func: A Func object that represents the join function. Result of the
+            function will be a string that is the concatenation of the elements
+            of the array, separated by the given separator.
+
+    Example:
+        ```py
+        dc.mutate(
+            contains1=func.array.join("signal.values", ":"),
+            contains2=func.array.join([1, 2, 3, 4, 5], "/"),
```
shcheklein, line 274: do we cast elements to str? will it always work fine?

amritghimire (approved):

src/datachain/func/array.py
dreadatour: Fixed the docstrings. Only strings are allowed to be joined.

src/datachain/func/array.py
dreadatour: Unfortunately no, no `C` syntax allowed in any `dc.func` functions as it can be quite complicated. We can try to add this later if needed.

PR #1115, opened 2025-05-21
Dataset namespaces

ilongin:
Adding dataset namespaces and projects.
Namespace can have multiple projects and each project can have multiple datasets.
Default namespace `local` and project `local` is automatically created.

Adding methods to create, get and list project and namespaces.
Adding `Project` as optional arg alongside name for all methods that are working with specific dataset. If it's not defined, default namespace and project is used.

Each dataset now has fully qualified name with schema `<namespace_name>.<project_name>.<dataset_name>`, e.g `dev.my-project.cats`. There cannot be multiple datasets with same fully qualified name.

- [x] Adding `Namespace` entity and function to create / get
- [x] Adding `Project` entity and function to create / get
- [ ] Adding delete project and namespace
- [x] Connecting dataset models to projects and namespaces and using them on creation


src/datachain/data_storage/metastore.py
```diff
@@ -322,11 +414,62 @@ def close(self) -> None:
     def cleanup_tables(self, temp_table_names: list[str]) -> None:
         """Cleanup temp tables."""
 
+    @classmethod
+    def _namespaces_columns(cls) -> list["SchemaItem"]:
+        """Namespace table columns."""
+        return [
+            Column("id", Integer, primary_key=True),
+            Column("uuid", Text, nullable=False, default=uuid4()),
```
dreadatour, line 439: Why do we need both `id` and `uuid`?

src/datachain/data_storage/metastore.py
```diff
@@ -322,11 +414,62 @@ def close(self) -> None:
...
+            for c in self._namespaces_columns()
+            if c.name  # type: ignore [attr-defined]
+        ]
+
+    @classmethod
+    def _projects_columns(cls) -> list["SchemaItem"]:
+        """Project table columns."""
+        return [
+            Column("id", Integer, primary_key=True),
+            Column("uuid", Text, nullable=False, default=uuid4()),
```
dreadatour, line 458: Same, do we need both `id` and `uuid`? 🤔

src/datachain/data_storage/metastore.py
```diff
@@ -322,11 +414,62 @@ def close(self) -> None:
     def cleanup_tables(self, temp_table_names: list[str]) -> None:
         """Cleanup temp tables."""
 
+    @classmethod
+    def _namespaces_columns(cls) -> list["SchemaItem"]:
+        """Namespace table columns."""
+        return [
+            Column("id", Integer, primary_key=True),
+            Column("uuid", Text, nullable=False, default=uuid4()),
+            Column("name", Text, nullable=False),
```
dreadatour, line 440: Just wonder if we need uniqueness on `name` here?

src/datachain/data_storage/metastore.py
```diff
@@ -510,13 +691,141 @@ def _datasets_dependencies_update(self) -> "Update":
...
+        query = self._namespaces_insert().values(
+            name=name,
+            uuid=uuid or str(uuid4()),
+            created_at=datetime.now(timezone.utc),
+            description=description,
+        )
+        if ignore_if_exists and hasattr(query, "on_conflict_do_nothing"):
+            # SQLite and PostgreSQL both support 'on_conflict_do_nothing',
+            # but generic SQL does not
+            query = query.on_conflict_do_nothing(index_elements=["name"])
```
dreadatour, line 732: 👍

src/datachain/data_storage/metastore.py
```diff
@@ -510,13 +691,141 @@ def _datasets_dependencies_update(self) -> "Update":
...
+        self.db.execute(query)
+
+        return self.get_namespace(name)
+
+    def get_namespace(self, name: str, conn=None) -> Namespace:
+        """Gets a single namespace by name"""
+        n = self._namespaces
+        if not self.db.has_table(self._namespaces.name):
+            raise TableMissingError
```
dreadatour: Not sure if this is a good idea to always check if table is exists on every `get` call 🤔
Can we at least not do it in SaaS?

Or may be there is an exception if case table is not exists and we can catch it?

Also why we don't make this check on `create_namespace`? 🤔

src/datachain/data_storage/metastore.py
```diff
@@ -510,13 +691,141 @@ def _datasets_dependencies_update(self) -> "Update":
...
+        rows = list(self.db.execute(query, conn=conn))
+        if not rows:
+            raise NamespaceNotFoundError(f"Namespace {name} not found.")
+        return self.namespace_class.parse(*rows[0])
+
+    def list_namespaces(self, conn=None) -> list[Namespace]:
+        """Gets a list of all namespaces"""
+        n = self._namespaces
+        if not self.db.has_table(self._namespaces.name):
+            raise TableMissingError
```
dreadatour: Same here and below, do we need to check if table exists every time?

src/datachain/data_storage/metastore.py
```diff
@@ -510,13 +691,141 @@ def _datasets_dependencies_update(self) -> "Update":
...
+        return [self.namespace_class.parse(*r) for r in rows]
+
+    #
+    # Projects
+    #
+
+    def create_project(
+        self,
+        name: str,
+        namespace: Namespace,
```
dreadatour: Why do we need whole `Namespace` object here as input param?

I can see we are using `namespace.name` below, but in terms of API this looks a bit inconsistent IMO. `create_project` requires whole namespace object, while `get_project` requires namespace name only.

Also passing whole objects as input params in metastore methods will cause additional complexity all other the code (both datachain and Studio), since we don't always have objects in the callers.

src/datachain/data_storage/metastore.py
```diff
@@ -510,13 +691,141 @@ def _datasets_dependencies_update(self) -> "Update":
...
+        self.db.execute(query)
+
+        return self.get_project(name, namespace.name)
+
+    def get_project(self, name: str, namespace_name: str, conn=None) -> Project:
+        """Gets a single project inside some namespace by name"""
+        n = self._namespaces
+        p = self._projects
+        if not self.db.has_table(self._projects.name):
+            raise TableMissingError
```
dreadatour: Why do we check `projects` table, but did not check `namespaces` table but use both tables below? 🤔 Both tables implemented in the same PR (release), but if you don't know this it looks a bit odd and may cause questions in future.

src/datachain/data_storage/metastore.py
```diff
@@ -761,31 +1073,51 @@ def _get_dataset_query(
...
+            *(getattr(p.c, f) for f in project_fields),
             *(getattr(d.c, f) for f in dataset_fields),
             *(getattr(dv.c, f) for f in dataset_version_fields),
         )
-        j = d.join(dv, d.c.id == dv.c.dataset_id, isouter=isouter)
+        j = (
+            n.join(p, n.c.id == p.c.namespace_id)
+            .join(d, p.c.id == d.c.project_id)
+            .join(dv, d.c.id == dv.c.dataset_id, isouter=isouter)
+        )
```
dreadatour, line 1102: Oh, my, 3 joins 🫣 What about performance here, especially in SaaS? 🤔

src/datachain/data_storage/metastore.py
```diff
@@ -872,23 +1215,20 @@ def update_dataset_status(
     #
     def add_dataset_dependency(
         self,
-        source_dataset_name: str,
+        source_dataset: "DatasetRecord",
         source_dataset_version: str,
-        dataset_name: str,
-        dataset_version: str,
+        dep_dataset: "DatasetRecord",
```
dreadatour, line 1233: Same, I am not sure passing objects here is a good idea, especially in SaaS 🤔

src/datachain/data_storage/sqlite.py
```diff
@@ -394,10 +401,33 @@ def _init_tables(self) -> None:
...
+        """
+        Creates local namespace and local project connected to it.
+        In local environment user cannot explicitly create other namespaces and
+        projects and all datasets user creates will be stored in those.
+        When pulling dataset from Studio, then other namespaces and projects will
+        be created implicitly though, to keep the same fully qualified name with
+        Studio dataset.
+        """
+        local_namespace = self.create_namespace("local", "Local namespace")
+        self.create_project("local", local_namespace, "Local project")
```
dreadatour: Do we want to move `"local"' to the constants somewhere (for both namespace and project)? 🤔

src/datachain/data_storage/metastore.py
```diff
@@ -107,13 +113,95 @@ def cleanup_for_tests(self) -> None:
...
+    def is_studio(self) -> bool:
+        """Returns True if this code is ran in Studio"""
+
+    def is_local_dataset(self, dataset_namespace: str) -> bool:
+        """
+        Returns True if this is local dataset i.e. not pulled from Studio but
+        created locally. This is False if we ran code in CLI mode but using dataset
+        names that are present in Studio.
+        """
+        return self.is_studio or dataset_namespace == "local"
```
dreadatour: What if dataset in Studio will have `"local"` namespace? 🤔

src/datachain/data_storage/metastore.py
```diff
@@ -510,13 +691,141 @@ def _datasets_dependencies_update(self) -> "Update":
...
+        return [self.project_class.parse(*r) for r in rows]
+
     #
     # Datasets
     #
 
     def create_dataset(
         self,
         name: str,
+        project: Optional[Project] = None,
```
dreadatour: I really don't like we now have instances everywhere as input params 😢

src/datachain/dataset.py
```diff
@@ -422,8 +498,9 @@ def parse(  # noqa: PLR0913
         )
 
         return cls(
-            id,
+            dataset_id,
             name,
+            project,
```
dreadatour, line 495: Oh, my, this object is so heavy right now 😢 Do we really need all these data everywhere? What is the performance impact of these changes? 🤔

src/datachain/dataset.py
```diff
@@ -27,6 +29,8 @@
 LISTING_PREFIX = "lst__"
 
 DEFAULT_DATASET_VERSION = "1.0.0"
+DATASET_NAME_RESERVED_CHARS = ["."]
```
dreadatour, line 32: Do we use dots (`.`) in table names in metastore DB (SQLite, PostgreSQL) now? 🤔

src/datachain/dataset.py
```diff
@@ -358,10 +388,39 @@ def parse_schema(
...
         }
 
+    @staticmethod
+    def validate_name(name: str) -> None:
+        """Throws exception if name has reserved characters"""
+        for c in DATASET_NAME_RESERVED_CHARS:
+            if c in name:
+                raise InvalidDatasetNameError(
+                    f"Character {c} is reserved and not allowed in dataset name"
+                )
```
dreadatour, line 398: I would rather make a whitelist of characters we can use in dataset names.

src/datachain/dataset.py
```diff
@@ -647,14 +761,19 @@ def parse(  # noqa: PLR0913
         )
 
         return cls(
-            id,
+            dataset_id,
             name,
+            project,
```
dreadatour, line 758: Do we really need these heavy dependencies in each object? 😢

dreadatour (approved):
I have questions about implementation, API and performance, some I have asked below, other we will handle in the follow-up PRs.
Approving this to do not block development and testing.

src/datachain/data_storage/metastore.py
ilongin: It is true that we don't use `uuid` atm but I decided to add it regardless since we made similar mistake with datasets before not to have it. In general `id` is good to unique identify object in one DB but we now have more and more Studio <-> CLI communication where we can only identify if some object in CLI is the same as the one from Studio by checking `uuid` as they will have different `id`s.

src/datachain/data_storage/metastore.py
ilongin: Same answer. We don't use it atm but we might in the near future so to avoid migrations later on (specially on CLI). In general, I think all objects that are shared between Studio and CLI should have `uuid`.

src/datachain/data_storage/metastore.py
ilongin: We do have uniqueness on `name` but it's not set in abstract class since unique constraint is different in Studio (there we have `team` in addition) so it's set in specific implementations, in this example it's [here](https://github.com/iterative/datachain/pull/1115/files#diff-ead1114e5797d3ea60eb4d560593ab9af1f009757bd5bc785576b5e3310c30a4R424)

src/datachain/data_storage/metastore.py
ilongin: You are 100% right, it ended up here because I was c/p stuff and forgot to remove it. I will remove it from every method as it's not needed anywhere

src/datachain/data_storage/metastore.py
ilongin: Answered above in similar comment, removing it altogether from anywhere.

src/datachain/data_storage/metastore.py
ilongin: I don't think it should affect performance at all. In general, we don't have that many metadata specially comparing to actual dataset rows  (it's not even comparable). Here tables are small in terms of column number, won't have much data (I don't expect people having more than couple of namespaces / projects) so I really don't see this as a bottleneck specially when we look at all those UDF magic and creating million rows tables multiple times :D

src/datachain/data_storage/metastore.py
ilongin: `local` is reserved name and cannot be created anywhere, not even in Studio.  https://github.com/iterative/datachain/pull/1115/files#diff-637fdc4d8445f827befda06c8e204ecfc036ae04e7d2f3ec1b8b4421cd15f292R23

src/datachain/dataset.py
ilongin: I don't think it's heavy, specially considering `DatasetRecord` has all those versions objects inside of it anyway. `Project` has 5 small fields and embedded `Namespace` which has also 5 small fields. Only one field in one dataset version (e,g `schema`) is probably **always** **much** bigger than these 2 small objects added so the size it brings to the `DatasetRecord` object is IMO not something we should worry about. On the other side it brings much convenience by having it there as we must know to which project / namespace it belongs and that information is used across the codebase.

src/datachain/dataset.py
ilongin: It's not prohibited so I assume yes. In Studio migrations I migrated it to `_`  . Not sure if there's anything smarter we can do about it.

src/datachain/dataset.py
ilongin: I explained in details above but just to repeat, it's not heavy at all (specially comparing to what's already in these objects). I would be **extremely** surprised that this is affecting performance in any regards.  On the other hand it's needed to have it here.

src/datachain/data_storage/metastore.py
ilongin: Can you explain why it's not good idea in SaaS?

src/datachain/data_storage/sqlite.py
ilongin: Good idea, I will create it

src/datachain/catalog/catalog.py
```diff
@@ -1351,15 +1396,42 @@ def _instantiate(ds_uri: str) -> None:
...
+                f"Invalid fully qualified dataset name {remote_ds_name}, namespace"
+                f" or project missing"
+            )
+
+        if local_ds_name:
+            local_namespace, local_project, local_ds_name = parse_dataset_name(
+                local_ds_name
+            )
+            if local_namespace and local_namespace != remote_namespace:
+                raise DataChainError(
```
amritghimire, line 1424: Lets add tests for these missed parts

src/datachain/cli/__init__.py
```diff
@@ -171,19 +168,13 @@ def handle_dataset_command(args, catalog):
             args.name,
             version=args.version,
             force=args.force,
-            studio=args.studio,
```
amritghimire, line 174: Why are we removing these?

src/datachain/cli/parser/__init__.py
```diff
@@ -221,26 +221,6 @@ def get_parser() -> ArgumentParser:  # noqa: PLR0915
         nargs="+",
         help="Dataset attributes",
     )
-    parse_edit_dataset.add_argument(
```
amritghimire, line 224: Same

src/datachain/data_storage/metastore.py
amritghimire: Lets move this to constant then

src/datachain/data_storage/metastore.py
ilongin: Haven't we had similar discussion where we decided to have instances as inputs instead of raw fields? Or it was other way around? :thinking:  
I agree that it's questionable and I don't have strong opinion.  Advantage of having instances is that we pass only one argument instead of many if multiple fields from that instance is needed in the method. It's more similar to object oriented design. On the other hand, we do pass some fields then that are not needed and usually we need only `id` field.

TL;DR I will try to change it here (and in other methods) to accept only `project_id` instead, but in higher level codebase (`DataChain`, `DatasetQuery`, `Catalog` etc.) I would leave whole instances as there there is much higher chance to use other fields from it and we usually work with whole instances in general there so to stay consistent. WDYT?

src/datachain/data_storage/metastore.py
ilongin: Answered below https://github.com/iterative/datachain/pull/1115#discussion_r2137499332

src/datachain/dataset.py
ilongin: I do agree it's maybe better as it's safer, i.e. you 100% know what will go inside, but currently I'm not sure what we will put in those whitelisted chars (we need to decide on that)... To not waste time on this I suggest to leave it as is and do whitelist as a followup.

src/datachain/cli/__init__.py
ilongin: On of the important motivation for introducing namespaces / projects was to reduce this `studio` boolean flags across the codebase. Here, user should provide fully qualified name that has namespace and project names in it or it can just provide dataset name in which case default namespace / project is used. If user provides fully qualified name with namespace / project that is different thant `local` - then it's studio dataset as locally users can only create datasets in `local` namespace and `local` project.
So:
```
datachain ds rm cats -> removes local cats dataset
datachain ds rm local.local.cats -> removes local cats dataset
datachain ds rm dev.animals.cats -> removes Studio cats dataset in dev namespace and animals project
```

src/datachain/cli/parser/__init__.py
ilongin: Same answer as above

src/datachain/data_storage/metastore.py
dreadatour: Can we use `uuid` as primary key may be? 🤔 [Same way as for Job model](https://github.com/iterative/datachain/blob/main/src/datachain/data_storage/metastore.py#L1002-L1008)?

src/datachain/data_storage/metastore.py
dreadatour: Added a comment in SaaS PR 🤗

dreadatour:
@ilongin thank you for answering all the questions 🙏❤️

src/datachain/data_storage/metastore.py
dreadatour: One more reason: these objects are not immutable, and it is very often you see they are modified inside the method which cause a lot of issues and troubles, then you do not expect this.

Me personally I am "OK"-ish with this if objects are immutable, but would prefer to do not do that in these methods (metastore and warehouse) at least.

src/datachain/catalog/catalog.py
ilongin: Actually, the idea is to completely remove this `local_ds_name` thing (I spoke to Dmitry about it and he agrees) as it's not needed and it just complicates system. Pulling Studio dataset can be considered as "caching" it locally and there is no reason to use different name for it. If user want's to use different name, it can create new dataset out of it. For that reason I would avoid writting tests for it now not to waste time on something which will be removed soon in followup.

src/datachain/data_storage/metastore.py
ilongin: Yea, that makes sense as well. I will do the same

.github/workflows/tests-studio.yml
```diff
@@ -98,6 +98,7 @@ jobs:
       - name: Run tests
         # Generate `.test_durations` file with `pytest --store-durations --durations-path ../.github/.test_durations ...`
         run: >
+          DATACHAIN_METASTORE_ARG_USERNAME=john
```
shcheklein: what is so unique about this env var that we set it here? (and only that var). Let's clean this up please

src/datachain/catalog/catalog.py
```diff
@@ -675,7 +676,11 @@ def _row_to_node(d: dict[str, Any]) -> Node:
             listing: Optional[Listing]
             if src.startswith("ds://"):
                 ds_name, ds_version = parse_dataset_uri(src)
-                dataset = self.get_dataset(ds_name)
+                ds_namespace, ds_project, ds_name = parse_dataset_name(ds_name)
+                assert ds_namespace
+                assert ds_project
+                project = self.metastore.get_project(ds_project, ds_namespace)
```
shcheklein, line 682: why do we need an extra call here? (it seems that ds_namespace is availlable already, project name as well)

why do we need it in general - can `get_dataset` call return all the details at once? it seems too complicated otherwise - too many granular calls ...

src/datachain/catalog/catalog.py
```diff
@@ -254,9 +257,7 @@ def do_task(self, urls):
         import pandas as pd
 
         # metastore and warehouse are not thread safe
-        with self.metastore.clone() as metastore, self.warehouse.clone() as warehouse:
```
shcheklein, line 257: why is it safe to remove `metastore.clone()`?

src/datachain/cli/parser/__init__.py
```diff
@@ -311,26 +291,6 @@ def get_parser() -> ArgumentParser:  # noqa: PLR0915
...
-    rm_dataset_parser.add_argument(
-        "--studio",
-        action="store_true",
-        default=False,
-        help="Remove dataset from Studio",
-    )
-    rm_dataset_parser.add_argument(
-        "-L",
-        "--local",
-        action="store_true",
```
shcheklein, line 323: do we still need a command to remove local version / cache of a dataset from Studio? (how do we distinguish these operations Studio / non-Studio overall) for locally cached ds from Studio and remote ones? E.g. what happens when we do rename?

src/datachain/dataset.py
shcheklein: we should not be using dots as table names in both (if understand what is happening correctly)

table names schema should be consistent across databases

src/datachain/lib/namespaces.py
```diff
@@ -0,0 +1,69 @@
...
+from datachain.error import NamespaceCreateNotAllowedError
+from datachain.namespace import Namespace
+from datachain.query import Session
+
+
+def create(
+    name: str, description: Optional[str] = None, session: Optional[Session] = None
+) -> Namespace:
+    """
+    Creates new custom namespace.
```
shcheklein: a new ...

(run description via chat gpt, etc to fix mistakes, syntax, etc)

src/datachain/lib/namespaces.py
```diff
@@ -0,0 +1,69 @@
...
+    Note that creating namespaces is not allowed for local environment, unlike in
+    Studio where it is allowed.
+    In local environment all datasets are created under default `local` namespace.
+
+    Parameters:
+        name : namespace name.
+        description : namespace description.
```
shcheklein: consistency - start the description with a capital everywhere or lower case ... (see Session below)

check how we docs in other place ... clean this up please carefully

src/datachain/lib/namespaces.py
shcheklein: add some meaningful one sentence what namespaces are for with some examples

src/datachain/lib/namespaces.py
```diff
@@ -0,0 +1,69 @@
...
+        ```py
+        import datachain as dc
+        namespace = dc.get_namespace("local")
+        ```
+    """
+    session = Session.get(session)
+    return session.catalog.metastore.get_namespace(name)
+
+
+def ls(session: Optional[Session] = None) -> list[Namespace]:
```
shcheklein, line 60: can we list remote namespaces? do we need that? do we need `ls` at all?

src/datachain/namespace.py
```diff
@@ -0,0 +1,41 @@
...
+from typing import Any, Optional, TypeVar
+
+N = TypeVar("N", bound="Namespace")
+
+
+@dataclass(frozen=True)
+class Namespace:
+    id: int
+    uuid: str
+    name: str
```
shcheklein, line 16: do we validate names in some way at all? length, allowed symbols, etc, etc?

tests/utils.py
```diff
@@ -187,7 +187,9 @@ def wait_for_condition(
 def assert_row_names(
     catalog: Catalog, dataset: DatasetRecord, version: str, expected_names: set
 ) -> None:
-    dataset_rows = catalog.ls_dataset_rows(dataset.name, version, limit=20)
+    dataset_rows = catalog.ls_dataset_rows(
+        dataset.name, version, dataset.project, limit=20
```
shcheklein: why can't DatasetRecord include the project? (while Project includes Namespace)?

tests/unit/test_session.py
```diff
@@ -9,17 +9,31 @@
 from datachain.sql.types import String
 
 
-def test_ephemeral_dataset_naming(catalog):
+@pytest.fixture
+def project(catalog):
+    namespace = catalog.metastore.create_namespace("dev")
```
shcheklein, line 14: how does it work locally?

tests/unit/test_session.py
```diff
@@ -9,17 +9,31 @@
...
+def test_ephemeral_dataset_naming(catalog, project):
     session_name = "qwer45"
 
     with pytest.raises(ValueError):
         Session("wrong-ds_name", catalog=catalog)
 
     with Session(session_name, catalog=catalog) as session:
         ds_name = "my_test_ds12"
-        session.catalog.create_dataset(ds_name, columns=(sa.Column("name", String),))
+        # namespace = session.catalog.metastore.create_namespace("dev")
```
shcheklein: dead code? remove?

tests/test_cli_studio.py
```diff
@@ -265,27 +267,29 @@ def test_studio_edit_dataset(capsys, mocker):
...
+            "namespace": "dev",
+            "project": "animals",
             "new_name": "new-name",
             "description": "description",
             "attrs": ["attr1"],
             "team_name": "team_name",
         }
 
 
+@skip_if_not_sqlite
```
shcheklein, line 286: should it be `skip_if_sqlite`? should we actually name it - `skip_if_local`? (sqlite is an implementation detail)

src/datachain/lib/projects.py
```diff
@@ -0,0 +1,81 @@
...
+
+    Example:
+        ```py
+        import datachain as dc
+        project = dc.create_project("my-project", "dev", "My personal project")
+        ```
+    """
+    session = Session.get(session)
+
+    if not session.catalog.metastore.project_allowed_to_create:
```
shcheklein, line 37: these checks should be handled by the create_project below (we are leaking too much for now reason, all these `session.catalog.metastore.project_allowed_to_create` look too complicated)

src/datachain/listing.py
```diff
@@ -66,7 +66,9 @@ def uri(self):
     @cached_property
     def dataset(self) -> "DatasetRecord":
         assert self.dataset_name
-        return self.metastore.get_dataset(self.dataset_name)
+        return self.metastore.get_dataset(
+            self.dataset_name, self.metastore.default_project
```
shcheklein: what is the default_project in Studio?

I feel we need a completely separate namespace / project for the listings ...

src/datachain/query/dataset.py
```diff
@@ -1128,55 +1130,45 @@ def __init__(
...
-                # exact version
-                self._set_starting_step(self.catalog.get_dataset(name))
-            else:
-                # not setting query step yet as listing dataset might not exist at
-                # this point
-                self.list_ds_name = name
+        namespace_name = namespace_name or self.catalog.metastore.default_namespace_name
+        project_name = project_name or self.catalog.metastore.default_project_name
+
+        if is_listing_dataset(name) and not version:
```
shcheklein, line 1136: this makes it less readable, more magical with some implicit assumptions (I mean the `not version` part) ... it is not clear at all from the code why this if is written specifically this way, why it fine, etc

src/datachain/query/dataset.py
```diff
@@ -1676,26 +1673,36 @@ def generate(
...
                 for dep in self.catalog.get_dataset_dependencies(
-                    dep_name, dep_version, indirect=False
+                    dep_dataset.name,
+                    dep_dataset_version,
+                    dep_dataset.project,
+                    indirect=False,
                 ):
                     if dep:
-                        dependencies.add((dep.name, dep.version))
+                        dep_project = self.catalog.metastore.get_project(
```
shcheklein, line 1685: it is an extra call? 

why this logic has become so more complicated and with 1-2 extra calls? (??)

src/datachain/query/dataset.py
```diff
@@ -1770,11 +1784,25 @@ def save(
 
             if dependencies:
                 # overriding dependencies
-                self.dependencies = {(dep.name, dep.version) for dep in dependencies}
```
shcheklein, line 1773: why doesn't `dependencies` contain all the information we need? (project, etc)

src/datachain/dataset.py
ilongin: aha sorry, I missunderstood. I thought we are speaking about dataset names. We **don't** use dots in table names though.

src/datachain/data_storage/metastore.py
ilongin: Decided to not do this atm as it would take me too much time. We can do it in follow-up.

src/datachain/data_storage/metastore.py
ilongin: Resolved, I'm not sending whole objects any more into metastore functions

src/datachain/data_storage/metastore.py
ilongin: I've actually put `Namespace` and `Project` to be immutable, but regardless I'm not sending whole objects into metastore any more.

.github/workflows/tests-studio.yml
ilongin: SaaS metastore must have user's username in it to be able to have default project name. In CLI this is not a problem as we just have constant `local`, but we decided to have username for SaaS. This is just for tests to work

src/datachain/catalog/catalog.py
ilongin: Metastore accepts raw fields like `project_name` and `namespace_name` but higher level code like `Catalog`, `DataChain` (in this case `Catalog.get_dataset()`) etc. works with `Project` instance (it has `Namespace` in it). This is because in general we work with instances in higher level code since we usually need multiple fields from it, function signatures looks better since we are sending just one argument instead of many etc.
I agree that here we have one more call that's not really needed but we need to decide should we have cleaner code and have few DB calls more or otherwise ... here (in "metadata" code) I'm leaning to the first option as these parts of the code are not affecting the performance (comparing to actual "work" which is done with processing large number of rows in chain this is not relevant)

src/datachain/catalog/catalog.py
ilongin: It's not that `metastore.clone()` is removed but we don't use metastore any more in this part of the code at all, just warehouse. This is because `DatasetRowsFetcher` now accepts dataset instance in constructor so no need to fetch it one more time in `do_task` function.

src/datachain/cli/parser/__init__.py
ilongin: You are right. I totally missed this cache removal .. I will think about what is the best way to do this. 
I think we should disallow renaming these "cached" datasets since they are just that - cache and there is no need for that. WDYT? As I spoke with Dmitry, we should even remove that special "local name" argument in `datachain pull` to save it under different local name. If user really wants to have it under some different name it can just create another dataset out of it.

src/datachain/lib/namespaces.py
ilongin: Done

src/datachain/lib/namespaces.py
ilongin: Currently we cannot list remove namespaces. We needed listing of namespaces for some Studio logic so I thought of adding it in public interface as well .. I figure i could be useful.

src/datachain/namespace.py
ilongin: Currently we just check for reserved name (`local` is not allowed). We actually don't do much of these kind of validations in general in our codebase .. maybe we should but, I don't think it's a big priority ATM.

tests/utils.py
ilongin: `DatasetRecord` does include `Project` so yes, we should refactor `ls_dataset_rows()` to accept just `DatasetRecord` instead. This aligns with discussion in above [comment](https://github.com/iterative/datachain/pull/1115#discussion_r2143690460) to have these higher level methods / code accept instances instead of a raw fields.

ilongin: done

tests/unit/test_session.py
ilongin: By default I've allowed creating namespaces and projects in tests as it's really needed there to test everything. I also added ability to disallow it where specific local experience is needed to test. Here are the fixtures that are making that happen [link](https://github.com/iterative/datachain/pull/1115/files#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128R545-R566)

tests/unit/test_session.py
ilongin: I think I removed it in mean time as it's not there any more.

tests/test_cli_studio.py
ilongin: We could, but I would do that in separate issue / PR as it's general thing

shcheklein: was `skip_if_not_sqlite` introduced by you? 

also, again, why do we skip it at all, in what cases?

src/datachain/lib/projects.py
ilongin: I actually think they should be here (on user interface level). Usually those auth checks / validations are close to interface (e.g in REST API). For example, we **do** need to create project some time in CLI, like when pulling dataset form Studio which has custom project / namespace name. 
I agree `session.catalog.metastore` looks too complicated, but the reason is that `catalog` is probably the excess here (and all around the codebase) and should be completely removed in future (something to discuss)

.github/workflows/tests-studio.yml
shcheklein: question is why we define it this way? right in CI, and no other vars are defined this way

it looks weird

also, name - let's make it then DATACHAIN_USER_NAME? Why is it about metastore?

re the default namespace: how do we pass in Studio a different default namespace? how will we show in UI which namespace is used now?

src/datachain/cli/parser/__init__.py
shcheklein: yes, we should not be renaming "cached" datasets. On the other hand we probably should allow brining dataset right into local namespace/project? why not ... then it can be renamed, etc

src/datachain/namespace.py
shcheklein: what is that about that reserved `.` env var? is it part of the name vaildation?

src/datachain/catalog/catalog.py
shcheklein: hmm, not sure I would this code looks cleaner tbh 🤔  

can we override get_dataset to also work with names? (that we already have here) .. what would be wrong with that approach?

src/datachain/catalog/catalog.py
```diff
@@ -1026,57 +1049,71 @@ def create_dataset_from_sources(
...
+            )
 
             self.pull_dataset(
                 remote_ds_uri=remote_ds_uri,
                 local_ds_name=name,
                 local_ds_version=version,
             )
-            return self.get_dataset(name)
+            return self.get_dataset(
+                name, self.metastore.get_project(project_name, namespace_name)
```
shcheklein, line 1103: same here, this code pattern looks weird to me ...

src/datachain/cli/__init__.py
shcheklein: how do we remove locally cached version of the dataset?

tests/test_cli_studio.py
ilongin: I'm not sure by whom it was introduced but it was a while ago (def not in this PR). There are multiple tests that are skipped, in this PR we skip tests that are testing disabling creating namespace / project in CLI

src/datachain/data_storage/metastore.py
```diff
@@ -107,13 +113,95 @@ def cleanup_for_tests(self) -> None:
...
+    def get_namespace(self, name: str, conn=None) -> Namespace:
+        """Gets a single namespace by name"""
+
+    @abstractmethod
+    def list_namespaces(self, conn=None) -> list[Namespace]:
+        """Gets a list of all namespaces"""
+
+    @property
+    @abstractmethod
+    def is_studio(self) -> bool:
```
shcheklein, line 149: do we use it enywhere beside allow_create_projects ?

I would try to remove it (have less studio specific code - just make metastore implement decide ... if it is not implemented let those methods raise?)

src/datachain/listing.py
ilongin: It's `User.usernmae` . I was thinking about this and decided to put listing into default namespace (`users`) and default project (e.g `ilongin`). If there is a listing created in one user's default project it should be "visible" by other users scripts and reused but I'm not sure this actually works (need to double check).
About new completely separate namespace / project for listing - I'm not against that as well ... didn't want to go with that at beginning not to complicate but maybe it's the best option .. need to think about it.

src/datachain/data_storage/metastore.py
```diff
@@ -107,13 +113,95 @@ def cleanup_for_tests(self) -> None:
...
+    def is_local_dataset(self, dataset_namespace: str) -> bool:
+        """
+        Returns True if this is local dataset i.e. not pulled from Studio but
+        created locally. This is False if we ran code in CLI mode but using dataset
+        names that are present in Studio.
+        """
+        return self.is_studio or dataset_namespace == Namespace.default()
+
+    @property
+    def namespace_allowed_to_create(self):
```
shcheklein, line 161: can we just raise if it is not allowed? this all set of methods looks a bit weird / hacked to just fit current immediate need - can it be simplified?

src/datachain/data_storage/metastore.py
```diff
@@ -107,13 +113,95 @@ def cleanup_for_tests(self) -> None:
...
+    @abstractmethod
+    def list_namespaces(self, conn=None) -> list[Namespace]:
+        """Gets a list of all namespaces"""
+
+    @property
+    @abstractmethod
+    def is_studio(self) -> bool:
+        """Returns True if this code is ran in Studio"""
+
+    def is_local_dataset(self, dataset_namespace: str) -> bool:
```
shcheklein, line 152: this also super weird name - `metastore.is_local_dataset` ... just to check if namespace name is local?

src/datachain/query/dataset.py
ilongin: `DatasetDependency` contains only those fields that are needed, not all the objects like `DatasetRecord`. Maybe we should refactor that....not sure though, need to think about it and see where it's used as it would increase the size of it. Not sure what is the max dependencies is realistically to expect for some dataset (if we include those non direct ones as well) as this information is fetched in Studio through that REST API so don't want to slow things.  

In this particular example the answer is the same as above, `Catalog` accepts instances and not raw fields but here we don't have instance.

ilongin: btw this is really special case here which happens only if someone uses `.persist()` inside the chain

src/datachain/namespace.py
ilongin: Oh, you are right, I totally forgot that dots should be prohibited here as well (and for projects), not only for datasets. I will add this.

ilongin: I've fixed this

src/datachain/data_storage/sqlite.py
```diff
@@ -60,6 +70,14 @@
 quote_schema = sqlite_dialect.identifier_preparer.quote_schema
 quote = sqlite_dialect.identifier_preparer.quote
 
+# NOTE! This should be manually increased when we change our DB schema in codebase
+SCHEMA_VERSION = 1
+
+OUTDATED_SCHEMA_ERROR_MESSAGE = (
+    "You have old version of database schema. Please refer to the documentation"
```
shcheklein: most likely:

an old ... the database schema 

(please run it by gpt again)

tests/unit/test_data_storage.py
```diff
@@ -75,3 +77,32 @@ def test_db_defaults(col_type, default_value, catalog):
...
     warehouse.db.drop_table(table)
+
+
+@skip_if_not_sqlite
+def test_outdated_schema(catalog):
+    from datachain.data_storage.sqlite import SCHEMA_VERSION
+
+    metastore = catalog.metastore
+
+    metastore._check_schema_version()  # should not raise exception
```
shcheklein: you don't really test this way that we actually call this internal method

it is better to call metastore init in different configurations and let is raise and test actually behavior, not internal business logic

src/datachain/dataset.py
```diff
@@ -358,10 +388,39 @@ def parse_schema(
...
+    def validate_name(name: str) -> None:
+        """Throws exception if name has reserved characters"""
+        for c in DATASET_NAME_RESERVED_CHARS:
+            if c in name:
+                raise InvalidDatasetNameError(
+                    f"Character {c} is reserved and not allowed in dataset name"
+                )
+
+    @staticmethod
+    def clean_name(name: str) -> str:
```
shcheklein: so, do we validate or replace? (not sure when we one thing and when the second one)

src/datachain/lib/dc/datachain.py
```diff
@@ -490,6 +499,22 @@ def listings(
         )
         return listings(*args, **kwargs)
 
+    @property
+    def namespace_name(self) -> str:
+        """Current namespace name in which the chain is running"""
+        return (
+            self._settings.namespace
+            or self.session.catalog.metastore.default_namespace_name
```
shcheklein, line 507: it feels that things like `default_namespace_name` belong to catalog, not particular metastore ... e.g. let's sy weill support duckdb and sqlite in the future. default_namespace must be the same for both ... os unless I'm missing something it doesn't really depend on a particular metastore implementation

src/datachain/lib/dc/datachain.py
```diff
@@ -490,6 +499,22 @@ def listings(
...
+        )
+
+    @property
+    def project_name(self) -> str:
+        """Current project name in which the chain is running"""
+        return (
+            self._settings.project
```
shcheklein, line 514: project -> project_name (for consistency and to distinguish from cases when we use Project)

shcheklein: not sure though, I see that is is coming for the user level API 🤔 (we need to check that we are consistent across ue facing APIs )

src/datachain/lib/listing.py
```diff
@@ -195,5 +196,6 @@ def get_listing(
         list_path = f"{ds_name.strip('/').removeprefix(listing.name)}/{list_path}"
 
     ds_name = listing.name if listing else ds_name
+    ds_name = DatasetRecord.clean_name(ds_name)
```
shcheklein: this logic with cleaning listing names seems to be off ... 

e.g. 

`ds_name, list_uri, list_path = parse_listing_uri(uri)`

probably returns "dirty" version?

but then

```
ls for ls in catalog.listings() if not ls.is_expired and ls.contains(ds_name)
```

will be comparing it with cleaned versions?

also, are there other places in code where we construct listing ds names (e.g. via `parse_listing_uri`)?

src/datachain/lib/namespaces.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+    Gets a list of all namespaces.
+
+    Parameters:
+        session : Session to use for getting project.
```
shcheklein: minor: fix - projec -> namespaces

src/datachain/lib/namespaces.py
```diff
@@ -0,0 +1,73 @@
...
+
+    Example:
+        ```py
+        import datachain as dc
+        namespace = dc.namespaces.create("dev", "Dev namespace")
+        ```
+    """
+    session = Session.get(session)
+
+    if not session.catalog.metastore.namespace_allowed_to_create:
```
shcheklein, line 33: let just `create_namespace` itself raise (and take care or the check)

src/datachain/lib/projects.py
shcheklein: Hmm ... if `session.catalog.metastore.create_project` allows dataset to be created, then it is a very easy bypass and someone has as dataset in a local database. Do we want to allow that? This logic (allow or not) should be deeper, even to enforce the behavior we need. Unless I'm missing something.

tests/unit/lib/test_datachain.py
```diff
@@ -3206,13 +3210,18 @@ def test_delete_dataset_versions_all(test_session):
 
 
 @pytest.mark.parametrize("force", (True, False))
+@skip_if_not_sqlite
```
shcheklein, line 3213: I still don't quite understand - why don't we test it against CH? we kinda do a lot of tests in way that go against usual mode, and at the same we don't test CH usage ... where it be actually used. Why is that an issue to run them with CH or both metastores?

shcheklein (commented):
okay, I think I've done the first pass finally. I've outline some 3 important (to my mind items) -shared them in Slack. Quite a lot of smaller things here and there - it would be nice to cleanup, but probably they don't affect the API ...

src/datachain/lib/projects.py
ilongin: True, if someone really wants to "hack" the system and bypass, it can be done regardless of what we do, but since this is open source it's users responsibility to use it in proper way.

The fact is that we need to create project implicitly at some point (when pulling dataset from studio), and we also need to limit users of doing so in explicit way.

src/datachain/query/dataset.py
ilongin: The real reason this whole logic about listings, starting steps etc in `DatasetQuery` is complex and "magical" lays in the fact that we have core logic of running step inside legacy subclass (`DatasetQuery`) and not in `DataChain`. This is causing issues in other places as well, e.g in delta updates. I think we really need to prioritize removing `DatasetQuery` class altogether.

src/datachain/query/dataset.py
ilongin: When developing dependencies I wanted to make `DatasetDependency` class lightweight since it's nested (forms a tree structure). Also, before we didn't have projects etc. Maybe we can re-consider this as these meta objects are not that big regardless .. not sure.

src/datachain/query/dataset.py
shcheklein: I mean - the previous version was simpler to read + probably had a comment or something

I'm not asking to refactor it everywhere ... I'm just asking to write it in more explicit / easy to read way

src/datachain/query/dataset.py
shcheklein: make it accept non-instnance? it is weird a bit that we do a call just to convert ... do we need anything from the Project object in the logic in the Catalog. Again, the design smells a bit to me here (I understand why is it written this way)

src/datachain/lib/dc/datachain.py
ilongin: `Catalog` should be refactored / removed IMO in the future. We should have business logic inside "domain" classes like `datasets`, `jobs` etc.
`default_namespace_name` is in abstract class so it doesn't need to be re-implemented for other DBs if not needed.

tests/unit/test_data_storage.py
ilongin: Refactored

src/datachain/lib/listing.py
ilongin: Yea, this was a mistake. I've now completely removed `DatasetRecord.clean_name()` and move it's logic to `parse_listing_uri()` as that was the only place using that method which was also very tied to constructing listing names anyway .. now I think it's much cleaner and simpler.

.github/workflows/tests-studio.yml
ilongin: Because this is the way we define arguments to metastore, see https://github.com/iterative/studio/blob/main/backend/datachain_server/scripts/datachain-local ... We didn't set arguments before since they were optional, but now we have `username` which is mandatory as we need it to create custom user project in default `users` namespace. It needs to start with `DATACHAIN_METASTORE` as that's our "contract"

src/datachain/dataset.py
ilongin: Removed this whole method at the end

src/datachain/lib/dc/datachain.py
ilongin: On user facing API I was using `project` and `namespace` instead of `project_name` and `namespace_name` as I think it's better / cleaner.. inside our code it's `project_name` on lot of places as we also often have instances of `Project` which are named `project` so just to be explicit and not shadow variables.

src/datachain/data_storage/metastore.py
ilongin: This whole thing about `is_studio`, `is_local_dataset` , `is_allowed_to_create_` can be refactored afterwards. I agree it could be better then it's now .. but we do need to have some kind of flag saying if we are running in CLI or in Studio. We cannot just rely on implementation / not implementation as we need that information in some other places above data storage and we don't want to have Studio implementation of each class that needs that information since that would mess the codebase... right now we only have specific implementations of metastore, warehouse and those base classes like `Dataset` (since they have additional fields in studio like `team_id`) and I think we should keep it that way

shcheklein: sounds good, definitely not a blocker ... let's put a list / ticket of refactorings ... while we are testing we can start in a separate PR addressing those (it is easier to do when we have fresh context)

src/datachain/lib/dc/datachain.py
shcheklein: yep, that's fine for now if it is consistent! thanks

PR #1116, opened 2025-05-22
improve setup docs to make it clear that callable is expected

shcheklein:
Replaces https://github.com/iterative/datachain/pull/1070


dreadatour (approved):
Thank you for the improvement ❤️

PR #1117, opened 2025-05-24
fix(video): append models only if they exist

shcheklein:
Fixes issue on prod: 

```
Error while waiting for UDF d1add18f-9e34-4c31-a4ac-03108bcdebfb: Pose3D must be a dictionary with coordinates.
```

probably in new Utlralytics we can get None as keypoints when we have some bounding boxes returned.

This code:

```
def pose_estimation(yolo: YOLO, file: VideoFile) -> Iterator[tuple[VideoFrame, ImageFile, YoloPoses]]:
    for frame in file.get_frames(step=10):
        image = frame.save("gs://datachain-cattle-care/frames")
        results = yolo(image.read())
        yield frame, image, YoloPoses.from_results(results)
```

## TODO

- [x] check if segments can contain empty bboxes with non-empty segments
- [x] add tests for segments

### DIFF
--- a/src/datachain/model/ultralytics/pose.py
+++ b/src/datachain/model/ultralytics/pose.py
@@ -58,12 +58,12 @@ class YoloPose(DataModel):
         name = summary[0].get("name", "")
         box = (
             BBox.from_dict(summary[0]["box"], title=name)
-            if "box" in summary[0]
+            if summary[0].get("box")
             else BBox()
         )
         pose = (
             Pose3D.from_dict(summary[0]["keypoints"])
-            if "keypoints" in summary[0]
+            if summary[0].get("keypoints")
             else Pose3D()
         )
         return YoloPose(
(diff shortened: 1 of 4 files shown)


src/datachain/model/ultralytics/pose.py
```diff
@@ -102,8 +102,10 @@ def from_results(results: list["Results"]) -> "YoloPoses":
                 cls.append(s["class"])
                 names.append(name)
                 confidence.append(s["confidence"])
-                box.append(BBox.from_dict(s.get("box", {}), title=name))
-                pose.append(Pose3D.from_dict(s.get("keypoints", {})))
+                if s.get("box"):
```
dmpetrov, line 105: How about `if "box" in s:` instead?

dmpetrov (approved):
Looks good

src/datachain/model/ultralytics/pose.py
shcheklein: apparently ultralytics returns `None` for some (or all of these objects). Easier to ignore them in such cases

PR #1119, opened 2025-05-26
More func tests

dreadatour:
More tests for sql functions.


0x2b3bfa0 (approved):
Not a "func"[^1] of the path names, but happy to have those tests.

[^1]: Read "fan"

dreadatour:
> Not a "func" of the path names, but happy to have those tests.

Renamed to `func/functions` 🥰

Issue #1103, opened 2025-05-16
Read S3 Storage Without Object Versions


dmpetrov:
### Discussed in https://github.com/iterative/datachain/discussions/1102

<div type='discussions-op-text'>

<sup>Originally posted by **dvnatanael** May 15, 2025</sup>
Is it possible to use `datachain.read_storage` to read from an S3 bucket without the `ListObjectVersions` permission? I am just starting to check out DataChain, and I currently do not have this permission for the bucket in question.

The [documentation](https://docs.datachain.ai/references/remotes/?h=permissions#aws-s3-compatible-storage) on remote storage suggests that I should be able to pass `{"version_aware": False}` to the `client_config` parameter. However, the implementation of the S3 client always sets the pagination method to `list_object_versions`:

https://github.com/iterative/datachain/blob/b57275a36c552081b7e5d1f6bfb58b2a135e42fe/src/datachain/client/s3.py#L104-L114

Finally, are there other permissions that DataChain requires which are not listed in the documentation?

Thanks in advance!</div>

dmpetrov:
It's fixed by #1127. Closing.

PR #1127, opened 2025-06-02
fix(client): avoid version ids if version aware is false

shcheklein:
Fixing a few issue where even setting `version_aware=False` was not enough to force it not use Version APIs + do all the version id related checks.

We now don't fetch `version_id` field in commands like `read_storage` for the Files if `version_aware` False is passed. Column value will be None and we won't be enforcing or using `version_id` downstream (e.g. in prefetch, etc).

To make code like this:

```python
datachain.read_storage("s3://ivan-test-versioned", client_config={"version_aware": False}, update=True).show(10, include_hidden=True)
datachain.read_csv("s3://ivan-test-versioned/test.json", client_config={"version_aware": False}, update=True).show(10)
```

work w/o additional permissions:

```
s3:ListBucketVersions
s3:GetObjectVersion
```

(and the same should be for the GCS and Azure)

Note: testing this is complicated. Azure test FS doesn't support versioning. Main changes are related to the permissions and we can't check them on emulators properly AFAIU.

### DIFF
--- a/src/datachain/client/azure.py
+++ b/src/datachain/client/azure.py
@@ -15,7 +15,7 @@ class AzureClient(Client):
     protocol = "az"
 
     def info_to_file(self, v: dict[str, Any], path: str) -> File:
-        version_id = v.get("version_id")
+        version_id = v.get("version_id") if self._is_version_aware() else None
         return File(
             source=self.uri,
             path=path,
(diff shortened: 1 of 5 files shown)


dreadatour (approved):
Looks good to me 👍

dreadatour:
Related issue: https://github.com/iterative/datachain/issues/1103

PR #1133, opened 2025-06-03
basic AI generated version for retry

shcheklein:
Add flags to source commands (`read_storage`, etc) to retry failed / missing records.

```python
   retry_chain = (
        dc.read_dataset(
            "sample_data",
            # Enable retry processing
            retry=True,
            # Match records based on the data.id field
            match_on="data.id",
            # Retry records where error field is not None
            retry_on="error",
        )
        .map(result=process_data)
        .save(name="processed_data")
    )
```


dmpetrov:
Great idea! A comment on the high level design:

> ```python
>             # Match records based on the data.id field
>             match_on="data.id",
> ```

this part is not obvious. Why should this matter? Can we get rid of this? 🙂

dmpetrov:
That's great change - retry, finally!

I've look at the example, not implementation. A few comments are below.

**First**, it looks like there is an opportunity to simplify user experience - user should not catch exceptions by themself, this is retry's job.

This part is not needed for users:
```python
        except Exception as e:
            # Return an error that will trigger reprocessing next time
            return {
                "content": None,
                "result": None,
                "error": str(e)  # Error field will trigger retry
            }
```

An ideal, API should have only a single flag `retry=True` without any additional fields and exception logic:
```python
def read_class(file) -> str:
   return json.loads(file.read())["class"]

ds = dc.read_dataset("s3://bkt", retry=True)
res = ds.map(cls=read_class)

# User realized not everything was processed and fixes a but in the code
def read_class_fixed(file) -> str:
   return json.loads(file.read()).get("class", "")

res = res.map(cls=read_class_fixed)
```

If user needs to see error they can specify the column using the same attr like `retry="my_error"`.

It would be great if retry is enabled by default - it requires a dead-simple API.

**Second**, retry feels like map-er level logic, not read_dataset() level. Some mapper might need retry while the other might not within the same chain of commands.

**Third**, we need a function to check if there are no issues with retry-s like `if ds.needs_retry(): ...`. 
Also, `ds.save("my_ds")` should probably fail if retry is needed (and force=True can still save it).

**Fourth**, `retry_missing` is hard to understand - it becomes clear when connection with delta is establish. SHould it be renamed to `retry_delta` or similar?

shcheklein:
Good comments @dmpetrov . Thanks for reviewing this. I like the suggestions, some of them I might incorporate now, some of them I had in mind for the next iterations, some of them I feel we need a bit more clarity demand.

> First, it looks like there is an opportunity to simplify user experience - user should not catch exceptions by themself, this is retry's job.

💯  but it is more an mechanism for mappers to kinda skip-failed records or return some default on errors. I also think we need some syntax sugar for this no matter if we have retries or not. I'll look into this as a next step. Probably better create a ticket for this. I'm not sure this is "retries" job though necessarily (depends I guess on how you see retry itself). 

> An ideal, API should have only a single flag retry=True without any additional fields and exception logic:

Yep, I can do a single flag that takes "boolean | str". I like the simplicity and naming (just a single `retry`). I feel though we might be hurting UX a bit (explicit is better than some implicit mix of things). I also prefer a single flag though ...

> It would be great if retry is enabled by default - it requires a dead-simple API.
> Second, retry feels like map-er level logic, not read_dataset() level. Some mapper might need retry while the other might not within the same chain of commands.

this seems goes into a different kind of retries (checkpoints / cached results) - this indeed should be enabled by default for all mappers most likely (if something is calculated before - we are reusing that next time). Or more like real-time retries (e.g. some intermittent error of quick fix available).

`retry` implemented here is a bit different - more on the "whole" query level, like delta mode. It simple and covers clear scenario - some record in the result are missing or failed - recalculate the whole block for them.

We will need both most likely. If later mappers cover all we need we can revisit this.

> Third, we need a function to check if there are no issues with retry-s like if ds.needs_retry(): ....
Also, ds.save("my_ds") should probably fail if retry is needed (and force=True can still save it).

I would wait on this ... again, seems a bit different semantics, different scenario.

> Fourth, retry_missing is hard to understand - it becomes clear when connection with delta is establish. SHould it be renamed to retry_delta or similar?

I'll make a single flag. Yes, it is now actually tightly coupled with delta (w/o delta we essentially run on the dataset anyways, recalculating everything - so, it is not clear yet to me if there is a value in just pure retry w/o delta enabled). I guess I can just clearly make it a single flag like `delta_retry` so it is 100% clear that is just a subset of that for now.

dmpetrov:
> `retry` implemented here is a bit different - more on the "whole" query level, like delta mode.

It looks like you have two retires in mind 🙂  It might be better to avoid two different functionalities without a really good reason.


Also, after some discussion with ChatGPT... Retry name might not perfectly reflect the meaning of the option. Better options might be `partial`, `tolerant`, `soft_fail` or `resilient`? Like:
- `read_dataset(partial=True, partial_delta=True)`
- `read_dataset(soft_fail=True, soft_fail_delta=True)`

Naming idea for `ds.needs_retry()` --> `ds.is_complete()` or `ds.is_partial()` (is the best if we stick with partial=True)

shcheklein:
> It looks like you have two retires in mind 🙂 It might be better to avoid two different functionalities without a really good reason.

I don't think the second covers the first and other way around. But for this to answer exactly we need to even spec the second one better (with more real examples).

shcheklein:
> read_dataset(partial=True, partial_delta=True)
> read_dataset(soft_fail=True, soft_fail_delta=True)

hmm, no .. these names are not clear to me at all ...

shcheklein:
Renamed (it's a single param now) / also update the description to make it clear that it is an extension to delta atm (not implicit caching of results, not real-time retries, etc, etc).

dmpetrov:
It should not have manual exception handling - it makes the approach complicated for users.

This partial update should automate 2 steps (delata update implements the same steps):
1. Automatically identify that to re-compute
2. Do the recompute
3. Union the re-computed part to the original dataset

The delta update does all the 3 and API is pretty straight forward because it solves problem end-to-end. The current implementation solves only 2 and 3 and dictates on what user should do about 1 - it's complicated for users. We can clearly simplify it for users by automating 1.

I strongly recommend implementing the full solution.

shcheklein:
> It should not have manual exception handling - it makes the approach complicated for users

Yep, as I mentioned. `map`, `gen`, etc should be modified to catch exceptions and ignore records. It can be a default mode (not sure tbh) or set by some flag in settings. This `delta_retry` flag will cover handling for that case. In some cases people will use that, in some cases (to return precise and custom error message or when they have some advanced error handling) it will be explicit. Which is fine. This can be a followup (and should be a followup for ops reasons).

I'm not sure I understand everything else tbh in the last comment :( Was more an expansion on the same issue ^^. Then again, I fully agree, but I strongly recommend doing this as a followup.

dreadatour (approved):
Not sure about API and naming, code looks good to me 🔥

PR #1135, opened 2025-06-04
Fix pickling generated model in 'read_json'

dreadatour:
Kind of hack found [here](https://github.com/cloudpipe/cloudpickle/issues/206#issuecomment-555939172), but it works.

Would love to have a better solution.

Using [pickletools](https://docs.python.org/3/library/pickletools.html) to check:

<details>
  <summary>Before</summary>

```
    0: \x80 PROTO      5
    2: \x95 FRAME      77
   11: \x8c SHORT_BINUNICODE 'datachain.lib.meta_formats'
   39: \x94 MEMOIZE    (as 0)
   40: \x8c SHORT_BINUNICODE 'Modeljsonl272c131de3ba4cec9f751b5be2b7facb'
   84: \x94 MEMOIZE    (as 1)
   85: \x93 STACK_GLOBAL
   86: \x94 MEMOIZE    (as 2)
   87: .    STOP
```
</details>

<details>
  <summary>After</summary>

```
    0: \x80 PROTO      5
    2: \x95 FRAME      4019
   11: \x8c SHORT_BINUNICODE 'cloudpickle.cloudpickle'
   36: \x94 MEMOIZE    (as 0)
   37: \x8c SHORT_BINUNICODE '_make_skeleton_class'
   59: \x94 MEMOIZE    (as 1)
   60: \x93 STACK_GLOBAL
   61: \x94 MEMOIZE    (as 2)
   62: (    MARK
   63: \x8c     SHORT_BINUNICODE 'pydantic._internal._model_construction'
  103: \x94     MEMOIZE    (as 3)
  104: \x8c     SHORT_BINUNICODE 'ModelMetaclass'
  120: \x94     MEMOIZE    (as 4)
  121: \x93     STACK_GLOBAL
  122: \x94     MEMOIZE    (as 5)
  123: \x8c     SHORT_BINUNICODE 'Modeljsonl272c131de3ba4cec9f751b5be2b7facb'
  167: \x94     MEMOIZE    (as 6)
  168: \x8c     SHORT_BINUNICODE 'datachain.lib.meta_formats'
  196: \x94     MEMOIZE    (as 7)
  197: \x8c     SHORT_BINUNICODE 'UserModel'
  208: \x94     MEMOIZE    (as 8)
  209: \x93     STACK_GLOBAL
  210: \x94     MEMOIZE    (as 9)
  211: \x85     TUPLE1
  212: \x94     MEMOIZE    (as 10)
  213: }        EMPTY_DICT
[...]

PR #1136, opened 2025-06-04
Fix pickling generated model in 'read_json'

dreadatour:
Better fix for https://github.com/iterative/datachain/pull/1135

See also: https://github.com/cloudpipe/cloudpickle/issues/175

### DIFF
--- a/src/datachain/lib/dc/json.py
+++ b/src/datachain/lib/dc/json.py
@@ -1,18 +1,13 @@
 import os
 import os.path
 import re
-from typing import (
-    TYPE_CHECKING,
-    Optional,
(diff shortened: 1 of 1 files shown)


src/datachain/lib/dc/json.py
```diff
@@ -88,4 +83,7 @@ def jmespath_to_name(s: str):
     }
     # disable prefetch if nrows is set
     settings = {"prefetch": 0} if nrows else {}
+
+    cloudpickle.register_pickle_by_value(meta_formats)
```
dreadatour, line 87: This is the fix.

PR #1138, opened 2025-06-04
fix(sqlite): use dynamic chunksize for large inserts

shcheklein:
On some systems (even newer one like Ubuntu 24 LTS), sqlite is still old and has quite restrictive settings. We detect and set the chunksize for bulk inserts based on that value.


src/datachain/data_storage/sqlite.py
```diff
@@ -150,6 +150,10 @@ def _connect(
             db.execute("PRAGMA journal_mode = WAL")
             db.execute("PRAGMA synchronous = NORMAL")
             db.execute("PRAGMA case_sensitive_like = ON")
+            # On some old machines is it set to ~100 which breaks us pulling
+            # datasets from Studio. Trying to bump it up ... Added in Python 3.11
+            if hasattr(db, "setlimit"):
+                db.setlimit(sqlite3.SQLITE_LIMIT_FUNCTION_ARG, 1000)  # type: ignore [attr-defined]
```
ilongin: `SQL_LIMIT_FUNCTION_ARG` is not relevant here as it defines how much variables we can set into SQL function. What is relevant is `SQLITE_LIMIT_VARIABLE_NUMBER` as it defines what is the number of placeholders in single SQL insert.
Also, I wasn't been able to increase `SQLITE_LIMIT_VARIABLE_NUMBER` dynamically in the code which makes sense as it says it's compile time constant var anyway.

What we need to do is dynamically calculate `chunksize` based on number of columns in dataframe, example:
```python
sqlite_max_variables = conn.getlimit(sqlite3.SQLITE_LIMIT_VARIABLE_NUMBER)
num_columns = df.shape[1]
chunksize = sqlite_max_variables // num_columns
```

Here there is a full example to demonstrate this (note the line **39** which can be un-commented to generate error)
```python
import sqlite3
import pandas as pd
import numpy as np
import random
import string

# Constants
NUM_FIELDS = 40    # Number of columns
NUM_ROWS = 50000   # Number of rows

# Helper to generate random column names
def random_column_name(length=5):
    return ''.join(random.choices(string.ascii_lowercase, k=length))

def create_df(num_rows, num_fields):
    # Generate random column names
    column_names = [random_column_name() for _ in range(num_fields)]
    data = {
        col: np.random.randint(1, 100, size=num_rows) for col in column_names
    }
    return pd.DataFrame(data)


conn = sqlite3.connect('example.db')
conn.setlimit(sqlite3.SQLITE_LIMIT_VARIABLE_NUMBER, 1000000)  # this doesn't work!

# limit of how much variables can be set in one insert function
sqlite_max_variables = conn.getlimit(sqlite3.SQLITE_LIMIT_VARIABLE_NUMBER)

# create random df with exact number of fields and rows
df = create_df(NUM_ROWS, NUM_FIELDS)

num_columns = df.shape[1]

chunksize = sqlite_max_variables // num_columns

### UNCOMENT TO GENERATE MAX VARIABLE ERROR ###
# chunksize += 1
###############################################

print(
    f"max variables: {sqlite_max_variables}, "
    f"num columns: {num_columns}, "
    f"chunksize: {chunksize}"
)

df.to_sql(
    'random_data',
    conn,
    if_exists='replace',
    index=False,
    method='multi',
    chunksize=chunksize
)

result_df = pd.read_sql('SELECT * FROM random_data', conn)
print(result_df.head())

conn.close()
```

shcheklein: > SQL_LIMIT_FUNCTION_ARG is not relevant here as it defines how much variables we can set into SQL function.

it is fixing / reproducing the issue though

while `SQLITE_LIMIT_VARIABLE_NUMBER` is quite large already and doesn't affect much actual example

ilongin: Hm, I'm not sure how can that fix / reproduce the issue .., but regardless we should still calculate `chunksize` dynamically as in my comment since it's clear it can generate error if `SQLITE_LIMIT_VARIABLE_NUMBER` is not big enough (sometimes it can be quite small though e.g `999`). If it's high (like in my example - 250000) then `chunksize` can be bigger as well which will speed up the inserts so we have safety and performance increase in one

shcheklein: yes, in this PR chucksize is also dynamic, just based off the different variable (and it helps, probably because SQLITE_LIMIT_VARIABLE_NUMBER < sqlite_max_variables // num_columns)

so, I feel both are important - we need to expand the experiment to play with different values

src/datachain/data_storage/sqlite.py
```diff
@@ -231,13 +231,20 @@ def execute_str(self, sql: str, parameters=None) -> sqlite3.Cursor:
         return self.db.execute(sql, parameters)
 
     def insert_dataframe(self, table_name: str, df) -> int:
+        # Dynamically calculates chunksize by dividing max variable limit in a
+        # single SQL insert with number of columns in dataframe.
+        # This way we avoid error: sqlite3.OperationalError: too many SQL variables,
+        chunksize = (
+            self.db.getlimit(sqlite3.SQLITE_LIMIT_VARIABLE_NUMBER) // df.shape[1]
```
shcheklein: Should we add some limit (1000?) ... to not hit some other limits (size of the query)?

src/datachain/data_storage/sqlite.py
```diff
@@ -150,14 +152,21 @@ def _connect(
             db.execute("PRAGMA journal_mode = WAL")
             db.execute("PRAGMA synchronous = NORMAL")
             db.execute("PRAGMA case_sensitive_like = ON")
+
+            max_variable_number = 999  # minimum in old SQLite versions
+            for row in db.execute("PRAGMA compile_options;").fetchall():
+                option = row[0]
+                if option.startswith("MAX_VARIABLE_NUMBER"):
```
shcheklein: why startswith, not exact match?

src/datachain/data_storage/sqlite.py
```diff
@@ -150,14 +152,21 @@ def _connect(
             db.execute("PRAGMA journal_mode = WAL")
             db.execute("PRAGMA synchronous = NORMAL")
             db.execute("PRAGMA case_sensitive_like = ON")
+
+            max_variable_number = 999  # minimum in old SQLite versions
```
shcheklein, line 156: where does it come from? I mean 999 ... (see too small)

src/datachain/data_storage/sqlite.py
```diff
@@ -231,13 +243,18 @@ def execute_str(self, sql: str, parameters=None) -> sqlite3.Cursor:
         return self.db.execute(sql, parameters)
 
     def insert_dataframe(self, table_name: str, df) -> int:
+        # Dynamically calculates chunksize by dividing max variable limit in a
+        # single SQL insert with number of columns in dataframe.
+        # This way we avoid error: sqlite3.OperationalError: too many SQL variables,
+        chunksize = self.max_variable_number // df.shape[1]
```
shcheklein: can `df.shape[1]` be 0?

can result of this operation be 0? e.g. self.max_variable_number < df.shape[1]

src/datachain/data_storage/sqlite.py
ilongin: That's the lowest ever value that was present for that option. It was in SQLite versions lower than `3.32.0`.
But I think we will "catch" it always with that `PRAGMA` query below it

src/datachain/data_storage/sqlite.py
ilongin: I don't think `df.shape[1]` will ever be 0 as I think we cannot have completely empty datasets in Studio. I think at least ID will be present, but I will add check for it just in case.

Regarding other question, yes I think you are right. It's really really small possibility I think but still present. In that case we need to throw exception

src/datachain/data_storage/sqlite.py
ilongin: options are string in format `MAX_VARIABLE_NUMBER=250000` ... I added `=` in matched string as well.

src/datachain/data_storage/sqlite.py
shcheklein: Can we just fall back to 1 and in case (if that happens) of an empty ds just return?

shcheklein: We might need to create the table though if ds is empty ...

shcheklein: If that possible at all (not in Studio now, in general)

src/datachain/data_storage/sqlite.py
shcheklein: If we always catch it - let's remove and raise (it is better vs silently using such a small value)

src/datachain/data_storage/sqlite.py
shcheklein: Ah, I see

src/datachain/data_storage/sqlite.py
ilongin: Isn't it better to use that smaller number then to raise Exception? ... dataset will still be pulled, just slower.

shcheklein: Yeah, but it might become always slow silently in case we make mistake below ... better to be aware about this

shcheklein: If pragma exists everywhere / always - what is the issue with it then?

src/datachain/data_storage/sqlite.py
ilongin: I'm pretty sure it's not possible at all. Table must have at least one column. But  I will add default 1.

PR #1140, opened 2025-06-08
fix(hf): initialize kwargs in default ls_dir

shcheklein:
Fixes `ls_dir` default implementation. It used only in HF at the moment. HF supposed to be (and was covered by a test) - but it was disabled since it very unstable.


shcheklein:
Windows tests are unrelated and will be fixed upstream - https://github.com/achimnol/aiotools/pull/84

PR #1141, opened 2025-06-08
fix(hf): init with async set to false to always init loop

shcheklein:
W/o this we are getting `RuntimeError("Loop is not running")` from the `sync()` wrapper:

```python
if loop is None or loop.is_closed():
        raise RuntimeError("Loop is not running")
```

`loop` is None since we have this piece of in the `AsyncFileSystem` `init`:

```python
        if not asynchronous:
            self._loop = loop or get_loop()
        else:
            self._loop = None
```

and `asynchronous` flag is set in the `AsyncWrapper` this way:

```python
        if asynchronous is None:
            asynchronous = running_async()
```

So `running_async()` async defines if loop is created or not. And if it is not creted, `sync` doesn't work. In Jupyter `running_async()` returns True. So, all our code that does `sync` is breaking`. I'm not sure if there is a better fix for all of this tbh. cc @skshetry @martindurant

If this sounds reasonable, we can add `*args` to the  official wrapper code I think.

### DIFF
--- a/src/datachain/client/hf.py
+++ b/src/datachain/client/hf.py
@@ -15,6 +15,24 @@ class classproperty:  # noqa: N801
         return self.fget(owner)
 
 
+def _wrap_class(sync_fs_class):
+    """
+    Analog of `AsyncFileSystemWrapper.wrap_class` from fsspec, but sets
+    asynchronous to False by default. This is similar to other Async FS
+    we initialize. E.g. it means we don't break things in Jupyter where code
+    run in async.
+    """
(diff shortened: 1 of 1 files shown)


martindurant:
Indeed it's a problem. The question asked should perhaps not be "was this invoked in the context of an event loop", but "was the calling code a coroutine". This, given in python async code can call sync code but not the reverse.

jupyter is also it's own little box, since everything is normally sync, but called from async. That's why top-level `await`s can work, but the same code cannot be used in a python script.

dreadatour (approved):
🤔👀

PR #1142, opened 2025-06-08
fix(tar): prefetching parent for vfiles

shcheklein:
Fixes prefetch for Tar (and VFile in general) + some cleanup.

Tar processing is broken at the moment by default since prefetch attempts do download the VFile itself which breaks on asserts in downloader:

```python
async def _put_in_cache(self, file: "File", *, callback: "Callback" = None) -> None:
        assert not file.location
        if file.etag:
            etag = await self.get_current_etag(file)
            if file.etag != etag:
                raise FileNotFoundError(
                    f"Invalid etag for {file.source}/{file.path}: "
                    f"expected {file.etag}, got {etag}"
                )
        await self.cache.download(file, self, callback=callback)
```

mind: 

`assert not file.location`

### DIFF
--- a/src/datachain/lib/file.py
+++ b/src/datachain/lib/file.py
@@ -127,10 +127,7 @@ class TarVFile(VFile):
     @classmethod
     def open(cls, file: "File", location: list[dict]):
         """Stream file from tar archive based on location in archive."""
-        if len(location) > 1:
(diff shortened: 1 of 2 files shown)


dreadatour (approved):
Looks good to me 👍

PR #1144, opened 2025-06-10
Pass cluster id as a parameter in CLI jobs run

amritghimire:
This uses https://github.com/iterative/studio/pull/11698 to pass compute
cluster id to studio to create a job


docs/commands/job/run.md
```diff
@@ -24,6 +24,7 @@ This command runs a job in Studio using the specified query file. You can config
 * `--team TEAM` - Team to run job for (default: from config)
 * `--env-file ENV_FILE` - File with environment variables for the job
 * `--env ENV` - Environment variables in KEY=VALUE format
+* `--cluster-id CLUSTER_ID` - Compute cluster ID to run the job on
```
shcheklein, line 27: probably we should be accepting cluster name as well

amritghimire: Yeah, but right now, we don't have any name field in compute cluster table. Will update once we have it.

PR #1150, opened 2025-06-11
Introduce read_listing_dataset for optimized reading of listing dataset

amritghimire:
For the case of listing dataset, that starts with lst__ , we have fixed set of the columns and listing. For the same reason, we can skip the autoload with the engine that fetches the metadata from the warehouse. This skips a few connection to the warehouse and also it skips the whole get_table part to get the tables from the warehouse.

Taken from https://github.com/iterative/studio/pull/11694

Relates to https://github.com/iterative/studio/issues/11704


src/datachain/__init__.py
```diff
@@ -73,6 +74,7 @@
     "read_dataset",
     "read_hf",
     "read_json",
+    "read_listing_dataset",
```
dreadatour: 🤔 to be honest I am not sure this should be exposed to the `datachain` root init file.

src/datachain/lib/dc/__init__.py
```diff
@@ -27,6 +27,7 @@
     "read_dataset",
     "read_hf",
     "read_json",
+    "read_listing_dataset",
```
dreadatour: Also here 🤔

src/datachain/lib/dc/listings.py
```diff
@@ -3,19 +3,67 @@
...
+                    sa.Column("sys__id", UInt64()),
+                    sa.Column("sys__rand", UInt64()),
+                    sa.Column("file__source", String()),
+                    sa.Column("file__path", String()),
+                    sa.Column("file__size", Int64()),
+                    sa.Column("file__version", String()),
+                    sa.Column("file__etag", String()),
+                    sa.Column("file__is_latest", Boolean()),
+                    sa.Column("file__last_modified", DateTime()),
+                    sa.Column("file__location", JSON()),
```
dreadatour: I think this should belongs to the warehouse class. like `dataset_listing_columns` or something like that. Should we move it there? 🤔

src/datachain/lib/dc/listings.py
```diff
@@ -41,3 +89,74 @@ def listings(
...
+    """
+    # Configure and return a DataChain for reading listing dataset files
+    # Uses ReadOnlyQueryStep to avoid warehouse metadata lookups
+    from datachain.lib.dc import Sys
+    from datachain.lib.file import File
+
+    from .datachain import DataChain
+
+    if not name.startswith(LISTING_PREFIX):
+        name = LISTING_PREFIX + name
```
dreadatour, line 123: Add test for this case?

src/datachain/data_storage/schema.py
```diff
@@ -272,6 +281,19 @@ def sys_columns(cls):
             ),
         ]
 
+    @classmethod
+    def listing_columns(cls):
+        return [
+            sa.Column("file__source", String()),
```
shcheklein, line 287: Let's generate this from the File class (and cache as a cached property).

I would check if we have some helpers to get column names list like this cc @ilongin

src/datachain/lib/dc/listings.py
```diff
@@ -41,3 +80,75 @@ def listings(
         output={column: ListingInfo},
         **{column: catalog.listings()},  # type: ignore[arg-type]
     )
+
+
+def read_listing_dataset(
```
shcheklein, line 85: we don't want / need this to be a public method (it competes with a regular `read_storage`) 

let's think how we can meaningfully incorporate these patches into datachain so that it makes sense to end users

e.g. can we automatically identify when we need to read table meta and when not? can it be "lazy" on-demand

let's think how to do this properly please

amritghimire: > we don't want / need this to be a public method (it competes with a regular `read_storage`) 
> 

This is not exposed publicly in datachain main import path. We need to import it from the path.

> 
> let's think how we can meaningfully incorporate these patches into datachain so that it makes sense to end users
> 
> e.g. can we automatically identify when we need to read table meta and when not? can it be "lazy" on-demand
> 
I am not sure how we can achieve those. For example in our use case, we knew exactly what we were striving for and we didn't do any edit or update. I am not sure how complicated it could be. It can be a separate followup IMO.

shcheklein: It can be after the release, yes. But we need to do this. Most likely we need a completely different approach.

> This is not exposed publicly in datachain main import path. We need to import it from the path.

is it visible in docs? in autocomplete? - we need to remove it from there

Issue #1139, opened 2025-06-05
collect() should return values, not iterator


dmpetrov:
### Description

I see `list(xxxx.collect())` pattern all the times. We use to have `iterate()` for iterators which I do not see now.

- [ ] rename `collect()` to  `iterate(...)`
- [ ] implement a proper `collect()` that returns values
- [ ] (optionally) implement `__iter__()` using iterator()
     - it's convenient to have default iterator when params are not needed `for file in ds.read_storage(...): `

It's a breaking change and it's better to implement it sooner rather than later.

dmpetrov:
| Framework  | Streaming (out-of-memory)| In-Memory Access |
|------------|-----------------------------------------------------|-----------------------------------------------------------------------------------|
| **Spark**  | ✅ `toLocalIterator()`                              | ✅ `collect()` / `df.toPandas()["col"].tolist()`            |
| **Polars** | ✅ `collect(streaming=True)`                | ✅ `to_dicts()` / `df["col"].to_list()`             |
| **Daft**   |  ⚠️ `to_arrow_table().to_batches()`                  | ✅ `to_pandas().to_dicts()` / `to_pandas()["col"].tolist()`                      |
| **LanceDB**| ⚠️ `to_arrow().to_batches()`                       | ✅ `to_dicts()` / `to_arrow()["col"].to_pylist()`                                 |
| **DuckDB** | ⚠️ `pyarrow.RecordBatchReader`                     | ✅ `fetchall()` / `fetchnumpy()["col"].tolist()`                                  |

dmpetrov:
A more safer option is:
1. Depreciate `collect()` as well as `results()` (it's a flatened version of `list(collect(*))`). It will work with Warning for a couple of months before the removal.
2. Introduce `to_list()` as in-memory version (in addition to `to_pandas()["col"].tolist()`)
3. Introduce `to_iter()` as the streaming option

Note, both `to_list()` and `to_iter()` have the same params as our current `collect()` which works nicely with individual columns as well as a column collections.

PR #1156, opened 2025-06-19
defaults for delta update

### LINKED ISSUE #1155: Defaults for diff and delata
diff() and delata mode are file column specific and require good file defaults instead of None.

dmpetrov:
Improvements:
- set default columns for delta updates
- simplify docstring for delta
- simplify docstring for diff
- rename `diff()` to `file_diff()` and `compare()` to `diff()`

It closes #1155

### DIFF
--- a/src/datachain/lib/dc/datachain.py
+++ b/src/datachain/lib/dc/datachain.py
@@ -1495,7 +1495,7 @@ class DataChain:
         """Calculate differences between two chains.
 
         This method identifies records that are added, deleted, modified, or unchanged
-        between two chains. It adds a status column with values: A=added, D=deleted, 
+        between two chains. It adds a status column with values: A=added, D=deleted,
         M=modified, S=same.
 
         Parameters:


dmpetrov:
Can anyone explain why `Tests / check (pull_request)` fails?  🙂

ilongin:
@dmpetrov since we are changing the interface, maybe this was a good opportunity to rename those `DataChain` functions as we spoke in the past.
I think we come up with something like:
1. `DataChain.compare(...)` -> `DataChain.diff(...)`
2. `DataChain.diff(...)` -> `DataChain.file_diff(...)`

WDYT?

dmpetrov:
@ilongin good idea. Renamed and simplified the docstring.

dmpetrov:
@ilongin Studio tests failed. Could you please help me investigate this? 

I'm not even sure if it's result of this change or recent merge to main.

ilongin:
> @ilongin Studio tests failed. Could you please help me investigate this?
> 
> I'm not even sure if it's result of this change or recent merge to main.

It has to do with recent changes in main with namespaces. I will take a more detail look in the evening.
EDIT: tests are passing now

dmpetrov:
Another thought... with this significant change in API. Why don't we change `collect()` #1139 ? 🙂

dmpetrov:
> Why don't we change `collect()` #1139 ? 🙂

PR: #1159

src/datachain/lib/dc/datasets.py
```diff
@@ -53,41 +57,25 @@ def read_dataset(
...
-            the final dataset and have not been renamed.
-        delta_compare: A list of fields used to check if the same row has been modified
-            in the new version of the source.
-            If not defined, all fields except those defined in delta_on will be used.
-        delta_retry: Specifies retry behavior for delta processing. If a string,
-            it's the name of a field in the result dataset that indicates an error
-            when not None - records with errors will be reprocessed. If True,
-            records that exist in the source dataset but not in the result dataset
-            will be reprocessed.
+        delta: If True, only process new or changed files instead of reprocessing
```
shcheklein, line 60: (minor) this looks better, but less functional - as user I would prefer to have more information (vs something that looks good) tbh for (clearly) quite advanced and complicated functionality

shcheklein (approved):

src/datachain/lib/dc/datasets.py
dmpetrov: The original version it's really hard to read - too much technical details.
And it should not be advance functionality if we won't expose too much technicalities.

shcheklein: we can move some of this into user guide btw, and also please check if we need to update it as well (if delta_on is mentioned there at all in some form or diff, or compare) aslo an example in README 

just to double check

PR #1159, opened 2025-06-20
Deprecate collect() introduce to_iter() and to_list()

### LINKED ISSUE #1139: collect() should return values, not iterator

dmpetrov:
closes #1139

See for details https://github.com/iterative/datachain/issues/1139#issuecomment-2989273928

It also make DataChain iterable (__iter__). So, user can just run `for file, data in ds: ...` without `ds.to_iter()`

## Summary by Sourcery

Replace the legacy collect API with new iterator and list-based methods and deprecate collect to unify DataChain output flows

New Features:
- Introduce DataChain.to_iter to yield rows as an iterator
- Add DataChain.to_list to retrieve all rows as a list
- Add DataChain.to_values to extract a flat list from a single column
- Make DataChain iterable via __iter__ to use directly in for loops

Enhancements:
- Deprecate the collect() method with a warning and forward its behavior to to_iter
- Rename internal collect_flatten to _leaf_values and update downstream calls accordingly

Tests:
- Update test suite to use to_iter(), to_list(), and to_values() instead of collect() across unit and functional tests

### DIFF
--- a/src/datachain/lib/dc/datachain.py
+++ b/src/datachain/lib/dc/datachain.py
@@ -430,7 +430,7 @@ class DataChain:
 
         from datachain.lib.arrow import schema_to_output
 
-        json_values = list(self.limit(schema_sample_size).collect(col))
+        json_values = self.limit(schema_sample_size).to_list(col)
         json_dicts = [
             json.loads(json_value) if isinstance(json_value, str) else json_value
             for json_value in json_values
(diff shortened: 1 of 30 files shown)


dmpetrov:
PR is ready. But Studio tests fail again. I'd appreciate your help investigating this. I guess, it's broken in main, not in this PR.

shcheklein:
@dmpetrov seems this test can be fixed here ... it is a matter of order not being guaranteed it seems?

src/datachain/lib/dc/datachain.py
```diff
@@ -2378,3 +2405,59 @@ def chunk(self, index: int, total: int) -> "Self":
...
+
+    def to_list(self, *cols: str) -> list[Union[DataValue, tuple[DataValue, ...]]]:  # type: ignore[overload-overlap,misc]
+        """Returns a list of rows of values, optionally limited to the specified
+        columns.
+
+        Args:
+            *cols: Limit to the specified columns. By default, all columns are selected.
+
+        Returns:
+            list[DataType]: Returns a list of single items if a column is selected.
```
dreadatour: What do you think about using `flat=True` argument to return flat list instead of varying return format based on number of input arguments?

Like:
```py
dc.to_list("file") -> list[tuple[File]]
```

But:
```py
dc.to_list("file", flat=True) -> list[File]
```

Also can be useful to flatten the list in case of multiple columns:
```py
dc.to_list("file1", "file2" flat=True) -> list[File]
```

for example, getting multiple columns as a flat list:

```py
files = dc.to_list("file1.path", "file2.path")
print(files)
```

Same suggestion for `to_iter` method.

A little bit more verbose, but more explicit and predictable.

dreadatour (approved):
Looks good to me! Good improvement IMO.

src/datachain/lib/dc/datachain.py
dmpetrov: > Also can be useful to flatten the list in case of multiple columns:

Flat makes sense only for a single column.

> ```python
> dc.to_list("file1", "file2" flat=True) -> list[File]
> ```

Two columns should return `[[f1, f2], [f3, f4], ...]` - flatness does not makes sense for 2+ params.

So, flat=True would work only for 1 param when it's needed in 100% cases. It's better to have that implicitly.

dmpetrov: I tried this - yes, it's a cleaner approach. I'll change the to_iter/to_list to always return list of list. flat=True is a syntax sugar that works only for 1 params - I'll play with it and see if this is needed.

THank for the comment!

dreadatour: Just my two cents’ worth of thoughts on this 😅

> Flat makes sense only for a single column.

This pitfall has caught me a couple of times during development and debugging, when I changed the number of parameters in the `collect` method, the script failed because it returned a different response format (list of signals instead list of list of signals). I needed to update the script accordingly and then update it back again.

> So, flat=True would work only for 1 param when it's needed in 100% cases. It's better to have that implicitly.

A lot of libraries have `flatten` implementation, like [numpy](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.flatten.html), [pytorch](https://docs.pytorch.org/docs/stable/generated/torch.flatten.html#torch.flatten), [tensorflow (reshape)](https://www.tensorflow.org/api_docs/python/tf/reshape), etc.

dmpetrov: fyi: I'm introducing `to_values(col: str)` instead of flat=True. 

Reasons:
- it's impossible to generalize variadic args `cols` with `flat`: `def to_list(self, *cols: str, flat: bool=False)` - it mixes up flat with cols. I've spend a couple hours figuring this out - no proper solution.
- to_values() looks better and shorter in code. to_list() is also simpler - returns only list of tuples.

dmpetrov:
fyi: I also renamed an internal `collect_flatten()` to `_leaf_values()`. It not suppose to be public facing method and mostly used inside to_records() and such.

dmpetrov:
what the heck is sourcery-ai? 🙂

PR #1160, opened 2025-06-20
fix(hf): fix fs wrapper to handle write

shcheklein:
Fixes HF filesystem (writes were failing silently), also updates and fixes HF failing example tests.

I'm not 100% sure what was the intention to skip `open` in the original `fsspec` wrapper. This fixes it (otherwise all writes were silently ignored pretty much). I'll follow up with fsspec's maintainers to see if it should be fixed upstream instead.

Also, update example to fix a bunch of other issues in the pipeline.

### DIFF
--- a/src/datachain/client/hf.py
+++ b/src/datachain/client/hf.py
@@ -21,6 +21,9 @@ def _wrap_class(sync_fs_class):
     asynchronous to False by default. This is similar to other Async FS
     we initialize. E.g. it means we don't break things in Jupyter where code
     run in async.
+
+    This also fixes write operations by ensuring they are properly forwarded
+    to the underlying filesystem without async buffering issues.
     """
     from fsspec.implementations.asyn_wrapper import AsyncFileSystemWrapper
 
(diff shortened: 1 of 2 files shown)


src/datachain/client/hf.py
```diff
@@ -29,6 +32,19 @@ def __init__(self, *args, **kwargs):
             sync_fs = sync_fs_class(*args, **kwargs)
             super().__init__(sync_fs, asynchronous=False)
 
+        def open(self, path, mode="rb", **kwargs):
```
shcheklein, line 35: @martindurant could you PTAL. Should we port it back (along with `asynchronous=False`) to the `GeneratedAsyncFileSystemWrapper` inside fsspec?

We avoid wrapping `open`:

```python
        excluded_methods = {"open"}
        for method_name in dir(self.sync_fs):
            if method_name.startswith("_") or method_name in excluded_methods:
                continue
...
```

Means we hit `_open` in the `AbstractFileSystem` and use `AbstractBufferedFile`  which doesn't implement _upload_chunk (and doesn't raise `NotImplemented`) -> writes were failing silently.

Unless I'm missing something, should we just redirect open to sync_fs.open?

Also,  should we raise in the AbstractFile if someone is trying to use it?

dreadatour (approved):
Looks good to me 👍

src/datachain/client/hf.py
martindurant: You should check that the method is a couroutine-function. Some methods are simply internal.

shcheklein: Hmm, I'm not sure I understand tbh. Can `open` be internal? Could you please give a bit more color / explanation / example?

martindurant: I believe you are trying to find methods here which have both async variants with a "_" and sync variants without. These would the the same set as can be automatically generated in https://github.com/fsspec/filesystem_spec/blob/master/fsspec/asyn.py#L928 ; the code for the reverse is what you are referencing ( https://github.com/fsspec/filesystem_spec/blob/master/fsspec/implementations/asyn_wrapper.py#L76 ) is the inverse of that.

All this is to say: yes, _open can and should be "internal", a method which is not a coroutine and therefore should not be async-wrapped. So you did the right thing! I think you should be able to trigger this in upstream fsspec, so yes a fix there would be appreciated.

Note that I'm not certain how much sense it makes to open() on an async filesystem at all, given that the file-like API is only blocking functions...

PR #1163, opened 2025-06-21
Added `move_dataset` function to be able to move datasets to different project

ilongin:
TODO


src/datachain/data_storage/metastore.py
```diff
@@ -1008,6 +1030,11 @@ def update_dataset(
                 else:
                     values[field] = json.dumps(value)
                     dataset_values[field] = DatasetRecord.parse_schema(value)
+            elif field == "project_id":
+                if not value:
+                    raise ValueError("Cannot set empty project_id for dataset")
+                dataset_values["project"] = self.get_project_by_id(value)
```
shcheklein, line 1036: is it needed only to return the new value? 

since (if) we have to do a call, let's do it at the end and just fetch the new record?

src/datachain/data_storage/warehouse.py
```diff
@@ -373,6 +374,24 @@ def rename_dataset_table(
 
         self.db.rename_table(old_ds_table_name, new_ds_table_name)
 
+    def rename_dataset_tables(
```
shcheklein, line 359: can I move only a particular version? why not?

shcheklein (commented):
A few important questions:

- transactional semantics - what happens if it fails in the middle, e.g. we are renaming tables 50 out of 100 succeeded - what happens? Will people be able to run it again?
- what happens if I in a single session:
  1. created a dataset
  2. renamed it
  3. sessions failed with an exception

will the created dataset be deleted (we have a special logic that handles this - will it handle renames?)


(same btw with deletions - are we going to restore them?)

- do we need lock the operation if other jobs are running that are accessing this dataset?

src/datachain/data_storage/metastore.py
ilongin: Yes, it's just to return latest / newest updated version of `DatasetRecord` without need to fetch it from DB again. 
If you meant on fetching new dataset record at the end, I think current implementation is better / faster as dataset record is much much bigger object and requires 4 joins to be constructed so fetching just a project is better from performance perspective.

src/datachain/data_storage/warehouse.py
ilongin: This task was about moving whole dataset from project to project. What you suggest is moving particular version across different datasets. We can create separate issue for that if it's really needed?

ilongin:
> A few important questions:

Good questions. So we were speaking about this transactional issues before but somehow it was not really a priority. 
Anyway, I think there are 2 problems:
1. Not using DB transactions for metastore where we can -> for this we also probably need to be able to start transaction above metastore, e.g in `Catalog` class but currently that is not possible - we need to refactor code and analyze where transactions are actually missed - maybe there is not much that's missed.
2. Multi DB transactions -> this would need to be implemented in code which doesn't solve main issue - what if process just dies...unless we have really complex mechanism which is able to re-continue after process comes live again


Regarding your questions:
> * transactional semantics - what happens if it fails in the middle, e.g. we are renaming tables 50 out of 100 succeeded - what happens? Will people be able to run it again?

We first update metadata and then we start to rename warehouse tables. If it fails in the middle, those tables that are renamed will be usable, and those that are not will be in some limbo state. If user starts to use dataset version for which table is not renamed error will be thrown. It's impossible to run it again without some manual intervention. BTW this could all be much easier if we can just use `uuid` in dataset table names instead, and actually we had that before but then requirement were to have human readable names.

> * what happens if I in a single session:
>   
>   1. created a dataset
>   2. renamed it
>   3. sessions failed with an exception
> 
> will the created dataset be deleted (we have a special logic that handles this - will it handle renames?)

What do you mean by session? Our `datachain.query.session.Session` object? All those operations are separate from DB perspective, i.e. there is no single high level transaction or any kind of that. So in you example dataset will be created and if something fails when updating dataset tables in warehouse the same scenario from above example will happen so no, dataset won't be deleted.

> (same btw with deletions - are we going to restore them?)
> 

Deletions are also not under transaction, but situation is little bit better as dataset will just stay present with less number of dataset versions (those that are not removed). The worst that can happen is that some dataset table in warehouse is not removed and it's metadata is which means it will hang there forever, but it's easy to create some job cleanup of hanging dataset tables which are not attached to any metadata.

> * do we need lock the operation if other jobs are running that are accessing this dataset?

Yes, probably something like that needs to happen, but we need to brainstorm about it. These are all general questions which needs to be addressed in some separate issue IMO as it's all around the codebase - at the time we didn't want to deal with it to not loose time but maybe now time has come to refactor everything.

src/datachain/data_storage/warehouse.py
shcheklein: what is the difference? why not allow this right away? E.g. I want to make only a particular version of the dataset a production version? Is it only about accepting (limiting the operation to version?)

(probably we need also duplicate operation)

tests/func/test_datasets.py
```diff
@@ -391,8 +383,37 @@ def test_edit_dataset(cloud_test_catalog, dogs_dataset):
...
+    )
+    assert expected_table_row_count
+    assert dataset.get_version("1.0.0").num_objects == expected_table_row_count
+
+
+def test_move_dataset(cloud_test_catalog, dogs_dataset, project):
+    session = cloud_test_catalog.session
+    catalog = cloud_test_catalog.catalog
+
+    dc.move_dataset(
```
shcheklein: I thought we wanted to do `datasets.move` ?

tests/func/test_datasets.py
```diff
@@ -391,8 +383,37 @@ def test_edit_dataset(cloud_test_catalog, dogs_dataset):
...
+        dogs_dataset.name,
+        namespace=dogs_dataset.project.namespace.name,
+        project=dogs_dataset.project.name,
+        new_namespace=project.namespace.name,
+        new_project=project.name,
+        session=session,
+    )
+
+    dataset = catalog.get_dataset(dogs_dataset.name, project)
```
shcheklein: let's use `dc.read_dataset` instead of an internal calls

tests/func/test_datasets.py
```diff
@@ -391,8 +383,37 @@ def test_edit_dataset(cloud_test_catalog, dogs_dataset):
...
-    expected_table_row_count = get_table_row_count(
+    assert table_row_count(catalog.warehouse.db, old_dataset_table_name) is None
+    expected_table_row_count = table_row_count(
+        catalog.warehouse.db, new_dataset_table_name
+    )
+    assert expected_table_row_count
+    assert dataset.get_version("1.0.0").num_objects == expected_table_row_count
+
+
+def test_move_dataset(cloud_test_catalog, dogs_dataset, project):
```
shcheklein: let's add a test / test when next query after the move fails (e.g. some UDF and it breaks)

tests/unit/lib/test_datachain.py
```diff
@@ -3571,3 +3579,99 @@ def test_save_create_project_not_allowed(test_session, allow_create_project):
...
+
+@pytest.mark.parametrize(
+    "old_namespace_name,old_project_name,new_namespace_name,new_project_name",
+    [
+        ("old", "old", "new", "new"),
+        ("old", "old", "old", "new"),
+        ("old", "old", "old", "old"),
+    ],
+)
+def test_move_dataset(
```
shcheklein: it doesn't look like a unit test, what is the difference with the func test above?

tests/unit/lib/test_datachain.py
```diff
@@ -3571,3 +3579,99 @@ def test_save_create_project_not_allowed(test_session, allow_create_project):
...
+        new_table_name = catalog.warehouse.dataset_table_name(dataset_updated, version)
+        if old_project == new_project:
+            assert old_table_name == new_table_name
+        else:
+            assert table_row_count(catalog.warehouse.db, old_table_name) is None
+
+        assert table_row_count(catalog.warehouse.db, new_table_name) == 3
+
+
+def test_move_dataset_wrong_old_project(test_session, project):
```
shcheklein: this is not unit test, this is regular func test

src/datachain/data_storage/warehouse.py
dmpetrov: In my impression it was about moving versions 🙂  

The use case I have in mind - I was massaging `my_ds`, created 13 versions and I'd like to "promote" this 13th to the latest major version of `prod/actions/animal_planet` which might be, let say animal_planet@4.2.2

I need something like dc.datasets.move("my_ds", "prod.actions.animal_planet") to create animal_planet@5.0.0

Note, there is an assumption that it increases major version by default.

shcheklein:
> So we were speaking about this transactional issues before but somehow it was not really a priority.

we are going into more complicated operations I think - we didn't have much happening before

> Multi DB transactions

we should try to design things in way that doesn't require that (implementation is too complicated). E.g. don't rename tables - use UUIDs as names, etc, etc

> Not using DB transactions for metastore where we can 

we can really try to structure code / keep it simple so that it is also not required or everything is happening within a single query

>  BTW this could all be much easier if we can just use uuid in dataset table names instead, and actually we had that before but then requirement were to have human readable names.

yes, exactly. And in general on UUID more (vs names, or internal DB ids like ids from RDS)

src/datachain/data_storage/warehouse.py
dmpetrov: Moving the whole dataset seems like a separate operation. We do need it but I'm not sure it's common.

shcheklein: I think we need both. Moving the whole dataset is much needed - just to rename, reorganize things.

My2cs: I feel the promotion scenario is much needed, but will more niche.

> Note, there is an assumption that it increases major version by default.

this seems too complicated, seems like a different operation (to rename / move that we talked about in the first place) 🤔

tests/func/test_datasets.py
ilongin: We still have all the methods in the root, like `dc.create_dataset(...)`. I wanted to reafactor whole public API (together with other objects like `projects`) in separate issue to have consistent API on all fronts... here, I've created an issue (and there are some open questions there right away) https://github.com/iterative/datachain/issues/1197

dmpetrov:
> > BTW this could all be much easier if we can just use uuid in dataset table names instead, and actually we had that before but then requirement were to have human readable names.
> 
> yes, exactly. And in general on UUID more (vs names, or internal DB ids like ids from RDS)

UUIDs can defiantly improve robustness. But it feels like it's a part of a bigger issue. In pure SQL it's nearly impossible to solve - you need extensions like T-SQL. So, this will need to support it in our code "manually".

Using UUIDs can simplify our code. But it's better to keep dataset tables with transparent naming when people can navigate through dataset and dataset versions in DB. I personally look up to the dataset tables pretty often. Additional level of indirection will ruin this experience. In any other tables - that's not problem they are not as important as data tables.

tests/unit/lib/test_datachain.py
ilongin: Sorry, I forgot to move these tests to correct place. The func one is duplicate but these from `test_datachain.py` should've been moved to `test_datasets.py`. Doing it now

tests/func/test_datasets.py
ilongin: What exactly should be tested by this?

shcheklein: that it doesn't fail at least with some weird errors ... that it either keep rename in place (doesn't roll it back) or rolls it back properly ... etc

I would test it also in things like:

-> rename happens
-> some code that save() into the new dataset and bumps the version
-> then some code that fails after 

so, `save()` is rolled back by our rules, 

we just need to sure that everything else works fine

src/datachain/data_storage/warehouse.py
ilongin: I think this signature should cover every possible use case:
```python
def move_dataset(
    namespace: str,
    project: str,
    name: str,
    version: Optional[str] = None,
    dest_namespace: Optional[str] = None,
    dest_project: Optional[str] = None,
    dest_name: Optional[str] = None,
    dest_version: Optional[str] = None,
    dest_update_version: Optional[str] = "patch",
    session: Optional[Session] = None
) -> None:
```

Examples:
```python
dc.move_dataset("dev", "analytics", "metrics", dest_name="final_metrics")  # moves whole dataset in the same project, basically just renames it
dc.move_dataset("dev", "analytics", "metrics", dest_namespace="prod")  # moves whole dataset into new namespace (and project with the same name) prod.analytics.metrics
dc.move_dataset("dev", "analytics", "metrics", dest_project="new-project", dest_name="final_metrics")  # moves whole dataset into new project under different name
dc.move_dataset("dev", "analytics", "metrics", version="2.0.1",  dest_name="final_metrics", dest_version="4.0.0")  # moves specific version 2.0.1 to new dataset in the same project by version 4.0.0
dc.move_dataset("dev", "analytics", "metrics", version="2.0.1",  dest_name="final_metrics", dest_update_version="major")  # moves specific version to new dataset in the same project as new major version
...
```

Alternative is to have different methods to move whole dataset and specific dataset version but that's too complicated IMO

tests/func/test_datasets.py
ilongin: I've added a test that saves to newly moved  dataset and tries to save to old one (fails) and I think that's pretty much what is possible to test.

shcheklein: it's not the same, I'm trying to test if UDF that has some successful saves, then fails (and has to cleanup some tables) - just to make sure that cleanup / rollback mechanism plays fine with the move operation

ilongin: Can you please write simple example of that as I really don't understand 100%. Move is happening before running chain and doesn't affect it. The only thing that can happen is that moving doesn't work properly (e.g dataset table is not renamed correctly etc) and then all chains using that dataset will fail afterwards. Regarding rollback, we just have cleaning up temp tables that are created for UDF and it's not connected to moving dataset that happened before at all...so I wouldn't even call that a rollback as we are not returning to any previous state.

src/datachain/data_storage/warehouse.py
ilongin: @shcheklein @dmpetrov what do you think about above?

shcheklein: this signature is fine for API, but looks too verbose and complicated for users.

Can I do something like:

`dc.move_dataset("dogs", "cats")` # should rename in local all versions
`dc.move_dataset("dogs", "cats")` # in Studio if default project is set to "global.dev" should rename it within that project

`dc.move_dataset("dogs", "test.test.cats")` # in Studio if default project is set to "global.dev" should rename `global.dev.dogs` to "test.test.cats" (all versions). Also create test.test if it doesn't exist

Version:

Ideally we need to support @1.2.3 syntax I think ... we can also do:

a single param `version` that accepts:

- just a single version (e.g. "1.0.0" - then we just move it) # This will be by far the most useful thing
- a list of versions [....] - then we move all of them as is 
- a list of tuples (map version to version) - we rename versions # This is fine to keep it complicated, most people won't use it. We can also include second param - dest_version in this case.


> dest_update_version

can we come up with a better name for this ? looks bad and complicated now

tests/func/test_datasets.py
shcheklein: > Move is happening before running chain and doesn't affect it.

doesn't use Session at all?

what happens if I do:

 - chain first that saves a dataset `test`
 - then I move it to `test2`
 - then I do chain that fails

Session will try to delete the `test` dataset (that doesn't exist anymore?)

I talk about scenarios like this (I'm not sure about movements right before - it was just to check that we have some reasonable semantics - e.g. we keep movements not matter what happens with the session), does it make sense?

shcheklein:
> I personally look up to the dataset tables pretty often. 

out of curiosity - what is your reason to do this? is it a sign that DataChain is not mature enough?

> But it feels like it's a part of a bigger issue. In pure SQL it's nearly impossible to solve - you need extensions like T-SQL. So, this will need to support it in our code "manually".

not sure I understand this - could you please clarify?

tests/func/test_datasets.py
ilongin: I totally forgot that we actually remove created datasets in session cleanup if exception happened. Anyway, in this situation moved dataset will stay in another namespace / project i.e it won't be removed. As you mentioned, session will try to remove `test` dataset but since it doesn't exist any more, nothing will happen.
If we want to change this and improve other "transactional" use cases I suggest to open a new issue because this affect other scenarios like updating dataset, removing it etc. Move dataset falls into the same scenario as updating dataset.

src/datachain/data_storage/warehouse.py
ilongin: Yes, this was other alternative I was thinking about as well. Ok, we can do that - to expect full dataset names together with version(in a followup). For now, since this issue is only about moving whole dataset from one project to another, signature will be:

```python
def move_dataset(
    name: str,
    destination_name: str,
    session: Optional[Session] = None
) -> None:
```

version part in name is not allowed yet, and special `version` parameter doesn't exist - this will be added when we do move of specific dataset version (separate issue)

tests/func/test_datasets.py
shcheklein: > Move dataset falls into the same scenario as updating dataset.

did / do we have update dataset? (was it for renaming)?

> If we want to change this and improve other "transactional" use cases I suggest to open a new issue because this affect other scenarios like updating dataset, removing it etc

yep, please create an issue for this - let's discuss what we can do (use UUID not names to track datasets in Session?)

ilongin: I was wrong, dataset is actually removed becasue we do it with dataset `id` and move just updates existing dataset (`name`, `project_id` etc) so dataset is removed, even though it's moved to another project. I've added test for this case.

ilongin: `id` or `uuid` is basically the same as long we are not using name + version.

src/datachain/data_storage/warehouse.py
dmpetrov: I'd suggest using more :
- `name`--> `src` (or `src_name`)
- `destination_name` --> `dest`(or `dest_name`)

It's better to use just src/dest.

shcheklein:
@ilongin tests fail

tests/func/test_datasets.py
```diff
@@ -391,14 +384,108 @@ def test_edit_dataset(cloud_test_catalog, dogs_dataset):
...
+
+    if old_name != new_name:
+        # check that old dataset doesn't exist any more
+        with pytest.raises(DatasetNotFoundError):
+            dc.read_dataset(old_name).save("wrong")
+
+    dataset_updated = dc.read_dataset(new_name).dataset
+
+    # check if dataset tables are renamed correctly as well
+    for version in [v.version for v in dataset.versions]:
```
shcheklein, line 430: why do we have to test this? (if we can read the new dataset, and not the old one with datachain - that shoould be enough) - let's try to avoid asserting on internals as much as possible, especially in func tests

PR #1170, opened 2025-06-24
Make datachain job run fail (non zero exit) if job fails in Studio or canceled

### LINKED ISSUE #1169: Make datachain job run fail (non zero exit) if job fails in Studio or canceled
When we run the job in studio, make it return status code 1 when the job fails.

amritghimire:
This returns the status code 1 if the job is in failed or canceled
status at the end.

Closes #1169

## Summary by Sourcery

Make the `datachain job run` command return a non-zero exit code when the associated Studio job fails or is canceled by returning the job status from the logging helper and using it in the CLI dispatcher.

New Features:
- Return non-zero exit code when a Studio job ends in FAILED or CANCELLED status

Enhancements:
- Propagate job terminal status from show_logs_from_client back to the CLI as exit codes
- Update CLI command dispatcher to use handler return values for exit codes instead of always returning zero

Tests:
- Add end-to-end test to verify non-zero exit code for failed Studio jobs

### DIFF
--- a/src/datachain/studio.py
+++ b/src/datachain/studio.py
@@ -290,7 +290,11 @@ def show_logs_from_client(client, job_id):
     else:
         print("\n\nNo dataset versions created during the job.")
 
-    return 1 if latest_status in ["FAILED", "CANCELLED"] else 0
+    exit_code_by_status = {
+        "FAILED": 1,
+        "CANCELLED": 2,
+    }
+    return exit_code_by_status.get(latest_status.upper(), 0) if latest_status else 0
 
 
 def create_job(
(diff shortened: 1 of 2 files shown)


src/datachain/studio.py
```diff
@@ -287,6 +290,8 @@ async def _run():
     else:
         print("\n\nNo dataset versions created during the job.")
 
+    return 1 if latest_status in ["FAILED", "CANCELLED"] else 0
```
shcheklein: Let's return different codes here: 1 and 2

tests/test_cli_studio.py
```diff
@@ -440,3 +440,60 @@ def test_studio_run(capsys, mocker, tmp_dir):
         "priority": 5,
         "compute_cluster_name": "default",
     }
+
+
+def test_studio_run_non_zero_exit_code(capsys, mocker, tmp_dir):
```
shcheklein: we can parametrize and test FAILED, CANCELED with different exit codes - 1 and 2

PR #1174, opened 2025-06-24
Removing creating local namespace / project by default

ilongin:
Before we would create default namespace and project which would cause issues in Studio as there is default project for each user. This PR is removing that and creating default project / namespace on the fly when needed.

### DIFF
--- a/src/datachain/data_storage/metastore.py
+++ b/src/datachain/data_storage/metastore.py
@@ -176,7 +176,9 @@ class AbstractMetastore(ABC, Serializable):
 
     @cached_property
     def default_project(self) -> Project:
-        return self.get_project(self.default_project_name, self.default_namespace_name)
+        return self.get_project(
+            self.default_project_name, self.default_namespace_name, create=True
+        )
 
     @cached_property
     def listing_project(self) -> Project:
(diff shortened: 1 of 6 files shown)


src/datachain/lib/dc/datasets.py
```diff
@@ -136,9 +140,15 @@ def read_dataset(
             # all 2.* dataset versions). If dataset doesn't have any versions where
             # major part is equal to that input, exception is thrown.
             major = int(version)
-            dataset = session.catalog.get_dataset(
-                name, get_project(project_name, namespace_name, session=session)
-            )
+            try:
+                ds_project = get_project(project_name, namespace_name, session=session)
+            except ProjectNotFoundError:
```
shcheklein, line 145: Q: may be keep it ProjectNotFound? To make it more explicit to users - that there is no project (e.g. they made a mistake in its name)

shcheklein (approved):

src/datachain/lib/dc/datasets.py
ilongin: I would leave `DatasetNotFoundError` as that is the main cause of the issue. Maybe project is missing because it's not created yet (default one for example) but that's just a side-effect of missing dataset so I don't think we should point user into wrong direction.

PR #1177, opened 2025-06-25
Namespaces and local db migration docs

ilongin:
## Summary by Sourcery

Add a new Namespaces guide to the user documentation and update the main index to include it

Documentation:
- Introduce a namespaces.md guide detailing dataset organization with namespaces and projects, defaults for Studio and CLI, project creation, and dataset saving/reading workflows
- Update docs/guide/index.md to link the new Namespaces section

### DIFF
--- a/docs/guide/db_migrations.md
+++ b/docs/guide/db_migrations.md
@@ -25,7 +25,7 @@ The recommended fix is to **delete the local database** and let the CLI recreate
 Before deleting the file, we strongly recommend making a backup of your current database:
 
 ```bash
-cp ~/.datachain/db ~/.datachain/db.backup
+cp .datachain/db .datachain/db.backup
 ```
 
 This allows you to recover data manually if needed later.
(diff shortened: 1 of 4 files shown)


shcheklein (approved):
Amazing! thanks

yathomasi (approved):
Looks great.

I see we have separate page regarding db migration, I think this should go together with namespaces as it's only namespaces related or at least we can add a link from namespaces page.

I think we can add new `API` on the `Datachain` API reference as well.

May not be important but I noticed  attr `namespace_name` . Didn't we migrated to `namespace` or it's unrelated?
Screenshot 2025-06-25 at 20 43 22

ilongin:
> Looks great.
> 
> I see we have separate page regarding db migration, I think this should go together with namespaces as it's only namespaces related or at least we can add a link from namespaces page.
> 
> I think we can add new `API` on the `Datachain` API reference as well.
> 
> May not be important but I noticed attr `namespace_name` . Didn't we migrated to `namespace` or it's unrelated? Screenshot 2025-06-25 at 20 43 22

I will look into the API reference in follow-up. 
Regarding separate page for db migration, it's correct to split it from namespaces as that is general doc about CLI DB migrations and how to handle them. In future there will be similar updates like namespaces.

PR #1178, opened 2025-06-26
feat(datasets): accept pep version spec in read_dataset

shcheklein:
Add a way to specify version spec in `read_dataset` to find and match the most recent compatible version.

Among other tings:

- I've dropped `fallback_to_studio` flag as part of this PR (because I needed to touch how we fetch and resolve dataset versions against Studio)
- Added a `read_dataset(update=True)` to always try to get the latest version from Studio that satisfies the version spec

## TODO:

- [x] Add for local dataset version resolving
- [x] Add more tests for `update=True` and remote dataset version resolving
- [x] Update docs

### DIFF
--- a/src/datachain/lib/dc/datasets.py
+++ b/src/datachain/lib/dc/datasets.py
@@ -77,8 +77,10 @@ def read_dataset(
               (error mode)
             - True: Reprocess records missing from the result dataset (missing mode)
             - None: No retry processing (default)
-        update: If True, it checks updates for the updates of the dataset on the Studio
-            side.
+        update: If True always checks for newer versions available on Studio, even if
+            some version of the dataset exists locally already. If False (default), it
+            will only fetch the dataset from Studio if it is not found locally.
+
 
     Example:
         ```py
@@ -103,6 +105,11 @@ def read_dataset(
         ```py
         # Legacy integer version support (finds latest in major version)
         chain = dc.read_dataset("my_cats", version=1)  # Latest 1.x.x version
+        ```
+
+        ```py
+        # Always check for newer versions matching a version specifier from Studio
+        chain = dc.read_dataset("my_cats", version=">=1.0.0", update=True)
         ```
 
         ```py


src/datachain/lib/dc/datasets.py
```diff
@@ -92,11 +92,17 @@ def read_dataset(
         ```
 
         ```py
-        chain = dc.read_dataset("my_cats", fallback_to_studio=False)
+        chain = dc.read_dataset("my_cats", version="1.0.0")
```
dmpetrov, line 97: why don't we make this a part of dataset name? like `my_cats@1.0.0` or `package_name>=1.0,<2.0` 

So, we can give up a whole parameter from almost every API call 🙂

src/datachain/lib/dc/datasets.py
```diff
@@ -79,6 +77,8 @@ def read_dataset(
               (error mode)
             - True: Reprocess records missing from the result dataset (missing mode)
             - None: No retry processing (default)
+        update: If True, it checks updates for the updates of the dataset on the Studio
```
ilongin: I'm not sure I understand this docs. And also what should `update` flag do in the first place?

ilongin (changes requested):

src/datachain/lib/dc/datasets.py
shcheklein: it will just check if there is a newer version of a dataset available on Studio (even if there is already a cached one)

shcheklein: does it make sense?

src/datachain/lib/dc/datasets.py
```diff
@@ -113,14 +119,15 @@ def read_dataset(
             version="1.0.0",
             session=session,
             settings=settings,
-            fallback_to_studio=True,
```
dmpetrov, line 116: that's nice!

src/datachain/lib/dc/datasets.py
ilongin: Hm...I would expect it to work the same as for local datasets, which means it will always try to pull / use the latest version. If user really want's to use the version that is cached (maybe to save time on pulling?), it can set the specific version of it.  Caching dataset is just optimization IMO and I feel like we shouldn't have specific flags dealing with it explicitly. WDYT?

shcheklein: > If user really want's to use the version that is cached (maybe to save time on pulling?), it can set the specific version of it

I think it can be annoying - it is an expensive operation ...

and I think it is how actually current implementation works - it doesn't check for updates I think

dmpetrov (approved):
LGTM - but I checked only users API, not implementation.
Approving to not to block anyone

src/datachain/lib/dc/datasets.py
ilongin: Yes, current implementation doesn't check for updates. Anyway, I would at least update this docs for `update` field as it looks strange, speically this part `it checks updates for the updates`

ilongin (approved):

src/datachain/lib/dc/datasets.py
shcheklein: yes, I'll update it

PR #1181, opened 2025-06-26
Fix creating default project

ilongin:
Fixing issue on creating missing default project on empty DB.

### DIFF
--- a/src/datachain/data_storage/metastore.py
+++ b/src/datachain/data_storage/metastore.py
@@ -799,12 +799,29 @@ class AbstractDBMetastore(AbstractMetastore):
 
         return self.get_project(name, namespace.name)
 
+    def _is_listing_project(self, project_name, namespace_name) -> bool:
+        return (
+            project_name == self.listing_project_name
+            and namespace_name == self.system_namespace_name
+        )
+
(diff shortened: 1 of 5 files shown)


src/datachain/data_storage/metastore.py
```diff
@@ -799,12 +799,29 @@ def create_project(
 
         return self.get_project(name, namespace.name)
 
+    def _is_listing_project(self, project_name, namespace_name) -> bool:
```
dreadatour:
```suggestion
    def _is_listing_project(self, project_name: str, namespace_name: str) -> bool:
```

src/datachain/data_storage/metastore.py
```diff
@@ -799,12 +799,29 @@ def create_project(
...
+        return (
+            project_name == self.listing_project_name
+            and namespace_name == self.system_namespace_name
+        )
+
+    def _is_default_project(self, project_name, namespace_name) -> bool:
```
dreadatour:
```suggestion
    def _is_default_project(self, project_name: str, namespace_name: str) -> bool:
```

dreadatour (approved):
Looks good to me 👍 Thank you for the fix ❤️

PR #1184, opened 2025-06-27
Improving error logs when getting dataset fails

ilongin:
This avoids vague error messages like `datachain.error.DatasetNotFoundError: Dataset dogs-catss not found in project 1501`. Now full namespace name and project name is in the error message to help user debug.

### DIFF
--- a/src/datachain/data_storage/metastore.py
+++ b/src/datachain/data_storage/metastore.py
@@ -206,6 +206,10 @@ class AbstractMetastore(ABC, Serializable):
         """
 
     @abstractmethod
+    def get_project_by_id(self, project_id: int, conn=None) -> Project:
+        """Gets a single project by id"""
+
+    @abstractmethod
     def list_projects(self, namespace_id: Optional[int], conn=None) -> list[Project]:
         """Gets list of projects in some namespace or in general (in all namespaces)"""
 
@@ -840,6 +844,24 @@ class AbstractDBMetastore(AbstractMetastore):
             )
         return self.project_class.parse(*rows[0])
 
+    def get_project_by_id(self, project_id: int, conn=None) -> Project:
+        """Gets a single project by id"""
+        n = self._namespaces
+        p = self._projects
+
+        query = self._projects_select(
+            *(getattr(n.c, f) for f in self._namespaces_fields),
+            *(getattr(p.c, f) for f in self._projects_fields),
+        )
+        query = query.select_from(n.join(p, n.c.id == p.c.namespace_id)).where(
+            p.c.id == project_id
+        )
+
+        rows = list(self.db.execute(query, conn=conn))
+        if not rows:
+            raise ProjectNotFoundError(f"Project {project_id} not found.")
+        return self.project_class.parse(*rows[0])
+
     def list_projects(self, namespace_id: Optional[int], conn=None) -> list[Project]:
         """
         Gets a list of projects inside some namespace, or in all namespaces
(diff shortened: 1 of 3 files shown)


shcheklein:
@ilongin is it ready for review? shipping?

src/datachain/data_storage/metastore.py
```diff
@@ -1182,15 +1204,29 @@ def get_dataset(
         """
         Gets a single dataset in project by dataset name.
         """
-        project_id = project_id or self.default_project.id
+        project = None
+        if not project_id:
+            project = self.default_project
+            project_id = project.id
```
shcheklein: is default project id predefined?

shcheklein (commented):
seems quite complicated just to improve the error message ...

where was it happening? where were we using the `get_dataset` in the first place, do we have names in the context to pass as extra fields instead of doing the extra call?

shcheklein (commented):
so, get_dataset is used almost exclusively in places where we have the full project (namespace and project names). I would strongly try to avoid such complicated implementation just for the sake of improving the error message. I think it would be easier to accept Project in this case, not project id.

Overall we have too many options across different (but similar methods) - names, Project, project.id - please let's try to keep it clean ...

ilongin:
> so, get_dataset is used almost exclusively in places where we have the full project (namespace and project names). I would strongly try to avoid such complicated implementation just for the sake of improving the error message. I think it would be easier to accept Project in this case, not project id.
> 
> Overall we have too many options across different (but similar methods) - names, Project, project.id - please let's try to keep it clean ...

We had this discussion in my first original namespaces PR where I got feedback that it's better for metastore not to accept whole objects, rather to accept specific fields like `id`. This is because in Studio workers we need to call these metastore methods using HTTP (using special metastore API implementation) and in some places (metastore API accepting those requests) we don't have whole objects ready so we need to do additional DB call which is going to affect performance. Here is the example of it: https://github.com/iterative/studio/blob/main/backend/dqlapp/metastore/views/datasets.py#L324-L326 .. in this example we would need to do another call to get project by id before calling `.get_dataset()`. Regarding our nature of this metastore over HTTP I also think it's better to continue to avoid putting whole object instances as arguments even if that means more complex logic in some places like this one.

src/datachain/catalog/catalog.py
```diff
@@ -1124,7 +1124,7 @@ def get_dataset_with_remote_fallback(
             raise DatasetNotFoundError(
                 f"Dataset {name}"
                 + (f" version {version} " if version else " ")
-                + "not found"
+                + f"not found in namespace {namespace_name} and project {project_name}"
```
shcheklein, line 1136: will it look like `local and project local` in some (all) cases?

src/datachain/data_storage/metastore.py
```diff
@@ -851,6 +855,24 @@ def get_project(
             )
         return self.project_class.parse(*rows[0])
 
+    def get_project_by_id(self, project_id: int, conn=None) -> Project:
```
shcheklein: a bit separate topic - btw, why are we using IDs still, not UUIDs for all these operations?

shcheklein:
okay, if metastore doesn't operate with objects and is concerned with only ids (we should be really using uuids), then it should not be throwing user facing error in the first place ... it should be handled by the metastore clients then. They should be taking it and wrapping it accordingly, depending on the context. And they most likely have enough context for this. Let's please review this one more time ...

src/datachain/data_storage/metastore.py
ilongin: ids are good enough. `uuid` will be needed if we are going to pull projects from Studio which is not used yet. I'm thinking to remove `uuid` completely for now though.

src/datachain/catalog/catalog.py
ilongin: Yes. Do you have another suggestion? ... Maybe I can put `not found in project local.local`?

ilongin:
> okay, if metastore doesn't operate with objects and is concerned with only ids (we should be really using uuids), then it should not be throwing user facing error in the first place ... it should be handled by the metastore clients then. They should be taking it and wrapping it accordingly, depending on the context. And they most likely have enough context for this. Let's please review this one more time ...

Yes, I think you are right. I've removed this logic for now and re-raising exception in calling methods like `catalog.get_dataset()`

src/datachain/data_storage/metastore.py
shcheklein: no, we should be using uuids I think ... it's a pitty we didn't start using them originally ... we'll have a very major and heavy migration later

src/datachain/data_storage/metastore.py
```diff
@@ -207,6 +207,10 @@ def get_project(
         It also creates project if not found and create flag is set to True.
         """
 
+    @abstractmethod
+    def get_project_by_id(self, project_id: int, conn=None) -> Project:
```
shcheklein: so, do we still need it?

src/datachain/catalog/catalog.py
shcheklein: I think we should try to avoid `local` / `local` when people deal with local setup. May we should have a helper to convert all these errors into something human readable - e.g. skip local completely, or say something like "Can't find in local namespace" ...

src/datachain/data_storage/metastore.py
ilongin: No need, removing.

src/datachain/catalog/catalog.py
shcheklein: can we handle it here (and in other places) somehow after all? to avoid local / local in local env? any other ideas?

PR #1186, opened 2025-06-27
Adding ability to set namespace and project using env variables

ilongin:
# Setting Namespace and Project via Environment Variables

In addition to using `.settings()`, you can configure the namespace and project using environment variables:

- `DATACHAIN_NAMESPACE` sets the namespace.
- `DATACHAIN_PROJECT` sets the project name, or both the namespace and project using the format `namespace.project`.

### Examples

```
# Set namespace only
export DATACHAIN_NAMESPACE=dev
# Set project only
export DATACHAIN_PROJECT=analytics
# Set both namespace and project
export DATACHAIN_PROJECT=dev.analytics
```

##  How Namespace and Project Are Resolved

When determining which namespace and project to use, Datachain applies the following precedence:

1. **Fully qualified dataset name**
   If the dataset name includes both the namespace and project, these values take highest precedence.
   ```python
   dc.read_dataset("dev.analytics.metrics")

2. **Explicit settings in code**
   Values provided via `.settings()` or passed directly to `read_dataset()` or similar methods.
   ```python
   dc.settings(namespace="dev", project="analytics")
   dc.read_dataset("metrics", namespace="dev", project="analytics")
   ```
3. **Environment variables**
   Namespace and project set using environment variables:
   ```console
   export DATACHAIN_PROJECT=dev.analytics
   ```

## Summary by Sourcery

Enable setting default namespace and project via environment variables and centralize dataset name resolution through Catalog.get_full_dataset_name

New Features:
- Allow configuring default namespace with DATACHAIN_NAMESPACE environment variable
- Allow configuring default project with DATACHAIN_PROJECT environment variable

Enhancements:
- Introduce Catalog.get_full_dataset_name to unify namespace, project, and dataset name resolution
- Refactor save, read, delete functions and CLI dataset commands to leverage the unified name resolution

Documentation:
- Add DATACHAIN_NAMESPACE and DATACHAIN_PROJECT entries to the environment variables guide

Tests:
- Add parameterized test to verify namespace and project resolution via environment variables and settings during save

## Summary by Sourcery

Enable environment variable support for namespace and project defaults and centralize dataset name resolution through Catalog.get_full_dataset_name, updating dataset operations, CLI commands, documentation, and tests accordingly

New Features:
- Allow configuring default namespace via DATACHAIN_NAMESPACE environment variable
- Allow configuring default project or combined namespace.project via DATACHAIN_PROJECT environment variable

Enhancements:
- Introduce Catalog.get_full_dataset_name to centralize namespace, project, and dataset name resolution
- Refactor save, read_dataset, delete_dataset, and CLI commands to use unified name resolution
- Return datasets by full_name in record insertion to ensure consistent lookup

Documentation:
- Document DATACHAIN_NAMESPACE and DATACHAIN_PROJECT in environment variable and namespaces guides
- Update namespaces guide with resolution precedence including environment variables and defaults

Tests:
- Add parameterized tests for namespace and project resolution via environment variables, settings, and explicit dataset names


docs/guide/env.md
```diff
@@ -15,4 +15,8 @@ List of environment variables used to configure DataChain behavior.
 - `DATACHAIN_STUDIO_TOKEN` – Authentication token for Studio.
 - `DATACHAIN_STUDIO_TEAM` – Studio team name.
 
+### Namespaces and projects
+- `DATACHAIN_NAMESPACE` – Namespace name to use as default.
+- `DATACHAIN_PROJECT` – Project name to use as default.
```
shcheklein: Can project name include namespace?

dmpetrov: yes. I'd expect there is a single variable, not two.

ilongin: So you would like for example this: `DATACHAIN_PROJECT=dev.analytics`?

shcheklein: yes, but we still need a way to specify only NAMESPACE

any ideas for a good var name?

(I'm fine to keep both vars and make sure that DATACHAIN_PROJECT can accept dev.analytics`)

ilongin: Yea, I agree that we need to keep `NAMESPACE`. I will do it as you propose, project can include namespace and that's it.

dmpetrov: yes: 
```
# I assume the 1st one (dev) is namespace and analytics is project
DATACHAIN_NAMESPACE=dev.analytics 

# OR only namespace without project
DATACHAIN_NAMESPACE=dev
```

Re the naming, chatted briefly, - it's better to use NAMESPACE rather than PROJECT since project is a vague term.

ilongin: Second example is strange to me. `DATACHAIN_PROJECT` env clearly states it's about projects but then we set `dev` namespace with it and no project? I'm more in favor of what Ivan suggested above, examples:
```
DATACHAIN_PROJECT=dev.analytics -> dev namespace and analytics project
DATACHAIN_PROJECT=analytics  -> analytics project and default namespace
DATACHAIN_NAMESPACE=dev  -> dev namespace and default project
DATACHAIN_NAMESPACE=dev & DATACHAIN_PROJECT=analytics -> dev namespace and analytics project
```

shcheklein (approved):

docs/guide/env.md
dmpetrov: sorry. I meant DATACHAIN_NAMESPACE. modifying...

dmpetrov: on a separate note... 
1. does `DATACHAIN_` prefix too long? Why not `DC_`?
2. why don't we operate in datachain config instead of env variables. it's more common practice. User might work on several projects with different namespaces (or other env) at the same time.

ilongin: @dmpetrov 
1. We already use `DATACHAIN_` prefix so I wanted to be consistent. We can do alias for all of those variables to `DC_` as well if we want, but I would do it in a separate issue
2. Main motivation for this was to enable Studio users to set default values for namespace / project. Locally user can only have `local.local` project and cannot create another one so dataset config doesn't really helps.

dmpetrov: Re env vs config. It's a general question. It feels like we are too heavy on env while in general config is preferred option.

shcheklein (approved):

dmpetrov:
@ilongin it feels like we introduce 2 variables to public API without a need.

We can handle this using a single namespace params. Let's try this instead.

shcheklein:
@dmpetrov I think we need then a good idea on the name. If you have a good suggestion please share. (keep in mind please - we need to release it today).

dmpetrov:
> I think we need then a good idea on the name.

The idea:
```
DATACHAIN_NAMESPACE=ns1
# or
DATACHAIN_NAMESPACE=ns1.pr1
```

I cannot imaging cases when user needs to set only project. Also, we should avoid using such vague names asa **project** especially because we already have project in experiments.

I'm pretty sure we won't find additional space in UI to see two params - it will be a single box for both. So, why we push them on users in the API?

PR #1187, opened 2025-06-27
fix(mutate): save literals properly

shcheklein:
Things like this was failing for me:

```python
    ds = dc.read_values(data=[1,2,3]).mutate(str_col="test_string").save("test")
```

then:

```python
dc.read_dataset("test").to_values("str_col")
```

with a Signal not found error.

## Summary by Sourcery

Fix schema handling in the mutate method for literal values and add a functional test to verify that primitive columns persist correctly through save/load and filtering operations

Bug Fixes:
- Fix mutate to pass the correct mutated mapping to schema.mutate, ensuring literal values are saved properly

Tests:
- Add test_mutate_with_primitives_save_load to validate schema persistence, data integrity, and filtering for primitive literal columns across save and load cycles

### DIFF
--- a/src/datachain/lib/dc/datachain.py
+++ b/src/datachain/lib/dc/datachain.py
@@ -1103,7 +1103,7 @@ class DataChain:
                 mutated[name] = value
 
         return self._evolve(
-            query=self._query.mutate(**mutated), signal_schema=schema.mutate(kwargs)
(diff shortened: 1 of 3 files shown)


shcheklein:
> PS: I'm sure the unit test is AI generated 🙂 - too detailed.

:) redacted heavily after, yep ... it was 2x initially

PR #1192, opened 2025-06-29
Fix 'aggregate' function if limit or offset was used

dreadatour:
Fix `aggregate` function (`.agg()`) if `limit` or `offset` was used.

Example of failed query:

```
from typing import Iterator

import datachain as dc
from datachain import func, File


def my_agg(p: list[str], s: list[int]) -> Iterator[tuple[str, int]]:
    yield str(p), sum(s)

chain = (
    dc.read_storage("s3://bucket/")
    .limit(20)
    .agg(
        my_agg,
        params={"file.path": str, "file.size": int,},
        output={"res": str, "total": int},
        partition_by=dc.C("file.path")
    )
    .save("tmp")
)
```

### DIFF
--- a/src/datachain/data_storage/warehouse.py
+++ b/src/datachain/data_storage/warehouse.py
@@ -234,13 +234,13 @@ class AbstractWarehouse(ABC, Serializable):
                 # Cursor results are not thread-safe, so we convert them to a list
                 results = list(wh.dataset_rows_select(paginated_query.offset(offset)))
 
-                processed = False
+                processed = 0
                 for row in results:
-                    processed = True
+                    processed += 1
                     yield row
                     num_yielded += 1
 
-                if not processed:
+                if not processed or processed < page_size:
                     break  # no more results
                 offset += page_size
 
@@ -345,9 +345,15 @@ class AbstractWarehouse(ABC, Serializable):
 
         if is_batched:
             for batch in ids:
-                yield list(self.dataset_rows_select(query.where(id_col.in_(batch))))
+                yield list(
+                    self.dataset_rows_select(
+                        query.offset(None).limit(None).where(id_col.in_(batch))
+                    )
+                )
         else:
-            yield from self.dataset_rows_select(query.where(id_col.in_(ids)))
+            yield from self.dataset_rows_select(
+                query.offset(None).limit(None).where(id_col.in_(ids))
+            )
 
     @abstractmethod
     def get_dataset_sources(
--- a/src/datachain/query/dataset.py
+++ b/src/datachain/query/dataset.py
@@ -11,6 +11,7 @@ from collections.abc import Generator, Iterable, Iterator, Sequence
 from copy import copy
 from functools import wraps
 from secrets import token_hex
+from types import GeneratorType
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -557,8 +558,8 @@ class UDFStep(Step, ABC):
         """
         assert self.partition_by is not None
 
-        if isinstance(self.partition_by, Sequence):
-            list_partition_by = self.partition_by
+        if isinstance(self.partition_by, (list, tuple, GeneratorType)):
+            list_partition_by = list(self.partition_by)
         else:
             list_partition_by = [self.partition_by]
 
@@ -575,7 +576,10 @@ class UDFStep(Step, ABC):
             f.dense_rank().over(order_by=partition_by).label(PARTITION_COLUMN_ID),
         ]
         self.catalog.warehouse.db.execute(
-            tbl.insert().from_select(cols, query.with_only_columns(*cols))
+            tbl.insert().from_select(
+                cols,
+                query.offset(None).limit(None).with_only_columns(*cols),
+            )
         )
 
         return tbl
@@ -602,12 +606,20 @@ class UDFStep(Step, ABC):
             partition_tbl = self.create_partitions_table(query)
             temp_tables.append(partition_tbl.name)
 
-            subq = query.subquery()
+            # Limit and offset are not applied to the partition table,
+            # so we need to apply them to the main query later.
+            limit = query._limit
+            offset = query._offset
+
+            subq = query.limit(None).offset(None).subquery()
             query = (
                 sqlalchemy.select(*subq.c)
                 .outerjoin(partition_tbl, partition_tbl.c.sys__id == subq.c.sys__id)
                 .add_columns(*partition_columns())
             )
+
+            # Apply limit and offset to the main query.
+            query = query.offset(offset).limit(limit)
 
         query, tables = self.process_input_query(query)
         temp_tables.extend(t.name for t in tables)
--- a/tests/func/test_datachain.py
+++ b/tests/func/test_datachain.py
@@ -2324,3 +2324,61 @@ def test_agg(catalog_tmpfile, parallel):
         ],
         "parent",
     )
+
+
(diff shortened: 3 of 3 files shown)


src/datachain/data_storage/warehouse.py
```diff
@@ -234,13 +234,13 @@ def dataset_select_paginated(
...
-                processed = False
+                processed = 0
                 for row in results:
-                    processed = True
+                    processed += 1
                     yield row
                     num_yielded += 1
 
-                if not processed:
+                if not processed or processed < page_size:
```
dreadatour: Do not make unnecessary SQL query.

src/datachain/data_storage/warehouse.py
```diff
@@ -345,9 +345,15 @@ def dataset_rows_select_from_ids(
 
         if is_batched:
             for batch in ids:
-                yield list(self.dataset_rows_select(query.where(id_col.in_(batch))))
+                yield list(
+                    self.dataset_rows_select(
```
dreadatour: Reset `limit` and `offset` here since we are fetching rows by IDs.

src/datachain/data_storage/warehouse.py
```diff
@@ -345,9 +345,15 @@ def dataset_rows_select_from_ids(
...
+                        query.offset(None).limit(None).where(id_col.in_(batch))
+                    )
+                )
         else:
-            yield from self.dataset_rows_select(query.where(id_col.in_(ids)))
+            yield from self.dataset_rows_select(
+                query.offset(None).limit(None).where(id_col.in_(ids))
```
dreadatour: Reset `limit` and `offset` here since we are fetching rows by IDs.

src/datachain/query/dataset.py
```diff
@@ -557,8 +558,8 @@ def create_partitions_table(self, query: Select) -> "Table":
         """
         assert self.partition_by is not None
 
-        if isinstance(self.partition_by, Sequence):
-            list_partition_by = self.partition_by
+        if isinstance(self.partition_by, (list, tuple, GeneratorType)):
```
dreadatour, line 561: Fix for signal name as string is used in `partition_by` argument.

src/datachain/query/dataset.py
```diff
@@ -557,8 +558,8 @@ def create_partitions_table(self, query: Select) -> "Table":
...
+            list_partition_by = list(self.partition_by)
```
dreadatour, line 562: Use the same `list` type.

src/datachain/query/dataset.py
```diff
@@ -575,7 +576,10 @@ def create_partitions_table(self, query: Select) -> "Table":
             f.dense_rank().over(order_by=partition_by).label(PARTITION_COLUMN_ID),
         ]
         self.catalog.warehouse.db.execute(
-            tbl.insert().from_select(cols, query.with_only_columns(*cols))
+            tbl.insert().from_select(
+                cols,
+                query.offset(None).limit(None).with_only_columns(*cols),
```
dreadatour, line 581: Reset `limit` and `offset` here since we are inserting all the partitions and we don't know which one will be used later.

src/datachain/query/dataset.py
```diff
@@ -602,13 +606,21 @@ def apply(
...
+
+            subq = query.limit(None).offset(None).subquery()
             query = (
                 sqlalchemy.select(*subq.c)
                 .outerjoin(partition_tbl, partition_tbl.c.sys__id == subq.c.sys__id)
                 .add_columns(*partition_columns())
             )
 
+            # Apply limit and offset to the main query.
+            query = query.offset(offset).limit(limit)
```
dreadatour: Another fix for `limit` and `offset` (see comments).

tests/func/test_datachain.py
```diff
@@ -2324,3 +2324,61 @@ def process(files: list[str]) -> Iterator[tuple[str, int]]:
...
+            partition_by="filename",
+        )
+        .save("my-ds")
+    )
+
+    assert sorted_dicts(ds.to_records(), "filename") == sorted_dicts(
+        [
+            {"filename": "file0", "count": 1},
+            {"filename": "file1", "count": 1},
+            {"filename": "file10", "count": 1},
```
dreadatour: I don't know why these rows were selected (`file0`, `file1` and `file10`) 🤔
This is the main unknown part for me here.

shcheklein (approved):

tests/func/test_datachain.py
shcheklein: Don't we need to figure this out?

shcheklein (commented):
Just to double check - do we apply limit (and friends) before we do agg in these examples?

src/datachain/query/dataset.py
```diff
@@ -602,13 +606,21 @@ def apply(
             partition_tbl = self.create_partitions_table(query)
             temp_tables.append(partition_tbl.name)
 
-            subq = query.subquery()
+            # Limit and offset are not applied to the partition table,
```
shcheklein: Not sure I understand this- are we changing semantics here by chance?

tests/func/test_datachain.py
dreadatour: We do, I just wanted to create a PR to share the results. Let me continue investigating this weird issue (also it is worth mentioning, is the same in SQLite and ClickHouse).

dreadatour: Hm, it looks like offset does not works in this case 🤔 Weird. Double checking and updating the test (also adding more tests, it looks like this part is not fully covered).

dreadatour:
> Just to double check - do we apply limit (and friends) before we do agg in these examples?

It's complicated, we are updating it before, and also kind of during the aggregation 👀

src/datachain/data_storage/warehouse.py
```diff
@@ -218,7 +218,7 @@ def dataset_select_paginated(
         limit = query._limit
         paginated_query = query.limit(page_size)
 
-        offset = 0
+        offset = query._offset or 0
```
dreadatour, line 221: `.offset()` was broken 😢

src/datachain/data_storage/warehouse.py
```diff
@@ -343,6 +343,8 @@ def dataset_rows_select_from_ids(
         if (id_col := get_query_id_column(query)) is None:
             raise RuntimeError("sys__id column not found in query")
 
+        query = query._clone().offset(None).limit(None).order_by(None)
```
dreadatour, line 346: The main change is here. Do not apply any limit, offset or order by if selecting by ids.

src/datachain/query/dataset.py
```diff
@@ -601,13 +605,10 @@ def apply(
...
-            subq = query.subquery()
-            query = (
-                sqlalchemy.select(*subq.c)
-                .outerjoin(partition_tbl, partition_tbl.c.sys__id == subq.c.sys__id)
-                .add_columns(*partition_columns())
-            )
+            query = query.outerjoin(
+                partition_tbl,
+                partition_tbl.c.sys__id == query.selected_columns.sys__id,
+            ).add_columns(*partition_columns())
```
dreadatour, line 611: Another major important change is here.
Instead of using subquery, use query itself.
This will allow us to later reset offset, limit and order by, otherwise it will be tricky (to traverse through the whole query with all subqueries any modify each one if needed is not very stable and also may leads to random bugs in future).

dreadatour (commented):
More highlights.

PR #1196, opened 2025-07-02
Added new reserved char and fixing studio tests

ilongin:
Adding `@` to reserved chars and fixing studio tests due to Studio companion PR changes: https://github.com/iterative/studio/pull/11823

## Summary by Sourcery

Add '@' to reserved characters for dataset, namespace, and project names, and update the CI workflow variable to fix Studio tests.

Enhancements:
- Include '@' in the list of reserved characters for dataset, namespace, and project naming

CI:
- Rename CI environment variable from DATACHAIN_METASTORE_ARG_PROJECT to DATACHAIN_METASTORE_ARG_NAMESPACE in tests workflow


.github/workflows/tests-studio.yml
```diff
@@ -98,7 +98,7 @@ jobs:
       - name: Run tests
         # Generate `.test_durations` file with `pytest --store-durations --durations-path ../.github/.test_durations ...`
         run: >
-          DATACHAIN_METASTORE_ARG_PROJECT=john
+          DATACHAIN_METASTORE_ARG_NAMESPACE=john
```
shcheklein: now that we have a bit more time - can we review please why we specify this here (and not other DATACHAIN vars)

ilongin: When I run tests locally I have all of them set which is correct. Here we don't put for example `team_id` which is then set to `None` by default and since `team_id` is nullable in our tables where we reference team all datachain tests work fine **magically**  although I think we should put some value to be more correct and don't rely on accidents. The same is with `user_id`. Unlike those two, `namespace` cannot be `None` and is required to set for tests to pass.

My suggestion is to add all those variables here (except `credentials_id` as that one can probably be removed from postgresql metastore in Studio as I don't see it being used anymore).

Here is how my local test run script looks like (I put some random numbers for `team_id` and `user_id`):
```bash
DATACHAIN_METASTORE=datachain_server.postgresql.PostgreSQLMetastore \
DATACHAIN_WAREHOUSE=datachain_server.clickhouse.ClickHouseWarehouse \
DATACHAIN_CH_DSN=clickhouse+native://dvcx:ch123456789\!@localhost:9000/dvcx_local_db \
DATABASE_URL=psql://postgres:pwd111@localhost:5432/dvcx_local_db \
DATACHAIN_METASTORE_ARG_USER_ID=42 \
DATACHAIN_METASTORE_ARG_NAMESPACE=john \
DATACHAIN_METASTORE_ARG_CREDENTIALS_ID=47 \
DATACHAIN_WAREHOUSE_ARG_TEAM_ID=5 \
DATACHAIN_AUTH_LOCAL_MODE=1 \
python -m pytest
```

shcheklein: are they automatically picked from some .env.test ? or some CI/CD secrets?

it's not the right way to embed these long env vars setups right into YAML to my mind ... run CI/CD with `env` command to see actual values, let's figure out if they set somewhere ... may be @0x2b3bfa0 or @dreadatour have this knowledge right away

src/datachain/namespace.py
```diff
@@ -6,7 +6,7 @@
 from datachain.error import InvalidNamespaceNameError
 
 N = TypeVar("N", bound="Namespace")
-NAMESPACE_NAME_RESERVED_CHARS = ["."]
+NAMESPACE_NAME_RESERVED_CHARS = [".", "@"]
```
shcheklein, line 9: just to make sure - are those replaced / escaped automatically if username contains them? do we check it when we create the default namespace from the username?

ilongin: They are replaced automatically for default namespace and validated when creating random namespace (error is thrown if name contains some of those)

.github/workflows/tests-studio.yml
ilongin: I don't think we have anything setup with env vars, but tests are working magically  / by accident as `None` values for those are still ok for tests, as I mentioned above. We should set all of those vars in whatever way and I think for now it's ok to put in YAML what's required (`username`). In separate issue we can improve this

0x2b3bfa0: @ilongin, is there any reason why we don't add that environment variable to https://github.com/iterative/studio/blob/f01323edf59941b0d408358a3dd68189a025f3bb/backend/datachain_server/.env.test, automagically sourced by https://github.com/iterative/studio/blob/f01323edf59941b0d408358a3dd68189a025f3bb/backend/datachain_server/pyproject.toml#L69?

0x2b3bfa0: > _may be @0x2b3bfa0 or @dreadatour have this knowledge right away_

(Apologies from being the lazy [crocodile](https://en.wikipedia.org/wiki/How_Doth_the_Little_Crocodile))

0x2b3bfa0: Opened these pull requests as a follow-up:
* https://github.com/iterative/datachain/pull/1203
* https://github.com/iterative/studio/pull/11846

ilongin: Thanks, I wasn't aware of that file.

Issue #1198, opened 2025-07-02
sys__id error when mutate before persist


dmpetrov:
### Description

```python
import datachain as dc

chain = (
    dc.read_values(fib=[1, 1, 2, 3, 5, 8, 13, 21])
    .map(mod3=lambda fib: fib % 3, output=int)
    .group_by(
        cnt=dc.func.count(),
        partition_by="mod3",
    )
    .mutate(x=1)
    .persist()
).show()
```

--> sqlite3.OperationalError: no such column: sys__id

### Version Info

```Text

```

ilongin:
@dreadatour is this fixed with your related PR? Can we close it?

dreadatour:
Yes, this issue can be closed. There are other issues, but I'll address them in separate issues/PRs.

PR #1199, opened 2025-07-03
Fix table copy: do not use batching if no 'sys.id' field found

dreadatour:
Fix for the https://github.com/iterative/datachain/issues/1198

When we do table copy where is an optimization to copy rows in batches. If no `sys__id` field found in table, this optimization does not works.

In this PR I am fixing this issue: copy whole table at once if no `sys__id` field found in table.

### DIFF
--- a/src/datachain/data_storage/sqlite.py
+++ b/src/datachain/data_storage/sqlite.py
@@ -774,7 +774,15 @@ class SQLiteWarehouse(AbstractWarehouse):
         query: Select,
         progress_cb: Optional[Callable[[int], None]] = None,
     ) -> None:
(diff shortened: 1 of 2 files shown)


dmpetrov:
@dreadatour do you have any ideas how we end up in a situation when `sys` was not created? it suppose to be present in any table, isn't it?

We might need to dig into this as an upstream issue.

dreadatour:
> @dreadatour do you have any ideas how we end up in a situation when `sys` was not created? it suppose to be present in any table, isn't it?
> 
> We might need to dig into this as an upstream issue.

You're absolutely right, this is a correct question. I'll dig into it in the following issue 👍

PR #1200, opened 2025-07-04
feat(audio): audio file and fragment with streaming

shcheklein:
Add Audio and AudioFragment support, to make code like below work with full streaming (no file download happening):

```python

import torch
from transformers import Pipeline, pipeline

import datachain as dc
from datachain import AudioFile, AudioFragment


def fragments(file: AudioFile) -> Iterator[AudioFragment]:
    yield file.get_fragment(300, 400)


def process(fragment: AudioFragment, pipeline: Pipeline) -> str:
    """Process audio fragment using direct numpy conversion."""
    audio_array, sample_rate = fragment.get_np()

    # Convert to mono if stereo (average the channels)
    if len(audio_array.shape) > 1 and audio_array.shape[1] > 1:
        audio_array = audio_array.mean(axis=1)

    # Pass the numpy array with exact sampling rate from fragment
    result = pipeline({
        "array": audio_array,
        "sampling_rate": sample_rate
    })
    return str(result)

(
    dc
    .read_storage("s3://ivan-test-versioned/sounds", type="audio")
    .settings(cache=False, prefetch=False)
    .gen(fragment=fragments)
    .setup(pipeline=lambda: pipeline(
        "audio-classification",
        model="superb/wav2vec2-base-superb-ks",
        torch_dtype=torch.float32,
        device="cpu"
    ))
    .map(type=process)
    .show()
)
```

## TODO:

- [x] Add tests
- [x] Add example like above
- [ ] Add FE support for AudioFragment model to show preview it from the middle point cc @yathomasi

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -143,3 +143,5 @@ cython_debug/
 
 # pt files produced by ultralytics examples
 *.pt
+
+.DS_Store/
--- a/src/datachain/lib/audio.py
+++ b/src/datachain/lib/audio.py
@@ -1,7 +1,5 @@
-from __future__ import annotations
-
 import posixpath
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Optional, Union
 
 from datachain.lib.file import FileError
 
@@ -20,17 +18,8 @@ except ImportError as exc:
     ) from exc
 
 
-def audio_info(file: File | AudioFile) -> Audio:
-    """
-    Returns audio file information.
-
-    Args:
-        file (AudioFile): Audio file object.
-
-    Returns:
-        Audio: Audio file information.
-    """
-    # Import here to avoid circular imports
+def audio_info(file: "Union[File, AudioFile]") -> "Audio":
+    """Extract metadata like sample rate, channels, duration, and format."""
     from datachain.lib.file import Audio
 
     file = file.as_audio_file()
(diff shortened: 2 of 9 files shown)


.gitignore
```diff
@@ -143,3 +143,5 @@ cython_debug/
 
 # pt files produced by ultralytics examples
 *.pt
+
+.DS_Store/
```
dreadatour, line 147: 🙏

src/datachain/lib/file.py
```diff
@@ -851,6 +859,223 @@ def get_fragments(
...
+        Returns:
+            bytes: The encoded audio segment as bytes.
+        """
+        from .audio import audio_segment_bytes
+
+        duration = None if self.end is None else self.end - self.start
+        return audio_segment_bytes(self.audio, self.start, duration, format)
+
+
+class AudioFragment(DataModel):
```
dreadatour, line 945: What is the difference between `AudioSegment` and `AudioFragment`?

dreadatour (commented):
Code looks good to me! 👍
One question though, have checked docstrings and implementation, but can't see the difference between `AudioSegment` and `AudioFragment` (may be I need to take a closer look).

src/datachain/lib/file.py
shcheklein: Good catch, thanks! One of them is no needed.

src/datachain/lib/audio.py
```diff
@@ -0,0 +1,151 @@
...
+        channels=channels,
+        duration=duration,
+        samples=frames,
+        format=format_name,
+        codec=codec_name,
+        bit_rate=bit_rate,
+    )
+
+
+def audio_segment_np(
```
dreadatour: Should we rename this to `audio_fragment_np`? 👀

src/datachain/lib/audio.py
```diff
@@ -0,0 +1,151 @@
...
+                audio_np = audio_np.squeeze()
+
+            return audio_np, int(sr)
+    except Exception as exc:
+        raise FileError(
+            "unable to read audio segment", audio.source, audio.path
+        ) from exc
+
+
+def audio_segment_bytes(
```
dreadatour: Should we rename this to `audio_fragment_bytes`? 👀

src/datachain/lib/audio.py
shcheklein: yep, will do. Minor since it is not user exposed.

src/datachain/lib/audio.py
shcheklein: yep, will do. Minor since it is not user exposed.

PR #1201, opened 2025-07-04
Display cluster name on `datachain job run`

0x2b3bfa0:
DataChain counterpart of https://github.com/iterative/studio/pull/11834

### DIFF
--- a/src/datachain/studio.py
+++ b/src/datachain/studio.py
@@ -348,7 +348,13 @@ def create_job(
         raise DataChainError("Failed to create job")
 
     job_id = response.data.get("job", {}).get("id")
-    print(f"Job {job_id} created")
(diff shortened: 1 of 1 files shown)


src/datachain/studio.py
```diff
@@ -348,7 +348,12 @@ def create_job(
         raise DataChainError("Failed to create job")
 
     job_id = response.data.get("job", {}).get("id")
```
0x2b3bfa0, line 350: @sourcery-ai is right: all this `response.data.get("job", {}).get(...)` code seems unnecessarily complicated; we should be using indexing, unless I missed something.

0x2b3bfa0:
Closing until new notice

PR #1202, opened 2025-07-04
read dataset: always look for new version with update flag

shcheklein:
Before only this would check for newer versions available (if version spec is provided):

```python
dc.read_dataset("some.remote.dataset", version=">=1.0.0", update=True)
```

now, this would also check for updates and pull the latest version:

```python
dc.read_dataset("some.remote.dataset", update=True)
```


src/datachain/catalog/catalog.py
```diff
@@ -1120,6 +1120,9 @@ def get_dataset_with_remote_fallback(
         pull_dataset: bool = False,
         update: bool = False,
     ) -> DatasetRecord:
+        if version:
+            update = False
```
amritghimire, line 1129: Changing the user argument doesn't seem like a good idea. We could either raise error or assert that having both version and update to true is not advisable.

Still, how would this work if we specify version>=0.2 as spec and update? Shouldn't this still update the version?

amritghimire (dismissed):

src/datachain/catalog/catalog.py
shcheklein: yes, it will work. It is not user facing method. It is an internal method and it expects the exact version (e.g. `1.0.0`, no version specs, etc, etc). So, it we ask it to give us an exact version - there is no reason at this point to do an update - we already figured out by now what version we need.

amritghimire: Then why not adding assertion that `version` and `update` should both not be provided or assign it to a different argument instead of modifying the function argument?

ilongin (approved):
LGTM

src/datachain/catalog/catalog.py
shcheklein: the can be provided - that's fine, we just ignore it then (from the upstream sometimes we want to pass exact version and don't want to track update flag). It is fine that we just ignore it here I think.

The same btw for the users:

```
dc.read_dataset("some.remote.dataset", version="1.0.0", update=True)
```

we just ignore the update flag. Is it the best semantics - don't know, but seems reasonable. Let me know WDYT.

amritghimire: Semantics feels little weird tbh. But dismissing the change request to move forward

PR #1214, opened 2025-07-08
Simplify permission checks for creating namespaces

ilongin:
WIP

### DIFF
--- a/src/datachain/catalog/catalog.py
+++ b/src/datachain/catalog/catalog.py
@@ -527,6 +527,7 @@ class Catalog:
             Callable[["AbstractWarehouse"], None]
         ] = None,
         in_memory: bool = False,
+        is_cli: Optional[bool] = True,
     ):
         datachain_dir = DataChainDir(cache=cache_dir, tmp=tmp_dir)
         datachain_dir.init()
@@ -540,6 +541,11 @@ class Catalog:
         }
         self._warehouse_ready_callback = warehouse_ready_callback
         self.in_memory = in_memory
+        self._is_cli = is_cli
+
+    @property
+    def is_cli(self) -> bool:
+        return self._is_cli  # type: ignore[return-value]
 
     @cached_property
     def warehouse(self) -> "AbstractWarehouse":
@@ -1128,7 +1134,12 @@ class Catalog:
         if version:
             update = False
 
-        if self.metastore.is_local_dataset(namespace_name) or not update:
+        # local dataset is the one that is in Studio or in CLI but has default namespace
+        is_local = (
+            not self.is_cli or namespace_name == self.metastore.default_namespace_name
+        )
+
+        if is_local or not update:
             try:
                 project = self.metastore.get_project(project_name, namespace_name)
                 ds = self.get_dataset(name, project)
@@ -1137,7 +1148,7 @@ class Catalog:
             except (NamespaceNotFoundError, ProjectNotFoundError, DatasetNotFoundError):
                 pass
 
-        if self.metastore.is_local_dataset(namespace_name):
+        if is_local:
             raise DatasetNotFoundError(
                 f"Dataset {name}"
                 + (f" version {version} " if version else " ")
--- a/src/datachain/catalog/loader.py
+++ b/src/datachain/catalog/loader.py
@@ -137,10 +137,13 @@ def get_catalog(
     we can provide DATACHAIN_METASTORE_ARG_TEAM_ID=12345 env variable.
     """
     from datachain.catalog import Catalog
+    from datachain.data_storage.sqlite import SQLiteMetastore
 
+    metastore = get_metastore(in_memory=in_memory)
     return Catalog(
-        metastore=get_metastore(in_memory=in_memory),
+        metastore=metastore,
         warehouse=get_warehouse(in_memory=in_memory),
         client_config=client_config,
         in_memory=in_memory,
+        is_cli=isinstance(metastore, SQLiteMetastore),
     )
--- a/src/datachain/cli/commands/datasets.py
+++ b/src/datachain/cli/commands/datasets.py
@@ -137,7 +137,8 @@ def rm_dataset(
 ):
     namespace_name, project_name, name = catalog.get_full_dataset_name(name)
 
-    if not catalog.metastore.is_local_dataset(namespace_name) and studio:
+    if catalog.is_cli and studio:
+        # removing Studio dataset from CLI
         from datachain.studio import remove_studio_dataset
 
         token = Config().read().get("studio", {}).get("token")
@@ -147,11 +148,24 @@ def rm_dataset(
             )
         remove_studio_dataset(team, name, namespace_name, project_name, version, force)
     else:
+        # if catalog.metastore.is_local_dataset(namespace_name):
         try:
             project = catalog.metastore.get_project(project_name, namespace_name)
             catalog.remove_dataset(name, project, version=version, force=force)
         except DatasetNotFoundError:
             print("Dataset not found in local", file=sys.stderr)
+
+    """
+    else:
+        from datachain.studio import remove_studio_dataset
+
+        token = Config().read().get("studio", {}).get("token")
+        if not token:
+            raise DataChainError(
+                "Not logged in to Studio. Log in with 'datachain auth login'."
+            )
+        remove_studio_dataset(team, name, namespace_name, project_name, version, force)
+    """
 
 
 def edit_dataset(
@@ -164,13 +178,27 @@ def edit_dataset(
 ):
     namespace_name, project_name, name = catalog.get_full_dataset_name(name)
 
-    if catalog.metastore.is_local_dataset(namespace_name):
+    if catalog.is_cli and namespace_name != catalog.metastore.default_namespace_name:
+        from datachain.studio import edit_studio_dataset
+
+        token = Config().read().get("studio", {}).get("token")
+        if not token:
+            raise DataChainError(
+                "Not logged in to Studio. Log in with 'datachain auth login'."
+            )
+        edit_studio_dataset(
+            team, name, namespace_name, project_name, new_name, description, attrs
+        )
+    else:
+        # if catalog.metastore.is_local_dataset(namespace_name):
         try:
             catalog.edit_dataset(
                 name, catalog.metastore.default_project, new_name, description, attrs
             )
         except DatasetNotFoundError:
             print("Dataset not found in local", file=sys.stderr)
+
+    """
     else:
         from datachain.studio import edit_studio_dataset
 
@@ -182,3 +210,4 @@ def edit_dataset(
         edit_studio_dataset(
             team, name, namespace_name, project_name, new_name, description, attrs
         )
+    """
--- a/tests/unit/lib/test_datachain.py
+++ b/tests/unit/lib/test_datachain.py
@@ -16,7 +16,6 @@ from pydantic import BaseModel
 
 import datachain as dc
 from datachain import Column
-from datachain.data_storage import AbstractMetastore
 from datachain.error import (
     DatasetInvalidVersionError,
     DatasetNotFoundError,
@@ -3224,6 +3223,7 @@ def test_delete_dataset_versions_all(test_session):
 
 
 @pytest.mark.parametrize("force", (True, False))
+@pytest.mark.parametrize("is_cli", (True,))
 @skip_if_not_sqlite
 def test_delete_dataset_from_studio(test_session, studio_token, requests_mock, force):
     requests_mock.delete(f"{STUDIO_URL}/api/datachain/datasets", json={"ok": True})
@@ -3236,6 +3236,7 @@ def test_delete_dataset_from_studio(test_session, studio_token, requests_mock, f
     )
 
 
+@pytest.mark.parametrize("is_cli", (True,))
 @skip_if_not_sqlite
 def test_delete_dataset_from_studio_not_found(
     test_session, studio_token, requests_mock
@@ -3378,8 +3379,8 @@ def test_semver_preview_ok(test_session):
     assert sorted([p["num"] for p in dataset.get_version("1.0.1").preview]) == [3, 4]
 
 
-@pytest.mark.parametrize("allow_create_project", [True, False])
-def test_save_to_default_project(test_session, allow_create_project):
+@pytest.mark.parametrize("is_cli", [True, False])
+def test_save_to_default_project(test_session, is_cli):
     catalog = test_session.catalog
     ds_name = "fibonacci"
     dc.read_values(fib=[1, 1, 2, 3, 5, 8], session=test_session).save(ds_name)
@@ -3387,10 +3388,8 @@ def test_save_to_default_project(test_session, allow_create_project):
     assert ds.dataset.project == catalog.metastore.default_project
 
 
-@pytest.mark.parametrize("allow_create_project", [True, False])
-def test_save_to_default_project_with_read_storage(
-    tmp_dir, test_session, allow_create_project
-):
+@pytest.mark.parametrize("is_cli", [True, False])
+def test_save_to_default_project_with_read_storage(tmp_dir, test_session, is_cli):
     catalog = test_session.catalog
     ds_name = "parquet_ds"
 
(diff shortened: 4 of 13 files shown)


src/datachain/catalog/catalog.py
```diff
@@ -527,6 +527,7 @@ def __init__(
             Callable[["AbstractWarehouse"], None]
         ] = None,
         in_memory: bool = False,
+        is_cli: Optional[bool] = True,
```
shcheklein: just to double check is it really optional?

src/datachain/catalog/catalog.py
```diff
@@ -1111,7 +1117,12 @@ def get_dataset_with_remote_fallback(
         if version:
             update = False
 
-        if self.metastore.is_local_dataset(namespace_name) or not update:
+        # local dataset is the one that is in Studio or in CLI but has default namespace
+        is_local = (
```
shcheklein: I'm not sure I understand this condition - how can we make it simpler, easier to read? (I'm also not sure I understand the comment above)

src/datachain/catalog/loader.py
```diff
@@ -137,10 +137,13 @@ def get_catalog(
...
+    from datachain.data_storage.sqlite import SQLiteMetastore
 
+    metastore = get_metastore(in_memory=in_memory)
     return Catalog(
-        metastore=get_metastore(in_memory=in_memory),
+        metastore=metastore,
         warehouse=get_warehouse(in_memory=in_memory),
         client_config=client_config,
         in_memory=in_memory,
+        is_cli=isinstance(metastore, SQLiteMetastore),
```
shcheklein: there should be a better (explicit way) to set this up ... also, it means not only CLI but local Python, right? Why is it called CLI then?

src/datachain/cli/commands/datasets.py
```diff
@@ -164,21 +167,22 @@ def edit_dataset(
 ):
     namespace_name, project_name, name = catalog.get_full_dataset_name(name)
 
-    if catalog.metastore.is_local_dataset(namespace_name):
+    if catalog.is_cli and namespace_name != catalog.metastore.default_namespace_name:
```
shcheklein: also, not sure I understand (too complicated) .. what exactly we are detecting here?

src/datachain/catalog/catalog.py
ilongin: Before namespaces / projects we didn't have any kind of check here and in Studio we would still try to fetch from remote (Studio) if missing locally, which was wrong (it would fail with strange error like missing token).
Now we can determine if dataset we are fetching is from it's own (local) DB or can be fetched from remote / Studio if missing locally.

We never fallback to Studio if:
1. Script is already ran in Studio
2. Dataset starts with `local.local.*`

I agree this whole function is way to complex and confusing and I will try to refactor it.

src/datachain/catalog/loader.py
ilongin: We usually call CLI for anything other than Studio so that's why I called it that way. It can be renamed. 
I was weighing between what you mentioned, only set this explicitly, or have implicit (if explicit flag doesn't exist) determination as it's implemented now. If you have strong opinion **only** explicit is better I can do that way.

src/datachain/cli/commands/datasets.py
ilongin: If someone does this: `datachain ds edit dev.my-project.cats --new-name "dogs"` this means it wants to edit Studio dataset and not local one. What's extra here, and should be removed is `catalog.is_cli` as by default this is called in CLI

ilongin: Removed `catalog.is_cli` condition

src/datachain/catalog/loader.py
ilongin: Removed implicit logic and leaved only explicit arg

src/datachain/catalog/catalog.py
ilongin: I've added better comment for that variable and changed name. I think it should be clear now.

src/datachain/catalog/catalog.py
shcheklein: so, it is really optional? 

let's also come up with a better name please, `is_cli` is confusing ... you can do `is_studio` for example, False by default.

src/datachain/cli/commands/datasets.py
```diff
@@ -164,21 +167,22 @@ def edit_dataset(
...
+        if Config().read().get("studio", {}).get("token"):
+            edit_studio_dataset(
+                team, name, namespace_name, project_name, new_name, description, attrs
+            )
+        else:
+            raise DataChainError(
+                "Not logged in to Studio. Log in with 'datachain auth login'."
+            )
+    else:
+        # if catalog.metastore.is_local_dataset(namespace_name):
```
shcheklein: dead code?

src/datachain/cli/commands/datasets.py
```diff
@@ -164,21 +167,22 @@ def edit_dataset(
 ):
     namespace_name, project_name, name = catalog.get_full_dataset_name(name)
 
-    if catalog.metastore.is_local_dataset(namespace_name):
+    if namespace_name != catalog.metastore.default_namespace_name:
```
shcheklein: let's introduce a descriptive var - studio_dataset = ...

to make condisiton descriptive:

if studio_dataset:
    ....

shcheklein: (tbh still don't a lot this very non obvious way to detect it by analyzing namespaces, comparing it with default - one needs to know a lot about namespaces to understand this code and why it is correct - it is not obvious)

src/datachain/catalog/catalog.py
ilongin: Removed optional and changed `is_cli` to `is_studio`

src/datachain/cli/commands/datasets.py
ilongin: Added `is_studio_dataset` var as you suggested.

Regarding second comment, the only way to fix this and make explicitly is to re-introduce `studio` flag but that's exactly what we tried to avoid in the first place with namespaces / projects - having those flags everywhere ... we wanted to be able to decide implicitly if some dataset is in Studio or local and that's what this code does .. trade-of is that code is less clean.

dreadatour: I just wonder if we can use another approach for this, like set env variable or use one of env variables we already have (like `DATACHAIN_COMPUTE_CLUSTER_ID`).

This is because it is a general task: to know if query was running from within Studio or CLI, @dmpetrov was asking about this specifically, and it would be great to have something like `dc.is_studio` flag/variable without instantiating whole `Catalog`, and `Catalog` itself can get this knowledge from the save variable seamless, without additional param in constructor.

src/datachain/cli/commands/datasets.py
```diff
@@ -164,21 +167,24 @@ def edit_dataset(
 ):
     namespace_name, project_name, name = catalog.get_full_dataset_name(name)
 
-    if catalog.metastore.is_local_dataset(namespace_name):
+    # studio dataset is if it has non "local" namespace
+    is_studio_dataset = namespace_name != catalog.metastore.default_namespace_name
```
shcheklein: how about Studio dataset with a default `global` namespace?

tests/unit/lib/test_datachain.py
```diff
@@ -4062,6 +4061,7 @@ def test_save_to_non_default_namespace_and_project(
     if use_settings:
         ds = ds.settings(namespace="dev", project="numbers").save("fibonacci")
     else:
+        print("before save")
```
shcheklein: needed?

tests/func/test_read_dataset_remote.py
```diff
@@ -445,6 +449,7 @@ def test_read_dataset_remote_update_flag_no_version(
 
 
 @skip_if_not_sqlite
+@pytest.mark.parametrize("is_studio", (False,))
```
shcheklein, line 452: how does it exactly work?  

`pytest.mark.parametrize("is_studio", (False,))`

what do we need this for?

src/datachain/cli/commands/datasets.py
```diff
@@ -164,21 +167,24 @@ def edit_dataset(
...
+
+    if is_studio_dataset:
+        from datachain.studio import edit_studio_dataset
+
+        if Config().read().get("studio", {}).get("token"):
```
shcheklein, line 184: unrelated: don't we have some util in StudioClient to do this? to get the token?

shcheklein (commented):
Looks good overall .. some minor questions

src/datachain/cli/commands/datasets.py
ilongin: Yes, I was thinking about this as well. Didn't want to introduce new env variable for now as I didn't know we need this in general as you mentioned but if we do need I can add as well.

tests/func/test_read_dataset_remote.py
ilongin: By default, all tests run with `Catalog.is_studio` flag set to `True`. This is to allow creating namespaces / projects etc .. i.e. "studio" datachain allows more things, so to avoid the need to mock it explicitly to `True` all the time it's set to `True` by default in tests.
We can also change that with this fixture - set it to `False`, or  both `True` and `False`, depending on a test need. This particular test is testing CLI so we need to set flag to `False`. 
Logic around that fixture is in conftest.py: `is_studio` and `mock_is_studio` fixtures.

src/datachain/cli/commands/datasets.py
ilongin: Not sure but we def need to refactor this file as it duplicates these lines multiple times.

src/datachain/cli/commands/datasets.py
ilongin: You mean if this will still work correctly? It should as it's different from `local` which is in `catalog.metastore.default_namespace_name`

tests/func/test_read_dataset_remote.py
shcheklein: Initial reaction - I'm lost :)

Second thought - two fixtures with negative condition, there gotta be a better / simpler way to do this that is clear when you read the code ...

Third thought (after reading conftest a bit):

- why autouse for this fixture? can we just pass `mock_is_studio` where it is appropriate? That would a straightforward way to use mocks ... am I missing something?
- `@skip_if_not_sqlite` can this one also define this ... or can it be replace with a single decorator?
- why do we need to set False (studio) explicitly - shouldn't be it default if run it on sqlite only already?

src/datachain/cli/commands/datasets.py
shcheklein: `catalog.metastore.default_namespace_name` is it local even for Studio? how do we redefine it?

tests/func/test_read_dataset_remote.py
ilongin: Ok, so this is the problem:
1. Default value for `Catalog.is_studio` is `False`
2. There are a lot of scenarios, mostly related to having custom namespace / project in tests, for which we need `Catalog.is_studio` to be `True`
3. There are rare scenarios / tests where we need `Catalog.is_studio` to be explicitly `False` (CLI methods testing)
4. With 2) and 3) in mind it's better to have that set to `True` by default and have special fixture to set it to `False` on those rare tests when needed then other way around (we would then need to repeat fixture in a lot of tests to set that flag to `True`)
5. We don't want to go into direction of splitting tests for CLI and Studio like we did with `@skip_if_not_sqlite` fixture.

Solution was to by default set `Catalog.is_studio` to be `True` so that we are not limited in terms of creating custom namespaces / projects in tests and this is achieved with:
`mock_is_studio` is autoused and it sets `Catalog.is_studio` to `True` **unless** value of other fixture `is_studio` is set to `False` - then we just yield which results in not setting flag to `True` which then stays `False` (default). This way we can set `Catalog.is_studio` to `False` just by adding `is_studio` fixture to be `False` on those rare tests where we need this scenario (CLI methods testing as mentioned above).

I know this is little bit complex but at the time I didn't have better idea. Maybe this will be simplified when I introduce ENV variable to control this `is_studio` thing, as Vlad suggested .. but on first thought I don't think it will change much. 
If someone has better idea how to handle this I'm open to suggestion ... this is actually a general problem of having slightly different logic on Studio vs local and wanting to have non separated tests for them both.

src/datachain/cli/commands/datasets.py
ilongin: Ok, now I see this better. So this is part of the code that will only run in local (CLI). In Studio we don't need CLI, right. 
Here we need to check if the dataset user want's to edit is his local dataset from `.datachain` folder or it's a remote Studio dataset which happens only if dataset's namespace is different from default one which is always `local` in local / CLI run. I could have hardcoded `local` here instead of `catalog.metastore.default_namespace_name` as well.

Note: this code would break if someone runs CLI in Studio - it would try to go to another Studio if namespace is different from `global` (default in Studio) ... but this code is never going to run from Studio in the first place - at least that wasn't ever the plan.

shcheklein: can we put some assert that it is not Studio? (to document and prevent it from running)?

src/datachain/cli/commands/datasets.py
ilongin: Added env variable called `DATACHAIN_IS_STUDIO` and function to check it, e.g
```python
import datachain as dc
dc.is_studio()  # True or False
```

It's function as it's easier to handle and write tests (mock)

src/datachain/cli/commands/datasets.py
ilongin: Added check to not allow CLI from Studio environment.

tests/func/test_read_dataset_remote.py
shcheklein: thanks @ilongin for a good explanation!

> There are a lot of scenarios, mostly related to having custom namespace / project in tests, for which we need Catalog.is_studio to be True


do we need primarily for tests that pull datasets from Studio one way or another? or what are other major examples

src/datachain/catalog/catalog.py
```diff
@@ -1129,7 +1131,14 @@ def get_dataset_with_remote_fallback(
         if version:
             update = False
 
-        if self.metastore.is_local_dataset(namespace_name) or not update:
+        # we don't do Studio fallback is script is already ran in Studio, or if we try
+        # to fetch dataset with local.local namespace & project as that one cannot
+        # exist in Studio in the first place
+        no_fallback = (
+            is_studio() or namespace_name == self.metastore.default_namespace_name
```
shcheklein: Let,s please add a helper is-namespae-local with assert that it is “local” and with an explanation and with this namespace check

shcheklein: I mean instead of the second part only … to make more explanatory

src/datachain/cli/commands/datasets.py
shcheklein: Let’s use the helper I mentioned above

src/datachain/cli/__init__.py
```diff
@@ -78,8 +79,12 @@ def main(argv: Optional[list[str]] = None) -> int:
 
 def handle_command(args, catalog, client_config) -> int:
     """Handle the different CLI commands."""
+    from datachain.lib.dc.utils import is_studio
     from datachain.studio import process_auth_cli_args, process_jobs_args
 
+    if is_studio():
+        raise DataChainError("CLI is not allowed in Studio environment")
```
dreadatour: [from](https://github.com/iterative/studio/pull/11859#issuecomment-3208896539) @dmpetrov:
> Please use term local instead of cli since cli might be in studio mode.

Just wanted to everyone to be on the same page.

src/datachain/cli/commands/datasets.py
dreadatour: > Added check to not allow CLI from Studio environment.

https://github.com/iterative/datachain/pull/1214/files#r2296718451

dreadatour: CLI can run in Studio env, at least in shell scripts 🤔

tests/func/test_read_dataset_remote.py
ilongin: Not just those, for example creating normal dataset but in custom namespace / project, check if namespace is created implicitly etc. Those are not local scenarios as they cannot happen in CLI but as mentioned above, I think we should avoid splitting tests for Studio and local - that would be another level of complexity... so now we are testing those with SQLite DB adapter as well.

src/datachain/catalog/catalog.py
ilongin: Done

src/datachain/cli/commands/datasets.py
ilongin: Hm, it's really getting complicated with this local vs Studio vs CLI things :D .. anyway, I removed check from all CLI commands to run on Studio and just changed problematic `edit_dataset` function

src/datachain/cli/__init__.py
ilongin: This line is removed after your other comment

PR #1215, opened 2025-07-08
partition_by with complex signals - includes agg and group_by

dmpetrov:
Allow complex signals (e.g., `File` type) in `agg`'s `partition_by` parameter for improved usability (resolves #1211).

ToDo:
- [x] in unit tests, replace to_records() logic with double underscore code to more native to_values()
- [x] simplify code a bit
- [x] reduce amount of unit tests
- [x] linter and tests

PS: it turn out, the problem was much more complicated since group_by changes schema while the changes become tricky when a nested column is used in `partition_by` - see to_partial() and related

---

[Slack Thread](https://iterativeai.slack.com/archives/C04A9RWEZBN/p1752007968694989?thread_ts=1752007968.694989&cid=C04A9RWEZBN)

## Summary by Sourcery

Add support for using complex signal types (DataModel subclasses) directly in the `partition_by` parameter of the `agg` method.

New Features:
- Allow DataModel subclasses (e.g., File, Image) to be passed directly to `partition_by` in `agg`.
- Introduce `_process_complex_signal_partition` to resolve complex signal types into their unique identifier columns.

Enhancements:
- Extend `PartitionByType` to include `type` for complex signals in the dataset query definitions.
- Update `agg` logic and docstrings to integrate complex signal handling alongside existing column-based partitioning.

Documentation:
- Enhance `agg` method documentation with examples demonstrating file-based partitioning.

Tests:
- Add functional tests for partitioning by complex signals, mixed type partitioning, and error scenarios for invalid or missing DataModel types.

## Summary by Sourcery

Enable using complex DataModel signals directly in partition_by for both agg and group_by by expanding nested model fields into SQL columns and extending schema partials, enrich documentation with examples, and add comprehensive tests for various partitioning scenarios

New Features:
- Support passing DataModel subclasses (e.g., File, ImageFile) directly to agg’s partition_by parameter
- Extend complex signal partitioning support to group_by, expanding nested signals into their identifier columns

Enhancements:
- Add _process_complex_signal_column to recursively resolve and expand complex signal paths into SQL columns
- Enhance SignalSchema.to_partial and group_by to handle full and field-level inclusion of custom types and deduplicate nested signals
- Adjust dataset query builder to select unique columns and correctly alias partition_by and group_by results

Documentation:
- Update agg and group_by docstrings with examples demonstrating file-based and nested complex signal partitioning

Tests:
- Add extensive unit tests for SignalSchema.to_partial covering entire, nested, mixed, and error scenarios
[...]

Issue #1211, opened 2025-07-08
Allow complex signals in partition_by


dreadatour:
Allow aggregation by `file` or any other complex signal, e.g.:

```python
from typing import Iterator

import datachain as dc
from datachain import func, File


def my_agg(f: list[File]) -> Iterator[tuple[File, int]]:
    yield f, sum(f.size)

chain = (
    dc.read_storage("s3://bucket/")
    .agg(
        my_agg,
        params={"file": File},
        output={"file": File, "total": int},
        partition_by="file",
    )
    .save("tmp")
)
```

dreadatour:
Fixed in https://github.com/iterative/datachain/pull/1215

PR #1220, opened 2025-07-09
Implement aggregate 'partition_by' by complex signals

dreadatour:
Fix for https://github.com/iterative/datachain/issues/1211
Based on https://github.com/iterative/datachain/pull/1215

## Summary by Sourcery

Enable the `partition_by` parameter to accept complex signals (DataModel subclasses) by expanding them into their unique key columns via a new `db_columns` API and a `unique_only` filter, update core schema traversal methods and aggregation logic accordingly, and add comprehensive tests for these scenarios.

New Features:
- Enable aggregation partitioning by complex signal types (DataModel) by expanding them into their unique key attributes
- Introduce SignalSchema.db_columns method to retrieve typed Column objects with optional name and unique_only filtering

Enhancements:
- Extend SignalSchema.get_column_type, db_signals, and get_flat_tree to accept a unique_only flag for filtering unique key fields
- Update agg() to handle BaseModel signals by expanding them into their unique key columns
- Switch mutate() and read_records() to use db_columns instead of db_signals for column operations
- Update DataModel and File classes to support per-class unique key field definitions

Tests:
- Add functional tests for partition_by with File, mixed types, nested signals, and inherited DataModel
- Rename and expand unit tests to cover db_columns behavior


src/datachain/lib/file.py
```diff
@@ -240,17 +240,16 @@ class File(DataModel):
...
-    _unique_id_keys: ClassVar[list[str]] = [
-        "source",
-        "path",
-        "size",
-        "etag",
-        "version",
-        "is_latest",
-        "location",
-        "last_modified",
-    ]
```
dreadatour, line 253: Unused, was removed in https://github.com/iterative/datachain/pull/90

src/datachain/lib/signal_schema.py
```diff
@@ -597,6 +607,35 @@ def db_signals(
 
         return signals  # type: ignore[return-value]
 
+    def db_columns(
+        self,
+        name: Optional[str] = None,
+        include_hidden: bool = True,
+        unique_only: bool = False,
+    ) -> list[Column]:
```
dreadatour, line 615: Have split `db_signals` into two methods: `db_columns` and `db_signals`.
Different return type, different logic.

tests/func/test_datachain.py
```diff
@@ -2299,113 +2299,3 @@ def func(key: str) -> str:
     for _ in range(4):
         with pytest.raises(Exception, match="Test Error!"):
             chain.map(res=func).exec()
-
-
-@pytest.mark.parametrize("parallel", [1, 2])
-def test_agg(catalog_tmpfile, parallel):
```
dreadatour, line 2305: Moved few tests to `test_aggregate`

src/datachain/lib/signal_schema.py
```diff
@@ -554,36 +554,46 @@ def _set_file_stream(
...
-        for path, _type, has_subtree, _ in self.get_flat_tree():
-            if (with_subtree or not has_subtree) and DEFAULT_DELIMITER.join(
-                path
-            ) == col_name:
+        for path, _type, has_subtree, _ in self.get_flat_tree(unique_only=unique_only):
+            if has_subtree and not with_subtree:
+                continue
+            if DEFAULT_DELIMITER.join(path) == col_name:
                 return _type
         raise SignalResolvingError([col_name], "is not found")
```
dreadatour, line 580: Tiny refactoring for better readability.

src/datachain/lib/file.py
```diff
@@ -240,17 +240,16 @@ class File(DataModel):
...
+    @classmethod
+    def __pydantic_init_subclass__(cls, **kwargs: Any) -> None:
+        """
+        Reset `_unique_key_fields` for every subclass
+        since additional fields can be added.
+        """
+        super().__pydantic_init_subclass__(**kwargs)
+        cls._unique_key_fields = None
```
dreadatour, line 252: This is needed in case one will inherit `File` model and add additional unique fields.

src/datachain/lib/file.py
```diff
@@ -240,17 +240,16 @@ class File(DataModel):
         "last_modified",
         "location",
     ]
+    _unique_key_fields: ClassVar[Optional[list[str]]] = ["source", "path"]
```
dreadatour, line 243: Not sure if this is enough and/or needed. Added as an example of such option.

src/datachain/lib/dc/records.py
```diff
@@ -51,14 +51,13 @@ def read_records(
...
-    columns: list[sqlalchemy.Column] = []
+    feature_schema: Optional[dict[str, DataType]] = None
 
     if schema:
         signal_schema = SignalSchema(schema)
+        feature_schema = signal_schema.clone_without_sys_signals().serialize()
         columns = [
-            sqlalchemy.Column(c.name, c.type)  # type: ignore[union-attr]
-            for c in signal_schema.db_signals(as_columns=True)
+            sqlalchemy.Column(c.name, c.type) for c in signal_schema.db_columns()
```
dreadatour, line 60: Tiny refactoring here for better readability.

src/datachain/lib/file.py
dmpetrov: absolutely not

tests/func/test_aggregate.py
```diff
@@ -0,0 +1,316 @@
+from collections.abc import Iterator
```
dmpetrov, line 1: This file should be in unit test, not functional. And @pytest.mark.parametrize("parallel", [1, 2]) might be not needed here. The new logic seems not related to parallel stuff.

We should prioritize unit test when it's possible.

tests/func/test_aggregate.py
```diff
@@ -0,0 +1,316 @@
...
+        yield files[0], sum(amounts)
+
+    # Test partitioning by File type directly
+    result = chain.agg(
+        test_agg,
+        params=("file", "amount"),
+        output={"file": File, "total": int},
+        partition_by="file",
+    )
+    records = result.to_records()
```
dmpetrov, line 150: please avoid using to_records() in tests since it requires double underscore logic. It's better to use `to_list()` and such and operate with dots in names.

tests/func/test_aggregate.py
```diff
@@ -0,0 +1,316 @@
...
+        output={"file": File, "total": int},
+        partition_by="signal.file",
+    )
+    records = result.to_records()
+
+    # We should have 3 groups (file1.txt appears twice, so should be grouped)
+    assert len(records) == 3
+
+    # Check that files with same unique attributes are grouped together
+    file_paths = {r["file__path"] for r in records}
```
dmpetrov, line 252: Same - to_records() that leads to the underscores

src/datachain/lib/file.py
dmpetrov: We should avoid using _unique_key_fields if possible - too specific to File class

src/datachain/lib/signal_schema.py
```diff
@@ -554,36 +554,46 @@ def _set_file_stream(
...
+    ) -> DataType:
         """
         Returns column type by column name.
 
         If `with_subtree` is True, then it will return the type of the column
         even if it has a subtree (e.g. model with nested fields), otherwise it will
         return the type of the column (standard type field, not the model).
 
+        If `unique_only` is True, then it will return only columns used as unique key.
+        Unique key columns defined by `_unique_key_fields` attribute in the type.
```
dmpetrov, line 571: Too technical - needs to be removed

src/datachain/lib/signal_schema.py
```diff
@@ -743,29 +782,56 @@ def _build_tree(
...
                     field: info
-                    for field, info in substree.items()
+                    for field, info in subtree.items()
                     if field not in hidden_fields
                 }
 
-            has_subtree = substree is not None
+            unique_key_fields: Optional[list[str]] = None
+            if unique_only:
+                unique_key_fields = getattr(type_, "_unique_key_fields", None)
```
dmpetrov, line 823: This logic does not work in general since it's too File specific. Cursor generated this mostly because the issue was originally defined through "file" which is oversimplification.

You can find tests in my PR when this logic breaks - like multi-level nested with not-file data models.

dreadatour:
Implemented in https://github.com/iterative/datachain/pull/1215

PR #1221, opened 2025-07-09
Add cli support to move, remove and copy file to storage using Studio

amritghimire:
This adds the support for following command

### Local to local 

```bash
datachain cp /tmp/test-2.py /tmp/test_folder/test-1.py
datachain cp -r /tmp/test_folder /tmp/testing_folder
```
### local to remote

```bash
datachain cp /tmp/test-2.py gs://amrit-datachain-test/playground/test-5.py
datachain cp /tmp/test-2.py gs://amrit-datachain-test/playground/test-6.py --studio-cloud-auth
```

```bash
datachain cp --recursive /tmp/test_folder gs://amrit-datachain-test/playground
datachain cp --recursive /tmp/test_folder gs://amrit-datachain-test/playground --studio-cloud-auth
```

### Remote to local

```bash
datachain cp gs://amrit-datachain-test/playground/test-5.py /tmp/output_folder/
datachain cp gs://amrit-datachain-test/playground/test-5.py /tmp/output_folder/ —studio-cloud-auth
datachain cp -r gs://amrit-datachain-test/playground/test_folder /tmp/output_folder/ 
```

### Remote to remote

```bash
datachain cp gs://amrit-datachain-test/playground/test_folder/test-3.py gs://amrit-datachain-test/playground/test_folder/test-4.py
datachain cp gs://amrit-datachain-test/playground/test_folder/test-3.py gs://amrit-datachain-test/playground/test_folder/test-4.py —studio-cloud-auth
```

## MV
```bash
datachain mv gs://amrit-datachain-test/playground/test.py gs://amrit-datachain-test/playground-2/test.py
datachain mv gs://amrit-datachain-test/playground/test.py gs://amrit-datachain-test/playground-2/test.py --studio-cloud-auth
```

```bash
datachain mv gs://amrit-datachain-test/playground gs://amrit-datachain-test/playground-2  --recursive
datachain mv gs://amrit-datachain-test/playground gs://amrit-datachain-test/playground-2  --recursive
```

## RM 
```bash
datachain rm gs://amrit-datachain-test/playground/test.py --recursive 
datachain rm gs://amrit-datachain-test/playground/test.py —studio-cloud-auth
```

```bash
datachain rm gs://amrit-datachain-test/playground --recursive 
datachain rm gs://amrit-datachain-test/playground --recursive --studio-cloud-auth
```
Please check the documentation on more details on this.

## Summary by Sourcery

Enable storage file management via Studio by adding backend methods and CLI support for `rm`, `mv`, and `cp` operations, along with corresponding documentation.

New Features:
- Add datachain storage rm, mv, and cp CLI commands for deleting, moving, and copying files via Studio
- Introduce StudioClient methods for storage operations including delete, move, copy, presigned URL batch requests, download URL retrieval, and upload logging

Enhancements:
- Extend CLI command routing and parser to support the new storage subcommands
- Refactor datachain CLI to route storage commands through process_storage_command

Documentation:
- Update mkdocs configuration and add user-facing documentation for storage rm, mv, and cp commands

## Summary by Sourcery

Add CLI support for managing remote storage through Studio by introducing storage subcommands and backend methods for file operations

New Features:
- Add datachain storage rm, mv, and cp CLI commands for deleting, moving, and copying files via Studio
- Introduce StudioClient methods for delete_storage_file, move_storage_file, copy_storage_file, batch_presigned_urls, download_url, and save_upload_log

Enhancements:
- Refactor CLI command routing to dispatch storage operations and handle local vs Studio workflows
- Add remote/storage utilities for uploading and downloading files through fsspec with Studio integration
- Update mkdocs configuration to include new storage commands

Documentation:
- Add user documentation for cp, mv, and rm storage commands in mkdocs

Tests:
- Add functional tests for storage rm, mv, and cp covering local-to-local, local-to-remote, remote-to-local, and remote-to-remote scenarios


shcheklein:
@amritghimire let's check first the existing APIs - `datachain cp` and `datachain ls` - they should be taking this role

amritghimire:
> @amritghimire let's check first the existing APIs - `datachain cp` and `datachain ls` - they should be taking this role

Yes, I am looking into that too. I wanted to implement the studio specific part and merge those.

amritghimire:
> @amritghimire let's check first the existing APIs - `datachain cp` and `datachain ls` - they should be taking this role

@shcheklein What do you propose of the syntax on how to handle both studio or local with this approach? 
I mean how to identify if we need to use studio or local?

amritghimire:
@sourcery-ai review

amritghimire:
@sourcery-ai guide

src/datachain/cli/__init__.py
```diff
@@ -108,15 +115,23 @@ def handle_command(args, catalog, client_config) -> int:
...
-        update=bool(args.update),
-        recursive=bool(args.recursive),
-        no_glob=args.no_glob,
-    )
+def handle_cp_command(args, catalog: "Catalog"):
+    from datachain.config import Config
+
+    config = Config().read().get("studio", {})
+    token = config.get("token")
+    local = True if not token else args.local
```
amritghimire: This is what I am not quite satisfied about. How should we distinguish between the call that should use the cp from catalog or through studio? cc. @shcheklein 

https://github.com/iterative/datachain/pull/1221#issuecomment-3057437853

docs/commands/cp.md
```diff
@@ -0,0 +1,175 @@
+# cp
+
+Copy storage files and directories between cloud and local storage.
+
+## Synopsis
+
+```usage
+usage: datachain cp [-h] [-v] [-q] [-r] [--team TEAM] [--local] [--anon] [--update] [--no-glob] [--force] source_path destination_path
```
shcheklein: unreadable on the screen

docs/commands/cp.md
```diff
@@ -0,0 +1,175 @@
...
+# Download single file
+datachain cp s3://my-bucket/data/file.txt /path/to/local/file.txt
+
+# Download to directory (filename preserved)
+datachain cp s3://my-bucket/data/file.txt /path/to/directory/
+```
+
+#### 4. Remote to Remote (`s3://` → `s3://`, `gs://` → `gs://`, etc.)
+**Operation**: Copy within cloud storage
+- Copies files between locations in the same bucket
```
shcheklein: why this limitation?

docs/commands/cp.md
```diff
@@ -0,0 +1,175 @@
...
+### Directory Operations
+- **Recursive flag required**: Copying directories requires the `--recursive` flag
+- **Directory structure preservation**: Directory structure is preserved during copy operations
+- **Empty directories**: Empty directories may not be copied in some scenarios
+
+
+### Error Handling
+- **File not found**: Missing source files result in operation failure
+- **Permission errors**: Insufficient permissions cause operation failure
+- **Network issues**: Network problems are reported with appropriate error messages
```
shcheklein: generated? please review, remove all the stuff that is not meaningful

docs/commands/cp.md
```diff
@@ -0,0 +1,175 @@
...
+
+### Team Configuration
+- **Default team**: If no team is specified, uses the team from your configuration
+- **Team-specific storage**: Each team has its own storage namespace
+
+## Notes
+
+* Use the `--verbose` flag to get detailed information about the copy operation
```
shcheklein: Notes should include a lot of explanations above, not repeat again the same info

shcheklein (commented):
* Review docs carefully and with attention

src/datachain/cli/__init__.py
shcheklein: can we always use catalog and just notify Studio if token is set about changes after it is done?

src/datachain/remote/storages.py
```diff
@@ -0,0 +1,228 @@
...
+
+if TYPE_CHECKING:
+    from argparse import Namespace
+
+    from fsspec import AbstractFileSystem
+
+    from datachain.remote.studio import StudioClient
+
+
+def get_studio_client(args: "Namespace"):
```
shcheklein: don't we have already some helpers like this?

src/datachain/remote/storages.py
```diff
@@ -0,0 +1,228 @@
...
+    from datachain.config import Config
+    from datachain.remote.studio import StudioClient
+
+    if Config().read().get("studio", {}).get("token"):
+        return StudioClient(team=args.team)
+
+    raise DataChainError("Not logged in to Studio. Log in with 'datachain auth login'.")
+
+
+def upload_to_storage(args: "Namespace", local_fs: "AbstractFileSystem"):
```
shcheklein: we should not be using argparse stuff at this stage

src/datachain/cli/__init__.py
amritghimire: > can we always use catalog and just notify Studio if token is set about changes after it is done?

That defeats the whole purpose of using credentials from Studio. That will only work as adding the activity logs.

shcheklein: yes, I'm just trying to simplify the scope (let me know if that doesn't make much difference). The request was to be able to have audit log in Studio (less about using Studio managed credentials).

We can add additional explicit feature `--studio-cloud-auth` to enable actual cloud credentials.

I think even if we keep we should decouple audit log from the actual mechanism how we actually perform the operation.

amritghimire: Yes, but I don't know, I don't feel comfortable with having a call to add activity log when the activity is performed completely locally. 

We already had `datachain cp` for activities user want to make the changes using local credentials. I think keeping that as it is and using following umbrella or following structure justify the change much more:
```
datachain studio cp
datachain studio rm
datachain studio mv
```

to explicitly mention that the activities are being performed through Studio.

docs/commands/cp.md
amritghimire: Because of the limitation in the respective clients. We don't download the files to Studio. We use the s3 endpoints or s3 features to copy file between s3 and so on.

docs/commands/cp.md
amritghimire: > generated? please review, remove all the stuff that is not meaningful

Yes, before updating the docs, first lets figure out https://github.com/iterative/datachain/pull/1221#discussion_r2211894683 and then we can go across the documentation changes.

src/datachain/remote/storages.py
amritghimire: No, we don't. We usually generate this directly when needed.

amritghimire:
@shcheklein PTAL

docs/commands/cp.md
```diff
@@ -0,0 +1,155 @@
...
+
+## Arguments
+
+* `source_path` - Path to the source file or directory to copy
+* `destination_path` - Path to the destination file or directory to copy to
+
+## Options
+
+* `-r`, `-R`, `--recursive` - Copy directories recursively
+* `--team TEAM` - Team name to copy storage contents to
```
shcheklein: what does it mean? how can one copy storage content to a team?

docs/commands/cp.md
```diff
@@ -0,0 +1,155 @@
...
+* `-s`, `--studio-cloud-auth` - Use credentials from Studio for cloud operations (Default: False)
+* `--anon` - Use anonymous access to storage
```
shcheklein: source or dest?

docs/commands/cp.md
```diff
@@ -0,0 +1,155 @@
...
+* `--update` - Update cached list of files for the source for local approach
```
shcheklein: what is `local` approach?

docs/commands/cp.md
```diff
@@ -0,0 +1,155 @@
...
+```
+
+4. Copy with force overwrite:
+```bash
+datachain cp --force s3://my-bucket/data /path/to/local/
+```
+
+5. Copy with update and no glob expansion:
+```bash
+datachain cp --update --no-glob s3://my-bucket/data/*.txt /path/to/local/
```
shcheklein: does it really work? (will shell expand it?)

docs/commands/cp.md
```diff
@@ -0,0 +1,155 @@
...
+* `destination_path` - Path to the destination file or directory to copy to
+
+## Options
+
+* `-r`, `-R`, `--recursive` - Copy directories recursively
+* `--team TEAM` - Team name to copy storage contents to
+* `-s`, `--studio-cloud-auth` - Use credentials from Studio for cloud operations (Default: False)
+* `--anon` - Use anonymous access to storage
+* `--update` - Update cached list of files for the source for local approach
+* `--no-glob` - Do not expand globs (such as * or ?) for local approach
```
shcheklein: what if I need avoid expanding ?, but expand *? is there an official way to escape glob?

docs/commands/cp.md
```diff
@@ -0,0 +1,155 @@
...
+* `-h`, `--help` - Show the help message and exit
+* `-v`, `--verbose` - Be verbose
+* `-q`, `--quiet` - Be quiet
+
+## Copy Operations
+
+The command supports two main modes of operation:
+
+### Default Mode
+By default, the command operates directly with local storage access, supporting various copy scenarios between local and remote storage.
```
shcheklein: `local storage access` this is very confusing, considering that we have already things like "_local_ and remote storage"

docs/commands/cp.md
```diff
@@ -0,0 +1,155 @@
...
+# Copy within same bucket
+datachain cp gs://my-bucket/data/file.py gs://my-bucket/archive/file.py
+
+# Copy within same bucket with Studio cloud auth
+datachain cp gs://my-bucket/data/file.py gs://my-bucket/archive/file.py --studio-cloud-auth
+```
+
+### Additional Examples
+
+1. Copy with specific team:
```
shcheklein: style: either use comments like above, or lists above to describe examples

overall, reduce the amount (no need to generate obvious examples that just repeat the description, be mindful clean it up)

src/datachain/cli/__init__.py
```diff
@@ -108,15 +115,36 @@ def handle_command(args, catalog, client_config) -> int:
...
     )
+    from datachain.config import Config
+
+    config = Config().read().get("studio", {})
+    token = config.get("token")
+    studio = False if not token else args.studio_cloud_auth
+    return (
+        StudioStorageImplementation(args, catalog)
+        if studio
+        else LocalStorageImplementation(args, catalog)
```
shcheklein: LocalStorage is a very confusing name. The same goes to StudioStorage .. they are not really storages

docs/commands/cp.md
```diff
@@ -0,0 +1,155 @@
...
+## Arguments
+
+* `source_path` - Path to the source file or directory to copy
+* `destination_path` - Path to the destination file or directory to copy to
+
+## Options
+
+* `-r`, `-R`, `--recursive` - Copy directories recursively
+* `--team TEAM` - Team name to copy storage contents to
+* `-s`, `--studio-cloud-auth` - Use credentials from Studio for cloud operations (Default: False)
```
shcheklein, line 32: is it actually default: false or what is the actual logic to determine which one is used? (it seems from the implementation below this is not true)

src/datachain/cli/commands/storage/base.py
```diff
@@ -0,0 +1,80 @@
...
+    def copy_local_to_local(self, source_cls: "Client"):
+        source_fs = source_cls.create_fs()
+        source_fs.copy(
+            self.args.source_path,
+            self.args.destination_path,
+            recursive=self.args.recursive,
+        )
+        print(f"Copied {self.args.source_path} to {self.args.destination_path}")
+
+    def upload_to_remote(self, source_cls: "Client", destination_cls: "Client"):
```
shcheklein: are all of these methods protected?

src/datachain/cli/commands/storage/base.py
```diff
@@ -0,0 +1,80 @@
...
+    from argparse import Namespace
+
+    from fsspec import AbstractFileSystem
+
+    from datachain.catalog import Catalog
+    from datachain.client.fsspec import Client
+
+
+class StorageImplementation:
+    def __init__(self, args: "Namespace", catalog: "Catalog"):
```
shcheklein: let's not use argparse concepts down the stack. It should be used only on the level of parsing args

src/datachain/cli/commands/storage/base.py
```diff
@@ -0,0 +1,80 @@
...
+    def upload_to_remote(self, source_cls: "Client", destination_cls: "Client"):
+        raise NotImplementedError("Upload to remote is not implemented")
+
+    def download_from_remote(self, destination_cls: "Client"):
+        raise NotImplementedError("Download from remote is not implemented")
+
+    def copy_remote_to_remote(self, source_cls: "Client"):
+        raise NotImplementedError("Copy remote to remote is not implemented")
+
+    def save_upload_logs(
```
shcheklein, line 53: how about rm / mv? 

also name it "save_activity_logs"

is it also protected?

src/datachain/cli/commands/storage/base.py
```diff
@@ -0,0 +1,80 @@
...
+
+        try:
+            studio_client = get_studio_client(self.args.team)
+        except DataChainError:
+            return
+
+        uploads = [
+            {
+                "path": dst,
+                "size": source_fs.info(src).get("size", 0),
```
shcheklein: does it mean an extra call? we def don't want to do this for all objects again ...

src/datachain/cli/commands/storage/base.py
```diff
@@ -0,0 +1,80 @@
...
+    def download_from_remote(self, destination_cls: "Client"):
+        raise NotImplementedError("Download from remote is not implemented")
+
+    def copy_remote_to_remote(self, source_cls: "Client"):
+        raise NotImplementedError("Copy remote to remote is not implemented")
+
+    def save_upload_logs(
+        self,
+        destination_path: str,
+        file_paths: dict,  # {destination_path: source}
```
shcheklein: I'm not sure I understand the input args here ...

src/datachain/cli/commands/storage/local.py
```diff
@@ -0,0 +1,127 @@
...
+from datachain.cli.commands.storage.utils import build_file_paths, validate_upload_args
+from datachain.error import DataChainError
+
+if TYPE_CHECKING:
+    from datachain.client.fsspec import Client
+
+
+class LocalStorageImplementation(StorageImplementation):
+    def upload_to_remote(self, source_cls: "Client", destination_cls: "Client"):
+        from tqdm import tqdm
```
shcheklein: can we use Datachain for all operations? instead of implementing these manually?

Datachain has read_strorage ... to_storage ... we should be able to utilize those ... otherwise we keep reimplementing the safe stuff again and again and again

src/datachain/cli/commands/storage/local.py
```diff
@@ -0,0 +1,127 @@
...
+                        while True:
+                            chunk = source_file.read(8192)
+                            if not chunk:
+                                break
+                            dest_file.write(chunk)
+                            pbar.update(len(chunk))
+
+        self.save_upload_logs(self.args.destination_path, file_paths, source_fs)
+
+    def download_from_remote(self, destination_cls: "Client"):
```
shcheklein: remote == cloud in this case?

let's use `cloud` then ... `remote` is used for too many things now, and we should stop using it for all of them really

shcheklein (commented):
It seems a bit too complicated. Let's review and try to use DataChain itself to do these operations. Our high level commands should be just wrappers on top, not custom implementations (we'll have otherwise to do tons of optimization, then support all of this, etc, etc)

Also, we need to review docs, cleanup, etc.

Also, in some case it seems it is not optimal - doing extra calls to APIs which is expensive.

docs/commands/cp.md
amritghimire: Yes, it is by default false. i.e. If user doesn't specify it, it use local credentials.

docs/commands/cp.md
amritghimire: For source, when downloading the file.  This implementation was carried on from already existing one.

docs/commands/cp.md
amritghimire: Using datachain with local credentials.

docs/commands/cp.md
amritghimire: This is also when using the catalog.cp. I carried over its existing implementation as it is.

docs/commands/cp.md
amritghimire: Updated.

src/datachain/cli/__init__.py
amritghimire: Renamed it, let me know if it works or feel free to suggest one that works,

src/datachain/cli/commands/storage/base.py
amritghimire: This is not down the stack. This is the helper functions inside the cli commands module itself.

src/datachain/cli/commands/storage/base.py
amritghimire: Protected in which essense? In Studio, it is rate limited along with proper authentication and authorization.

src/datachain/cli/commands/storage/base.py
amritghimire: We needed upload logs for both studio based and local based implementation since the upload logic is based on presigned url. The rm and move part are only needed in local credentials based approach. So, they are included only in the necessary class. 

It is not save_activity_logs because it saves the upload logs only. 

About protected, yes it is.

src/datachain/cli/commands/storage/base.py
amritghimire: This is based on what is implemented between frontend and backend in Studio. It basically gets the file paths where it is uploaded mapped to where the file was uploaded from to get the source.

src/datachain/cli/commands/storage/base.py
amritghimire: This is for local path as you see from name of function, we are uploading the files. It won't make extra api call.

src/datachain/cli/commands/storage/local.py
amritghimire: Sure, I will look into that.

src/datachain/cli/commands/storage/local.py
amritghimire: Updated.

amritghimire:
> It seems a bit too complicated. Let's review and try to use DataChain itself to do these operations. Our high level commands should be just wrappers on top, not custom implementations (we'll have otherwise to do tons of optimization, then support all of this, etc, etc)

Updated it to use datachain when possible. 

> Also, we need to review docs, cleanup, etc.
Updated docs to cleanup little. Let me know wordings that you feel like needs changing. 


> Also, in some case it seems it is not optimal - doing extra calls to APIs which is expensive.

I am not sure we have any extra api calls IIRC

docs/commands/cp.md
shcheklein: what if token is set for Studio? will it start using it automatically?

amritghimire: No, it was by that previous but with the comments I changed it so that it uses local unless -s is specified explicitly

docs/commands/cp.md
```diff
@@ -0,0 +1,129 @@
...
+* `-h`, `--help` - Show the help message and exit
+* `-v`, `--verbose` - Be verbose
+* `-q`, `--quiet` - Be quiet
+
+## Copy Operations
+
+The command supports two main modes of operation:
+
+### Default Mode
+By default, the command operates directly with clouds using credentials in ypur system, supporting various copy scenarios between local and remote storage.
```
shcheklein: typos (run some tool / plugin to catch those)

docs/commands/cp.md
```diff
@@ -0,0 +1,129 @@
...
+* `--update` - Update cached list of files for the source when downloading from cloud using local credentials.
+* `-h`, `--help` - Show the help message and exit
+* `-v`, `--verbose` - Be verbose
+* `-q`, `--quiet` - Be quiet
+
+## Copy Operations
+
+The command supports two main modes of operation:
+
+### Default Mode
```
shcheklein: Two many headers, when we have a single sentence inside it doesn't make much sense. Use some other style please, otherwise it looks noisy

docs/commands/cp.md
```diff
@@ -0,0 +1,129 @@
...
+
+The command supports the following storage protocols:
+- **Local file system**: Direct paths (e.g., `/path/to/directory` or `./relative/path`)
+- **AWS S3**: `s3://bucket-name/path`
+- **Google Cloud Storage**: `gs://bucket-name/path`
+- **Azure Blob Storage**: `az://container-name/path`
+
+## Examples
+
+The command automatically determines the operation type based on the source and destination protocols:
```
shcheklein: put it into description, but overall it is not very informative (AI generated?) ...  let's still keep simplifying the structure and the content here

docs/commands/cp.md
```diff
@@ -0,0 +1,129 @@
...
+- Uses the local filesystem's native copy operation
+- Fastest operation as no network transfer is involved
+- Supports both files and directories
+
+```bash
+datachain cp /path/to/source/file.py /path/to/destination/file.py
+datachain cp -r /path/to/source/directory /path/to/destination/directory
+```
+
+### Local to Remote
```
shcheklein, line 58: I don't think we need so many examples ...

docs/commands/cp.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Additional Examples
+
+```bash
+# Copy with specific team:
+datachain cp -s --team other-team /path/to/file.txt s3://my-bucket/data/file.txt
+```
+
+
+## Notes
```
shcheklein: this should be somewhere before examples

docs/commands/cp.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Arguments
+
+* `source_path` - Path to the source file or directory to copy
+* `destination_path` - Path to the destination file or directory to copy to
+
+## Options
+
+* `-r`, `-R`, `--recursive` - Copy directories recursively
+* `--team TEAM` - Team name to use the credentials from.
```
shcheklein: what is default?

src/datachain/cli/__init__.py
```diff
@@ -108,15 +115,36 @@ def handle_command(args, catalog, client_config) -> int:
...
 
-def handle_cp_command(args, catalog):
-    catalog.cp(
-        args.sources,
-        args.output,
-        force=bool(args.force),
-        update=bool(args.update),
-        recursive=bool(args.recursive),
-        no_glob=args.no_glob,
+def _get_storage_implementation(args: "Namespace", catalog: "Catalog"):
```
shcheklein: rename the function

src/datachain/cli/__init__.py
```diff
@@ -108,15 +115,36 @@ def handle_command(args, catalog, client_config) -> int:
...
+        LocalCredentialsBasedFileHandler,
+        StorageCredentialFileHandler,
     )
+    from datachain.config import Config
+
+    config = Config().read().get("studio", {})
+    token = config.get("token")
+    studio = False if not token else args.studio_cloud_auth
+    return (
+        StorageCredentialFileHandler(args, catalog)
```
shcheklein: name is misleading (or I don't understand it) - why is it `StorageCredentialFileHandler` ... why `Storage` is used there?

src/datachain/cli/__init__.py
```diff
@@ -108,15 +115,36 @@ def handle_command(args, catalog, client_config) -> int:
...
+        if studio
+        else LocalCredentialsBasedFileHandler(args, catalog)
+    )
+
+
+def handle_cp_command(args, catalog):
+    storage_implementation = _get_storage_implementation(args, catalog)
```
shcheklein: `storage_implementation`  - rename

src/datachain/cli/commands/storage/base.py
shcheklein: hmm, i think it is fine to use in `_get_storage_implementation` ... but this looks like actual business logic (no matter if it CLI or not). All the logic can be contained inside CLI, doesn't mean we can push it down too much. It should be staying strictly in the first layer where we parse args and before we send it to the next layers (handlers)

src/datachain/cli/commands/storage/base.py
shcheklein: > About protected, yes it is.

address it?

> The rm and move part are only needed in local credentials based approach.

I'm not sure I understand this ... could you clarify please?

src/datachain/cli/commands/storage/local.py
```diff
@@ -0,0 +1,111 @@
+from pathlib import Path
+
+from datachain.cli.commands.storage.base import CredentialBasedFileHandler
+from datachain.error import DataChainError
+from datachain.lib.dc.storage import read_storage
+
+
+class LocalCredentialsBasedFileHandler(CredentialBasedFileHandler):
+    def cp(self):
+        from datachain.client.fsspec import Client
```
shcheklein, line 9: this is still way too low level  ... what do we need from to_storage to make it look as simple as:


`read_storage(source_path, update=update).to_storage(....)`

?

docs/commands/mv.md
```diff
@@ -0,0 +1,74 @@
+# mv
+
+Move storage files and directories through Studio.
```
shcheklein: why is it only "through" Studio?

docs/commands/mv.md
```diff
@@ -0,0 +1,74 @@
...
+This command moves files and directories within storage. The move operation is performed within the same bucket - you cannot move files between different buckets. The command supports both individual files and directories, with the `--recursive` flag required for moving directories.
+
+## Arguments
+
+* `path` - Path to the storage file or directory to move
+* `new_path` - New path where the file or directory should be moved to
+
+## Options
+
+* `--recursive` - Move recursively
```
shcheklein, line 23: is it a regular thing to do require `--recursive` for mv?

docs/commands/mv.md
```diff
@@ -0,0 +1,74 @@
...
+* `--team TEAM` - Team name to move storage contents from
```
shcheklein: what does it mean?

please review and cleanup all docs carefully

src/datachain/cli/commands/storage/base.py
```diff
@@ -0,0 +1,54 @@
...
+        self,
+        destination_path: str,
+        file_paths: dict,  # {destination_path: size}
+    ):
+        from datachain.remote.storages import get_studio_client
+
+        try:
+            studio_client = get_studio_client(self.args.team)
+        except DataChainError:
+            return
```
shcheklein, line 61: why can it happen? should we log / print something here?

src/datachain/cli/commands/storage/base.py
shcheklein: > It is not save_activity_logs because it saves the upload logs only.

I mean something like "save_upload_activilty_logs"

src/datachain/remote/storages.py
```diff
@@ -0,0 +1,193 @@
...
+from datachain.cli.commands.storage.utils import build_file_paths, validate_upload_args
+from datachain.error import DataChainError
+
+if TYPE_CHECKING:
+    from fsspec import AbstractFileSystem
+
+    from datachain.remote.studio import StudioClient
+
+
+def get_studio_client(team: Optional[str] = None):
```
shcheklein: do we have something like this already / for other operations (e.g. when we read datasets from Studio)

src/datachain/remote/storages.py
```diff
@@ -0,0 +1,193 @@
...
+    recursive: bool = False,
+    local_fs: "AbstractFileSystem" = None,
+):
+    studio_client = get_studio_client(team)
+
+    is_dir = validate_upload_args(source_path, recursive, local_fs)
+    file_paths = build_file_paths(source_path, destination_path, local_fs, is_dir)
+    response = _get_presigned_urls(studio_client, destination_path, file_paths)
+
+    for dest_path, src_path in file_paths.items():
```
shcheklein: this will be very slow ... and we'll have to reinvent multithreading here, etc, etc .. how can we avoid this?

(again, I really don't like the amount of reimplementation, custom logic that we do for "what-should-be-basic" operations for DataChain)

tests/func/test_storage_commands.py
```diff
@@ -0,0 +1,200 @@
...
+    )
+    assert result == 0
+    assert (tmp_dir / "file1.txt").read_text() == "file1"
+
+    history = requests_mock.request_history
+    assert len(history) == 2
+    assert history[1].url == "https://example.com/download"
+
+
+def test_cp_s3_to_s3(requests_mock, capsys, studio_token, tmp_dir):
```
shcheklein: it is probably not a functional test

shcheklein: (and review other tests here)

tests/func/test_storage_commands.py
shcheklein, line 1: I'm not sure we are testing much in this case tbh ... we just test some implementation details (that a particular API is called with some specific arguments, right? Those are not functional tests. We have actual cloud implementations, and can use those to test things

amritghimire: Yes, this is to test everything is working for Studio based credentials which is why this checks that a particular API is called. If you have any idea on how to test that otherwise Please let me know?

tests/func/test_storage_commands.py
amritghimire: Why not? From perspective of datachain, it starts with the cli invocation (main) and reaches to studio (third party).

shcheklein: function == actually doing some action (e.g. actually creates files). In this case everything is mocked, even CLI is called via direct main, not a subprocess or something

(usually, in most cases if you have mocks you are not doing a functional test)

src/datachain/remote/storages.py
amritghimire: I can't see other option because we are uploading files one by one as we already do in FE. We can implement multithreading at most to improve this. 

>  I really don't like the amount of reimplementation

What reimplementation? This is to upload the files using studio credentials. So, this is the only way. Otherwise we need to pull the actual credentials first. I don't think we want that.

src/datachain/remote/storages.py
amritghimire: No we don't.

src/datachain/cli/commands/storage/base.py
amritghimire: If user has not logged in to Studio. We don't want to throw error here. We just want to add the logs in studio if only they are logged in

tests/func/test_storage_commands.py
shcheklein: if we don functional tests - use actual files and storages. In Studio mode - generate actual signed URLs

(not saying we have to do this, but current tests are not very meaningful to my mind)

docs/commands/mv.md
amritghimire: I am not sure, we do have it with the call to fs.mv . LMK if we want to remove this.,

src/datachain/remote/storages.py
shcheklein: check `is_token_set`, check studio.py in general, check the login inside StudioClient

even if we need this call ... it belongs to studio.py most likely. I hope we don't need it though

src/datachain/cli/commands/storage/base.py
shcheklein: do you need to raise an exception then? can it be Optional[StudioClient] ? It is not really an "exceptional" situation in your case it seems

src/datachain/remote/storages.py
shcheklein: let's brainstorm in Slack, each thing - one at a time ...

src/datachain/cli/commands/storage/local.py
amritghimire: A few things:
1. read_storage(source_path, update=update).to_storage(....) doesn't work effectively with a single file. Copying test.py to test2.py makes a directory named test.py and create test2.py. 
2. It doesn't return the files it copied from to the files it copied to. We need that to save the upload logs. 


Changing them are out of the scope for this PRs.

shcheklein: > Changing them are out of the scope for this PRs.

I think it is better to improve datachain carefully to do this, instead of growing the amount of custom implementatiins

docs/commands/cp.md
amritghimire: I think otherwise, There are exactly 4 different types of examples to show what are supported. Within each type, we have file and folders each using local and studio authentication. 

If you think otherwise, let me know which examples to keep.

shcheklein: just keep 2-3 examples, command and options should be more or less self descriptive

or make examples more comprehensive - e.g. do Studio auth, etc ...

again, otherwise we just repeat the option descritptions

tests/func/test_storage_commands.py
amritghimire: > use actual files and storages

We already have one for actual files and storages under `test_cli_e2e` with actual storage s3. This is one for studio.

tests/func/test_storage_commands.py
amritghimire: I am not sure I agree, if thats the case every tests we have that tests we are calling correct api calls to third party should be moved elsewhere. Just a quick search shows other tests as we have https://github.com/search?q=repo%3Aiterative%2Fdatachain+mock+path%3Atests%2Ffunc&type=code. 

It is still functional because it is doing some function i.e. calling correct API IMO

src/datachain/cli/commands/storage/base.py
amritghimire: > address it? 

Address it how? 

> The rm and move part are only needed in local credentials based approach.

We don't need to send the activity logs for removing or moving using studio credentials because they are already handled in Studio. Bottomline is we do have helper functions for saving activity logs for removing and moving. Just they are at subclass where it actually is used.

tests/func/test_storage_commands.py
shcheklein: > We already have one for actual files and storages under test_cli_e2e with actual storage s3. This is one for studio.

does it test what we need or some legacy stuff? again, I don't think it is very valuable to see that studio API is called. Even if we keep it - let's at least keep one per command (simplify), let's move them to unit

src/datachain/cli/commands/storage/base.py
amritghimire: Removed the helper function altogether.

docs/commands/cp.md
amritghimire: > just keep 2-3 examples, command and options should be more or less self descriptive

Which ones? Asking before making any changes to avoid back and forth.

tests/func/test_storage_commands.py
amritghimire: > does it test what we need or some legacy stuff?

It tests what we want. It copies all subdirectories and file tree from s3 bucket to local and check if the tree resulting is correct one.

src/datachain/remote/storages.py
amritghimire: Used AsyncMapper from datachain to run the uploads in a asynchronous tasks.

tests/func/test_storage_commands.py
amritghimire: Moved the logic to unit tests folder. We have single tests for a each functions. 

I still don't believe this is unit test. Unit test usually isolates and tests a single function or code segment. Here we are testing complete E2E flow from starting entry to our codebase to the end segment of our test segment while mocking the part where we go to third party like studio. To make this functional according to you, we must actually call studio api and test its response i.e. we need to have a studio instance. 

But since this PR is already stuck for more than a month, I moved the tests to unit tests. We still have few tests that uses mock in functional tests like https://github.com/iterative/datachain/blob/main/tests/test_cli_e2e.py and as seen in above search. LMK if we want to move those to unit tests too. We can create a followup PR.

amritghimire:
@shcheklein Would love your review

shcheklein:
@amritghimire it is on my list, I'll get to it as soon as I can. Thanks!

docs/commands/cp.md
```diff
@@ -0,0 +1,109 @@
...
+
+```usage
+usage: datachain cp [-h] [-v] [-q] [-r] [--team TEAM]
+    [-s] [--anon] [--update]
+    source_path destination_path
+```
+
+## Description
+
+This command copies files and directories between local and/or remote storage. This uses the credentials in your system by default or can use the cloud authentication from Studio.
```
ilongin, line 15: Probably I'm missing something, but what was the motivation to develop this on our own in `datachain` CLI when there are already some tools doing the same thing and are more specialized (e.g https://github.com/rclone/rclone)? 

To me this almost pollutes our CLI / API as it adds completely different domain that doesn't look related to datasets (and other related things)  .. not to mention it's added at root level of CLI.

src/datachain/lib/dc/datachain.py
```diff
@@ -2517,6 +2517,7 @@ def to_storage(
         num_threads: Optional[int] = EXPORT_FILES_MAX_THREADS,
         anon: Optional[bool] = None,
         client_config: Optional[dict] = None,
+        relative_to: Optional[str] = None,
```
ilongin, line 2555: I don't understand what this actually does? Also, it's missing in docstring of this method.

docs/commands/cp.md
```diff
@@ -0,0 +1,109 @@
...
+```
+
+### Remote to Remote
+
+**Operation**: Copy within cloud storage
+- Copies files between locations between cloud storages
+- Requires `--recursive` flag for directories
+
+```bash
+# Copy within same bucket
```
ilongin, line 104: Is it possible to copy from one remote (bucket or even remote type) to another one?

src/datachain/cli/commands/storage/studio.py
```diff
@@ -0,0 +1,106 @@
+from typing import TYPE_CHECKING
+
+from datachain.cli.commands.storage.base import CredentialBasedFileHandler
+from datachain.error import DataChainError
+
+if TYPE_CHECKING:
+    from datachain.client.fsspec import Client
+
+
+class StudioAuthenticatedFileHandler(CredentialBasedFileHandler):
```
ilongin, line 10: Why do we need these long generic words like `Authenticated` or `CredentialBased` inside class names?

src/datachain/cli/commands/storage/studio.py
```diff
@@ -0,0 +1,106 @@
...
+            self.args.destination_path,
+            self.args.team,
+            destination_fs,
+        )
+
+    def copy_cloud_to_cloud(self, source_cls: "Client"):
+        from datachain.remote.storages import copy_inside_storage
+
+        copy_inside_storage(
+            self.args.source_path,
```
ilongin: What is `self.args`? Can we extract those `source_path` and other from `self.args` before it reaches this class?

docs/commands/cp.md
amritghimire: Yes.

docs/commands/cp.md
amritghimire: I believe the main requirements are these two:
1. To be able to track the changes made from CLI in studio file browser as requested by the consumer
2. To be able to use the credentials stored in Studio to perform file operations 

For that, this is used.

src/datachain/lib/dc/datachain.py
amritghimire: When you try to copy the files from a storage to another storage using `to_storage`, we used to be able to use two different relevant value to placement. One is filename, another is fullpath. With filename, all files are dumped in parent directory i.e. file tree is not maintained. With full path, the full path is also dumped. To implement cp, we needed to be able to maintain the tree relative to certain path. 

i.e. If we copy a file from `directory/deep/nested/test.txt`  using `datachain cp -r directory/ destination/` it should create directory under destination using deep/nested/ and create file there.

src/datachain/cli/commands/storage/studio.py
amritghimire: There were some discussions regarding this. Not sure what exactly to use. I am open to suggestions here.

src/datachain/cli/commands/storage/studio.py
amritghimire: Updated and extracted.

src/datachain/lib/dc/datachain.py
shcheklein: Is it only about single file case? Not sure why the full path mode was not enough tbh?

amritghimire: > Is it only about single file case? Not sure why the full path mode was not enough tbh?

Not single file but nested file tree.

amritghimire: > > Is it only about single file case? Not sure why the full path mode was not enough tbh?
> 
> Not single file but nested file tree. 
> 

Will add example demonstration tomorrow

amritghimire: Imagine a tree for the `gs://amrit-datachain-test/playground/test_folder` path as:
```
├── deep
│   └── nested
│       └── test.py
├── test_1.txt
├── test_2.txt
├── test-1.py
└── test-4.py

3 directories, 5 files
```

Now, if I copy the file to a output with following code:
```python
import datachain as dc
dc.read_storage("gs://amrit-datachain-test/playground/test_folder").to_storage("/tmp/copy_destination", placement="fullpath")
```

It gave the following tree:
```
/tmp/copy_destination
└── amrit-datachain-test
    └── playground
        └── test_folder
            ├── deep
            │   └── nested
            │       └── test.py
            ├── test_1.txt
            ├── test_2.txt
            ├── test-1.py
            └── test-4.py

6 directories, 5 files
```

What we wanted to achieve was as: 
```python
import datachain as dc
dc.read_storage("gs://amrit-datachain-test/playground/test_folder").to_storage(
    "/tmp/copy_destination_normpath",
    placement="normpath",
    relative_to="playground/test_folder"
)
```

This gave the following output:
```
/tmp/copy_destination_normpath
├── deep
│   └── nested
│       └── test.py
├── test_1.txt
├── test_2.txt
├── test-1.py
└── test-4.py

3 directories, 5 files
```

cc. @ilongin @shcheklein

src/datachain/remote/studio.py
```diff
@@ -471,3 +476,127 @@ def cancel_job(
...
+        data = {
+            "bucket": bucket,
+            "newPath": new_subpath,
+            "oldPath": subpath,
+            "recursive": recursive,
+            "remote": remote,
+            "team": self.team,
+        }
+
+        return self._send_request("datachain/storages/files/mv", data, method="POST")
```
shcheklein, line 519: We should probably unify this with `ls` command that we have for storages. It has a different API atm AFAIU from a different PR

src/datachain/remote/studio.py
```diff
@@ -471,3 +476,127 @@ def cancel_job(
...
+        bucket, _ = client.split_url(destination_path)
+
+        data = {
+            "bucket": bucket,
+            "paths": paths,
+            "remote": remote,
+            "team": self.team,
+        }
+        return self._send_request(
+            "datachain/storages/batch-presigned-urls", data, method="POST"
```
shcheklein, line 557: what scope btw is required for all these commands?

src/datachain/remote/storages.py
```diff
@@ -0,0 +1,230 @@
...
+
+    download_chain = read_values(
+        info=[
+            UploadFileInfo(dest_path=dest_path, src_path=src_path)
+            for dest_path, src_path in file_paths.items()
+        ],
+    )
+
+    download_chain = (
+        download_chain.settings(cache=True, parallel=10).map(
```
shcheklein: parallel - why 10? just put True? ideally there should be a setting per CLI itself to specify this

src/datachain/remote/storages.py
```diff
@@ -0,0 +1,230 @@
...
+    recursive: bool = False,
+    local_fs: "AbstractFileSystem" = None,
+):
+    studio_client = StudioClient(team=team)
+
+    is_dir = validate_upload_args(source_path, recursive, local_fs)
+    file_paths = build_file_paths(source_path, destination_path, local_fs, is_dir)
+    response = _get_presigned_urls(studio_client, destination_path, file_paths)
+
+    download_chain = read_values(
```
shcheklein, line 41: so, we have already abstractions like File and File.upload, etc in datachain ... we must do a maximum effort to reuse those (improve them if needed) ... we have datachain itlsef with File that can handle source path at least ... let's please think how / if we can use it in both modes

src/datachain/remote/storages.py
```diff
@@ -0,0 +1,230 @@
...
+    src_path: str
+
+
+class DownloadFileInfo(DataModel):
+    filename: str
+    url: str
+
+
+def upload_to_storage(
+    source_path: str,
```
shcheklein, line 29: can we operate with File object here? can we do the same for dest?

shcheklein: instead of creating new wrappers like UploadFileInfo, etc

src/datachain/remote/storages.py
```diff
@@ -0,0 +1,230 @@
...
+    destination_path: str,
+    team: Optional[str] = None,
+    recursive: bool = False,
+    local_fs: "AbstractFileSystem" = None,
```
shcheklein, line 33: can DataChain handle this internally via File abstraction? we must do a maximum effort to use it across both scenarios, instead of manipulating low level things like file systems directly

src/datachain/remote/storages.py
```diff
@@ -0,0 +1,230 @@
...
+                info.dest_path, info.src_path, response, local_fs
+            )
+        )
+    ).exec()
+
+    print(f"Successfully uploaded {len(file_paths)} file(s)")
+    return file_paths
+
+
+def download_from_storage(
```
shcheklein, line 60: can we use to_storage? let's try to scope what would it take for it to work through Studio? can it be for example some abstract Studio file system or something?

src/datachain/cli/commands/storage/local.py
```diff
@@ -0,0 +1,120 @@
...
+        is_file = source_fs.isfile(source_path)
+        if not self.recursive and source_fs.isdir(source_path):
+            raise ValueError("Source is a directory, but recursive is not specified")
+
+        chain = read_storage(source_path, update=update, anon=self.anon).settings(
+            cache=True, parallel=10
+        )
+        file_paths = {}
+
+        def _calculate_full_dst(file) -> str:
```
shcheklein, line 33: types

src/datachain/cli/commands/storage/local.py
```diff
@@ -0,0 +1,120 @@
...
+        source_cls = Client.get_implementation(source_path)
+        source_fs = source_cls.create_fs()
+        _, relative_to = source_cls.split_url(source_path)
+
+        update = self.update
+        if source_cls.protocol == "file":
+            update = True
+            relative_to = None
+
+        is_file = source_fs.isfile(source_path)
```
shcheklein, line 24: let's try to decide on this inside function (e.g. if this is a directory path ends in chain with / or we have many files for a single source, if this is a file we can potentially also see it?)

bottomline - DataChain should be providing proper abstractions / tools to work with this ...

src/datachain/cli/commands/storage/local.py
```diff
@@ -0,0 +1,120 @@
...
+        fs = client_cls.create_fs()
+        fs.rm(self.path, recursive=self.recursive)
+        if client_cls.protocol != "file":
+            _, path = client_cls.split_url(self.path)
+            self.save_deleted_logs(self.path, [path])
+
+        print(f"Deleted {self.path}")
+
+    def mv(self):
+        from datachain.client.fsspec import Client
```
shcheklein, line 74: can DataChain also be used here (e.g. to handle scale since we are saving files, properly parallellize, etc) ... also to unify with Studio code path

src/datachain/remote/studio.py
amritghimire: > We should probably unify this with ls command that we have for storages. It has a different API atm AFAIU from a different PR

This is for moving files, ls is for listing files. We don't have separate ls endpoint or feature implemented in this PR.

src/datachain/remote/studio.py
amritghimire: STORAGES scope.

src/datachain/remote/storages.py
amritghimire:
```suggestion
        download_chain.settings(cache=True, parallel=True).map(
```

src/datachain/cli/commands/storage/local.py
amritghimire: No, because we don't do move with datachain, but copying files. For move operation, it is always simple to use filesystem since it usually is just changing the references underneath. The file itself will not be physically moved, so the operation will be almost instantaneous in majority of local move cases.

src/datachain/cli/commands/storage/local.py
amritghimire: I agree with the bottomline, but this PR is already big enough to work on that, Can be followup, but for now, it doesn't have a way to copy the file properly for a single file. For file, we use file.save as we have in datachain and for folder, we use the to_storage part.

src/datachain/remote/storages.py
amritghimire: No, again because we need to pass the credentials stored in studio and that is not a good idea IMO. Adding studio file system is whole another thing IMO.

src/datachain/remote/storages.py
amritghimire: Local filesystem here is to identify if it is a directory or not and build the file paths necessary.

src/datachain/remote/storages.py
amritghimire: Not exactly at this moment because it can either be a directory or a file. We form the chain using this information below.

src/datachain/cli/commands/storage/local.py
```diff
@@ -0,0 +1,124 @@
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+from datachain.cli.commands.storage.base import CredentialBasedFileHandler
+from datachain.lib.dc.storage import read_storage
+
+if TYPE_CHECKING:
+    from datachain import File
```
amritghimire:
```suggestion
    from datachain.lib.file import File
```

amritghimire:
@shcheklein Can you please take a final look and let me know what we want to change if anything?

shcheklein:
@amritghimire yep, I remember about this one. It is the only PR probably where I'll be very conservative, so please be patient with me a bit :). As I mentioned I'm not comfortable merging quite a lot of stuff where it seems DataChain core should be handling things. I'll check one more asap. It just takes a bit more effort to give some meaningful feedback, since it's not an easy one.

shcheklein:
Closing this for now, @amritghimire we'll get back to it after a few other things hopefully and review it again. Apologies for shifting the priorities a bit - it seems it is less important now and I'm not we want to merge things that won't be actively used, especially if it is something pretty large. It is a good base to get back to it a bit later though. We'll definitely need it anyways.

PR #1223, opened 2025-07-09
migrate to 4.0 datasets from HF

shcheklein:
Fixing broken HF integration after 4.0 release https://github.com/huggingface/datasets/releases/tag/4.0.0

## Summary by Sourcery

Migrate the Hugging Face dataset integration to be compatible with datasets v4.0 by updating the dependency version and adapting feature conversion logic and data models to the new API structure.

Bug Fixes:
- Restore Hugging Face integration compatibility after upgrading to datasets v4.0

Enhancements:
- Replace deprecated Sequence feature handling with generic dict and List-based logic in convert_feature and _feature_to_chain_type
- Remove HFAudio.path field and adjust HFAudio instantiation to accept only array and sampling_rate

Build:
- Bump datasets[audio,vision] dependency to >=4.0.0 in pyproject.toml

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -34,9 +34,6 @@ jobs:
           enable-cache: true
           cache-suffix: lint
           cache-dependency-glob: pyproject.toml
-          # revert after this is fixed
-          # https://github.com/wntrblm/nox/issues/953
-          version: ">=0.6,<0.7"
 
       - name: Install nox
         run: uv pip install nox --system
@@ -81,9 +78,6 @@ jobs:
           fetch-depth: 0
           ref: ${{ github.event.pull_request.head.sha || github.ref }}
 
-      - name: Set up FFmpeg
-        uses: AnimMouse/setup-ffmpeg@v1
-
       - name: Set up Python ${{ matrix.pyv }}
         uses: actions/setup-python@v5
         with:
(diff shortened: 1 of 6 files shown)


.github/workflows/tests.yml
```diff
@@ -81,9 +78,6 @@ jobs:
           fetch-depth: 0
           ref: ${{ github.event.pull_request.head.sha || github.ref }}
 
-      - name: Set up FFmpeg
```
shcheklein, line 84: It doesn't install libraries properly, only binaries. There might be a way to do some exports, but it will depend on a platform, requires more research. It is caching it though ...

.github/workflows/tests.yml
```diff
@@ -95,9 +89,22 @@ jobs:
...
+
+      - name: Install FFmpeg on Windows
+        if: runner.os == 'Windows'
+        run: choco install ffmpeg
+
+      - name: Install FFmpeg on macOS
+        if: runner.os == 'macOS'
+        run: |
+          brew install ffmpeg
+          echo 'DYLD_FALLBACK_LIBRARY_PATH=/opt/homebrew/lib' >> "$GITHUB_ENV"
```
shcheklein, line 101: Need bc of this it seems https://github.com/pytorch/torchcodec/issues/570

PR #1230, opened 2025-07-11
fix(setup): allow empty values for args

shcheklein:
Allow sending empty (`{}`, `None`, `""`, etc) values to setup. Before it was failing with an index error.

### DIFF
--- a/src/datachain/lib/signal_schema.py
+++ b/src/datachain/lib/signal_schema.py
@@ -446,14 +446,14 @@ class SignalSchema:
                 res[db_name] = python_to_sql(type_)
         return res
 
-    def row_to_objs(self, row: Sequence[Any]) -> list[DataValue]:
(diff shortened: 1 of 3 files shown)


src/datachain/lib/signal_schema.py
```diff
@@ -446,14 +446,14 @@ def to_udf_spec(self) -> dict[str, type]:
...
 
-    def row_to_objs(self, row: Sequence[Any]) -> list[DataValue]:
+    def row_to_objs(self, row: Sequence[Any]) -> list[Any]:
         self._init_setup_values()
 
-        objs: list[DataValue] = []
+        objs: list[Any] = []
         pos = 0
         for name, fr_type in self.values.items():
-            if self.setup_values and (val := self.setup_values.get(name, None)):
```
shcheklein, line 455: Here: we skip it if value is not `boolen(val)`. So, we skip if we have value set to `{}` for example.

Issue #1212, opened 2025-07-08
Allow whole table aggregation into a single group


dreadatour:
By not setting the `partition_by` param or set `partition_by=None`.

PR #1231, opened 2025-07-12
Aggregate table into single group if no partition_by set

dreadatour:
Fix for the https://github.com/iterative/datachain/issues/1212

## Summary by Sourcery

Fix aggregation to group all rows into a single partition when partition_by is unset and add a corresponding unit test

Bug Fixes:
- Default to a single aggregation group when no partition_by is specified

Enhancements:
- Treat None partition_by as an empty list inside agg

Tests:
- Add test to verify aggregation behavior without partition_by combines all rows into one group

### DIFF
--- a/src/datachain/lib/dc/datachain.py
+++ b/src/datachain/lib/dc/datachain.py
@@ -823,9 +823,8 @@ class DataChain:
             chain.save("new_dataset")
             ```
         """
-        # Convert string partition_by parameters to Column objects
-        processed_partition_by = partition_by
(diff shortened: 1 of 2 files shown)


shcheklein (commented):
thanks @dreadatour ! let's please review and fix docs for this (also for complex signals)

dreadatour:
> let's please review and fix docs for this (also for complex signals)

Updated, please, take a look? 🙏

shcheklein (approved):

PR #1233, opened 2025-07-13
Update docs + add more tests

dreadatour:
Tiny improvements:

- Add `dc.func.not_` function + tests
- Add/update docs for DataChain methods: `reset_schema`, `add_schema`, `remove_file_signals`, `sum`, `avg`, `min`, `max` and `chunk` (+ very minor updates in some other methods)
- Fix bug in `SignalSchema._find_in_tree` + add more tests for this method
- Add more tests for DataChain methods: `count`, `distinct` and `filter`

## Summary by Sourcery

Add new conditional helper, improve schema resolution, document and implement aggregation methods, and expand test coverage across DataChain operations

New Features:
- Add dc.func.not_ conditional function

Bug Fixes:
- Fix SignalSchema._find_in_tree lookup logic and extend error handling

Enhancements:
- Enhance and unify docstrings for DataChain methods reset_schema, add_schema, remove_file_signals, sum, avg, min, max, and chunk
- Expose not_ in func module __all__

Tests:
- Add extensive unit tests for DataChain count, distinct, filter, and aggregation methods
- Add tests for the not_ function in mutates and conditional functions
- Parameterize resolve error scenarios in SignalSchema tests
- Add test_column_compute covering sum, avg, min, and max operations on nested data

### DIFF
--- a/src/datachain/lib/dc/datachain.py
+++ b/src/datachain/lib/dc/datachain.py
@@ -974,7 +974,7 @@ class DataChain:
         query_func = getattr(self._query, method_name)
 
         new_schema = self.signals_schema.resolve(*args)
-        columns = [C(col) for col in new_schema.db_signals()]
+        columns = new_schema.db_signals(as_columns=True)
         return query_func(*columns, **kwargs)
 
     @resolve_columns
--- a/src/datachain/lib/dc/datachain.py
+++ b/src/datachain/lib/dc/datachain.py
@@ -33,7 +33,13 @@ from datachain.func import literal
 from datachain.func.base import Function
 from datachain.func.func import Func
 from datachain.lib.convert.python_to_sql import python_to_sql
-from datachain.lib.data_model import DataModel, DataType, DataValue, dict_to_data_model
+from datachain.lib.data_model import (
+    DataModel,
+    DataType,
+    DataValue,
+    StandardType,
+    dict_to_data_model,
+)
 from datachain.lib.file import (
     EXPORT_FILES_MAX_THREADS,
     ArrowRow,
@@ -361,10 +367,20 @@ class DataChain:
         return self
 
     def reset_schema(self, signals_schema: SignalSchema) -> "Self":
+        """Reset the schema of the chain to the provided schema.
+
+        Parameters:
+            signals_schema: The new schema to set for the chain.
+        """
         self.signals_schema = signals_schema
         return self
 
     def add_schema(self, signals_schema: SignalSchema) -> "Self":
+        """Extend existing chain schema with the provided one.
+
+        Parameters:
+            signals_schema: The schema to add to the existing schema.
+        """
         self.signals_schema |= signals_schema
         return self
 
@@ -1446,6 +1462,7 @@ class DataChain:
         )
 
     def remove_file_signals(self) -> "Self":
+        """Remove file signals from the chain schema."""
         schema = self.signals_schema.clone_without_file_signals()
         return self.select(*schema.values.keys())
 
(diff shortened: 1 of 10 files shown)


src/datachain/lib/dc/datachain.py
```diff
@@ -361,10 +367,20 @@ def reset_settings(self, settings: Optional[Settings] = None) -> "Self":
...
+        """Reset the schema of the chain to the provided schema.
+
+        Parameters:
+            signals_schema: The new schema to set for the chain.
+        """
         self.signals_schema = signals_schema
         return self
 
     def add_schema(self, signals_schema: SignalSchema) -> "Self":
+        """Extend existing chain schema with the provided one.
```
shcheklein: does it mean it will be exposed in docs?

(I think we don't want that, these are internal methods it seems, I would even try to make the protected / private)

src/datachain/lib/dc/datachain.py
```diff
@@ -2268,21 +2292,49 @@ def from_records(
...
+        return self._extend_to_data_model("avg", col)
 
-    def max(self, fr: DataType):  # type: ignore[override]
-        """Compute the maximum of a column."""
-        return self._extend_to_data_model("max", fr)
+    def min(self, col: str) -> StandardType:  # type: ignore[override]
+        """Compute the minimum of a column.
+
+        Parameters:
+            col: The column to compute the minimum for.
```
shcheklein, line 2310: can we inline some example of how that column can look like?

src/datachain/lib/dc/datachain.py
```diff
@@ -1446,6 +1462,7 @@ def to_pytorch(
         )
 
     def remove_file_signals(self) -> "Self":
+        """Remove file signals from the chain schema."""
```
shcheklein: same here, I keep stumbling into this helper - but it should not be part of the user facing API, should be some protected / internal helper

src/datachain/lib/dc/datachain.py
dreadatour: Can not find any usage of this method in the source code, I was thinking it is user-facing API.
Should I remove this method?

src/datachain/lib/dc/datachain.py
dreadatour: This method is used only here: https://github.com/iterative/datachain/blob/main/src/datachain/lib/pytorch.py#L128
Since it is very simple, I can inline it there.

src/datachain/lib/dc/datachain.py
shcheklein: yes, absolutely! I'm not sure I understand what it was made for ...

src/datachain/lib/dc/datachain.py
dreadatour: Updated (for `sum`, `avg`, `min` and `max` methods), please, take a look.

shcheklein (approved):
(some modifications might be still needed - e.g. hiding removing weird methods)

src/datachain/lib/dc/datachain.py
dreadatour: I am going to remove the following unused methods:
- `reset_settings`
- `reset_schema`
- `add_schema`

shcheklein: `reset_settings` might be needed? (i can see a practical need I think)

dreadatour: Got it. `reset_schema` and `add_schema` were removed.

src/datachain/lib/dc/datachain.py
dreadatour: Removed `remove_file_signals` method, updated the code in [pytorch.py](https://github.com/iterative/datachain/blob/main/src/datachain/lib/pytorch.py#L128).

Issue #1213, opened 2025-07-08
Always add `sys` columns to all tables


dreadatour:
For now in few cases datasets tables have no `sys` columns (`sys__id` and `sys__rand`), for example, aggregation result dataset by default comes without these tables.

We need to:
1. Check all the codebase and always add `sys` columns.
2. Check all the codebase where we are expecting datasets to do not have `sys` columns and remove all hacks we have for this case.
3. Tests.

dreadatour:
Closing this as not relevant anymore.

PR #1234, opened 2025-07-13
Fix partition_by for queries without sys__id

dreadatour:
Part of the https://github.com/iterative/datachain/issues/1213

In case query have no `sys_id` field (for example, `group_by` query), aggregate does not works (see traceback below).

In this PR I am fixing this specific case by creating temp table and copying query result to it.

Test was also added.

```
  File "<string>", line 35, in <module>
  File "/tmp/local/datachain_venv/python3.11/733fe23f9c9da996e532eb256ede753e0e46c859e1fbff91b8ec2ac80ba6364a/lib/python3.11/site-packages/datachain/lib/dc/datachain.py", line 635, in save
    query=self._query.save(
          ^^^^^^^^^^^^^^^^^
  File "/tmp/local/datachain_venv/python3.11/733fe23f9c9da996e532eb256ede753e0e46c859e1fbff91b8ec2ac80ba6364a/lib/python3.11/site-packages/datachain/query/dataset.py", line 1749, in save
    query = self.apply_steps()
            ^^^^^^^^^^^^^^^^^^
  File "/tmp/local/datachain_venv/python3.11/733fe23f9c9da996e532eb256ede753e0e46c859e1fbff91b8ec2ac80ba6364a/lib/python3.11/site-packages/datachain/query/dataset.py", line 1259, in apply_steps
    result = step.apply(
             ^^^^^^^^^^^
  File "/tmp/local/datachain_venv/python3.11/733fe23f9c9da996e532eb256ede753e0e46c859e1fbff91b8ec2ac80ba6364a/lib/python3.11/site-packages/datachain/query/dataset.py", line 609, in apply
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/local/datachain_venv/python3.11/733fe23f9c9da996e532eb256ede753e0e46c859e1fbff91b8ec2ac80ba6364a/lib/python3.11/site-packages/datachain/query/dataset.py", line 578, in create_partitions_table
    query.selected_columns.sys__id,
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/local/datachain_venv/python3.11/733fe23f9c9da996e532eb256ede753e0e46c859e1fbff91b8ec2ac80ba6364a/lib/python3.11/site-packages/sqlalchemy/sql/base.py", line 1631, in __getattr__
    raise AttributeError(key) from err
AttributeError: sys__id
```

## Summary by Sourcery

Enable partitioning for queries that lack a sys__id column by copying the query results into a temporary table before applying partition logic, and add a test to verify aggregation after group_by.

Bug Fixes:
- Support partition_by for queries without a sys__id column by creating a temporary table and copying the query results before partitioning

Tests:
- Add test for aggregate operations following a group_by on queries that do not include sys__id

### DIFF
--- a/src/datachain/query/dataset.py
+++ b/src/datachain/query/dataset.py
@@ -559,7 +559,13 @@ class UDFStep(Step, ABC):
         """
         Create temporary table with group by partitions.
         """
+        # Check if partition_by is set, we need it to create partitions.
         assert self.partition_by is not None
+        # Check if sys__id is in the query, we need it to be able to join
+        # the partition table with the udf table later.
+        assert any(c.name == "sys__id" for c in query.selected_columns), (
+            "Query must have sys__id column to use partitioning."
+        )
 
         if isinstance(self.partition_by, (list, tuple, GeneratorType)):
             list_partition_by = list(self.partition_by)
@@ -606,6 +612,22 @@ class UDFStep(Step, ABC):
 
         # Apply partitioning if needed.
         if self.partition_by is not None:
+            if not any(c.name == "sys__id" for c in query.selected_columns):
+                # If sys__id is not in the query, we need to create a temp table
+                # to hold the query results, so we can join it with the
+                # partition table later.
+                columns = [
+                    c if isinstance(c, Column) else Column(c.name, c.type)
+                    for c in query.subquery().columns
+                ]
+                temp_table = self.catalog.warehouse.create_dataset_rows_table(
+                    self.catalog.warehouse.temp_table_name(),
+                    columns=columns,
+                )
+                temp_tables.append(temp_table.name)
+                self.catalog.warehouse.copy_table(temp_table, query)
+                _query = query = temp_table.select()
+
             partition_tbl = self.create_partitions_table(query)
             temp_tables.append(partition_tbl.name)
             query = query.outerjoin(
--- a/tests/unit/lib/test_partition_by.py
+++ b/tests/unit/lib/test_partition_by.py
@@ -550,3 +550,41 @@ def test_no_partition_by(test_session):
 
     assert len(result) == 1
     assert result[0] == ("f1.txt,f1.txt,f2.txt,f3.txt", 700, 100)
+
+
+def test_aggregate_after_group_by(test_session):
+    files = [
+        File(source="s3://bucket", path="f1.txt", size=100),
+        File(source="s3://bucket", path="f2.txt", size=200),
+        File(source="s3://bucket", path="f1.txt", size=100),  # duplicate
+        File(source="s3://bucket", path="f3.txt", size=300),
+    ]
+    amounts = [10, 20, 30, 40]
+
+    # Create a chain with File objects
+    chain = dc.read_values(
+        file=files,
+        amount=amounts,
+        session=test_session,
+    )
+
+    # Group by file and aggregate
+    grouped = chain.group_by(
+        total_size=dc.func.sum("file.size"),
+        total_amount=dc.func.sum("amount"),
+        partition_by="file",
+    )
+
+    assert sorted(grouped.to_list("file.path", "total_size", "total_amount")) == [
+        ("f1.txt", 200, 40),
+        ("f2.txt", 200, 20),
+        ("f3.txt", 300, 40),
+    ]
+
+    # Now aggregate over the grouped results
+    result = grouped.agg(
+        lambda file, total_amount: [(len(file), sum(total_amount))],
+        output={"files": int, "total": int},
+    ).to_list("files", "total")
+
+    assert result == [(3, 100)]


dreadatour:
> Seems a bit of an overkill tbh.

Agree, but I can not find a proper way to do this without materializing query to the temp table for the `group_by`.

Also the only case I know when query have no `sys__id` present is `group_by` result, which, in general, produce less columns and it should be "ok"-ish to copy this table.

Another possible option for query to do not have `sys__id` is when user select only limited set of columns, but in this case it is safe to manually add `sys__id` field.

> * we do an optimization to use ids
> * but now we have to copy tables to have those ids

No, this is different scenarios.

1. We do have an optimization to use IDs, but this is only for UDFs and for UDFs we always copy table anyway, so this optimization is only to pass rows to the UDF workers. Nothing was changed there.

2. We now have to copy table to have those IDs is because `group_by` query does not have IDs (in general) and it is very tricky to have those IDs in the query and moreover to have those IDs consistent across queries run.
 
> Q: can we avoid copying? (put assert to fail fast though if it's not present) - can all operations include sys_id at the end?

The reason we cannot avoid copying is that some SQL queries do not return IDs in their results, and I have not found a reliable way to add those IDs to the queries.

I could add something like `SELECT random() AS sys__id, ...`, but these IDs would not be idempotent, they would change with each query execution.

I could add something like `SELECT row_number() AS sys__id, ...`, but I can not make `GROUP BY` work with such a complex query.

More details on internals below (in the next comment).

> Q: are we relying on sys_id in some other operations - will we have to copy tables there as well eventually?

No, aggregation with `partition_by` is only place I know where it is crucial to have `sys__id` field in the query. See next comment.

dreadatour:
So why do we need `sys__id` to be set for aggregation with partitioning? Let me start from the very beginning, with obvious things.

DataChain in general follows classic ETL pattern: 

1. **E**xtract data from sources (datasets, s3 buckets, CSV, etc);
2. **T**ransform data with operations (on SQL level and via UDFs);
3. **L**oad (save) results (to datasets, s3 buckets, etc).

**Transformation** can be done on two levels:

1. Raw SQL queries: filtering, merging, limiting, **group by**, etc;
2. Python User-Defined Functions (UDFs): mappers, generators, aggregators, etc.

All transformations can be chained, and results of the previous transformation became an input to the next one.
For SQL queries we are basically using subqueries (it is more complicated, but for now let's assume we are using basic subqueries), and result of the previous step became a subquery to the next step. For UDFs we are "materializing" queries to the temp tables and passing set of IDs to each UDF worker, so it can then fetch basic rows from temp table by IDs and process the rows, saving result to another temp table with results. Next step will do `SELECT * FROM temp_results` and pass it to the next transformer as raw SQL query and so on and so on.

One special case is aggregation with partitioning. Since query from all previous transformers (steps) can be complex, we are doing partitioning the following way:

1. We do have query from all previous steps. This query should have `sys__id` (see below why).
2. Create new temp table with two fields: `sys__id` and `partition_id`.
3. Fill this table using `sys__id` from the query and `partition_id = DENSE_RANK() OVER (PARTITION BY ...)` (see [here](https://www.sqlitetutorial.net/sqlite-window-functions/sqlite-dense_rank/) for SQLite, same for ClickHouse).
4. Next when we do aggregation, we are selecting rows from the query, `JOIN`ed with the temp tables with `partition_id` by `sys__id` with `ORDER BY partition_id, ...`. We can then collect all the rows for each partition and pass them to aggregation UDF.

The reason why we do need to have `sys__id` in query, and it should be set, is because we are running query twice:

1. To fill temp table with `partition_by`
2. To fetch rows for the next step (transformer).

Thus `sys__id` can not be generated (like `SELECT random() AS sys__id`).

---

Moreover, I am now thinking we should always copy query to the temp table, because we are running query twice and we need to be 100% results will be idempotent. We can get rid of temp table with two fields: `sys__id` and `partition_id`, we can use one single temp table for both "materializing" the query into temp table and to store `partition_id`.

dreadatour:
I also did some unsuccessful experiments.

1. Using `row_number` as `sys__id`: `Column udf_3dsrZD.partition_id is not under aggregate function and not in GROUP BY keys`

    ```
    SELECT rowNumberInAllBlocks() + 1 AS sys__id,
           anon_2.file__path AS file__path,
           count(*) AS cnt,
           partition_id
    FROM
      (SELECT ds.sys__id AS sys__id,
              ds.sys__rand AS sys__rand,
              ds.file__source AS file__source,
              ds.file__path AS file__path,
              ds.file__size AS file__size,
              ds.file__version AS file__version,
              ds.file__etag AS file__etag,
              ds.file__is_latest AS file__is_latest,
              ds.file__last_modified AS file__last_modified,
              ds.file__location AS file__location
       FROM ds
       ORDER BY rand() ASC LIMIT 5000) AS anon_2
    LEFT JOIN udf_3dsrZD ON udf_3dsrZD.sys__id = (rowNumberInAllBlocks() + 1)
    GROUP BY file__path
    ```

2. Selecting `sys__id` from the original table when doing `GROUP_BY`: `Column anon_1.sys__id is not under aggregate function and not in GROUP BY keys```

    ```
    SELECT anon_1.sys__id,
           dense_rank() OVER () AS partition_id
    FROM
      (SELECT ds.sys__id AS sys__id,
              ds.sys__rand AS sys__rand,
              ds.file__source AS file__source,
              ds.file__path AS file__path,
              ds.file__size AS file__size,
              ds.file__version AS file__version,
              ds.file__etag AS file__etag,
              ds.file__is_latest AS file__is_latest,
              ds.file__last_modified AS file__last_modified,
              ds.file__location AS file__location
       FROM ds
       ORDER BY rand() ASC LIMIT 5000) AS anon_1
    GROUP BY file__path
    ```

---

I keep investigating, but for now I am thinking creating temp table in case of aggregation with partitioning is a best way to handle this issue and to avoid any possible issues in future with queries running twice (e.g. someone want to use `random()` or something like that).

PR #1241, opened 2025-07-15
Normalize HF dataset feature names

### LINKED ISSUE #1204: Reading from huggingface dataset
### Description

Some datasets on the HF hub have column names that are not valid python names.

Example reproduction:
```pytthon
import datachain as dc
from datasets import load_dataset
ds = load_dataset('toxigen/toxigen-data', 'annotated', split='test')
dc.read_hf(ds)
```

Resulting in:

```text
   ....<snipped>...
    anno = self.output_schema.model_fields[name].annotation
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
KeyError: 'factual?'
```

I believe support to handle these cases already exists and can be fixed by passing `original_names` from the dataset's features.

### Version Info

```Text
datachain = 0.21.1 
python = 3.12.10
datasets = 3.6.0
```

shcheklein:
Fixes #1204 

We convert complex (`dict`) features into Pydantic objects. HF dataset contain any (?) string.

We normalize the names (similar to other readers - e.g. JSON) now, to be able to run `read_hf` on top of such datasets.


tests/unit/lib/test_hf.py
```diff
@@ -50,26 +54,28 @@ def test_hf_class_label():
...
     row = next(iter(gen.process()))
     assert row.seq == [0, 1]
 
 
 def test_hf_sequence_dict():
+    # ? in the column name should be normalized
+    # Check if even nested names are not normalized we handle it correctly
     ds = Dataset.from_dict(
-        {"pokemon": [{"name": ["bulbasaur"]}, {"name": ["squirtle"]}]}
+        {"pokemon": [{"name?": ["bulbasaur"]}, {"name?": ["squirtle"]}]}
```
dreadatour, line 70: Do we also need to check double underscore in tests? Something like `na__me` should also use alias?

shcheklein: I think it is a question to our normalize_names function (that is used in many places) and has a bunch of unit tests, it should be indeed handling double underscores. I'll check that.

PR #1242, opened 2025-07-15
chore(pydantic): use class model fields to remove deprectated warning

shcheklein:
Minor fix to reduce noise from warnings in tests.


src/datachain/lib/signal_schema.py
```diff
@@ -550,7 +550,7 @@ def _set_file_stream(
     ) -> None:
         if isinstance(obj, File):
             obj._set_stream(catalog, caching_enabled=cache)
-        for field, finfo in obj.model_fields.items():
+        for field, finfo in obj.__class__.model_fields.items():
```
0x2b3bfa0: While `obj.__class__` reads nicely, I reckon that `type(obj)` would have been slightly more Pythonic:

```suggestion
        for field, finfo in type(obj).model_fields.items():
```

See e.g. [pydantic@d156ba0 → tests/benchmarks/basemodel_eq_performance.py:58](https://github.com/pydantic/pydantic/blob/d156ba08c140ee1e2b931120cb150080843476fe/tests/benchmarks/basemodel_eq_performance.py#L58) 💅🏼

0x2b3bfa0 (approved):
Thank you! (Linking to https://github.com/pydantic/pydantic/pull/11169 for future references)

src/datachain/lib/signal_schema.py
shcheklein: less Pythonic more ... Ruby? :) (Python has to be __init__(self) something .. :) ) 

I like it though, will update it

0x2b3bfa0: I requested GitHub Copilot's review after mine, and it agrees; not sure if it's because it parsed my comments, though. 📋 

* https://github.com/iterative/datachain/pull/1242#discussion_r2208709506

src/datachain/lib/udf.py
```diff
@@ -282,7 +282,7 @@ def _set_stream_recursive(
 
         # Check all fields for nested File objects, but only for DataModel objects
         if isinstance(obj, DataModel):
-            for field_name in obj.model_fields:
+            for field_name in obj.__class__.model_fields:
```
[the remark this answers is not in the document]
0x2b3bfa0: Duplicate of https://github.com/iterative/datachain/pull/1242#discussion_r2208702023, but thanks for confirming what I said. 😄

src/datachain/lib/convert/flatten.py
```diff
@@ -6,12 +6,14 @@
 
 
 def flatten(obj: BaseModel) -> tuple:
-    return tuple(_flatten_fields_values(obj.model_fields, obj))
+    return tuple(_flatten_fields_values(obj.__class__.model_fields, obj))
```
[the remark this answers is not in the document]
0x2b3bfa0: I almost-agree, but the current syntax doesn't bother me either. 🤷🏼‍♂️

PR #1249, opened 2025-07-18
Refactoring `Catalog.get_dataset()`

ilongin:
Refactoring `Catalog.get_dataset()` to accept `project_name` and `namespace_name` instead of already instantiated `Project`. This is because in more than one place we have project and namespace name available, but don't have whole `Project` instance and we needed to fetch it from DB because of that which was basically +1 SQL command that could've been avoided.

## Summary by Sourcery

Refactor dataset lookup to use namespace and project names instead of passing Project objects throughout the codebase

Enhancements:
- Change Catalog.get_dataset signature to accept namespace_name and project_name parameters
- Update Metastore.get_dataset interface and queries to use namespace and project names
- Modify all internal callers and tests to pass namespace_name and project_name rather than Project instances

### DIFF
--- a/src/datachain/catalog/catalog.py
+++ b/src/datachain/catalog/catalog.py
@@ -680,8 +680,9 @@ class Catalog:
                 ds_namespace, ds_project, ds_name = parse_dataset_name(ds_name)
                 assert ds_namespace
                 assert ds_project
-                project = self.metastore.get_project(ds_project, ds_namespace)
-                dataset = self.get_dataset(ds_name, project)
+                dataset = self.get_dataset(
+                    ds_name, namespace_name=ds_namespace, project_name=ds_project
+                )
                 if not ds_version:
                     ds_version = dataset.latest_version
                 dataset_sources = self.warehouse.get_dataset_sources(
@@ -807,7 +808,11 @@ class Catalog:
             )
         default_version = DEFAULT_DATASET_VERSION
         try:
-            dataset = self.get_dataset(name, project)
+            dataset = self.get_dataset(
+                name,
+                namespace_name=project.namespace.name if project else None,
+                project_name=project.name if project else None,
+            )
             default_version = dataset.next_version_patch
             if update_version == "major":
                 default_version = dataset.next_version_major
@@ -1016,7 +1021,11 @@ class Catalog:
             dc.save(name)
         except Exception as e:  # noqa: BLE001
             try:
-                ds = self.get_dataset(name, project)
+                ds = self.get_dataset(
+                    name,
+                    namespace_name=project.namespace.name,
+                    project_name=project.name,
+                )
                 self.metastore.update_dataset_status(
                     ds,
                     DatasetStatus.FAILED,
@@ -1033,7 +1042,11 @@ class Catalog:
             except DatasetNotFoundError:
                 raise e from None
 
-        ds = self.get_dataset(name, project)
+        ds = self.get_dataset(
+            name,
+            namespace_name=project.namespace.name,
+            project_name=project.name,
+        )
 
         self.update_dataset_version_with_warehouse_info(
             ds,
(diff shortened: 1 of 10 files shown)


src/datachain/catalog/catalog.py
```diff
@@ -1077,22 +1094,23 @@ def get_full_dataset_name(
         return namespace_name, project_name, name
 
     def get_dataset(
-        self, name: str, project: Optional[Project] = None
+        self,
+        name: str,
+        namespace_name: Optional[str] = None,
+        project_name: Optional[str] = None,
```
dmpetrov, line 1100: Let's use full project name instead of additional params - `ns1.ns2.ds3`

dmpetrov (changes requested):
Let's use fully qualified dataset names instead of 3 params: name, namespace, project.

src/datachain/catalog/catalog.py
ilongin: This is internal API where we already have it split this into 3 parts. In public API we use fully qualified name. Honestly I wouldn't change this atm as it would again require me some refactoring which would take time and we already have this kind of signature in other internal methods so not much would be changed. WDYT?

dmpetrov: Let's approve to move fast. But we are accumulating tech depth using different notations in internal and external APIs. Please create a followup issue.

dmpetrov (approved):
LG

src/datachain/catalog/catalog.py
```diff
@@ -1077,22 +1094,23 @@ def get_full_dataset_name(
...
     ) -> DatasetRecord:
         from datachain.lib.listing import is_listing_dataset
 
-        project = project or self.metastore.default_project
+        namespace_name = namespace_name or self.metastore.default_namespace_name
+        project_name = project_name or self.metastore.default_project_name
```
dreadatour, line 1105: Just wonder, if it make sense if `project_name` is set with `namespace_name` is empty (default one) and vise versa (`namespace_name` is set, but `project_name` is default one).

I assume yes, it make sense in some cases, but it also may leads to some weird behavior. Taking the full name (`ns1.ns2.ds3` as @dmpetrov suggests below) may solve this issue (it will not be possible to set namespace name only, without setting project name). Or may be I am wrong and this is a nice feature to have. Just something to keep in mind and think about, may be later.

**Note**: to be honest I am still confused with `namespace` and `project` — which one goes first 😥

dreadatour (approved):
Looks good to me overall 👍

src/datachain/catalog/catalog.py
dreadatour: Just an idea: we can store full name in dataset table as dataset name, this will simplify everything and we don't need anymore to pass namespace and project name everywhere.

ilongin: We don't need to do that as dataset table already have connection to project table which has connection to namespaces table, so it's normalized. Also, when fetching dataset from table object `DatasetRecord` already has `Project` object in it and property method `full_name` which returns namespace, project and dataset name so it's all ok.
We need explicit namespace / project when we are fetching dataset (as here) or saving the new one in DB. The only question is should we everywhere have it split into 3 arguments, as now, or should we have just dataset name which must, by convention, have namespace and project embedded in it, e.g `namespace.project.dataset`. This is what we already have in public API but not in internal ones.

src/datachain/catalog/catalog.py
ilongin: It def makes sense to have just project name set, as then default namespace will be used, but you are right about vice versa .. if someone just sets namespace name then default project will be used which can be little bit weird. I'm not 100% sure if that's ok or not TBH but for now we allow all combinations.

It's `namespace.project.dataset` :D That's hierarchy in DB as well.

PR #1257, opened 2025-07-24
audio conversion without fragment-ing

dmpetrov:
This code generalizes save_audio_fragment() to use cases when conversion to "mp3" is needed but fragmenting is not.

Also, extends AudioFile.save() and more docs.

## Summary by Sourcery

Add support for converting entire audio files without fragmenting, and refactor save_audio_fragment to unify full-file and segment conversion flows with improved validation and error handling.

New Features:
- Introduce audio_to_bytes helper to convert whole audio files to bytes.
- Allow save_audio_fragment to convert the entire file when start is negative.

Enhancements:
- Refactor save_audio_fragment logic to consolidate fragment and full-file conversion paths.
- Improve time-range validation and error handling in save_audio_fragment.

### DIFF
--- a/src/datachain/lib/audio.py
+++ b/src/datachain/lib/audio.py
@@ -160,6 +160,11 @@ def audio_fragment_bytes(
     return buffer.getvalue()
 
 
+def audio_to_bytes(audio: "AudioFile", format: str = "wav") -> bytes:
+    """Convert entire audio file to bytes in specified format."""
+    return audio_fragment_bytes(audio, start=0, duration=None, format=format)
+
+
 def save_audio_fragment(
     audio: "AudioFile",
     start: float,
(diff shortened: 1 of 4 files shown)


src/datachain/lib/file.py
```diff
@@ -964,6 +964,35 @@ def get_fragments(
             yield self.get_fragment(start, min(start + duration, end))
             start += duration
 
+    def save(  # type: ignore[override]
```
shcheklein, line 967: just to double check - I wonder if need some kind of `@overload` here (I don't have much experience with it though)

src/datachain/lib/audio.py
```diff
@@ -160,39 +163,82 @@ def audio_fragment_bytes(
...
+        # Extract from start to end of file
+        output_file = posixpath.join(
+            output, f"{audio.get_file_stem()}_{int(start * 1000):06d}_end.{format}"
+        )
+        try:
+            audio_bytes = audio_to_bytes(audio, format, start=start, duration=None)
+        except Exception as exc:
+            raise FileError(
+                "unable to save audio fragment", audio.source, audio.path
+            ) from exc
```
dreadatour, line 219: Looks like a duplicated code (above on line 204 and below, on line 236).

src/datachain/lib/audio.py
```diff
@@ -160,39 +163,82 @@ def audio_fragment_bytes(
     return buffer.getvalue()
 
 
-def save_audio_fragment(
+def save_audio(
```
dreadatour, line 166: I just wonder if we need to just save audio (fragment) to a file, will it be faster and easier to use, let's say, `ffmpeg` rather than to use audio_to_np and torchaudio 🤔

dreadatour (approved):
Looks good to me! 👍 Couple comments below.

PR #1258, opened 2025-07-24
Fix double mapper call when using to_storage method

dreadatour:
When using `to_storage` datachain method, if there are mappers in chain, they will be applied twice. This PR fixes that.

Also we were using `chain.to_list` method to pass files to file exporter, now we are using iterator (`to_iter`) method, this should help with performance and prevent memory issues.

Test case was added.

## Summary by Sourcery

Ensure to_storage applies mappers only once by using a persisted chain and iterator-based export, and add a test to validate this behavior

Bug Fixes:
- Prevent mappers from being applied twice when using to_storage

Enhancements:
- Persist the chain and compute its count once before storage export
- Switch file export in to_storage from list-based to iterator-based for improved performance and lower memory usage

Tests:
- Add a functional test to verify mappers are only called once during to_storage

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -177,7 +177,7 @@ plugins:
             - https://numpy.org/doc/stable/objects.inv
             - https://pandas.pydata.org/docs/objects.inv
             - https://arrow.apache.org/docs/objects.inv
-            - https://docs.sqlalchemy.org/objects.inv
+            # - https://docs.sqlalchemy.org/objects.inv  # SSL certificate issue
             - https://docs.pydantic.dev/latest/objects.inv
 
 watch:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,7 +49,7 @@ dependencies = [
   "Pillow>=10.0.0,<12",
   "msgpack>=1.0.4,<2",
   "psutil",
-  "huggingface_hub",
+  "huggingface_hub<0.34.0",
   "iterative-telemetry>=0.0.10",
   "platformdirs",
   "dvc-studio-client>=0.21,<1",
--- a/src/datachain/data_storage/job.py
+++ b/src/datachain/data_storage/job.py
@@ -12,10 +12,11 @@ class JobStatus(int, Enum):
     CANCELING = 7
     CANCELED = 8
     CANCELING_SCHEDULED = 9
+    TASK = 11
 
     @classmethod
     def finished(cls) -> tuple[int, ...]:
-        return cls.COMPLETE, cls.FAILED, cls.CANCELED
+        return cls.COMPLETE, cls.FAILED, cls.CANCELED, cls.TASK
 
 
 class JobQueryType(int, Enum):
--- a/src/datachain/studio.py
+++ b/src/datachain/studio.py
@@ -21,6 +21,8 @@ POST_LOGIN_MESSAGE = (
     "Once you've logged in, return here "
     "and you'll be ready to start using DataChain with Studio."
 )
+RETRY_MAX_TIMES = 10
+RETRY_SLEEP_SEC = 1
 
 
 def process_jobs_args(args: "Namespace"):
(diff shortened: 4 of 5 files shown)


src/datachain/lib/dc/datachain.py
```diff
@@ -2419,9 +2419,11 @@ def to_storage(
             ds.to_storage("gs://mybucket", placement="filename")
             ```
         """
+        chain = self.persist()
```
dreadatour, line 2422: We are now persist chain results in this method.

src/datachain/lib/dc/datachain.py
```diff
@@ -2419,9 +2419,11 @@ def to_storage(
...
+        count = chain.count()
```
dreadatour, line 2423: Cache `count` results (was called twice before).

src/datachain/lib/dc/datachain.py
```diff
@@ -2444,7 +2446,10 @@ def to_storage(
             max_threads=num_threads or 1,
             client_config=client_config,
         )
-        file_exporter.run(self.to_values(signal), progress_bar)
+        file_exporter.run(
+            (rows[0] for rows in chain.to_iter(signal)),
```
dreadatour, line 2450: Use iterator instead of loading all the signals to memory.

tests/func/test_datachain.py
```diff
@@ -358,15 +358,24 @@ def test_to_storage(
     file_type,
     num_threads,
 ):
+    mapper = Mock(side_effect=lambda file_path: len(file_path))
```
dreadatour, line 361: Mock to be able to count calls.

tests/func/test_datachain.py
```diff
@@ -358,15 +358,24 @@ def test_to_storage(
...
+
     ctc = cloud_test_catalog
     df = dc.read_storage(ctc.src_uri, type=file_type, session=test_session)
     if use_map:
-        df.settings(cache=use_cache).map(
-            res=lambda file: file.export(tmp_dir / "output", placement=placement)
-        ).exec()
+        (
+            df.settings(cache=use_cache)
+            .map(mapper, params=["file.path"], output={"path_len": int})
```
dreadatour, line 368: This mapper should be run once for the whole dataset.

tests/func/test_datachain.py
```diff
@@ -358,15 +358,24 @@ def test_to_storage(
...
+            .map(res=lambda file: file.export(tmp_dir / "output", placement=placement))
+            .exec()
+        )
     else:
-        df.settings(cache=use_cache).to_storage(
-            tmp_dir / "output", placement=placement, num_threads=num_threads
+        (
+            df.settings(cache=use_cache)
+            .map(mapper, params=["file.path"], output={"path_len": int})
```
dreadatour, line 375: This mapper should be run once for the whole dataset.

tests/func/test_datachain.py
```diff
@@ -387,6 +396,8 @@ def test_to_storage(
         with open(tmp_dir / "output" / file_path) as f:
             assert f.read() == expected[file.name]
 
+    assert mapper.call_count == 7
```
dreadatour: Check call count (it was `14` before this fix).

dreadatour (commented):
Couple highlights.

src/datachain/lib/dc/datachain.py
shcheklein: It means an extra copy in the case we do something like read_dataset.to_storage?

shcheklein: Do we need it only for tqdm? Is there a udf running to do actual instantiation? Will we do two copies one by one?

dreadatour: > It means an extra copy in the case we do something like read_dataset.to_storage?

It will create an extra copy indeed. Good catch! Let me take a look how to avoid this.

dreadatour: @shcheklein @dmpetrov I can not find quick and good solution for this, a bit of refactoring is needed to handle cases like this.

My suggestion is to create a separate issue to handle cases where `persist` can check if there was any changes and skip copying table if now, or to implement another internal method like `checkpoint` to do this.

I would like to merge this PR now as this "extra copy" will only happens in cases where `to_storage` is used and this PR fixes much worse things (multiple UDF runs) rather than bring in additional copy table at this moment.

PR #1264, opened 2025-07-27
Rebase path

dmpetrov:
Tow helper functions:
- rebase_path
- audio channel name


src/datachain/lib/file.py
```diff
@@ -1219,6 +1253,23 @@ class Audio(DataModel):
     codec: str = Field(default="")
     bit_rate: int = Field(default=-1)
 
+    @staticmethod
+    def get_channel_name(num_channels: int, channel_idx: int) -> str:
```
shcheklein, line 1257: (just to make sure - is it related?)

also, what is the use case for this? just curious ...

shcheklein (approved):

PR #1270, opened 2025-07-30
Introduce chunk_rows to settings

### LINKED ISSUE #1261: Expose batch size
### Description

When working with heavy data like audio or video matrices, loading a full batch (2K today) into memory can cause an OOM error.

To avoid this, users should set a batch size using something like:
`chain.settings(batch=300).gen(...)`

UPDATE:
- [ ] #1262
- [ ] expose both `batch_rows=2000, batch_mb=1000` in settings() - the current issue.

amritghimire:
This PR implements configurable batch sizing to prevent OOM errors when working with heavy data like audio or video matrices. Users can now control row count limits per batch.

- Added `chunk_rows` to `chain.settings()`

```python
chain = (
    chain
    .settings(chunk_rows=300)
    .map(process_heavy_data, params="file")
)

```

This addresses the OOM issues described in #1261 by giving users fine-grained control over batch sizes based on both row count and memory constraints.

Closes #1261

## Summary by Sourcery

Enable configurable dynamic batching based on both row count and memory constraints by introducing chunk_rows and chunk_mb settings, a new DynamicBatch strategy, and supporting memory estimation utilities throughout the datachain pipeline

New Features:
- Allow configuring batch size by row count and memory limit via new chunk_rows and chunk_mb settings
- Introduce DynamicBatch strategy for UDF batching that monitors row count and memory usage to prevent OOM

Enhancements:
- Extend batched and batched_it utilities to support dynamic memory-based batching
- Add memory_utils module for recursive memory estimation and high-memory detection
- Propagate new chunk_rows and chunk_mb settings through chain.settings, UDFAdapter, dispatch, and internal batch processing

Tests:
- Add unit tests for Settings validation, memory_utils functions, and dynamic batching behavior
- Update functional tests to verify default and overridden batch parameters and memory-based limits


src/datachain/lib/dc/datachain.py
```diff
@@ -341,12 +343,14 @@ def settings(
                       To disable prefetching, set it to 0.
             namespace: namespace name.
             project: project name.
+            batch_rows: number of rows per batch (default=2000)
```
shcheklein: not clear what batch do we mean here

src/datachain/lib/dc/datachain.py
```diff
@@ -341,12 +343,14 @@ def settings(
...
+            batch_mem: memory limit in MB per batch (default=1000)
 
         Example:
             ```py
             chain = (
                 chain
-                .settings(cache=True, parallel=8)
+                .settings(cache=True, parallel=8, batch_rows=300, batch_mem=500)
```
shcheklein: it is quite special, no need to change the example

src/datachain/lib/dc/datachain.py
```diff
@@ -341,12 +343,14 @@ def settings(
                       To disable prefetching, set it to 0.
             namespace: namespace name.
             project: project name.
+            batch_rows: number of rows per batch (default=2000)
+            batch_mem: memory limit in MB per batch (default=1000)
```
shcheklein: check / cleanup other entries in terms of style, compare with other docs (spaces before / after :, periods, etc)

src/datachain/lib/dc/datachain.py
```diff
@@ -897,7 +909,9 @@ def batch_map(
         func: Optional[Callable] = None,
         params: Union[None, str, Sequence[str]] = None,
         output: OutputType = None,
-        batch: int = 1000,
+        batch: Optional[int] = None,
```
shcheklein: is it expected? 3 different options? 

why do we need these and settings also?

src/datachain/lib/dc/datachain.py
amritghimire: Yes. Its in draft for a reason. I wanted to see the CI run.

shcheklein (commented):
@dmpetrov @amritghimire it seem quite complicated (it is fine if we need something like this quick), can it be simplified? can we rely on the existing logic that checks virtual mem ?

why can't user for now use batch parameter (does it limit the number of rows)?

dmpetrov (changes requested):
@amritghimire it's not exactly what's required.

`batch_map()` is needed if user needs to implement batching logic themselves. It's already working fine - batch=1234 param is enough for this.

What we need in the issue - change batch size for regular map(), gen() and agg(). It suppose to be a single settings for all.

PS: this setting will also effect batch_map().

dmpetrov:
@amritghimire regarding the param names, please take a look https://github.com/iterative/datachain/issues/1261#issuecomment-3138095432

src/datachain/lib/dc/datachain.py
amritghimire: Updated it to same conventions within this docs. The style varies from docs to docs in our codebase. Either we need to prepare a guideline for docstrings and change the same for all. For now, it is out of scope of this PR.

src/datachain/lib/dc/datachain.py
amritghimire: This is to maintain the backward compatibility. But will update the code to use settings only instead.

dmpetrov:
@amritghimire I see that you renamed to chunk, but `batch_mem` stays the same for some reason. We need to unify these.

Also, I'm not to deep in the code base of batching but my understanding was - it's happening due to INSERT_BATCH_SIZE. But I"m not an expert.
Please reproduce the OOM error and make sure the solution solves it.

amritghimire:
> @amritghimire I see that you renamed to chunk, but `batch_mem` stays the same for some reason. We need to unify these.

Yeah, sorry. I was off yesterday and was able to make some changes only while waiting for the boarding.

amritghimire:
@sourcery-ai dismiss

amritghimire:
@sourcery-ai review

amritghimire:
@sourcery-ai guide

amritghimire:
@sourcery-ai summary

src/datachain/lib/dc/datachain.py
```diff
@@ -2039,7 +2071,7 @@ def to_parquet(
         self,
         path: Union[str, os.PathLike[str], BinaryIO],
         partition_cols: Optional[Sequence[str]] = None,
-        chunk_size: int = DEFAULT_PARQUET_CHUNK_SIZE,
+        chunk_size: Optional[int] = None,
```
dmpetrov: Parquet chunk is not related to the chunk we are dealing with

src/datachain/data_storage/sqlite.py
```diff
@@ -801,7 +801,7 @@ def copy_table(
             .limit(None)
         )
 
-        for batch in batched_it(ids, 10_000):
+        for batch in batched_it(ids, 10_000, 1000):
```
dmpetrov: Is it related to the chenge?

src/datachain/lib/dc/datachain.py
```diff
@@ -919,9 +937,20 @@ def batch_map(
             ```
         """
         udf_obj = self._udf_to_obj(BatchMapper, func, params, output, signal_map)
+        chunk_rows = self._settings.chunk_rows
```
dmpetrov: we are mixing up `chunk` and user facing `batch` size - as I commented before. The chage should not affect `batch_map()` in any way.

src/datachain/lib/dc/datachain.py
```diff
@@ -1272,7 +1301,7 @@ def _leaf_values(self, *, row_factory=None, include_hidden: bool = True):
             yield from rows
 
     def to_columnar_data_with_names(
-        self, chunk_size: int = DEFAULT_PARQUET_CHUNK_SIZE
+        self, chunk_size: Optional[int] = None
```
dmpetrov: same, it's not related to the goal of the issue.

src/datachain/lib/dc/records.py
```diff
@@ -94,7 +94,11 @@ def read_records(
         {c.name: c.type for c in columns if isinstance(c.type, SQLType)},
     )
     records = (adjust_outputs(warehouse, record, col_types) for record in to_insert)
-    for chunk in batched(records, INSERT_BATCH_SIZE):
```
dmpetrov, line 97: NOt related as well. I guess, INSERT_BATCH_SIZE was imported from there just to have some limit. But these params are not really have the same meaning.

I'd introduce a separate constant in this file instead of importing INSERT_BATCH_SIZE.

src/datachain/query/batch.py
```diff
@@ -92,6 +93,86 @@ def __call__(
                 yield [r[0] for r in results] if ids_only else results
 
 
+class DynamicBatch(BatchingStrategy):
```
dmpetrov: why do we need a whole new strategy? The limit check was already introduced in `query/dataset.py`

src/datachain/query/dispatch.py
```diff
@@ -196,14 +196,17 @@ def run_udf_single(
...
     ) -> None:
         udf: UDFAdapter = loads(self.udf_data)
-
         if ids_only and not self.is_batching:
             input_rows = flatten(input_rows)
 
         def get_inputs() -> Iterable["RowsOutput"]:
             warehouse = self.catalog.warehouse.clone()
             if ids_only:
-                for ids in batched(input_rows, DEFAULT_BATCH_SIZE):
```
dmpetrov: Please correct me if I'm wrong, but this should not be connected to the issue.

src/datachain/query/dispatch.py
```diff
@@ -246,9 +250,15 @@ def run_udf_parallel(  # noqa: C901, PLR0912
         processed_cb: Callback = DEFAULT_CALLBACK,
         generated_cb: Callback = DEFAULT_CALLBACK,
     ) -> None:
+        udf: UDFAdapter = loads(self.udf_data)
         self.task_queue = self.ctx.Queue()
         self.done_queue = self.ctx.Queue()
 
+        chunk_rows = self.input_batch_size(
+            n_workers, udf.chunk_rows or DEFAULT_BATCH_SIZE
```
dmpetrov: same here

src/datachain/query/dispatch.py
```diff
@@ -427,7 +438,9 @@ def get_inputs(self, ids_only: bool) -> Iterable["RowsOutput"]:
         warehouse = self.catalog.warehouse.clone()
         while (batch := get_from_queue(self.task_queue)) != STOP_SIGNAL:
             if ids_only:
-                for ids in batched(batch, DEFAULT_BATCH_SIZE):
+                chunk_rows = self.udf.chunk_rows or DEFAULT_BATCH_SIZE
```
dmpetrov: same

src/datachain/utils.py
```diff
@@ -225,30 +238,65 @@ def get_envs_by_prefix(prefix: str) -> dict[str, str]:
...
 
-def batched(iterable: Iterable[_T_co], n: int) -> Iterator[tuple[_T_co, ...]]:
-    """Batch data into tuples of length n. The last batch may be shorter."""
-    # Based on: https://docs.python.org/3/library/itertools.html#itertools-recipes
-    # batched('ABCDEFG', 3) --> ABC DEF G
-    if n < 1:
-        raise ValueError("Batch size must be at least one")
-    it = iter(iterable)
-    while batch := tuple(islice(it, n)):
+def _dynamic_batched_core(
```
dmpetrov, line 229: Why is this needed? Why not to use the previos one?

src/datachain/utils.py
```diff
@@ -225,30 +238,65 @@ def get_envs_by_prefix(prefix: str) -> dict[str, str]:
...
+    chunk_mb: float,
+) -> Iterator[list[_T_co]]:
+    """Core batching logic that yields lists."""
+    max_memory_bytes = chunk_mb * 1024 * 1024
+
+    batch: list[_T_co] = []
+    current_memory = 0
+
+    for row_count, item in enumerate(iterable):
+        item_memory = estimate_memory_recursive(item)
```
dmpetrov: We cannot afford doing this for every item - too expensive. The logic need to be changes.

src/datachain/lib/memory_utils.py
```diff
@@ -0,0 +1,57 @@
...
+
+    if isinstance(item, (str, bytes, int, float, bool)):
+        return sys.getsizeof(item)
+    if isinstance(item, (list, tuple)):
+        total_size = sys.getsizeof(item)
+        for subitem in item:
+            total_size += sys.getsizeof(subitem)
+        return total_size
+    # For complex objects, use a conservative estimate
+    return sys.getsizeof(item) + OBJECT_OVERHEAD_BYTES
```
dmpetrov: This likely won't work since we mostly deal with complex objects. I can guess that logic will get here in 99% of cases.

src/datachain/lib/memory_utils.py
```diff
@@ -0,0 +1,57 @@
...
+
+def estimate_memory_recursive(item: Any) -> int:
+    if item is None:
+        return 0
+
+    if isinstance(item, (str, bytes, int, float, bool)):
+        return sys.getsizeof(item)
+    if isinstance(item, (list, tuple)):
+        total_size = sys.getsizeof(item)
+        for subitem in item:
```
dmpetrov: It seems duplicate the line above - getting the sizes one by one while we get a common size.

src/datachain/lib/memory_utils.py
```diff
@@ -0,0 +1,57 @@
...
+MEMORY_USAGE_THRESHOLD = 80
+
+# System memory check frequency (every N rows)
+MEMORY_CHECK_FREQUENCY = 100
+
+# Shared constant for object overhead estimation
+OBJECT_OVERHEAD_BYTES = 100
+
+
+def estimate_memory_recursive(item: Any) -> int:
```
dmpetrov: I have a feeling that it can be simplified given that we work with Pydentic.

dmpetrov (changes requested):
I've looked at the current code -  it's doing way more than needed.

All we need is to overwrite a constant in a single place (we just need to be sure INSERT_BATCH_SIEZE is the correct one). While we overwriting multiple constant somehow related to batching and caching.

Also, the logic with object size calculation raises a lot of questions - implementation as well as overall perf overhead.

For now, we need the simple possible way to solve the issue ASAP (it's really pressing!) without object sizes. Object sizes should be a separate PR when we can address the questions.

src/datachain/lib/dc/records.py
amritghimire: I am not sure, given that it also inserts the rows into the warehouse, it could be good to have chunk as well. But reverted it back for now.

src/datachain/query/dispatch.py
amritghimire: It could be helpful to read the rows in chunk from warehouse that can be customized,. But, reverted back for now,

amritghimire:
> I've looked at the current code - it's doing way more than needed.
> 
> All we need is to overwrite a constant in a single place (we just need to be sure INSERT_BATCH_SIEZE is the correct one). While we overwriting multiple constant somehow related to batching and caching.
> 
> Also, the logic with object size calculation raises a lot of questions - implementation as well as overall perf overhead.
> 
> For now, we need the simple possible way to solve the issue ASAP (it's really pressing!) without object sizes. Object sizes should be a separate PR when we can address the questions.

Updated the PR to support chunk rows for now. Will look for better alternatives for chunk mb.

src/datachain/lib/dc/datachain.py
```diff
@@ -324,29 +324,31 @@ def settings(
...
-            prefetch: number of workers to use for downloading files in advance.
+            min_task_size : minimum number of tasks. (default=1)
+            prefetch : number of workers to use for downloading files in advance.
                       This is enabled by default and uses 2 workers.
                       To disable prefetching, set it to 0.
-            namespace: namespace name.
-            project: project name.
+            namespace : namespace name.
+            project : project name.
+            chunk_rows : number of rows per batch. (default=2000)
```
dmpetrov: how about `row limit per insert to balance speed and memory usage. (default=2000)`
please avoid using batch since we decided to use it mostly in batch_map()

src/datachain/lib/dc/records.py
```diff
@@ -15,6 +15,8 @@
 
     P = ParamSpec("P")
 
+READ_RECORDS_BATCH_SIZE = 10000
```
dmpetrov, line 18: thanks! it was needed here.

src/datachain/utils.py
dmpetrov: Now i see - just code redesign

src/datachain/utils.py
```diff
@@ -225,30 +236,49 @@ def get_envs_by_prefix(prefix: str) -> dict[str, str]:
...
+) -> Iterator[list[_T_co]]:
+    """Core batching logic that yields lists."""
+
+    batch: list[_T_co] = []
+
+    for row_count, item in enumerate(iterable):
+        # Check if adding this item would exceed limits
+        # Also check system memory usage every 100 items
+        should_yield = len(batch) >= chunk_rows or (
+            row_count % 100 == 0 and is_memory_usage_high()
```
dmpetrov: We shouldn't check memory if `chunk_mb` is not defined - waste of resources. Also, it should be None by default - so no memory check at all.

So, you either introduce this as well or remove all code for memory check. Your call.

dmpetrov (changes requested):
Much cleaner now - thank you!

One comment about `chunk_mb`. Also, in the current implementation it's not limit to chunks it's limit to memory usage by the whole app and can be applited to other limits INSERT/SELECT etc... But I'd void introducing these logic now. But we might need to rename it to `mem_mb` (no chunk anymore). Or you can avoid introducing this functionality for now - a separate PR.

amritghimire:
> Much cleaner now - thank you!
> 
> One comment about `chunk_mb`. Also, in the current implementation it's not limit to chunks it's limit to memory usage by the whole app and can be applited to other limits INSERT/SELECT etc... But I'd void introducing these logic now. But we might need to rename it to `mem_mb` (no chunk anymore). Or you can avoid introducing this functionality for now - a separate PR.

Removed all logic about memory check for now. Will create a separate PR for that.

dmpetrov (approved):
I tested and it works as expected - really goot tool to manage memory usage! Thank you @amritghimire 

We need to rename it back to `batch_rows`. Apologies for the back-and-forth - I originally asked to rename batch_rows to chunk_rows, but we’ve decided to deprecate batch_map() #1272 and there are not naming conflicts anymore. So, batch_row is a netter name.

Approving but please do renaming before merging.

Issue #1276, opened 2025-08-05
Merge after delta update


dmpetrov:
### Description

Now we are getting an exception: `Delta update cannot be used with merge`. It prevents race cond issues. However, we need to enable it in cases when user accepts the risk or runs custom check before `merge()`.

- [ ] Introduce reset_delta(), So, `chain.reset_delta().merge(...)` just works.
- [ ] Extend error message like `Cannot merge two delta datasets - may cause inconsistency. Use reset_delta() to disable delta tracking.`
- [ ] Allow merges between delta dataset and not-delta datasets - it cannot lead to race condition 
- [ ] Make sure all affected commands follow this rules: merge, join, union(?), distinct, etc. (please list it in the docs)

Open question: should `chain.clone()` reset delta and why?

ilongin:
Should we maybe call it `delta_unsafe()` instead of `reset_delta()` ?
And should we put it just as an argument with other delta arguments?

dmpetrov:
> Should we maybe call it `delta_unsafe()`

Good idea! It clearly highlights taking risks.

ilongin:
> > Should we maybe call it `delta_unsafe()`
> 
> Good idea! It clearly highlights taking risks.

What about having `delta_unsafe` just as additional argument with others? To me it seems strange that this one is separate as special `DataChain` method ...

dmpetrov:
@ilongin I was thinking it's more granular and we have too many params for delta already. And checkpoints are coming with it's own params 🙀 

But you are right, for now it might be better to follow the existing pattern and introduce unsafe as a param. Later we can figure out how to extract all the delta params to a separate function.

Let's make it a parameter, not a method.

shcheklein:
Does delta usafe cover everything here?

dmpetrov:
We need to check item my item once unsafe is implemented

PR #1279, opened 2025-08-07
Unsafe delta

ilongin:
Added ability to use methods like `merge`, `union` etc. with delta updates, which are normally disabled.

## Summary by Sourcery

Enable unsafe datachain operations during delta updates by introducing a delta_unsafe flag to override method restrictions and ensure correct handling in chain evolution and dataset reads.

New Features:
- Add delta_unsafe parameter to DataChain to permit unsafe methods (merge, union, group_by, agg, distinct) during delta updates
- Extend read_dataset and read_storage APIs to accept delta_unsafe flag and propagate it through the chain setup
- Modify delta_disabled decorator to bypass method restrictions when delta_unsafe is enabled

Tests:
- Add functional test to verify delta updates with unsafe operations maintain correct dependencies and merge data as expected


shcheklein:
@ilongin I don't think it is a priority atm.

src/datachain/delta.py
```diff
@@ -30,7 +30,7 @@ def delta_disabled(
 
     @wraps(method)
     def _inner(self: T, *args: "P.args", **kwargs: "P.kwargs") -> T:
-        if self.delta:
+        if self.delta and not self._delta_unsafe:
             raise NotImplementedError(
                 f"Delta update cannot be used with {method.__name__}"
```
dmpetrov: Please extend the message. Text example is in #1276

src/datachain/lib/dc/datasets.py
```diff
@@ -80,6 +81,9 @@ def read_dataset(
         update: If True always checks for newer versions available on Studio, even if
             some version of the dataset exists locally already. If False (default), it
             will only fetch the dataset from Studio if it is not found locally.
+        delta_unsafe: If True, allows the use of methods that are normally unsafe
+            and forbidden during a delta update. The following methods will be
+            permitted: `merge`, `agg`, `union`, `group_by`, and `distinct`.
```
dmpetrov: It needs to be shorten - too many words. Or We can put more light on the reasons. Like: 
```
Allow restricted ops in delta: `merge`, `agg`, `union`, `group_by`, `distinct`.
Caller must ensure datasets are consistent and not partially updated.
```

src/datachain/lib/dc/storage.py
```diff
@@ -77,6 +78,9 @@ def read_storage(
               (error mode)
             - True: Reprocess records missing from the result dataset (missing mode)
             - None: No retry processing (default)
+        delta_unsafe: If True, allows the use of methods that are normally unsafe
+            and forbidden during a delta update. The following methods will be
+            permitted: `merge`, `agg`, `union`, `group_by`, and `distinct`.
```
dmpetrov: please simplify the comment - see above

tests/func/test_delta.py
```diff
@@ -96,6 +96,101 @@ def create_delta_dataset(ds_name):
...
+            delta_result_on=["file.source", "file.path"],
+            delta_compare=["file.version", "file.etag"],
+            delta_unsafe=True,
+        ).merge(merge_ds, on="file.path", inner=True).save(ds_name)
+
+    # first version of starting dataset
+    create_image_dataset(starting_ds_name, images[:2])
+    # first version of delta dataset
+    create_delta_dataset(ds_name)
+    assert sorted(_get_dependencies(catalog, ds_name, "1.0.0")) == sorted(
```
dmpetrov: please use set() instead of sorting - looks nicer

tests/func/test_delta.py
```diff
@@ -96,6 +96,101 @@ def create_delta_dataset(ds_name):
     create_delta_dataset(ds_name)
 
 
+def test_delta_update_unsafe(test_session, tmp_dir, tmp_path):
+    catalog = test_session.catalog
+    default_namespace_name = catalog.metastore.default_namespace_name
+    default_project_name = catalog.metastore.default_project_name
```
dmpetrov: Do we need to deal with the namespaces here? Can we just run it in the default namespace?

tests/func/test_delta.py
```diff
@@ -96,6 +96,101 @@ def create_delta_dataset(ds_name):
...
+            delta_compare=["file.version", "file.etag"],
+            delta_unsafe=True,
+        ).merge(merge_ds, on="file.path", inner=True).save(ds_name)
+
+    # first version of starting dataset
+    create_image_dataset(starting_ds_name, images[:2])
+    # first version of delta dataset
+    create_delta_dataset(ds_name)
+    assert sorted(_get_dependencies(catalog, ds_name, "1.0.0")) == sorted(
+        [(dependency_ds_name, "1.0.0"), (dependency_ds_merge_name, "1.0.0")]
```
dmpetrov: How dependency check is relevant to delta unsafe? can we avoid testing it here?

dmpetrov: Looked deeper - yes, the dependency deserved it's own test. But it should not be combined with the unsafe test. PLease extract or remove (if it's exist already)

tests/func/test_delta.py
```diff
@@ -96,6 +96,101 @@ def create_delta_dataset(ds_name):
...
+    assert sorted(_get_dependencies(catalog, ds_name, "1.0.1")) == sorted(
+        [(dependency_ds_name, "1.0.1"), (dependency_ds_merge_name, "1.0.0")]
+    )
+
+    assert (dc.read_dataset(ds_name, version="1.0.0").order_by("file.path")).to_list(
+        "file.path", "value"
+    ) == [
+        ("img1.jpg", 1),
+        ("img2.jpg", 2),
+    ]
```
dmpetrov: by using order_by you extend the test scope without a need. It's better to keep tests scop nerrow to the target function - unsafe.

tests/func/test_delta.py
```diff
@@ -96,6 +96,101 @@ def create_delta_dataset(ds_name):
...
+            ],
+            session=test_session,
+        ).save(ds_name)
+
+    def create_delta_dataset(ds_name):
+        dc.read_dataset(
+            starting_ds_name,
+            session=test_session,
+            delta=True,
+            delta_on=["file.source", "file.path"],
```
dmpetrov: Why not just keeping the defaults? Same for delta_result_on & delta_compare.

tests/func/test_delta.py
```diff
@@ -96,6 +96,101 @@ def create_delta_dataset(ds_name):
...
+    ]
+
+    merge_images = [
+        {
+            "name": f"img{i}.jpg",
+            "data": Image.new(mode="RGB", size=(64, 128)),
+            "value": i,
+        }
+        for i in range(1, 5)
+    ]
```
dmpetrov: This feels more complicated than necessary for testing unsafe. Why Image is needed? Why not just File? You also probably don't need to create an actual image.

Another thought - why not using simple structure like read_values(num=[1, 2, 3, 4, 5], status="ready1", "ready2", "ready3", "ready4", "ready5") and run delta on it?

dmpetrov (changes requested):
Great change. I know it's not fully done but provided some comments.

Main point: unit tests are missing. We should focus on unit tests over functional tests here. The current tests feel heavier than needed - they check merge dependencies and merge value correctness, while what we really need is to test when the exception is thrown and when it isn’t.

It might be a great idea to check the dependencies and merge - please create separate tests for this. It's not related to unsafe logic. Unit tests are strongly preffered.

tests/func/test_delta.py
ilongin: I need to check full dataset name and since this tests must run in Studio as well, I cannot just hardcode `local` as default namespace is different there.

tests/func/test_delta.py
ilongin: Refactored it to be more simple with just basic numbers and without files / images. I originally just c/p other test to be more quick but yes, it's not needed here.

tests/func/test_delta.py
ilongin: I would keep dependency check in this test since this is func test. I can create unit tests to test more granular details. This needs to be checked with unsafe as we are using things like merge that affects dependencies

src/datachain/lib/dc/datasets.py
shcheklein: We have also a guide for delta updates - I think we can put a separate paragpraph there ## Using delta with merge, agg, etc ... Where we can explain some reasoning behind it

src/datachain/delta.py
```diff
@@ -30,9 +30,10 @@ def delta_disabled(
 
     @wraps(method)
     def _inner(self: T, *args: "P.args", **kwargs: "P.kwargs") -> T:
-        if self.delta:
+        if self.delta and not self._delta_unsafe:
             raise NotImplementedError(
-                f"Delta update cannot be used with {method.__name__}"
+                f"Cannot use {method.__name__} with delta datasets - may cause"
+                " inconsistency. Use delta_unsafe flag to disable delta tracking."
```
shcheklein: `to disable delta tracking.` - what does it exactly mean?

shcheklein: rephrase to "Use ... to allow this operation" or something

src/datachain/lib/dc/datasets.py
```diff
@@ -80,6 +81,9 @@ def read_dataset(
         update: If True always checks for newer versions available on Studio, even if
             some version of the dataset exists locally already. If False (default), it
             will only fetch the dataset from Studio if it is not found locally.
+        delta_unsafe: Allow restricted ops in delta: merge, agg, union, group_by,
+            distinct. Caller must ensure datasets are consistent and not partially
```
shcheklein: can we clarify that we mean results datasets?

tests/func/test_delta.py
```diff
@@ -96,6 +96,76 @@ def create_delta_dataset(ds_name):
...
+    # second version of delta dataset
+    dc.read_dataset(
+        starting_ds_name,
+        session=test_session,
+        delta_on="id",
+        delta=True,
+        delta_unsafe=True,
+    ).merge(merge_ds, on="id", inner=True).save(ds_name)
+
+    assert set(_get_dependencies(catalog, ds_name, "1.0.1")) == {
```
shcheklein, line 124: probably we need a proper method for gependencies? or even `chain.dataset` should be returning a record that can be expanded to dependencies ...

tests/func/test_delta.py
```diff
@@ -96,6 +96,76 @@ def create_delta_dataset(ds_name):
...
+    default_namespace_name = catalog.metastore.default_namespace_name
+    default_project_name = catalog.metastore.default_project_name
+
+    starting_ds_name = "starting_ds"
+    merge_ds_name = "merge_ds"
+
+    dependency_ds_name = (
+        f"{default_namespace_name}.{default_project_name}.{starting_ds_name}"
+    )
+    dependency_ds_merge_name = (
```
shcheklein: is it only to use `(_get_dependencies(catalog, ds_name, "1.0.1")` ? (quite unfortunate tbh)

src/datachain/delta.py
ilongin: Not 100% sure, Dmitry wanted to put that message...but I agree, it sounds strange, I will change to something else

src/datachain/lib/dc/datasets.py
ilongin: I'm not sure I'm following what you meant by results dataset. Actually, when I read this sentence I'm not sure I completely understand it in general. @dmpetrov can you clarify what did you mean by it? 

These operations are not safe from different reasons, e.g
1. union / merge -> producing duplicates 
2. distinct -> must be called on whole dataset which conflicts with delta logic which runs delta part on new + updated rows and then merges results

tests/func/test_delta.py
ilongin: We already have method to return dependencies in `Catalog` but this one is just helper for tests to extract dataset name + version out of it as that ones are being asserted.

tests/func/test_delta.py
ilongin: It's to check if dependencies are ok but still you are right it's not needed. I don't think test of correct namespace / project is in scope of this tests so I will just return pure dataset name from that helper method and assert with that one in tests.

src/datachain/lib/dc/datasets.py
shcheklein: When you mention datasets in the message, what datasets are we talking about? It should be clear to end users

tests/func/test_delta.py
```diff
@@ -96,6 +96,101 @@ def create_delta_dataset(ds_name):
     create_delta_dataset(ds_name)
```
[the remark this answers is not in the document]
shcheklein: Resolve this?

tests/func/test_delta.py
shcheklein: Don’t u think users might need the same? Not sure why this can’t be accessed via chain itslef

dreadatour (approved):
Looks good to me 🤔👍

tests/func/test_delta.py
ilongin: We didn't have this kind of request yet. If needed, we can add it to chain (public API) but I would do it in separate issue / PR as it's really not related to this one.

src/datachain/lib/dc/datasets.py
ilongin: Yea, that's good idea. I've added new section in docs and simplified this function docs at the same time.

src/datachain/lib/dc/datasets.py
ilongin: Removed that sentence all together (see https://github.com/iterative/datachain/pull/1279#discussion_r2297732570)

shcheklein (approved):

shcheklein:
Just to I understand the terminology and be on the same page in the terminology we use:

https://github.com/iterative/datachain/pull/1279/files#diff-2a81289be4719523a37fa0369702d59af21a78b3522f155ebbbfb48c4e4ed2e8R382 - is this one unit or not, for example? (not matter what directory it is in).

dmpetrov:
> is this one unit or not, for example?

yes, these are unit and should be in the unit test dir. the previous one is not.

shcheklein:
> yes, these are unit and should be in the unit test dir.

yep, I don't call them unit (I call unit tests those that usually test only a very particular function, module in isolation ... usually with mocks, etc, etc), but that's fine ... I'm totally fine to call them whatever and place them in unit or func.

> the previous one is not.

you mean test_delta_update_unsafe or something else?

dmpetrov:
> > the previous one is not.
> 
> you mean test_delta_update_unsafe or something else?

yep

tests/func/test_delta.py
```diff
@@ -96,6 +82,66 @@ def create_delta_dataset(ds_name):
...
 
 
+def test_delta_update_unsafe(test_session, tmp_dir, tmp_path):
```
shcheklein: do we need tmp_dir / tmp_path now?

shcheklein:
> yep

okay, I see

my 2cs on this  - that one looks heavy indeed (some cleanup can be done?) - but it is a single test for the whole unsafe stuff and it is also using in-memory db AFAIU, no file systems, etc, etc. Some complexity comes from the delta itself - requires multiple versions of source / destinations, etc, etc. Not sure what would be a better alternative here tbh if we want to run delta + merge.

tests/func/test_delta.py
ilongin: No need, leftover

src/datachain/delta.py
```diff
@@ -155,25 +156,23 @@ def _get_source_info(
         indirect=False,
     )
 
-    dep = dependencies[0]
-    if not dep:
+    source_ds_dep = next((d for d in dependencies if d.name == source_ds.name), None)
```
shcheklein, line 159: was it catched by the test?

ilongin: yes, test was inconsistent and there was a bug when things like `merge` used - we didn't expect more than 1 dependencies before but I guess now it can happen

PR #1281, opened 2025-08-07
Introduce batch_size in mb limit

### LINKED ISSUE #1262: Introduce batch size in MB limit

amritghimire:
Similar to how we implemented batch_rows, this implements batch_mb
settings which when available, enables the batching based on size as
well as system memory.

Closes #1262

## Summary by Sourcery

Enable batching by memory usage by introducing the batch_mb parameter and MemoryTracker integration

New Features:
- Add optional batch_mb parameter for memory-based batching alongside batch_rows
- Implement MemoryTracker to measure object sizes and enforce memory thresholds

Enhancements:
- Propagate batch_mb through Settings, core batch functions, UDF adapters, dataset processing, and DataChain API methods
- Add validation, properties, and merging logic for batch_mb in Settings

Build:
- Add Pympler dependency for memory sizing

Tests:
- Extend unit tests for Settings and functional tests to cover batch_mb behavior

### DIFF
--- a/src/datachain/utils.py
+++ b/src/datachain/utils.py
@@ -8,6 +8,7 @@ import random
 import re
 import sys
 import time
+import weakref
 from collections.abc import Iterable, Iterator, Sequence
 from contextlib import contextmanager
 from datetime import date, datetime, timezone
@@ -19,6 +20,7 @@ import platformdirs
 from dateutil import tz
 from dateutil.parser import isoparse
 from pydantic import BaseModel
+from pympler import asizeof
 
 if TYPE_CHECKING:
     import pandas as pd
@@ -26,6 +28,7 @@ if TYPE_CHECKING:
 
 
 DEFAULT_CHUNK_ROWS = 2000
+MEMORY_USAGE_THRESHOLD = 80
 
 logger = logging.getLogger("datachain")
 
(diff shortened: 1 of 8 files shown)


src/datachain/utils.py
```diff
@@ -531,3 +552,41 @@ def safe_closing(thing: T) -> Iterator[T]:
...
+        self._type_cache: dict[type, int] = {}
+        self._object_cache = weakref.WeakValueDictionary()
+        self._cache_size = 1024
+
+    def get_size(self, obj: Any) -> int:
+        if obj is None:
+            return 0
+
+        obj_id = id(obj)
+        if obj_id in self._object_cache:
```
dmpetrov, line 568: `_object_cache` doubles amount of consumed memory - we cannot afford this.

src/datachain/utils.py
```diff
@@ -229,16 +232,27 @@ def get_envs_by_prefix(prefix: str) -> dict[str, str]:
...
+    batch_mb: Optional[float] = None,
 ) -> Iterator[list[_T_co]]:
     """Core batching logic that yields lists."""
 
     batch: list[_T_co] = []
+    memory_tracker = MemoryTracker() if batch_mb else None
+    total_size = 0
+
+    for i, item in enumerate(iterable):
+        if memory_tracker and batch_mb:
```
dmpetrov, line 244: `batch_mb` was checked in the above and not needed here

src/datachain/utils.py
```diff
@@ -229,16 +232,27 @@ def get_envs_by_prefix(prefix: str) -> dict[str, str]:
...
+            item_size = memory_tracker.get_size(item)
```
dmpetrov, line 245: I'm a little bit afraid doing this for every single object - it introduces too much computational overhead.
I'd try to do some stats based approach - do this in 5% cases of buffer size (row limit) only. I'm not sure this complication is needed.

src/datachain/utils.py
amritghimire: We save just the id and size in this cache with specified limit i.e. only 1024 numbers at max

dmpetrov: oh, limit is only 1024.

it optimizes the same objects, right? but how often UDF outputs the same objects? I can imaging only one case - gen() that produces multiple records for a single file record. Not sure it worth to optimize for. We're lacking real use cases, unfortunately.

amritghimire:
> I'd propose postponing this change until we get a good use case. WDYT folks?

I am okay with this. Decreasing the priority of the issue.

dmpetrov:
@amritghimire thank you for the effort and sorry for back and forth with this. But it's better to wait for use cases to evaluate agains.

PR #1287, opened 2025-08-11
Add option to pass credentials name when creating a job

amritghimire:
## Summary by Sourcery

Add support for specifying credentials by name when creating Studio jobs

New Features:
- Add --credentials-name CLI flag for datachain job run
- Pass credentials_name through process_jobs_args and include it in create_job API calls

Documentation:
- Document the --credentials-name option and update examples in job run guide

Tests:
- Add unit tests to verify credentials_name is parsed and sent in job creation

### DIFF
--- a/docs/commands/job/run.md
+++ b/docs/commands/job/run.md
@@ -32,6 +32,7 @@ This command runs a job in Studio using the specified query file. You can config
 * `--env-file ENV_FILE` - File with environment variables for the job
 * `--env ENV` - Environment variables in KEY=VALUE format
 * `--cluster CLUSTER` - Compute cluster to run the job on
+* `--credentials-name CREDENTIALS_NAME` - Name of the credentials to use for the job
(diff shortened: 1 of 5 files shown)


docs/commands/job/run.md
```diff
@@ -97,7 +98,12 @@ datachain job clusters
 datachain job run --cluster 1 query.py
 ```
 
-9. Schedule a job to run once at a specific time
+9. Run a job with specific credentials
```
shcheklein, line 101: again, I don't think we need an example for every option, when it is obvious

shcheklein (approved):
what is the default?

amritghimire:
> what is the default? 

None. We don't have default. Studio will select latest creds by itself

shcheklein:
> None. We don't have default. Studio will select latest creds by itself

yes, I put a comment in Slack - we must have a way to set the default one per cloud, so that people don't have to pass manually each time

PR #1296, opened 2025-08-17
fix(to_database): return number of rows, normalize conflict columns

shcheklein:
* Return number of inserted and updated rows
* Normalize conflict columns names to allow DataChain style columns names
* Add examples on how to use PG schema

### DIFF
--- a/src/datachain/lib/dc/database.py
+++ b/src/datachain/lib/dc/database.py
@@ -77,12 +77,15 @@ def to_database(
     on_conflict: Optional[str] = None,
     conflict_columns: Optional[list[str]] = None,
     column_mapping: Optional[dict[str, Optional[str]]] = None,
-) -> None:
+) -> int:
     """
     Implementation function for exporting DataChain to database tables.
 
     This is the core implementation that handles the actual database operations.
     For user-facing documentation, see DataChain.to_database() method.
+
+    Returns:
+        int: Number of rows affected (inserted/updated).
     """
     if on_conflict and on_conflict not in ("ignore", "update"):
         raise ValueError(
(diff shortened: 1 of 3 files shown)


src/datachain/lib/dc/database.py
```diff
@@ -249,7 +289,8 @@ def prepare_row(row_values):
             stacklevel=2,
         )
 
-    conn.execute(insert_stmt, rows_to_insert)
+    result = conn.execute(insert_stmt, rows_to_insert)
+    return result.rowcount
```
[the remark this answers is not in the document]
dreadatour: Agree, IMO it should be something like this, at least:
```suggestion
    result = conn.execute(insert_stmt, rows_to_insert)
    return result.rowcount if result.rowcount > 0 else 0
```

src/datachain/lib/dc/database.py
```diff
@@ -206,6 +245,7 @@ def prepare_row(row_values):
 
     supports_conflict = on_conflict and conn.engine.name in ("postgresql", "sqlite")
 
+    insert_stmt: Any  # Can be PostgreSQL, SQLite, or regular insert statement
```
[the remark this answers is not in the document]
dreadatour: `Insert` type from `sqlalchemy` should works here 🤔

src/datachain/lib/dc/database.py
```diff
@@ -117,14 +125,15 @@ def to_database(
...
+                    rows_affected = _process_batch(
                         conn,
                         table,
                         batch,
                         on_conflict,
-                        conflict_columns,
+                        normalized_conflict_columns,
                         column_indices_and_names,
                     )
+                    total_rows_affected += rows_affected
```
dreadatour: Or at least here we need a check if `rows_affected` is `<0` (possible case for some DBs):
```suggestion
                    if rows_affected > 0:
                        total_rows_affected += rows_affected
```

shcheklein: yep, I'll put a check and will return -1 then from the to_database (since we can't get the reliable value?)

dreadatour: > since we can't get the reliable value?

As far as I know it is sometimes possible, but with additional queries.

`-1` is fine, but we don't need to sum all `-1`'s, this will looks odds below in `print`. Also as an improvement we can show number of affected rows only if we have this number.

shcheklein: yes, -1 one time, not summing them up ... any -1 returned by the underlying layer means returning -1 from the top function (we can get the reliable number)

PR #1298, opened 2025-08-18
fix(mutate): simplify and drop mutated columns

shcheklein:
Fixes https://github.com/iterative/studio/issues/11989
Fixes https://github.com/iterative/studio/issues/12041

+ Simplifies code (removed redundant code in mutate)
+ Adds more tests and consolidates existing tests
+ Fixes a few other edge cases (e.g. dealing with nested columns)

## Summary by Sourcery

Simplify and correct the mutate clause implementation and reorganize its tests

Bug Fixes:
- Fix mutate to drop original columns and prevent old column names from leaking into the database

Enhancements:
- Simplify SQLMutate.apply_sql_clause by using an intermediate subquery and excluding mutated columns to avoid name collisions
- Rename create_tar_dataset_with_legacy_columns to create_tar_dataset and remove its legacy mutate step

Tests:
- Consolidate and expand mutate tests into a new test_mutate.py covering overwrites, ordering, multiple columns, chaining, persistence, renaming, and window functions
- Remove outdated mutate tests from test_datachain.py

### DIFF
--- a/src/datachain/lib/signal_schema.py
+++ b/src/datachain/lib/signal_schema.py
@@ -34,7 +34,7 @@ from datachain.lib.data_model import DataModel, DataType, DataValue
 from datachain.lib.file import File
 from datachain.lib.model_store import ModelStore
 from datachain.lib.utils import DataChainParamsError
-from datachain.query.schema import DEFAULT_DELIMITER, Column, ColumnMeta
+from datachain.query.schema import DEFAULT_DELIMITER, C, Column, ColumnMeta
 from datachain.sql.types import SQLType
 
 if TYPE_CHECKING:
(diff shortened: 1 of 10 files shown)


src/datachain/query/dataset.py
```diff
@@ -795,28 +794,36 @@ def apply_sql_clause(self, query: Select) -> Select:
 
 @frozen
 class SQLMutate(SQLClause):
-    args: tuple[Union[Function, ColumnElement], ...]
+    args: tuple[Label, ...]
```
dreadatour, line 798: Not sure why do we use `Label` here now and why we removed `datachain.func.base.Function`.

We have a lot of these cases in tests, for example: `mutate(t1=func.rand())`.

src/datachain/query/dataset.py
```diff
@@ -795,28 +794,36 @@ def apply_sql_clause(self, query: Select) -> Select:
...
-        ]
-        to_mutate = {c.name for c in args}
-
-        prefix = f"mutate{token_hex(8)}_"
-        cols = [
-            c.label(prefix + c.name) if c.name in to_mutate else c
+        to_mutate = {c.name for c in self.args}
+
+        # Collect columns that are being referenced by simple column renames
+        # e.g., mutate(new_col=C('old_col')) should exclude 'old_col' from base_cols
```
dreadatour: What if we are going to mutate column from a nested signal? Like `.mutate(source=C("file.source"))`?

dreadatour (commented):
Couple questions below.

src/datachain/query/dataset.py
shcheklein: that's a good question, I spent quite a few hours figuring our why we needed first lines of code in the function. Turns out in this particular case we are getting only Label. Function or not - doesn't matter. We do (more or less) something like `v.label(k)` which roughly means `SELECT v as k` for each `mutate(k=v)` AFAIU. And we are not getting Function or anything else directly there - thus we actullaly don't need some code ..

I still want to check btw that part where we duplicate select at the end - I really don't like it, I'm not 100% sure I understand why we need it (apparenly for window operations or something? is it a bug in sqlalchemy? I haven't had time to check - may be you rememeber :))

src/datachain/query/dataset.py
shcheklein: good question, we drop it now, means we have a partial model .. .we had already something like this for select or select_except? right ? need to check this ...

also there is a case I'm looking into - when we do other way around:

`.mutate(file__source=C("tmp))`

where file__source doesn't exist yet - do we allow this? also partial Pydantic model? 

quite complicated ...

dreadatour: Oh, one more good question! It is even more complicated than I thought.

1. In case of mutating field from nested model, I'd suggest to do not delete field from the model. In general for me it make no sense to leave model partial.

    Example: if we want to extract file extension from the file with `ext=func.file_ext("file.path")` and use this extension later, we 100% still want to have `file.path`.

    Another example: `distance=func.byte_hamming_distance("data.s1", "data.s2")`

2. In case of mutating to the inexistent nested field, may be we should raise an error? Like user needs to define correct model first? 🤔

dreadatour: I am now thinking may be it was a good idea to keep the field we are mutating 🤔

Like in SQL `UPDATE` query you're keeping previous values, this feels natural and intuitive.

IMO it is all about naming: `mutate` means converting signal to something else, `update` means adding new fields based on others (or even based on some constant values, — does not matter).

src/datachain/lib/signal_schema.py
```diff
@@ -680,35 +680,46 @@ def mutate(self, args_map: dict) -> "SignalSchema":
...
-                continue
-            if isinstance(value, Column):
+            elif isinstance(value, Column):
                 # adding new signal from existing signal field
-                try:
-                    new_values[name] = self.get_column_type(
-                        value.name, with_subtree=True
-                    )
-                    continue
-                except SignalResolvingError:
```
shcheklein, line 695: I'm not why we had it here. There was a single test, `test_mutate_new_signal` that failed after this was fixed. I fixed the test itself, not clear why would we allow referencing a non-existing column tbh. cc @dreadatour may be you have some ideas.

PR #1300, opened 2025-08-18
Fixing parallel

ilongin:
Companion PR of https://github.com/iterative/studio/pull/12038

We mixed multiprocess and multiprocessing modules when calling `get_context("spawn")`

## Summary by Sourcery

Bug Fixes:
- Replace get_context("spawn") with mp.get_context("spawn") to ensure the correct multiprocessing context is used

### DIFF
--- a/src/datachain/query/dispatch.py
+++ b/src/datachain/query/dispatch.py
@@ -1,4 +1,5 @@
 import contextlib
+import multiprocessing as mp
 from collections.abc import Iterable, Sequence
 from itertools import chain
 from multiprocessing import cpu_count
@@ -8,7 +9,6 @@ from typing import TYPE_CHECKING, Literal, Optional
 import multiprocess
 from cloudpickle import load, loads
 from fsspec.callbacks import DEFAULT_CALLBACK, Callback
-from multiprocess import get_context
 
 from datachain.catalog import Catalog
 from datachain.catalog.catalog import clone_catalog_with_cache
@@ -116,7 +116,7 @@ class UDFDispatcher:
         self.buffer_size = buffer_size
         self.task_queue = None
         self.done_queue = None
-        self.ctx = get_context("spawn")
+        self.ctx = mp.get_context("spawn")
 
     @property
     def catalog(self) -> "Catalog":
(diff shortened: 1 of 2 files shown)


src/datachain/query/dispatch.py
```diff
@@ -116,7 +116,7 @@ def __init__(self, udf_info: UdfInfo, buffer_size: int = DEFAULT_BATCH_SIZE):
         self.buffer_size = buffer_size
         self.task_queue = None
         self.done_queue = None
-        self.ctx = get_context("spawn")
+        self.ctx = mp.get_context("spawn")
```
dreadatour, line 119: I am not sure we can change `multiprocess` with `multiprocessing`, we did use `multiprocess` specifically, with a reasoning behind it.

Same time I am not really remember the reason, could you, please, explain why this is better and why this change fixes the issue? 🙏

src/datachain/query/queue.py
```diff
@@ -1,6 +1,7 @@
 import datetime
 from collections.abc import Iterable, Iterator
-from queue import Empty, Full, Queue
+from multiprocessing.queue import Queue
```
dreadatour, line 3: Same, not sure I understand the reasoning behind this change and why it fix the original issue 🤔 Would be great if you can explain this 🙏

dreadatour: By a quick search through the repos I've found this comment about `multiprocessing.queue`: https://github.com/iterative/dvcx/pull/1297#issuecomment-2026308773

Not sure it is related or not (but I think it is), just wanted to be sure we are fully understand all the reasonings behind all the changes 🙏

src/datachain/query/dispatch.py
dreadatour: Found this issue where we bring `multiprocess.get_context` in: https://github.com/iterative/dvcx/pull/735
It may not be related, just wanted to get more details on this change 🙏

src/datachain/query/queue.py
ilongin: You are right, I was wrongly used `multiprocessing` to get process name in child which ended up being always `MainProcess` and then I changed creating spawn with it but actually I needed to use `multiprocess` to get process name in child process :facepalm:  .. anyway this PR is not needed, closing it.

ilongin:
Changes from this PR turned out not needed.

src/datachain/query/queue.py
dreadatour: Got it! Thank you!
(I was trying to make almost the same changes twice for now 🤣)

PR #1306, opened 2025-08-24
Introduce 'batch_size' param in warehouse 'insert_rows' method

dreadatour:
Now `batch_size` datachain setting does not affects warehouse insert rows.

In this PR I am suggesting changes to respect `batch_size` param also in `insert_rows` implementation. Would be useful in Studio mostly with large scale datasets.

### DIFF
--- a/src/datachain/data_storage/sqlite.py
+++ b/src/datachain/data_storage/sqlite.py
@@ -37,6 +37,7 @@ from datachain import semver
 from datachain.data_storage import AbstractDBMetastore, AbstractWarehouse
 from datachain.data_storage.db_engine import DatabaseEngine
 from datachain.data_storage.schema import DefaultSchema
+from datachain.data_storage.warehouse import INSERT_BATCH_SIZE
 from datachain.dataset import DatasetRecord, StorageURI
 from datachain.error import DataChainError, OutdatedDatabaseSchemaError
 from datachain.namespace import Namespace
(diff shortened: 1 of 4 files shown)


shcheklein (approved):

dmpetrov:
we got quite a bit of different param names: `insert_rows`, `batch_rows`, `buffer_size`.
Why don't we start using a single name for all of these? Like `batch_size` or `insert_size`.

PS: we had a discussion about batch_size in Settings but end up introducing batch_rows and probably batch_size (MB) in the future. But it looks like batch_size in MB has bigger overhead that expected and might not be needed. So, we can get back to `batch_size` as the easiest option and use it everywhere including internals. 
WDYT?

dreadatour:
> we got quite a bit of different param names: `insert_rows`, `batch_rows`, `buffer_size`. Why don't we start using a single name for all of these? Like `batch_size` or `insert_size`.

Agree! 👍

> PS: we had a discussion about batch_size in Settings but end up introducing batch_rows and probably batch_size (MB) in the future. But it looks like batch_size in MB has bigger overhead that expected and might not be needed. So, we can get back to `batch_size` as the easiest option and use it everywhere including internals. WDYT?

Me personally I like `batch_size`, but we can use `batch_rows` to do not change existing user API (`settings` param). IMO `batch_rows` can be a bit confusing, especially in internal code, as it can looks like it is list of rows 🤔

dmpetrov:
> Me personally I like `batch_size`,

yes! let's stick with this. its ok renaming batch_row now

ilongin (approved):
LGTM

PR #1309, opened 2025-08-26
File patterns in read_storage: wildcard, globstar & braces

### LINKED ISSUE #1283: Filename filter shortcut in read_storage()
### Description

It's quite annoying that users need to apply filter in majority of their pipelines while the default filtering code is quite heavy: `.filter(dc.C("file.name").glob("*.mp3"))` 

We need to introduce a shortcuts to simplify it:
- ~~**Pattern based** `.read_storage("s3://mybkt/dir1/dir2/", patterns="*.mp3")` (note, plural - supports list)~~
- **Wildcards** `.read_storage("s3://mybkt/dir1/dir2/dir3/*.mp3")` Also, probably we need `?`.
- **Globstar based** `.read_storage("s3://mybkt/dir1/dir2/**/*.mp3")`
- **Brace Expansion**  `.read_storage("s3://mybkt/dir1/dir2/**/*.{mp3, wav}")`

All should just apply `.filter()` under the hood.

dmpetrov:
Closes #1283
AI generated mostly.

Result: user can do this 
```python
dc.read_storage("s3://mybkt/dir1/dir2/**/*.{png,jpg}")
```
without the need in this
```python
( 
    dc.read_storage("s3://mybkt/dir1/dir2/")
    .filter(dc.C("file.path").glob("*.jpg") | dc.C("file.path").glob("*.png"))
)
```

But it also applies to dirs and combination of these like `dc.read_storage("s3://mybkt/**/{march,april,may}/**/*.{png,jpg}")`

One side effect: `dir1/*` include only files in dir1 (like `dir1/file.txt`) and won't include nested files (like `dir1/dir2/song.mp4`) like we did before (old wildcard). It better aligns with Unix way of expanding file pattern and it's hard to support both.

## Summary by Sourcery

Enable full glob pattern support in read_storage by adding utilities for expanding braces, detecting and splitting patterns, converting globstars to SQL-compatible filters, and integrating pattern-based filtering into the listing process.

New Features:
- Support shell-like glob patterns in read_storage, including wildcards (*, ?), recursive globstars (**), and brace expansions

Enhancements:
- Introduce storage_pattern module with URI pattern splitting, brace expansion, recursion detection, and SQLite-compatible pattern conversion
- Extend read_storage to pre-expand braces, split URIs into base and pattern, and apply filtering via a new _apply_glob_filter helper

Tests:
- Add comprehensive unit and functional tests for wildcard, globstar, brace expansion, question mark, mixed patterns, and edge cases


dmpetrov:
Folks, it's ready to review - please take a look. I'm just fixing some tests/linter.

src/datachain/lib/dc/storage.py
```diff
@@ -138,13 +142,31 @@ def read_storage(
...
 
+    # First, expand all URIs that contain brace patterns
+    expanded_uris = []
+    for single_uri in uris:
+        uri_str = str(single_uri)
+        # Check if URI contains braces and expand them
+        if "{" in uri_str and "}" in uri_str:
+            expanded_uris.extend(expand_uri_braces(uri_str))
+        else:
+            expanded_uris.append(uri_str)
```
dreadatour: Optional: can be simplified, since we already have checks in `expand_uri_braces` and `"{" in uri_str and "}" in uri_str` check is quite a naive check anyway 👀:
```suggestion
        expanded_uris.extend(expand_uri_braces(uri_str))
```

src/datachain/lib/dc/storage.py
```diff
@@ -138,13 +142,31 @@ def read_storage(
...
+        base_uri, glob_pattern = split_uri_pattern(single_uri)
+
+        # If a pattern is found, use the base_uri for listing
+        # The pattern will be used for filtering later
+        list_uri_to_use = base_uri if glob_pattern else single_uri
+
         list_ds_name, list_uri, list_path, list_ds_exists = get_listing(
-            single_uri, session, update=update
+            list_uri_to_use, session, update=update
         )
```
dreadatour: Optional: same, we already have checks in `split_uri_pattern`, so we can simplify this here:
```python
        base_uri, glob_pattern = split_uri_pattern(single_uri)
        list_ds_name, list_uri, list_path, list_ds_exists = get_listing(
            base_uri, session, update=update
        )
```

src/datachain/lib/dc/storage.py
```diff
@@ -138,13 +142,31 @@ def read_storage(
...
+
+    # Now process each expanded URI
     chains = []
     listed_ds_name = set()
     file_values = []
 
-    for single_uri in uris:
+    for single_uri in expanded_uris:
+        # Check if URI contains glob patterns and split them
+        base_uri, glob_pattern = split_uri_pattern(single_uri)
```
dreadatour, line 154: If I am not mistaken, `glob_pattern` will be rewrited here, in case of few `expanded_uris`. Let's say, first one have globs, and second one — does now have, then result `glob_pattern` will be `None` and later check (line `220`) will now works, and globs will not be processed.

We should define `glob_pattern` with default `None` out of the `for` loop scope and set it only if `glob_pattern` from `split_uri_pattern` is not `None`.

Also it might be the case with multiple glob, but we are storing only one (from last URI) and ignoring the rest.

Also since we are storing glob pattern and apply it to all the results, it will affects other URIs in case, let's say, first URI was defined without glob pattern, and second URI have glob pattern. Results will be unpredictable.

dmpetrov: Could you please elaborate?

My understanding - `glob_pattern` is defined for each uri independently and not affecting each other. Could you please explain the cases when it can affect something.

The only thing I was able to find that changes over the loops - bucket update. The same bucket could be updated multiple times which we have to avoid. Change is comming.

dmpetrov (commented):
minor improvements

dmpetrov:
> Also a question: is it possible to use glob pattern in bucket name? What will be a result of: `dc.read_storage(s3://*/**/*)`?

Good catch - obviously it's not supported. Added the validation.

src/datachain/lib/dc/storage.py
dreadatour: Hm, you're absolutely right! Sorry for confusing message from me :(
I was "mislook" the indentation level 😢

dreadatour (approved):
Looks good to me! Great improvement 👍

Wild random controversial idea: implement additional `filter_file` chain method to filter chain by the same globs using the same functions you've added, without the need of using `C("file").glob(...)` (which syntax I don't remember and need to go to the docs all the time). With signature looks something like that: `.filter_file(glob: str, file_signal: str = "file")`.

tests/unit/lib/test_storage_pattern.py
```diff
@@ -0,0 +1,393 @@
...
+
+    # Verify appropriate filtering
+    assert mock_chain.filter.called or mock_ls.called
+
+
+@patch("datachain.lib.dc.storage.get_listing")
+@patch("datachain.lib.dc.storage.ls")
+@patch.object(sys.modules["datachain.lib.dc.datasets"], "read_dataset")
+def test_read_storage_brace_expansion_pattern(
+    mock_read_dataset, mock_ls, mock_get_listing, mock_session, mock_listing
```
shcheklein: Tests like this look very complicated but really test only that a function was called? `mock_chain.filter.called` - is that the whole purpose? To my mind we don't test much, just testing that function is called is not very interesting.

But also there will be quite heavy maintenance cost to such tests (all these Mocks expose internals and will require updates on each refactoring change + good understanding how internals work).

src/datachain/lib/dc/storage_pattern.py
```diff
@@ -0,0 +1,299 @@
...
+    """
+    Validate that cloud storage bucket names don't contain glob patterns.
+
+    Args:
+        uri: URI to validate
+
+    Raises:
+        ValueError: If a cloud storage bucket name contains glob patterns
+    """
+    if not any(uri.startswith(scheme) for scheme in ["s3://", "gs://", "az://"]):
```
shcheklein: No `hf`? Do we really want to hard code the list? Is there a better way to handle this?

src/datachain/lib/dc/storage_pattern.py
```diff
@@ -0,0 +1,299 @@
...
+        return
+
+    # Extract bucket name (everything between :// and first /)
```
shcheklein, line 24: Don't we have some helpers for this in Client / fsspec?

tests/unit/lib/test_storage_pattern.py
dmpetrov: Yeah, removed. Just a small drop in test coverage since it's covered by func tests pretty well.

src/datachain/lib/dc/storage_pattern.py
dmpetrov: good catch. 

1. I added hf with test.

2. I was not able to find a single check for this in fsspec/client - I extracted this to `fsspec.py`

src/datachain/lib/dc/storage_pattern.py
dmpetrov: nope. But I created one.

dmpetrov:
> it would be great to check if some libraries exist to support all of this. It is quite a lot of low level code for the change to maintain later

Yeah, I also was hoping that it will be much easier - that's why I took it. However, it's much more involved when it comes down to the details.

PR #1310, opened 2025-08-27
Fix for delta with retry when multiple error rows with same `id`

ilongin:
Currently, if there are multiple error rows generated with same `id` (name of the column for matching) then, for some unexplained reason, subtract filters out "duplicates" and leaves out only one error row in result. I've also added test for subtract that should've fail as well but for some, yet unknown, reason test passes which means that subtract behaves differently in test and in retry delta code.
Instead of `.subtract()`, `.diff()` was used which works as expected. Note that `.subtract()` was working fine with Clickhouse DB (also not explainable right now).

### DIFF
--- a/src/datachain/delta.py
+++ b/src/datachain/delta.py
@@ -124,7 +124,13 @@ def _get_retry_chain(
     # Subtract also diff chain since some items might be picked
     # up by `delta=True` itself (e.g. records got modified AND are missing in the
     # result dataset atm)
-    return retry_chain.subtract(diff_chain, on=on) if retry_chain else None
+    return (
+        retry_chain.diff(
+            diff_chain, on="id", added=True, same=True, modified=False, deleted=False
+        )
+        if retry_chain
+        else None
+    )
 
 
 def _get_source_info(
--- a/tests/func/test_retry.py
+++ b/tests/func/test_retry.py
@@ -425,3 +425,45 @@ def test_delta_and_delta_retry_no_duplicates(test_session):
     assert len(ids_in_result) == 4
     assert len(set(ids_in_result)) == 4  # No duplicate IDs
     assert set(ids_in_result) == {1, 2, 3, 4}
+
+
+def test_repeating_errors(test_session):
+    from collections.abc import Iterator
+
+    def create_input(num_values):
+        (
+            dc.read_values(
+                id=[i + 1 for i in range(num_values)],
+                name=[str(i + 1) for i in range(num_values)],
+                session=test_session,
+            ).save("input")
+        )
+
+    def run_delta():
+        def func(id) -> Iterator[tuple[int, str, str]]:
+            yield id, "name1", "error"
+            yield id, "name2", "error"
+
+        return (
+            dc.read_dataset(
+                "input",
+                delta=True,
+                delta_on="id",
+                delta_result_on="id",
+                delta_retry="error",
+                session=test_session,
+            )
+            .gen(func, output={"id": int, "name": str, "error": str})
+            .save("processed_data")
+        )
+        return dc.read_dataset("processed_data")
+
+    create_input(1)
+    ch1 = run_delta()
+    assert sorted(ch1.collect("id")) == [1, 1]
+    create_input(2)
+    ch2 = run_delta()
+    assert sorted(ch2.collect("id")) == [1, 1, 1, 1, 2, 2]
+    create_input(3)
+    ch3 = run_delta()
+    assert sorted(ch3.collect("id")) == [1, 1, 1, 1, 1, 1, 1, 1, 2, 2, 2, 2, 3, 3]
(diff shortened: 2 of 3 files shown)


src/datachain/delta.py
```diff
@@ -124,7 +124,13 @@ def _get_retry_chain(
     # Subtract also diff chain since some items might be picked
```
shcheklein, line 125: comment need an update

src/datachain/delta.py
```diff
@@ -124,7 +124,13 @@ def _get_retry_chain(
...
     # up by `delta=True` itself (e.g. records got modified AND are missing in the
     # result dataset atm)
-    return retry_chain.subtract(diff_chain, on=on) if retry_chain else None
+    return (
+        retry_chain.diff(
```
shcheklein, line 131: do we really want this semantics? do we really want to retry multiple times per item? it seems wrong in most case tbh

tests/func/test_retry.py
```diff
@@ -425,3 +425,45 @@ def test_delta_and_delta_retry_no_duplicates(test_session):
     assert len(ids_in_result) == 4
     assert len(set(ids_in_result)) == 4  # No duplicate IDs
     assert set(ids_in_result) == {1, 2, 3, 4}
+
+
+def test_repeating_errors(test_session):
+    from collections.abc import Iterator
+
+    def create_input(num_values):
```
shcheklein: can we reuse existing numerous helpers in this file:

_create_sample_data
_simple_process
_process_with_errors

etc etc ....

shcheklein:
So, does subtract itself work fine or not? is it expected behavior? I'm not sure I fully understand the description ...

src/datachain/delta.py
ilongin: This line is just replacing `.subtract()` with `.diff()` which should produce the same output (nothing should change). The reason I added this is because for some, still unexplained, reason in CLI (SQLite) subtract **here** works differently than expected (wrong). Because of this strange behavior it was **accidentally** removing duplicates in that client's example and I thought that Studio / CH code was the issue and CLI / SQLite was ok but it was other way around. 

So to break it down:
1. Client issue is not actually an issue / bug. Duplicates are expected because of their code, please look at my explanation to them [here](https://iterativeai.slack.com/archives/C08NDNSM80J/p1756299073891839?thread_ts=1756124980.092669&cid=C08NDNSM80J)
2. On the other hand, there **is** an issue in CLI / SQLite caused by this wrong subtract behavior that I accidentally found when debugging client "issue" and is fixed by using `.diff()` instead which works as expected and now CH and SQLite are working the same / consistent.
3. I couldn't reproduce this wrong `.subtract()` SQLite behavior in isolated test (added it in this PR) which means it's probably something related to specific context in `delta` logic but I don't think we should waste more time on that (already I waste too much IMO).

tests/func/test_retry.py
ilongin: Added `_create_sample_data`. Others are not useful for this test

src/datachain/delta.py
shcheklein: > Client issue is not actually an issue / bug. Duplicates are expected because of their code, please look at my explanation to them [here](https://iterativeai.slack.com/archives/C08NDNSM80J/p1756299073891839?thread_ts=1756124980.092669&cid=C08NDNSM80J)

I think I understand the reason, my question is - do we really want to keep that semantics or would it better to deduplicate records based on `on`. So, even if the target table has multiple errors witt the same measurement_ids, we take it only once. I think it is more expected.

This will solve the issue from the product perspective and remove this discrepancy anyways. (though it is an interested thing - might hurt us somewhere else down the road)

ilongin: Yea, it's interesting question but I think we should not do that as we might disrupt specific client logic with it. Note that those error rows from client issue are not completely the same (they differ in one column) and the first question is which one to discard and which one to leave. If they don't want this kind of behavior, they can easily fix it in their business logic.

shcheklein: Tbh, looking more into this, I think retrying once is the only right option here. We are _retrying_ input items. And it is a single one in this case. Unless im missing something. We are not retrying outputs.

They could indeed change logic, but it might be very inconvenient. It is fine to use gen and produce multiple items per a single row.

shcheklein: We need to discard all based on the 'on' value.

shcheklein: It also doesn't make sense that shape keeps changing on a retry run with a very regular gen.

ilongin: But we are not retrying input items, the whole implementation now gets error rows from result, which makes sense as that's where the errors will be. Output is not 1-1 with input, specially if you have `gen` which can generate totally random number of various rows (including error rows) and input rows are not kept after it. We cannot know from which input we got error at the end in that case.

ilongin: If they didn't generate 2 error rows from one input row, there wouldn't be any issue.

ilongin: Aha, sorry, we do merge source records with error rows based on `on` column and we use `source` schema ..

shcheklein: They might need multiple errors. Moreover it can be a mix (errors and not). It would wrong and complicated to change it to generate a single row in case of any errors

shcheklein (approved):

PR #1313, opened 2025-08-31
File open to make streaming uploads possible

shcheklein:
To make code like this work:

```python
    plt.figure(figsize=(6, 4))
    ...
    
    file_obj = File.at(dest_uri)
    with file_obj.open("wb") as fh:
        plt.savefig(fh, format="png")
    plt.close()

    return file_obj
```

vs:

```python
    buf = io.BytesIO()
    plt.savefig(buf, format='png')
    plt.close()
    buf.seek(0)
    return File.upload(buf.getvalue(), "s3://597088047692-uw2-skysound-datachain/plots/plot.png")
```

That we have to use now.

The same applies to parquet, we can now write it in a streaming fashion.


dmpetrov:
Love the change!

> ```python
>     file_obj = File.at(dest_uri)
>     with file_obj.open("wb") as fh:
> ```

Just a thought... it would be great to have static method `with File.open(dest) as fd:` But it's already taken by a non-static method with no params. I'm wondering if there is a way to make it work like `with File(source="ss", path="qwerty").open("wb") as fd:`. In this case source and path separation is not convenient. So, `.at()` solves the problem. 🙂  Please share ideas folks.

shcheklein:
> file_obj = File.at(dest_uri)

Is needed primarily to capture the result - we need an updated version of the File with all the information returned by a cloud. Unlike regular fs.open in Python - thus the difference. If we go with open like this then we will have to capture the object returned anyways AFAIU.

Happy to review other ideas.

I was considering an experimenting with providing a Stream to `File.upload` .. which can be also a solution, but then requires consumers being able to provide a stream which can be non-trivial, most likely we'll end up creating helpers for parquet, etc, etc. Writing to a file descriptor seems to be supported by a lot of types of libs. And already can help avoiding creating an in-memory buffer.

Anyways, it is still a draft - I'm experimenting with different options atm.

tests/func/test_file.py
```diff
@@ -91,3 +94,69 @@
...
+            # (some backends don't keep microsecond precision, we keep it simple here)
+            assert match.last_modified.timestamp() == pytest.approx(
+                file.last_modified.timestamp(), abs=1
+            )
+        else:
+            assert getattr(match, field_name) == getattr(file, field_name), (
+                f"Mismatch in field '{field_name}'"
+            )
+
+    catalog.get_client(src_uri).fs.rm(file_path)
```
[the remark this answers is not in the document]
shcheklein: Not related to this PR

shcheklein:
@sourcery-ai review

dreadatour:
Love the idea, I always wanted to refactor `File.open` method but it was no chance to do that.

> it would be great to have static method `with File.open(dest) as fd:` But it's already taken by a non-static method with no params.

Can we use overload here? Like to bring an additional `path` param and check `self` type to know if it was called as a class method or an instance method? 🤔 Or it will be too much magic?

src/datachain/lib/file.py
```diff
@@ -354,36 +383,78 @@ def parent(self) -> str:
...
+        writing = any(ch in mode for ch in "wax+")
+        if self.location and writing:
+            raise VFileError(
+                "Writing to virtual file is not supported",
+                self.source,
+                self.path,
+            )
+
+        if self._catalog is None:
+            raise RuntimeError("Cannot open file: catalog is not set")
```
ilongin, line 402: This is public API, right? I don't think this error message will make sense to the users .. they don't know what catalog is at all.

src/datachain/lib/file.py
```diff
@@ -341,6 +349,27 @@ def upload(
         file._set_stream(catalog)
         return file
 
+    @classmethod
+    def at(cls, uri: str, catalog: Optional["Catalog"] = None) -> "Self":
```
ilongin: I don't think we should be setting `Catalog` as argument to public API. This is our internal class and unknown and undocumented for users. In other public API we have `Session` which is also little bit questionable but is better than catalog. Is this added only for tests to inject right one?

dreadatour: Hm, sounds reasonable, I agree 👍
If I am not mistaken we also have `catalog` as an argument in many other places, we may want to cleanup all these places if we'll have agreement on this.

src/datachain/lib/file.py
dreadatour: In theory this should never happens and if it is happens, then something really wrong is going on and I think it is indeed a `RuntimeError`. Other option is to use custom `InternalError`-like exception, but this will be the same as `RuntimeError`.

shcheklein: yes, exactly, this is just defensive ... like `assert`

src/datachain/lib/file.py
shcheklein: yeah, picked up from the `upload` that we have

It was needed probably for tests (like in most other cases)

but yes, if anything it is better to use Session + even inside it is better to take catalog from session AFAIU

but overall, it would be nice to review sessions and catalogs at some point

shcheklein:
@ilongin PTAL

PR #1315, opened 2025-09-02
Update the changes to job API

amritghimire:
This uses the changes made to the job related apis https://github.com/iterative/studio/pull/12067 in an effort to make apis
public and uniform.

## Summary by Sourcery

Update Studio client and CLI to align with uniform job and cluster APIs by pluralizing endpoints, switching file uploads to multipart with retry, and adapting to new response structures

Enhancements:
- Add multipart/form-data upload support with retry logic in Studio client
- Change upload_file to use multipart upload at /datachain/jobs/files
- Standardize job and cluster routes to use pluralized endpoints
- Update JobListData and ClusterListData aliases to list of dicts
- Extend get_jobs to accept an optional job_id filter
- Adapt CLI to extract top-level id and url fields and show the new 'Is Default' cluster property

Tests:
- Update CLI tests to mock new pluralized endpoints and verify multipart/form-data uploads and query parameters

### DIFF
--- a/src/datachain/remote/studio.py
+++ b/src/datachain/remote/studio.py
@@ -240,7 +240,6 @@ class StudioClient:
 
         return Response(data, ok, message, response.status_code)
 
-    @retry_with_backoff(retries=3, errors=(HTTPError, Timeout))
     def _send_multipart_request(
         self, route: str, files: dict[str, Any], params: Optional[dict[str, Any]] = None
     ) -> Response[Any]:
@@ -263,12 +262,6 @@ class StudioClient:
             },
             timeout=self.timeout,
         )
-        try:
-            response.raise_for_status()
-        except HTTPError:
-            if _is_server_error(response.status_code):
-                # going to retry
-                raise
 
         ok = response.ok
         try:
--- a/src/datachain/remote/studio.py
+++ b/src/datachain/remote/studio.py
@@ -1,4 +1,3 @@
-import base64
 import json
 import logging
 import os
@@ -7,6 +6,7 @@ from datetime import datetime, timedelta, timezone
 from struct import unpack
 from typing import (
     Any,
+    BinaryIO,
     Generic,
     Optional,
     TypeVar,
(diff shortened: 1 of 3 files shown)


src/datachain/remote/studio.py
```diff
@@ -239,6 +240,52 @@ def _send_request(
 
         return Response(data, ok, message, response.status_code)
 
+    @retry_with_backoff(retries=3, errors=(HTTPError, Timeout))
```
dreadatour: I would be really careful with retries for requests sent with `POST` method by default 👀
May be we should add `retry` decorator in-place, for methods we are sure it is safe to retry (like `get_jobs`) and not add it by default for other methods (like `upload_file`, which can in theory fail with timeout error but file will be uploaded successfully)? What do you think?

src/datachain/remote/studio.py
```diff
@@ -239,6 +240,52 @@ def _send_request(
...
+
+        # Add team_name to params
+        request_params = {**(params or {}), "team_name": self.team}
+
+        response = requests.post(
+            url=f"{self.url}/{route}",
+            files=files,
+            params=request_params,
+            headers={
+                "Authorization": f"token {self.token}",
```
dreadatour, line 261: Just wonder, — not remember on Studio side, but why don't we use `Bearer` or `Api-Key` types here and using `token`?

Tiny note: not really make any difference here, depends on realization, but type is is usually capitalized: `Token ...`, `Bearer ...`, etc.

dreadatour (approved):
Looks good to me! 👍 Couple questions below.

src/datachain/remote/studio.py
dreadatour: Btw, in docs we are using `Bearer` type: https://a2d6c428.datachain-documentation.pages.dev/studio/api/#authorization 🤔

amritghimire: I am not sure I remember the reason, but we had always supported both token and bearer method.

src/datachain/remote/studio.py
amritghimire: Thanks, removed the retry for at least multipart upload for now.

PR #1317, opened 2025-09-03
Added `QueryStep`

ilongin:
`QueryStep` is new model that tracks each specific chain in job query script. We track when chain has began to process and when it ended. This will be needed later on with checkpoints, as we need to track chains somehow and skip those that have successfully ended on job  re-run.

## Summary by Sourcery

Introduce JobQueryStep tracking to record each query step’s lifecycle (start, finish, error) in the metastore and integrate it into the Datachain save workflow

New Features:
- Add JobQueryStep dataclass and JobQueryStepStatus enum
- Extend AbstractMetastore API and SQLite schema with CRUD operations for job query steps
- Instrument Datachain.save to create and update query steps on job start, completion, and failure

Tests:
- Add functional tests for success and error scenarios of job query step tracking

### DIFF
--- a/src/datachain/data_storage/__init__.py
+++ b/src/datachain/data_storage/__init__.py
@@ -1,4 +1,5 @@
 from .job import JobQueryType, JobStatus
+from .job_query_step import JobQueryStepStatus
 from .metastore import AbstractDBMetastore, AbstractMetastore
 from .warehouse import AbstractWarehouse
 
@@ -6,6 +7,7 @@ __all__ = [
     "AbstractDBMetastore",
     "AbstractMetastore",
     "AbstractWarehouse",
+    "JobQueryStepStatus",
     "JobQueryType",
     "JobStatus",
 ]
--- a/src/datachain/data_storage/job_query_step.py
+++ b/src/datachain/data_storage/job_query_step.py
@@ -1,0 +1,7 @@
+from enum import Enum
+
+
+class JobQueryStepStatus(int, Enum):
+    RUNNING = 10
+    FINISHED = 20
+    ERROR = 30
--- a/src/datachain/data_storage/metastore.py
+++ b/src/datachain/data_storage/metastore.py
@@ -23,7 +23,7 @@ from sqlalchemy import (
     select,
 )
 
-from datachain.data_storage import JobQueryType, JobStatus
+from datachain.data_storage import JobQueryStepStatus, JobQueryType, JobStatus
 from datachain.data_storage.serializer import Serializable
 from datachain.dataset import (
     DatasetDependency,
@@ -41,7 +41,7 @@ from datachain.error import (
     ProjectNotFoundError,
     TableMissingError,
 )
-from datachain.job import Job
+from datachain.job import Job, JobQueryStep
 from datachain.namespace import Namespace
 from datachain.project import Project
 from datachain.utils import JSONSerialize
@@ -72,6 +72,7 @@ class AbstractMetastore(ABC, Serializable):
     dataset_list_version_class: type[DatasetListVersion] = DatasetListVersion
     dependency_class: type[DatasetDependency] = DatasetDependency
     job_class: type[Job] = Job
+    job_query_step_class: type[JobQueryStep] = JobQueryStep
 
     def __init__(
         self,
@@ -400,6 +401,41 @@ class AbstractMetastore(ABC, Serializable):
     def get_job_status(self, job_id: str) -> Optional[JobStatus]:
         """Returns the status of the given job."""
 
+    #
+    # Jobs query steps
+    #
+
+    @abstractmethod
+    def list_jobs_query_steps(self, job_id: str, conn=None) -> Iterator["JobQueryStep"]:
+        """Returns all jobs queries related to some job"""
+
+    @abstractmethod
+    def get_job_query_step(
+        self, job_query_step_id: str, conn=None
+    ) -> Optional[JobQueryStep]:
+        """Gets single job query step by id"""
+
+    @abstractmethod
+    def create_job_query_step(
+        self,
+        job_id: str,
+        status: JobQueryStepStatus = JobQueryStepStatus.RUNNING,
+        conn: Optional[Any] = None,
+    ) -> str:
+        """Creates new job query step"""
+
+    @abstractmethod
+    def update_job_query_step(
+        self,
+        job_query_step_id: str,
+        status: Optional[JobQueryStepStatus] = None,
+        error_message: Optional[str] = None,
+        error_stack: Optional[str] = None,
+        finished_at: Optional[datetime] = None,
+        conn: Optional[Any] = None,
+    ) -> Optional["JobQueryStep"]:
+        """Updates job query step fields"""
+
 
 class AbstractDBMetastore(AbstractMetastore):
     """
@@ -415,6 +451,7 @@ class AbstractDBMetastore(AbstractMetastore):
     DATASET_VERSION_TABLE = "datasets_versions"
     DATASET_DEPENDENCY_TABLE = "datasets_dependencies"
     JOBS_TABLE = "jobs"
+    JOBS_QUERY_STEPS_TABLE = "jobs_query_steps"
 
     db: "DatabaseEngine"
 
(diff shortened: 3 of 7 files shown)


shcheklein:
is it in draft / ready to review?

src/datachain/data_storage/__init__.py
```diff
@@ -1,11 +1,13 @@
 from .job import JobQueryType, JobStatus
+from .job_query_step import JobQueryStepStatus
 from .metastore import AbstractDBMetastore, AbstractMetastore
 from .warehouse import AbstractWarehouse
 
 __all__ = [
     "AbstractDBMetastore",
     "AbstractMetastore",
     "AbstractWarehouse",
+    "JobQueryStepStatus",
```
shcheklein, line 10: we have too many job and query steps ... :(

src/datachain/lib/dc/datachain.py
```diff
@@ -593,88 +596,118 @@ def save(  # type: ignore[override]
...
         try:
-            project = self.session.catalog.metastore.get_project(
-                project_name,
-                namespace_name,
-                create=is_studio(),
-            )
-        except ProjectNotFoundError as e:
-            # not being able to create it as creation is not allowed
-            raise ProjectCreateNotAllowedError("Creating project is not allowed") from e
+            if job_id := os.getenv("DATACHAIN_JOB_ID"):
```
shcheklein, line 603: where do we set it? why we need this if?

shcheklein:
Tests fail, some naming concerns, otherwise we can go forward I think.

@dreadatour check please metastore API and context when we call it, etc - that we are not missing anything

@amritghimire check please in the corresponding Studio API that we do RBAC and other things properly

src/datachain/data_storage/__init__.py
shcheklein: Screenshot 2025-09-04 at 2 15 07 PM

we already have this for the job, it will become even more confusing

call them Chains or Stages (or rename those Steps into Stages - even better tbh)

ilongin: I was actually thinking to rename this to just `Chains` ... `QueryStep` is too generic and ambiguous name anyway. Also what you mentioned, there are too many "Steps".

src/datachain/lib/dc/datachain.py
ilongin: It's set in `Catalog.query(...)` when calling subprocess to run the script.

If you just run your python script with dc code, e.g `python my-script.py` then we don't create any job, chains created are not related to a specific jobs and this env var is not set. 
On the other hand, if you run `datachain query my-script.py` (uses `Catalog.query(...)` under) then `Job` is created even in open source and is sent to subprocess that is running the actual code via this env variable. When you run from UI `Job` is always created as we know and is using the same `Cataog.query(...)`

shcheklein: omg, I forgot about `datachain query my-script.py` ... who is using it? what is the purpose for this command to exist? what is the different with the `datachain job run`?

src/datachain/data_storage/job_query_step.py
```diff
@@ -0,0 +1,7 @@
+from enum import Enum
+
+
+class JobQueryStepStatus(int, Enum):
+    RUNNING = 10
+    FINISHED = 20
+    ERROR = 30
```
dreadatour, line 7: We use incremental ints everywhere, I like this approach with having gaps, considering future updates, but for consistency across all of the codebase in two repos let's use the same incremental int values? 🙏

src/datachain/data_storage/metastore.py
```diff
@@ -400,6 +401,41 @@ def set_job_status(
     def get_job_status(self, job_id: str) -> Optional[JobStatus]:
         """Returns the status of the given job."""
 
+    #
+    # Jobs query steps
+    #
+
+    @abstractmethod
+    def list_jobs_query_steps(self, job_id: str, conn=None) -> Iterator["JobQueryStep"]:
```
dreadatour, line 409:
```suggestion
    def list_jobs_query_steps(self, job_id: str, conn=None) -> Iterator[JobQueryStep]:
```
for consistency (see `get_job_query_step` typing below)

src/datachain/data_storage/metastore.py
```diff
@@ -400,6 +401,41 @@ def set_job_status(
...
+    @abstractmethod
+    def update_job_query_step(
+        self,
+        job_query_step_id: str,
+        status: Optional[JobQueryStepStatus] = None,
+        error_message: Optional[str] = None,
+        error_stack: Optional[str] = None,
+        finished_at: Optional[datetime] = None,
+        conn: Optional[Any] = None,
+    ) -> Optional["JobQueryStep"]:
```
dreadatour, line 436:
```suggestion
    ) -> Optional[JobQueryStep]:
```
also for consistency.

src/datachain/data_storage/metastore.py
```diff
@@ -1607,3 +1644,138 @@ def get_job_status(
         if not results:
             return None
         return results[0][0]
+
+    #
+    # Job query step
+    #
+
+    @staticmethod
+    def _jobs_query_steps_columns() -> "list[SchemaItem]":
```
dreadatour, line 1653:
```suggestion
    def _jobs_query_steps_columns() -> list["SchemaItem"]:
```

src/datachain/data_storage/metastore.py
```diff
@@ -1607,3 +1644,138 @@ def get_job_status(
...
+        return [
+            Column(
+                "id",
+                Text,
+                default=uuid4,
+                primary_key=True,
+                nullable=False,
+            ),
+            Column("job_id", Text, nullable=True),
```
dreadatour, line 1662: How it can be null? 🤔

src/datachain/data_storage/metastore.py
```diff
@@ -1607,3 +1644,138 @@ def get_job_status(
...
+    def _jobs_query_steps_insert(self) -> "Insert": ...
+
+    def _jobs_query_steps_select(self, *columns) -> "Select":
+        if not columns:
+            return self._jobs_query_steps.select()
+        return select(*columns)
+
+    def _jobs_query_steps_update(self, *where) -> "Update":
+        if not where:
+            return self._jobs_query_steps.update()
```
dreadatour, line 1694: I know it is not related to this PR, but should we reject updates without `where` clause? 🤔

src/datachain/data_storage/metastore.py
```diff
@@ -1607,3 +1644,138 @@ def get_job_status(
...
+        return self._jobs_query_steps.update().where(*where)
+
+    def _parse_job_query_step(self, rows) -> JobQueryStep:
+        return self.job_query_step_class.parse(*rows)
+
+    def _parse_jobs_query_steps(self, rows) -> Iterator["JobQueryStep"]:
```
dreadatour, line 1700:
```suggestion
    def _parse_jobs_query_steps(self, rows) -> Iterator[JobQueryStep]:
```
for the consistency with `_parse_job_query_step` above and all other methods. Also below 🙏

src/datachain/data_storage/metastore.py
```diff
@@ -1607,3 +1644,138 @@ def get_job_status(
...
+        for _, g in groupby(rows, lambda r: r[1]):
+            yield self._parse_job_query_step(*list(g))
+
+    def _jobs_query_steps_query(self):
```
dreadatour, line 1704: Return type?

src/datachain/lib/dc/datachain.py
```diff
@@ -593,88 +596,118 @@ def save(  # type: ignore[override]
...
+            error_message = str(e)
+            error_stack = traceback.format_exc()
+            raise
+        finally:
+            if job_id:
+                status = JobQueryStepStatus.FINISHED
+                finished_at: Optional[datetime] = datetime.now(timezone.utc)
+                if error_message:
+                    status = JobQueryStepStatus.ERROR
+                    finished_at = None
```
dreadatour, line 702: Why we don't set `finished` here? This run is finished for now.

src/datachain/lib/dc/datachain.py
```diff
@@ -593,88 +596,118 @@ def save(  # type: ignore[override]
...
-        namespace_name, project_name, name = catalog.get_full_dataset_name(
-            name,
-            namespace_name=self._settings.namespace,
-            project_name=self._settings.project,
-        )
-
+        error_message = ""
+        error_stack = ""
+        job_query_step_id = None
         try:
```
dreadatour, line 602: May be move everything from this `try` block to separate method, so we will not have huge function with deep nesting? 🤔

ilongin:
Closing in the favor of https://github.com/iterative/datachain/pull/1348

Issue #1302, opened 2025-08-19
Remove `name` and `version` properties from `DataChain`


ilongin:
Currently we have `name` and `version` properties in `DataChain` class which are not needed as we already have `dataset` property which points to underlying dataset if one exists.
Also, in addition we should think about `namespace_name` and `project_name` as well which are also properties. Probably we should remove those and think about exposing settings in general.

dmpetrov:
> `namespace_name` and `project_name`

Please combine these two to a single `namespace` that contains both.

ilongin:
> > `namespace_name` and `project_name`
> 
> Please combine these two to a single `namespace` that contains both.

I would avoid doing this atm. All around the code we have this split into two so if we want to have only namespace that contains both we should do one bigger refactoring to be consistent everywhere.

shcheklein:
I think Dmitry's point was to do changes on the public API level. What would be the scope for this?

ilongin:
> I think Dmitry's point was to do changes on the public API level. What would be the scope for this?

It should not be a problem to do that, will do it in this issue. Also, I will create a follow-up to use `namespace.project.dataset` naming convention everywhere in our codebase where we use dataset name (internal and external APIs) to avoid `namespace_name` and `project_name` arguments in a lot of functions.

shcheklein:
Can you scope it please before you do it?

ilongin:
> Can you scope it please before you do it?

From my short investigation in `datachain` we need to change:

- `lib.dc.datasets.read_dataset()`
- `lib.dc.datasets.delete_dataset()`
- `lib.dc.datachain.DataChain.settings()`

We need to check Studio usage of those and change them as well if needed.

Overall I think it's not a big task, but we do change public API - that's the biggest issue here

shcheklein:
How about env variables that they use now? What else are we missing? can we carefully grep and think what exactly we'll break?

how will the new API look like exactly (env vars, settings, read_dataset) - can you please describe it here ?

can we make it with some deprication to give some time to migrate first and then drop support later? 

please scope it e2e with a proper plan and ETA

ilongin:
New API will look the same for now, except that `namespace` argument we will now cover both `namespace` and `project` .. in future, after some deprecation time it would go from 
```python
def read_dataset(
    name: str,
    namespace: Optional[str] = None,
    project: Optional[str] = None,
    ...
)
```
to
```python
def read_dataset(
    name: str,
    namespace: Optional[str] = None,
[...]

PR #1318, opened 2025-09-05
Added `delete_namespace` and `delete_project` functions

ilongin:
Up until now we didn't have the way to remove namespace or project. This PR adds 2 functions: `delete_project` and `delete_namespace`

There are some rules:
1. Cannot delete default namespace or project
2. Cannot delete system namespace
3. Cannot delete listing project
4. Cannot delete non empty project (has datasets in it)
5. Cannot delete non empty namespace (has projects in it)

## Summary by Sourcery

Enable removal of projects and namespaces with corresponding metastore support, error handling, and comprehensive tests.

New Features:
- Add delete_project function to remove a project with safety checks
- Add delete_namespace function to remove a namespace with safety checks

Enhancements:
- Extend metastore with remove_project, remove_namespace, count_datasets, is_default_project, and is_listing_project methods

Tests:
- Add unit tests for delete_project covering success, not found, listing/default, and non-empty scenarios
- Add unit tests for delete_namespace covering success, not found, system/default, and non-empty scenarios

Chores:
- Define ProjectDeleteNotAllowedError and NamespaceDeleteNotAllowedError error classes
- Export delete_project in the public API

### DIFF
--- a/src/datachain/catalog/catalog.py
+++ b/src/datachain/catalog/catalog.py
@@ -144,26 +144,19 @@ def shutdown_process(
             return proc.wait()
 
 
-def process_output(stream: IO[bytes], callback: Callable[[str], None]) -> None:
+def _process_stream(stream: "IO[bytes]", callback: Callable[[str], None]) -> None:
     buffer = b""
+    while byt := stream.read(1):  # Read one byte at a time
+        buffer += byt
 
-    try:
-        while byt := stream.read(1):  # Read one byte at a time
-            buffer += byt
-
-            if byt in (b"\n", b"\r"):  # Check for newline or carriage return
-                line = buffer.decode("utf-8", errors="replace")
-                callback(line)
-                buffer = b""  # Clear buffer for the next line
-
-        if buffer:  # Handle any remaining data in the buffer
-            line = buffer.decode("utf-8", errors="replace")
+        if byt in (b"\n", b"\r"):  # Check for newline or carriage return
+            line = buffer.decode("utf-8")
             callback(line)
-    finally:
-        try:
-            stream.close()  # Ensure output is closed
-        except Exception:  # noqa: BLE001, S110
-            pass
+            buffer = b""  # Clear buffer for next line
+
+    if buffer:  # Handle any remaining data in the buffer
+        line = buffer.decode("utf-8")
+        callback(line)
 
 
 class DatasetRowsFetcher(NodesThreadPool):
@@ -1767,13 +1760,13 @@ class Catalog:
             recursive=recursive,
         )
 
-    @staticmethod
     def query(
+        self,
         query_script: str,
         env: Optional[Mapping[str, str]] = None,
         python_executable: str = sys.executable,
-        stdout_callback: Optional[Callable[[str], None]] = None,
-        stderr_callback: Optional[Callable[[str], None]] = None,
+        capture_output: bool = False,
+        output_hook: Callable[[str], None] = noop,
         params: Optional[dict[str, str]] = None,
         job_id: Optional[str] = None,
         interrupt_timeout: Optional[int] = None,
@@ -1788,18 +1781,13 @@ class Catalog:
             },
         )
         popen_kwargs: dict[str, Any] = {}
-
-        if stdout_callback is not None:
-            popen_kwargs = {"stdout": subprocess.PIPE}
-        if stderr_callback is not None:
-            popen_kwargs["stderr"] = subprocess.PIPE
+        if capture_output:
+            popen_kwargs = {"stdout": subprocess.PIPE, "stderr": subprocess.STDOUT}
 
         def raise_termination_signal(sig: int, _: Any) -> NoReturn:
             raise TerminationSignal(sig)
 
-        stdout_thread: Optional[Thread] = None
-        stderr_thread: Optional[Thread] = None
-
+        thread: Optional[Thread] = None
         with subprocess.Popen(cmd, env=env, **popen_kwargs) as proc:  # noqa: S603
             logger.info("Starting process %s", proc.pid)
 
@@ -1813,20 +1801,10 @@ class Catalog:
             orig_sigterm_handler = signal.getsignal(signal.SIGTERM)
             signal.signal(signal.SIGTERM, raise_termination_signal)
             try:
-                if stdout_callback is not None:
-                    stdout_thread = Thread(
-                        target=process_output,
-                        args=(proc.stdout, stdout_callback),
-                        daemon=True,
-                    )
-                    stdout_thread.start()
-                if stderr_callback is not None:
-                    stderr_thread = Thread(
-                        target=process_output,
-                        args=(proc.stderr, stderr_callback),
-                        daemon=True,
-                    )
-                    stderr_thread.start()
+                if capture_output:
+                    args = (proc.stdout, output_hook)
+                    thread = Thread(target=_process_stream, args=args, daemon=True)
+                    thread.start()
 
                 proc.wait()
             except TerminationSignal as exc:
@@ -1844,22 +1822,8 @@ class Catalog:
             finally:
                 signal.signal(signal.SIGTERM, orig_sigterm_handler)
                 signal.signal(signal.SIGINT, orig_sigint_handler)
-                # wait for the reader thread
-                thread_join_timeout_seconds = 30
-                if stdout_thread is not None:
-                    stdout_thread.join(timeout=thread_join_timeout_seconds)
-                    if stdout_thread.is_alive():
-                        logger.warning(
-                            "stdout thread is still alive after %s seconds",
-                            thread_join_timeout_seconds,
-                        )
-                if stderr_thread is not None:
-                    stderr_thread.join(timeout=thread_join_timeout_seconds)
-                    if stderr_thread.is_alive():
-                        logger.warning(
-                            "stderr thread is still alive after %s seconds",
-                            thread_join_timeout_seconds,
-                        )
+                if thread:
+                    thread.join()  # wait for the reader thread
 
         logger.info("Process %s exited with return code %s", proc.pid, proc.returncode)
         if proc.returncode in (
--- a/src/datachain/lib/file.py
+++ b/src/datachain/lib/file.py
@@ -35,6 +35,7 @@ if TYPE_CHECKING:
     from datachain.catalog import Catalog
     from datachain.client.fsspec import Client
     from datachain.dataset import RowDict
+    from datachain.query.session import Session
 
 sha256 = partial(hashlib.sha256, usedforsecurity=False)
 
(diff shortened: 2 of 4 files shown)


shcheklein:
Ci fails?

src/datachain/data_storage/metastore.py
```diff
@@ -190,10 +195,22 @@ def get_project(
         It also creates project if not found and create flag is set to True.
         """
 
+    @abstractmethod
+    def is_default_project(self, project_name: str, namespace_name: str) -> bool:
+        """Checks if some project in namespace is default"""
+
+    @abstractmethod
+    def is_listing_project(self, project_name: str, namespace_name: str) -> bool:
```
shcheklein, line 206: Can we prevent it in the db level as well? Have a flag for all special ovjects. To make it more reliable and not easy to hack

src/datachain/lib/namespaces.py
```diff
@@ -71,3 +74,49 @@ def ls(session: Optional[Session] = None) -> list[Namespace]:
...
+            as these cannot be removed.
+
+    Parameters:
+        name : The name of the namespace.
+        session : Session to use for getting project.
+
+    Example:
+        ```py
+        import datachain as dc
+        from datachain.lib.namespace import delete as delete_namespace
```
shcheklein, line 97: This looks bad. Like API is not named good and we have to rename it

src/datachain/lib/projects.py
```diff
@@ -86,3 +86,50 @@ def ls(
...
+        from datachain.lib.projects import delete as delete_project
+        delete_project("my-project", "local")
+        ```
+    """
+    session = Session.get(session)
+    metastore = session.catalog.metastore
+
+    project = metastore.get_project(name, namespace)
+
+    if metastore.is_listing_project(name, namespace):
```
shcheklein, line 117: This checks should be also happening on the API side + on the db level.

src/datachain/__init__.py
```diff
@@ -74,6 +76,8 @@
     "create_project",
     "datasets",
     "delete_dataset",
+    "delete_namespace",
+    "delete_project",
```
dmpetrov: Let's don't expose project in user's api. Instead you can do something like:

```python
def delete_namespace(name: str):
    parts = name.split(".")
    if len(parts) == 1:
        remove_namespace(parts[0])
    elif len(parts) == 2:
        delete_project(parts[1])
    else:
        raise ValueError(f"Invalid namespace format: {name}. Expected 'namespace' or 'ns1.ns2'.")
```

dmpetrov (changes requested):
LG but there is a request to change public API

src/datachain/data_storage/metastore.py
```diff
@@ -141,6 +142,10 @@ def create_namespace(
     def get_namespace(self, name: str, conn=None) -> Namespace:
         """Gets a single namespace by name"""
 
+    @abstractmethod
+    def remove_namespace(self, namespace_id: int, conn=None) -> None:
+        """Removes a single namespace by id"""
```
dreadatour, line 149: Just a heads-up: if we add new methods to the metastore in the Datachain project, we’ll also need a corresponding Studio PR. Otherwise, merging the Datachain PR first could potentially break the Studio release.

src/datachain/data_storage/metastore.py
```diff
@@ -735,6 +756,11 @@ def create_namespace(
 
         return self.get_namespace(name)
 
+    def remove_namespace(self, namespace_id: int, conn=None) -> None:
+        n = self._namespaces
+        with self.db.transaction():
+            self.db.execute(self._namespaces_delete().where(n.c.id == namespace_id))
```
dreadatour, line 779: I am not sure, just wonder, should we add another check here (so no datasets exists within this namespace), or is it enough to have this check on a higher level? May be consider to move the check here and raise an exceptions (like "default namespace can not be deleted" or "namespace with datasets can not be deleted) to be on the safe side? 🤔

src/datachain/data_storage/metastore.py
```diff
@@ -858,6 +884,11 @@ def get_project_by_id(self, project_id: int, conn=None) -> Project:
             raise ProjectNotFoundError(f"Project with id {project_id} not found.")
         return self.project_class.parse(*rows[0])
 
+    def remove_project(self, project_id: int, conn=None) -> None:
+        p = self._projects
+        with self.db.transaction():
+            self.db.execute(self._projects_delete().where(p.c.id == project_id))
```
dreadatour, line 912: Same question here with checks on this level 🤔

src/datachain/data_storage/metastore.py
```diff
@@ -1198,6 +1228,16 @@ def list_datasets(
             query = query.where(d.c.project_id == project_id)
         yield from self._parse_dataset_list(self.db.execute(query))
 
+    def count_datasets(self, project_id: Optional[int] = None) -> int:
+        d = self._datasets
+        query = self._base_list_datasets_query()
```
dreadatour: This is not optimal IMO, since it will JOIN all datasets versions. Should we have simpler `COUNT` query here to count datasets only?

src/datachain/lib/namespaces.py
```diff
@@ -71,3 +74,49 @@ def ls(session: Optional[Session] = None) -> list[Namespace]:
...
+        raise NamespaceDeleteNotAllowedError(
+            f"Namespace {metastore.system_namespace_name} cannot be removed"
+        )
+
+    if name == metastore.default_namespace_name:
+        raise NamespaceDeleteNotAllowedError(
+            f"Namespace {metastore.default_namespace_name} cannot be removed"
+        )
+
+    projects = metastore.list_projects(namespace.id)
```
dreadatour: May be add `count_projects` or `has_projects` method to do not fetch all the projects (similar to `count_datasets`)?
Otherwise in Studio this will pass all the projects objects via HTTP API.

Also for the consistency I think we should stick to either `has_datasets` and `has_projects` or `count_datasets` and `count_projects`.

shcheklein:
@ilongin what is the progress here?

src/datachain/data_storage/metastore.py
ilongin: We could,  yes, but I would like to do it in a followup as it requires more changes, DB migration etc.

src/datachain/lib/namespaces.py
ilongin: This was just a bad example. Correct one is:
```python
import datachain as dc
dc.delete_project(...)
```

shcheklein: Then we need to update the file?

src/datachain/__init__.py
ilongin: I didn't want to do this here as there is an open issue about this to change it everywhere (and be consistent) https://github.com/iterative/datachain/issues/1302  ..

src/datachain/lib/namespaces.py
ilongin: Yes, of course .. updated

src/datachain/data_storage/metastore.py
ilongin: Yea, I again totally forgot about it .. thanks for the comment. Added Studio PR

src/datachain/data_storage/metastore.py
ilongin: I don't think that check should be added here. It's a higher level business logic check and metastore, IMO should be de-coupled of that .. it should be just simple DB layer for other parts to use and that's it. For example, maybe we will have some management command to actually remove those default namespaces at some point and we want to be able to use metastore.

src/datachain/data_storage/metastore.py
ilongin: Same answer as above.... I personally wouldn't put these kind of things here. It's business logic

src/datachain/data_storage/metastore.py
ilongin: Fixed

src/datachain/lib/namespaces.py
ilongin: Yes, I agree. I didn't put too much effort into optimization as this functions will not be called often and are working on small number of objects, but I refactored it now to use new count method.

src/datachain/data_storage/metastore.py
dreadatour: I believe it not a business logic, but DB level constraints.

Since we don't have foreign keys at the DB level to prevent deletion of relations and if someone will use this `remove_namespace` metastore method to delete namespace, db will be broken — we will have records (datasets) with namespace IDs and no namespace exists in the DB.

Also if we'll want to use `remove_namespace` metastore method from somewhere else, we should implement these same sanity checks in multiple places. It is more practical to move it on a metastore level to do not duplicate the code.

ilongin: Aha, I thought you were referring to removing default / system namespaces and checks for that .. sorry. Anyway, yes, I added additional check for empty namespace / project on removal, that makes sense to me.

src/datachain/__init__.py
dmpetrov: Not sure I follow 🙂

We’re exposing an API that’ll be removed soon anyway. Skipping it would just mean ~5 extra lines of code.

Or am I missing something?

src/datachain/lib/namespaces.py
```diff
@@ -71,3 +74,49 @@ def ls(session: Optional[Session] = None) -> list[Namespace]:
...
+        raise NamespaceDeleteNotAllowedError(
+            f"Namespace {metastore.default_namespace_name} cannot be removed"
+        )
+
+    num_projects = metastore.count_projects(namespace.id)
+    if num_projects > 0:
+        raise NamespaceDeleteNotAllowedError(
+            f"Namespace cannot be removed. It contains {num_projects} project(s). "
+            "Please remove the project(s) first."
+        )
```
dreadatour, line 126: We can now remove this check from here.

src/datachain/lib/projects.py
```diff
@@ -86,3 +86,49 @@ def ls(
...
+        raise ProjectDeleteNotAllowedError(
+            f"Project {metastore.default_project_name} cannot be removed"
+        )
+
+    num_datasets = metastore.count_datasets(project.id)
+    if num_datasets > 0:
+        raise ProjectDeleteNotAllowedError(
+            f"Project cannot be removed. It contains {num_datasets} dataset(s). "
+            "Please remove the dataset(s) first."
+        )
```
dreadatour, line 132: We can now remove this check from here.

src/datachain/__init__.py
ilongin: I wanted to be consistent in all of our APIs and change them all together, but ok, we can do this one in correct way right away. I changed it, please approve if you think this is ok now.

Issue #722, opened 2024-12-18
It hangs for cleaning up tables


dmpetrov:
### Description

It hangs for cleaning up tables for a regular operations. A funny thing - it does not clean up the tables. I have about 60 tables, it "cleans up" for every commands but I'm having the same 60 tables.

```python
>>> ds1.show()
Processed: 1 rows [00:00, 209.64 rows/s]
Generated: 3 rows [00:00, 833.86 rows/s]
Cleanup: 36 tables [00:33,  1.08 tables/s]    # <-- THIS TAKES 2 minutes
    file file file    file file      file                      file     file score
  source path size version etag is_latest             last_modified location
0     s1   p1    0       1   e1         1 1970-01-01 00:00:00+00:00     None     1
1     s2   p2    0       1   e1         1 1970-01-01 00:00:00+00:00     None     2
2     s4   p4    0       1   e1         1 1970-01-01 00:00:00+00:00     None     4
```

### Version Info

```Text

```

shcheklein:
It's not very actionable. @dmpetrov please find a way to reproduce it so that we can address / research this.

PR #1321, opened 2025-09-07
fix(ds query): isolate temp table names

shcheklein:
Updated: might also fix https://github.com/iterative/datachain/issues/722

When we run multiple joins, within the same chain, due to a recursive line:

```python
temp_tables.extend(dq.temp_table_names)
```

in `SQLJoin` ([link](https://github.com/iterative/datachain/blob/91617c050a981d096b06e4d5185ccc12fa30458a/src/datachain/query/dataset.py#L933-L944))

we might end up with a list of 8K+ items, with a lot a lot of duplicates. 

It means query can run very long at the end.

Script to reproduce this. Mind we run show and save at the end, essentially also means we are doubling the list.

```python
from dotenv import load_dotenv

import datachain as dc
from datachain import C, func


load_dotenv("local/.env.test")

all_files = dc.read_storage("gs://datachain-demo", anon=False).mutate(s3_path=dc.C("file.path")).persist()

laion_files = all_files.filter(dc.C("file.path").glob("50k-laion-files/*")).mutate(laion_file=dc.C("file"))
aspset510_files = all_files.filter(dc.C("file.path").glob("aspset510/*")).mutate(aspset510_file=dc.C("file"))
coco2017_files = all_files.filter(dc.C("file.path").glob("coco2017/*")).mutate(coco2017_file=dc.C("file"))
datacomp_small_files = all_files.filter(dc.C("file.path").glob("datacomp-small/*")).mutate(datacomp_small_file=dc.C("file"))
open_images_v6_files = all_files.filter(dc.C("file.path").glob("open-images-v6/*")).mutate(open_images_v6_file=dc.C("file"))
nlp_cnn_stories_files = all_files.filter(dc.C("file.path").glob("nlp-cnn-stories/*")).mutate(nlp_cnn_stories_file=dc.C("file"))

raw_data = (
    laion_files
    .merge(aspset510_files, on="s3_path", inner=True)
    .merge(coco2017_files, on="s3_path", inner=True)
    .merge(datacomp_small_files, on="s3_path", inner=True)
    .merge(open_images_v6_files, on="s3_path", inner=True)
    .merge(nlp_cnn_stories_files, on="s3_path", inner=False)
    .select(
        "s3_path",
        "laion_file",
        "aspset510_file",
        "coco2017_file",
        "datacomp_small_file",
        "open_images_v6_file",
        "nlp_cnn_stories_file",
    )
)

raw_data.save("datachain-demo-merge")
raw_data.show(1000)
```

## TODO:

- [x] Run all tests
- [x] Add description
- [x] Add proper tests
- [x] Confirm semantics one more time

## Summary by Sourcery

Isolate temporary table names per dataset clone and enforce no duplicates to prevent runaway temp table list growth, and fix cleanup to target the cloned query instance.

Bug Fixes:
- Add assertion to ensure temp_table_names has no duplicates before cleanup
- Reset temp_table_names on dataset.clone to avoid carrying over names from previous queries
- Modify exec to clean up temp tables on the cloned query instance rather than the original

### DIFF
--- a/src/datachain/query/dataset.py
+++ b/src/datachain/query/dataset.py
@@ -1438,6 +1438,7 @@ class DatasetQuery:
         # This is needed to always use a new connection with all metastore and warehouse
         # implementations, as errors may close or render unusable the existing
         # connections.
+        assert len(self.temp_table_names) == len(set(self.temp_table_names))
         with self.catalog.metastore.clone(use_new_connection=True) as metastore:
             metastore.cleanup_tables(self.temp_table_names)
         with self.catalog.warehouse.clone(use_new_connection=True) as warehouse:
@@ -1531,6 +1532,7 @@ class DatasetQuery:
         obj.steps = obj.steps.copy()
         if new_table:
             obj.table = self.get_table()
+        obj.temp_table_names = []
         return obj
 
     @detach
@@ -1870,11 +1872,11 @@ class DatasetQuery:
 
     def exec(self) -> "Self":
         """Execute the query."""
+        query = self.clone()
         try:
-            query = self.clone()
             query.apply_steps()
         finally:
-            self.cleanup()
+            query.cleanup()
         return query
 
     def save(


dreadatour:
Fixes for the tests in separate PR: https://github.com/iterative/datachain/pull/1322

shcheklein:
> Fixes for the tests in separate PR: https://github.com/iterative/datachain/pull/1322

yep, thanks @dreadatour  ... I'll keep looking into this PR ... it is probably right for the current approach with temp tables, but I need to understand the whole temp table mechanics a bit better

src/datachain/query/dataset.py
```diff
@@ -1438,6 +1438,7 @@ def cleanup(self) -> None:
         # This is needed to always use a new connection with all metastore and warehouse
         # implementations, as errors may close or render unusable the existing
         # connections.
+        assert len(self.temp_table_names) == len(set(self.temp_table_names))
```
shcheklein, line 1441: C: trying to add this instead of tests (we should not be getting duplicates)

In tests we have additional check that no tables left behind.

shcheklein: cc @dreadatour not sure if it makes to add a specific complicated test ... this should be working better

PR #1323, opened 2025-09-08
Update chain settings docs

dreadatour:
Follow-up for the https://github.com/iterative/datachain/pull/1306

AI was slightly used to improve docstrings text.

**Update**: One important note: I have removed `workers` setting description from the docstrings (only), since it is not supported in Studio for now.

### DIFF
--- a/src/datachain/lib/dc/datachain.py
+++ b/src/datachain/lib/dc/datachain.py
@@ -349,31 +349,27 @@ class DataChain:
         sys: Optional[bool] = None,
     ) -> "Self":
         """
-        Update chain settings.
-
-        This method updates the specified settings without changing not specified ones.
-        It returns the chain itself, allowing method chaining for subsequent operations.
-        To restore all settings to their default values, use `reset_settings()`.
+        Set chain execution parameters. Returns the chain itself, allowing method
+        chaining for subsequent operations. To restore all settings to their default
+        values, use `reset_settings()`.
 
         Parameters:
             cache: Enable files caching to speed up subsequent accesses to the same
-                files from the same or different chains on the same worker node.
-                Defaults to False.
+                files from the same or different chains. Defaults to False.
             prefetch: Enable prefetching of files. This will download files in
                 advance in parallel. If an integer is provided, it specifies the number
-                of files to prefetch concurrently for each worker. Defaults to 2.
-                Set to 0 or False to disable prefetching.
+                of files to prefetch concurrently for each process on each worker.
+                Defaults to 2. Set to 0 or False to disable prefetching.
             parallel: Number of processes to use for processing user-defined functions
                 (UDFs) in parallel. If an integer is provided, it specifies the number
-                of CPUs to use. If True, all available CPUs are used.  Defaults to 1.
+                of CPUs to use. If True, all available CPUs are used. Defaults to 1.
             namespace: Namespace to use for the chain by default.
             project: Project to use for the chain by default.
             min_task_size: Minimum number of rows per worker/process for parallel
                 processing by UDFs. Defaults to 1.
-            batch_size: Maximum number of rows per insert by UDF to fine tune
-                and balance speed and memory usage. This might be useful when
-                processing large rows or when running into memory issues.
-                Defaults to 2000.
+            batch_size: Number of rows per insert by UDF to fine tune and balance speed
+                and memory usage. This might be useful when processing large rows
+                or when running into memory issues. Defaults to 2000.
 
         Example:
             ```py


src/datachain/lib/dc/datachain.py
```diff
@@ -348,24 +348,32 @@ def settings(
...
-            workers: number of distributed workers. Only for Studio mode. (default=1)
-            namespace: namespace name.
-            project: project name.
-            min_task_size: minimum number of tasks. (default=1)
-            batch_size: row limit per insert to balance speed and memory usage.
-                (default=2000)
+            cache: Enable files caching to speed up subsequent accesses to the same
+                files from the same or different chains on the same worker node.
+                Defaults to False.
+            prefetch: Enable prefetching of files. This will download files in
```
shcheklein, line 359: I think we use async, not threads

dreadatour: Updated, please, check 🙏
Also please, notice update in the PR description:

> One important note: I have removed workers setting description from the docstrings (only), since it is not supported in Studio for now.

src/datachain/lib/dc/datachain.py
```diff
@@ -348,24 +348,32 @@ def settings(
         batch_size: Optional[int] = None,
         sys: Optional[bool] = None,
     ) -> "Self":
-        """Change settings for chain.
+        """
+        Update chain settings.
```
shcheklein: let's remove this phrase completely, not informative

src/datachain/lib/dc/datachain.py
```diff
@@ -348,24 +348,32 @@ def settings(
...
 
-        This function changes specified settings without changing not specified ones.
-        It returns chain, so, it can be chained later with next operation.
+        This method updates the specified settings without changing not specified ones.
```
shcheklein: I think `without changing not specified ones.` this is expected and clear by default.

Also, don't start with `This method` ... just "Set chain execution parameters. It returns ...."

src/datachain/lib/dc/datachain.py
```diff
@@ -348,24 +348,32 @@ def settings(
...
-            parallel: number of thread for processors. True is a special value to
-                enable all available CPUs. (default=1)
-            workers: number of distributed workers. Only for Studio mode. (default=1)
-            namespace: namespace name.
-            project: project name.
-            min_task_size: minimum number of tasks. (default=1)
-            batch_size: row limit per insert to balance speed and memory usage.
-                (default=2000)
+            cache: Enable files caching to speed up subsequent accesses to the same
+                files from the same or different chains on the same worker node.
```
shcheklein: way too many tech detail ... e.g. same worker node - how can it be useful to users?

Essentially it means that cache is not useful / broken in a distributed mode, right?

src/datachain/lib/dc/datachain.py
```diff
@@ -348,24 +348,32 @@ def settings(
...
+                Defaults to False.
+            prefetch: Enable prefetching of files. This will download files in
+                advance in parallel. If an integer is provided, it specifies the number
+                of files to prefetch concurrently for each worker. Defaults to 2.
```
shcheklein: per worker of per process on a worker?

src/datachain/lib/dc/datachain.py
```diff
@@ -348,24 +348,32 @@ def settings(
...
+                Set to 0 or False to disable prefetching.
+            parallel: Number of processes to use for processing user-defined functions
+                (UDFs) in parallel. If an integer is provided, it specifies the number
+                of CPUs to use. If True, all available CPUs are used.  Defaults to 1.
```
shcheklein: double space at the end?

src/datachain/lib/dc/datachain.py
```diff
@@ -348,24 +348,32 @@ def settings(
...
+            namespace: Namespace to use for the chain by default.
+            project: Project to use for the chain by default.
+            min_task_size: Minimum number of rows per worker/process for parallel
+                processing by UDFs. Defaults to 1.
+            batch_size: Maximum number of rows per insert by UDF to fine tune
```
shcheklein: Just "Number of rows"

Does it affect now only result inserts and not anything else?

src/datachain/lib/dc/datachain.py
dreadatour: > way too many tech detail ... e.g. same worker node - how can it be useful to users?

Can be useful for Studio users.

> Essentially it means that cache is not useful / broken in a distributed mode, right?

Kind of 👀 Same time it is expected, and IMO should be in docs. Let me remove this phrase.

src/datachain/lib/dc/datachain.py
dreadatour: Oops, my bad 😥

src/datachain/lib/dc/datachain.py
dreadatour: Strictly saying it is maximum number of rows, but you're right, this is small details.

> Does it affect now only result inserts and not anything else?

Only inserts for now as we have discussed.

dreadatour:
@shcheklein thank you for the comments! Updated docstring, please, check 🙏

PR #1326, opened 2025-09-09
fix(delta): read qualified dataset name on no changes in delta

shcheklein:
Fixes the scenario in delta case:

* No changes / second run
* Output dataset name exists in the default namespace (as well as in the target namespace)
* Delta silently reading the dataset from de the default namespace instead of the proper one (in the target namespace)

Quite serious issue since it can lead to usage of some wrong data

## Summary by Sourcery

Ensure delta save with no changes returns the correct dataset version from the intended namespace and project instead of defaulting to the default namespace

Bug Fixes:
- Pass namespace and project parameters to read_dataset when handling no-change delta saves to avoid reading from the default namespace

Enhancements:
- Update dependency retrieval to include project and namespace context and generate properly qualified dataset names

Tests:
- Introduce helper _get_short_ds_name and update existing tests to use fully qualified names
- Add test to verify no-change delta behavior across multiple namespace and project combinations

### DIFF
--- a/tests/func/test_delta.py
+++ b/tests/func/test_delta.py
@@ -11,11 +11,26 @@ from datachain.lib.dc import C
 from datachain.lib.file import File, ImageFile
 
 
+def _get_short_ds_name(catalog, name, project_name, namespace_name) -> str:
+    if project_name == catalog.metastore.default_project.name:
+        return name
+    if namespace_name == catalog.metastore.default_project.namespace.name:
+        return f"{project_name}.{name}"
+    return f"{namespace_name}.{project_name}.{name}"
+
+
 def _get_dependencies(catalog, name, version) -> list[tuple[str, str]]:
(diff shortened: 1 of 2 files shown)


tests/func/test_delta.py
```diff
@@ -11,20 +11,35 @@
 from datachain.lib.file import File, ImageFile
 
 
+def _get_short_ds_name(catalog, name, project_name, namespace_name) -> str:
```
dreadatour, line 14: Why `_get_short_ds_name`? IMO it should be something like `_get_full_ds_name` 🤔

dreadatour (approved):
Looks good to me 👍

tests/func/test_delta.py
shcheklein: it is trying to get the shortest name (check if namespace passed is default and project is default and doesn't use them)

PR #1327, opened 2025-09-09
Implements `DataChain.hash()`

ilongin:
Implements `DataChain.hash()` that is calculating hash for current **definition** of chain (steps).  Two chains with identical steps should have the same hash. Ordering of steps is important. Hash calculation is fast and consistent - it's not making any DB calls or compiling to SQL queries and it always produce the same result for same chain definition. 

Hash for chain is calculated by combining hashes of each step that is added to the chain. Hash for each steps is generally calculated by hashing it's inputs (e.g expressions added to `.filter()`) and adding step name to the calculation.
Hash function is implemented for steps:

- [x] Select and SelectExcept
- [x] Mutate
- [x] Filter
- [x] OrderBy
- [x] Limit
- [x] Offset
- [x] Count
- [x] Distinct
- [x] Union
- [x] Join
- [x] GroupBy
- [x] Subtract
- [x] UDFSignal
- [x] RowGenerator

### Limitations:
Currently hash is inconsistent for chains that starts with `.read_values()`. This is because in order to be correct, we would need to calculate hash of whole input to `.read_values()` and that is not implemented yet. It will be added in followup.

### Important changes

- [Hash SQL `ColumnElement`s](https://github.com/iterative/datachain/pull/1327/files#diff-480637f95224f7895ea91ef4c4055c357320d4c7e8a2ffb3e2361ab258e704aeR20-R101) - this is used everywhere in hash calculation and is the most important function in the PR
- [Hash python function](http://github.com/iterative/datachain/pull/1327/files#diff-480637f95224f7895ea91ef4c4055c357320d4c7e8a2ffb3e2361ab258e704aeR104-R114) - used in UDF to serializize user's custom function
- [Hash of `SignalSchema`](https://github.com/iterative/datachain/pull/1327/files#diff-69a1fc2d77ff106392ad7649ed8cf7ec344dad230af2e4d0ea9c64b816cee15bR262-R265)
- [Hash of UDFAdapter](https://github.com/iterative/datachain/pull/1327/files#diff-bcaff551e415953e55fef366c56ffca72ed3ad969e38acea2e6360f1b412d3bfR159-R173) - this is needed to hash UDFStep. We hash user function, input's and outputs.
- [Main hashing function of `DatasetQuery`](https://github.com/iterative/datachain/pull/1327/files#diff-5cd0b1bd1147a2704c8c333fdb16b0fa21c66e6f5101e4889f7a624d5da4c7cfR1310-R1325) - this is just used by `DataChain` internally and is the core hashing logic. It just goes through steps and appends each step's hash to create final hash
- [Hash implementation of Step subclasses](https://github.com/iterative/datachain/pull/1327/files#diff-5cd0b1bd1147a2704c8c333fdb16b0fa21c66e6f5101e4889f7a624d5da4c7cfR187-R1156) - hash implementation for each of the `DatasetQuery` step classes which are listed above in the description, like `SQLFilter`, `SQLJoin` etc.

## Summary by Sourcery

Introduce a hashing framework by adding hash() methods to the query step interface and DataChain, enabling reproducible chain-level hashes based on step definitions.

New Features:
- Add DataChain.hash() to compute a unique SHA-256 hash for the entire chain definition
- Define hash() as an abstract method on Step and provide stub implementations on all SQL and UDF step classes
- Implement QueryStep.hash() to return the dataset version UUID from the catalog

## Summary by Sourcery

Introduce a hashing framework to compute fast, consistent SHA-256 hashes for DataChain definitions by combining step-level hashes without external dependencies.

New Features:
- Add DataChain.hash() to compute a unique hash of the entire chain definition
- Introduce hash_inputs and hash methods on the Step interface and implement them on all SQL and UDF step classes
- Add SignalSchema.hash() to generate a deterministic schema-level hash
- Create a new hash_utils module with functions to serialize and hash SQL expression elements and callables

Enhancements:
- Replace dynamic diff column name generation with fixed constants for deterministic behavior
- Add ensure_sequence helper in utils for consistent sequence handling
- Seed stable dataset version UUID in test fixtures to enable reproducible hash tests
- Extend UDFAdapter to propagate hash calls to underlying UDF wrappers

Tests:
- Add comprehensive parameterized tests for hashing behavior of each query step and UDF step
- Add unit tests for DataChain.hash(), SignalSchema.hash(), and all hash_utils functions to verify deterministic outputs


src/datachain/diff/__init__.py
```diff
@@ -23,6 +23,11 @@ def get_status_col_name() -> str:
     )
 
 
+STATUS_COL_NAME = "diff_7aeed3aa17ba4d50b8d1c368c76e16a6"
+LEFT_DIFF_COL_NAME = "diff_95f95344064a4b819c8625cd1a5cfc2b"
+RIGHT_DIFF_COL_NAME = "diff_5808838a49b54849aa461d7387376d34"
```
ilongin, line 17: This was needed to make hash function consistent. Before we were generating random names for these temporary columns that are added during diff process and removed before the end but that was causing `.hash()` function of chain having `.diff()` to be inconsistent since column names added with mutate are part of hash calculation. 

Workaround is not to always generate full random column name, but just set some constant that is already random enough not to cause collision with user's columns and since it's constant, hash will be consistent as well.

src/datachain/query/dataset.py
```diff
@@ -1213,6 +1307,23 @@ def __iter__(self):
...
+        """
+        Calculates hash of this class taking into account hash of starting step
+        and hashes of each following steps. Ordering is important.
+        """
+        hasher = hashlib.sha256()
+        if self.starting_step:
+            hasher.update(self.starting_step.hash().encode("utf-8"))
+        else:
+            assert self.list_ds_name
+            hasher.update(self.list_ds_name.encode("utf-8"))
```
ilongin, line 1316: When we start the chain with `read_storage()` we are not creating starting step right away as we may not have correct listing dataset created yet since we need to do lazy listing first, but listing process starts only when applying steps (calling `.save()`). On the other hand we want to be able to calculate hash of the chain before applying steps so in this case we are just using listing dataset name (that will be created in future with listing process if doesn't exist) for hashing starting step.

src/datachain/diff/__init__.py
dreadatour: I am sorry, but I can't resist to put it here 😅
image
Although the question is: how will it works with multiple diff operations, applied to the same chain? 🤔 Are there any cases when it might overlaps?

src/datachain/query/dataset.py
```diff
@@ -939,6 +1008,24 @@ class SQLJoin(Step):
     full: bool
     rname: str
 
+    def hash_inputs(self) -> str:
+        predicates = (
+            self.predicates
+            if isinstance(self.predicates, tuple)
+            else (self.predicates,)
+        )
```
dreadatour: We have `ensure_sequence` imported and used above, just wonder, why not to use it here?

src/datachain/utils.py
```diff
@@ -537,3 +537,9 @@ def getenv_bool(name: str, default: bool = False) -> bool:
     if val is None:
         return default
     return val.lower() in ("1", "true", "yes", "on")
+
+
+def ensure_sequence(x) -> Sequence:
+    if isinstance(x, Sequence) and not isinstance(x, (str, bytes)):
+        return x
+    return [x]
```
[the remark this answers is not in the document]
dreadatour: Sounds reasonable (it is about Sourcery comment for this method).

src/datachain/query/dataset.py
ilongin: Forgot to use it here, nice catch...added it.

src/datachain/diff/__init__.py
ilongin: I just didn't have any other ideas how to fix this but it shouldn't affect our diff logic. Good point about multiple diffs in the single chain - I've added the test for that and it works fine.

Regarding randomization, those randoms were used for column names to not collide with user columns and when you think about it, it's not very different than just setting large "random" uuid. It's the same chance to collision which is ~ 0 .  ... but it does look strange in code, I do agree with that :D

ilongin:
> Looks good to me overall.
> 
> One important note is: IMO we have way less tests for such an important feature, than we should have. I think it might be a good idea to ask AI to suggest more tests cases for these changes.
> 
> Other important question is: are there any cases, when we consider chain as "unhashable", and how do we behave in these cases? I can think of, at least, one example, when it will be almost impossible to calculate hash: using SQL `random()` function inside the chain. There are many more (like using Python `random` inside `from_values`, etc). How do we cover this scenario? Are there other scenarios like this?

I already used AI to add some additional test cases but I agree, we could probably add more. I will ask mister ChatGPT to try harder :D 

Regarding "unhashable" chains - currently I'm not able to find such a case. Note that we don't hash generated SQL query or it's params, just the "structure" of the chain which is basically a Python construct. I've added a test case with `func.rand()` and it works as expected.

src/datachain/diff/__init__.py
dreadatour: > Regarding randomization, those randoms were used for column names to not collide with user columns and when you think about it, it's not very different than just setting large "random" uuid.

Yes, I know 😅 This image is just a perfect fit for this case, I could not resist 🤣

> Good point about multiple diffs in the single chain - I've added the test for that and it works fine.

Awesome! Thank you! 🙏

dreadatour:
> Regarding "unhashable" chains - currently I'm not able to find such a case. Note that we don't hash generated SQL query or it's params, just the "structure" of the chain which is basically a Python construct. I've added a test case with `func.rand()` and it works as expected.

I meant not in terms of "getting a hash of some code which uses `random()` function", but rather in general, on a higher level — will it make sense at all to get hash from chains like this? Because during the next run results could be unpredictable, especially if chain broke in the middle and we want to re-run it.

So my question is more general: should we bring in a new "unhashable" term and consider some chains as "unhashable" for proper evaluation of such chains? If so, this "unhashable" feature (result) should be implemented here, in this PR, I believe.

This is kind of more generic question, and we might want to do it in a follow-up, but IMO we need to think about this case anyway, it is just a good time to start thinking on this.

ilongin:
> I meant not in terms of "getting a hash of some code which uses `random()` function", but rather in general, on a higher level — will it make sense at all to het hash from chains like this? Because during the next run results could be unpredictable, especially if chain broke in the middle and we want to re-run it.
> 
> So my question is more general: should we bring in a new "unhashable" term and consider some chains as "unhashable" for proper evaluation of such chains? If so, this "unhashable" feature (result) should be implemented here, in this PR, I believe.
> 
> This is kind of more generic question, and we might want to do it in a follow-up, but IMO we need to think about this case anyway, it is just a good time to start thinking on this.

Ok, I see your point now. But I still don't see the reason for this "unhashable" chains. Even if someone for example added random value column with mutate and code broke later in some UDF and results with the value of that column generated with random are taken on re-run, I don't see how it could break user's code / logic.. In any case, user's will be aware of how checkpoints work and if that's an issue for them they will have ability to run job from the scratch every time so I wouldn't complicate this for now by introducing "unhashable" chain ... maybe this will change in future though, cannot be sure.

dreadatour:
> Ok, I see your point now. But I still don't see the reason for this "unhashable" chains. Even if someone for example added random value column with mutate and code broke later in some UDF and results with the value of that column generated with random are taken on re-run, I don't see how it could break user's code / logic.. In any case, user's will be aware of how checkpoints work and if that's an issue for them they will have ability to run job from the scratch every time so I wouldn't complicate this for now by introducing "unhashable" chain ... maybe this will change in future though, cannot be sure.

What about `ORDER BY random()`? 🤔 We will have duplicates in such cases.

ilongin:
> What about `ORDER BY random()`? 🤔 We will have duplicates in such cases.

:thinking:  ... Could you explain how duplicates could happen? ... just to explain - if something breaks in UDF (let's say half UDF is done and half is not becasue error happened) we will have input udf table saved and use that to continue with that UDF. So if something is calculated with `order by random()` before UDF that will stay that way - inputs to UDF won't be re-calculated..

dreadatour:
> > What about `ORDER BY random()`? 🤔 We will have duplicates in such cases.
> 
> 🤔 ... Could you explain how duplicates could happen? ... just to explain - if something breaks in UDF (let's say half UDF is done and half is not becasue error happened) we will have input udf table saved and use that to continue with that UDF. So if something is calculated with `order by random()` before UDF that will stay that way - inputs to UDF won't be re-calculated..

I didn't know about "we will have input udf table saved" 🤔 Does this means we will save all the chain steps results in temp tables?

ilongin:
> I didn't know about "we will have input udf table saved" 🤔 Does this means we will save all the chain steps results in temp tables?

Not all the chain steps, but UDF ones since other steps are not that "heavey" and don't have real processing .. alongside checkpoint object we will save (actually just won't delete) input table, partial result table and special new table for generator called "processed inputs table" which is needed to know what rows are processed).
Basically we will put checkpoints in 2 places:

1. At the end of successfully saved dataset from chain - to know that we can completely skip calculation of this chain
2. In the UDFs where we will have 2 types of checkpoints - partial checkpoint created at beginning of UDF which will be active until UDF is completely successful -> then we save non partial (default / normal) checkpoint to know that we can skip completely this UDF in re-run.

dreadatour:
> > I didn't know about "we will have input udf table saved" 🤔 Does this means we will save all the chain steps results in temp tables?
> 
> Not all the chain steps, but UDF ones since other steps are not that "heavey" and don't have real processing .. alongside checkpoint object we will save (actually just won't delete) input table, partial result table and special new table for generator called "processed inputs table" which is needed to know what rows are processed). Basically we will put checkpoints in 2 places:
> 
> 1. At the end of successfully saved dataset from chain - to know that we can completely skip calculation of this chain
> 2. In the UDFs where we will have 2 types of checkpoints - partial checkpoint created at beginning of UDF which will be active until UDF is completely successful -> then we save non partial (default / normal) checkpoint to know that we can skip completely this UDF in re-run.

Not sure if I follow all the logic here but let's get back to this when we will actually implement checkpoints.

ilongin:
> > > I didn't know about "we will have input udf table saved" 🤔 Does this means we will save all the chain steps results in temp tables?
> > 
> > 
> > Not all the chain steps, but UDF ones since other steps are not that "heavey" and don't have real processing .. alongside checkpoint object we will save (actually just won't delete) input table, partial result table and special new table for generator called "processed inputs table" which is needed to know what rows are processed). Basically we will put checkpoints in 2 places:
> > 
> > 1. At the end of successfully saved dataset from chain - to know that we can completely skip calculation of this chain
> > 2. In the UDFs where we will have 2 types of checkpoints - partial checkpoint created at beginning of UDF which will be active until UDF is completely successful -> then we save non partial (default / normal) checkpoint to know that we can skip completely this UDF in re-run.
> 
> Not sure if I follow all the logic here but let's get back to this when we will actually implement checkpoints.

Yes, the logic is little big complex because of that "special" feature we want to give which is those partial checkpoints in UDF.  Without that it would've been much simpler

src/datachain/lib/signal_schema.py
```diff
@@ -257,6 +259,11 @@ def serialize(self) -> dict[str, Any]:
             signals["_custom_types"] = custom_types
         return signals
 
+    def hash(self) -> str:
+        """Create SHA hash of this schema"""
+        json_str = json.dumps(self.serialize(), sort_keys=True, separators=(",", ":"))
```
shcheklein, line 264: how often / when do we call it?

it is a very heavy operation AFAIU ... 

what if implementation becomes unstable (e.g. some model names are different each time, etc) -will we catch it with tests? how can we guarantee this?

src/datachain/hash_utils.py
```diff
@@ -0,0 +1,147 @@
...
+    if not callable(func):
+        raise TypeError("Expected a callable")
+
+    # Determine if it is a lambda
+    is_lambda = func.__name__ == "<lambda>"
+
+    if not is_lambda:
+        # Try to get exact source of named function
+        try:
+            lines, _ = inspect.getsourcelines(func)
```
shcheklein, line 122: why do we extract source code here? do we include it into hash? (this is not what is expected)

tests/unit/lib/test_signal_schema.py
```diff
@@ -1563,3 +1563,27 @@ def test_to_partial_complex_signal_error_invalid_field():
...
+            "26a08b3793e738814f199c89c4582f9bde052ff3dcba84c2020535063df4c36c",
+        ),
+        (
+            {},
+            "44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a",
+        ),
+    ],
+)
+def test_hash(schema, _hash):
+    assert SignalSchema(schema).hash() == _hash
```
shcheklein, line 1589: what do we really test here? why three tests? we should probably do some minor changes to the same schema and test that hash is different, also test some complex scenarios in schema

tests/unit/test_datachain_hash.py
```diff
@@ -0,0 +1,173 @@
...
+        "Hash of the chain started with read_values is currently inconsistent,"
+        " meaning it produces different hash every time. This happens because we"
+        " create random name dataset in the process. Correct solution would be"
+        " to calculate hash of all those input values."
+    )
+    assert dc.read_values(num=[1, 2, 3]).hash() == ""
+
+
+def test_read_storage(mock_get_listing):
+    assert dc.read_storage("s3://bucket").hash() == (
```
shcheklein, line 68: again, what are we actually testing here? what is the purpose of this kind of tests?

src/datachain/hash_utils.py
```diff
@@ -0,0 +1,147 @@
...
+)
+from sqlalchemy.sql.functions import Function
+
+T = TypeVar("T", bound=ColumnElement)
+ColumnLike = Union[str, T]
+
+
+def serialize_column_element(expr: Union[str, ColumnElement]) -> dict:  # noqa: PLR0911
+    """
+    Recursively serialize a SQLAlchemy ColumnElement into a deterministic structure.
```
shcheklein, line 24: can we use str(query) or something like this? seems too complicated atm

src/datachain/lib/signal_schema.py
ilongin: We call it for each UDF in the chain when we calculate the whole chain hash. Chain hash will be calculated:
1. On `.save()`
2. In each UDF itself

IMO we should not worry about this being bottleneck. I will test with some big schema to be 100% sure, but since this is only converting column and their types to string implementation (doesn't do any IO or similar) I would guess this is not even on the same page with UDF calculation, or applying all the steps in chain. 

Regarding instability, can you give example of something that could happen? I'm trying to come up with scenario but don't have much ideas .. `SignalsSchema.hash()` depends on schema serialization which should always be consistent , otherwise we have big issue in our system regardless checkpoints.

shcheklein: I don't tbh ... I remember we have things like random names for generated DataModels in some places ... SignalSchema overall and serialization is quite complex - I think it is worth reviewing it high level

shcheklein: > In each UDF itself

Once? or per each input?

src/datachain/hash_utils.py
ilongin: Yes, we use UDF function source to calculate hash. I know we spoke how user should be able to change UDF function code and continue on that UDF and this doesn't block that. 

This are implementation details / my plans about UDF checkpoints:
1. At beginning of UDF create special `partial` checkpoint of the chain that enters the UDF (without UDF itself). All partial results will be connected to this checkpoint and this will be active until UDF is 100% success
2. When UDF is 100% success / done we create another normal / **non** `partial` checkpoint that now includes UDF itself into calculation (here we use this function code for hash). This is to be able to continue from this UDF if something breaks in the future (avoid UDF re-calculation) and also to be able to recalculate UDF if user decides to change UDF code after even if UDF is successfully done (maybe user realized there is still something he missed in that UDF later on). I think this is important as we need to give user ability to re-run that UDF from beginning, without the need to re-run the whole job for example.

Let me know if you see some issues in the approach.

tests/unit/lib/test_signal_schema.py
ilongin: We test that certain schema always produce consistent hash. We have empty schema, one field schema with `File` and little bit more complex schema with multiple fields. I'm not sure what is wrong with these scenarios? I agree we could add more though.

tests/unit/test_datachain_hash.py
ilongin: This is about testing starting steps as there is little bit of special logic there compared to other steps. This is just simple test, there are more complex tests after. We test consistent hash calculation, as in other places.

src/datachain/hash_utils.py
shcheklein: yep, my main concern is complexity while it doesn't fully solve the issue - anything outside in the code can change that leads to change in the UDF - e.g. it is using some helper that changed? or some package? We won't be able to detect this - so unless I'm missing something I would keep it simple, skip recalculation no matter what and have some mechanism to trigger full recall (or partial on the UDF side - e.g. by introducing some version of the function as a param).

tests/unit/lib/test_signal_schema.py
shcheklein: right, but how do I know that hash is meaningful at all? what it possibly detects that if some logic changes we will be getting some new hash - but I can predict in most cases people will be just updating the value w/o paying attention

so, what is the intention here? to make that it works for some edge cases? 

do we need to test some actual logic (at least in other tests) - that it actually takes certain fields when it calculates the hash?

tests/unit/test_datachain_hash.py
shcheklein: > We test consistent hash calculation, as in other places.

what does this exactly mean? (what is "consistent" in this case?)

src/datachain/hash_utils.py
ilongin: It's not possible because:
1. It depends on a dialect. By default SQLAlchemy uses default dialect, but if table from which we generate expression is bound to engine it will create different string representation for different DBs
2. Bind parameter names are unstable - there is no guarantee that `col == 5` will always produce `"col = :param_1"` .. sometimes it can produce `"col = :param_2"`
3. SQLAlchemy version changes - The stringification format is not a stable API. A minor upgrade could produce different SQL text for the same expression → inconsistent hashes.

shcheklein: > It depends on a dialect. By default SQLAlchemy uses default dialect, but if table from which we generate expression is bound to engine it will create different string representation for different DBs

do we care about this for hash?

>SQLAlchemy version changes - The stringification format is not a stable API. A minor upgrade could produce different SQL text for the same expression → inconsistent hashes.

same - is it important? it seems not tbh ...

> Bind parameter names are unstable - there is no guarantee that col == 5 will always produce "col = :param_1" .. sometimes it can produce "col = :param_2"

what are the cases?

src/datachain/lib/signal_schema.py
ilongin: For each UDF we calculate it twice, for `params` and `outputs` - both are instances of `SignalSchema` in udf codebase.

src/datachain/hash_utils.py
ilongin: True, this hash will only catch function code changes itself, not the helper method code changes or lib changes. But my feeling was that it's good enough for the users (we can clearly communicate this to them) and it's def better than to only give option to recalculate the whole job. ... idk

BTW it's not that hard to add those helper methods changes affect the hash as well, the only problematic part are the external libs (for them we can only include name + type in the hash so hash will change if user stops using one or adds another one etc.)

tests/unit/lib/test_signal_schema.py
ilongin: Intention is to test that:
1. Make sure it works
2. Make sure it doesn't return random values every time which is basically the same as "2 identical schemas produce identical hashes"

I'm not sure what else can we test. Internal calculation of hash is "black box" to the tests.
We do have tests that are calculation hash of the whole meaningful chain that includes UDF with schemas with actual fields. This above is just simple unit test.

tests/unit/test_datachain_hash.py
ilongin: Same chains produce same hashes every time. In this case it's most simple chain possible - just reading storage and saving it;

src/datachain/hash_utils.py
ilongin: I think we should care about all 3 of those items. 

About the unstable param names - I think that param names depend on ordering which should not affect the hash (e,g orders of elements in `and_` clause etc.) .. but in addition, constants are not included in representation so these 2 expressions produce same hashes even though they are clearly different:

```python
>>> from datachain import C
>>> expr = C("age") == 20
>>> str(expr)
'age = :age_1'
>>> expr = C("age") == 30
>>> str(expr)
'age = :age_1'
```

shcheklein: > . but in addition, constants are not included in representation 

can we / don't we take all values separately in the Step? it should be simple I guess?

I don't think we should care about order - again, we are complicating this ahead of time I feel

>I think we should care about all 3 of those items.

could you elaborate please?

ilongin: > can we / don't we take all values separately in the Step? it should be simple I guess?
> 

Inputs to the `.filter()` (and a lot of other `DataChain` methods) is a list of `sqlalchemy.ColumnElement` which can be column, binary expression, unary expression, over, function (all those classes that are handled in questionable `serialize_column_element` function btw). 
So in this example: `chain.filter(C("numbers") > 3, C("numbers") < 10)` we have 2 arguments, both binary expressions. If we would do just `str(e)` for each expression we would not get constant values in that string -> hash is incorrect. In order to get constant values we would need to parse that binary expression down the tree and that is actually exactly what that `serialize_column_element` is doing but it's also returning correct hashable representation (dicts) which are hashed after that in another method.
TL;DR; It's impossible to achieve correct / consistent hash with just using `str()` on inputs.

BTW `str(dc.func.and_(C("numbers") > 3, C("file.path").glob("*.jpg")))` == `and()`.

> > I think we should care about all 3 of those items.
> 
> could you elaborate please?

1. If we don't have to I don't think we should be dialect depended in hash calculation. The more hash is generic, the better. If hash is coupled with dialect then 2 identical chains will produce different hashes locally and on Studio and we always want to avoid this.
2. Explained above why `str(expression)` is not enough - it's not consistent and won't work correctly for hash calculation
3. `str(expr)` is just string representation of query class and is more prone to changes - I don't think it's meant to be used for important parts of the code, let alone hash calculation but more like printing queries in logs etc.

src/datachain/hash_utils.py
shcheklein: > it's def better than to only give option to recalculate the whole job.

100%

>  this hash will only catch function code changes itself, not the helper method code changes or lib changes. 

the worst thing here - is that it will be unpredictable - sometimes it works, sometimes doesn't. I don't think we want that level of magic

let's please simplify this, at least for now

src/datachain/hash_utils.py
shcheklein: 2. Can we use something like `str(expr.compile(..., compile_kwargs={"literal_binds": True}))`?
1 and 3. Still not clear why it should be that stable. It is made to restart a job right away, but not to make it portable. If we need it - we can add it later.

I would say - Let's simplify unless we really have a blocker for a basic requirement.

shcheklein: Thanks for the detailed response!

src/datachain/query/dataset.py
```diff
@@ -876,12 +931,18 @@ def apply_sql_clause(self, query: Select) -> Select:
 class SQLOffset(SQLClause):
     offset: int
 
+    def hash_inputs(self) -> str:
+        return hashlib.sha256(str(self.offset).encode()).hexdigest()
```
shcheklein, line 935: move `hashlib.sha256(str(self.offset).encode()).hexdigest()` into a helper? it will make code cleaner a lot ... since it repeats so many times

tests/unit/lib/test_signal_schema.py
shcheklein: 🤔 

> Internal calculation of hash is "black box" to the tests.

100%

A more meaningful test is probably to see that some changes make hash different or some other changes (e.g. some order of parameters if we don't depend on them in some places) don't make it different

>Make sure it doesn't return random values every time which is basically the same as "2 identical schemas produce identical hashes"

btw - we don't really test it here - there should be two calls at least then ...

ilongin: I will add those tests. I thought that are not needed since we always expect specific hash to be asserted (meaning on every run must be consistent for same structure) but you might be right that it's better to explicitly add it

src/datachain/hash_utils.py
ilongin: My very first thought was to use compile with `literal_binds` but I quickly decided not to. Beside the fact that current implementation is also very straightfoward and not complex there are multiple reasons not to go with compile, each of those are deal breaker IMO:
1. Already mentioned consistency with compilers - maybe it's not direct requirement now but it's still nice to have consistent hashing across every backend and what is important now is that then we wouldn't be able to write tests that are expecting specific hash value as Studio (PG) and local (SQLite) tests would end up with different value which means we won't be able to test regression - this means I would also need to refactor current tests.
2. Alias naming - SQLAlchemy automatically adds aliases sometimes to compiled statements and those are really not stable, not sure what they depend on but I've read it can be due to compiler internal state. Also there might be issue with whitespaces, parenthesis, changing ordering of elements in AND  / OR  functions etc. 
3. I've read in the past that those `literal_binds` are not completely stable - they do not work for some cases .. though this would need further investigation .. in general that method doesn't guarantee stable results and that is the issue for hashing.
4. It's slower 30-40% than just python object traversal (this might not be that important though)

PR #1328, opened 2025-09-10
Add documentation for webhooks feature

amritghimire:
This adds the documentation page on how to test, implement and verify
the webhooks in Studio

### DIFF
--- a/docs/studio/webhooks.md
+++ b/docs/studio/webhooks.md
@@ -1,0 +1,233 @@
+# Webhooks in Studio
+
+# About webhooks
+
+Webhooks provide a way for the notifications to be delivered to an external web server whenever certain events occur in [Studio](https://studio.datachain.ai).
+
+Webhooks are used to receive data as it happens as opposed to polling an API to see if data is available. With webhooks, you can set a setting once that you want to hear about certain events or activities when you create the webhook.
+
+When you create a webhook, you specify a URL, and necessary information you want us to send to the URL along with the events that you want to listen on Datachain. When the event your webhook is subscribed to occurs, Datachain Studio will send an HTTP request with the data about the event to the URL that you specified. If your server is setup to listen for webhook deliveries at that URL, it can take action when it receives one.
+
+For example, you can subscribe your webhook to events that occur when a job is created, is complete, is failed, is running, and so on. You can then monitor whenever a job is failed through this webhook.
+
+### Alternative
+
+As opposed to webhooks, you can also use CLI command to get the job information
+
+or some of our available API endpoints but webhook requires less effort than polling an API since it allows near real time updates.
+
+# Available webhooks
+
+As of now, the webhooks available with us falls under following categories.
+
+## JOB
+
+Whenever any job is created or any status is changed to the job, you will receive the JOB webhook event. The payload you get with the job webhook is as:
+
+```jsx
+{
+        "event": "job_status",
+        "job": {
+            "id": "da59df47-d121-4eb6-aa76-dc452755544e",
+            "status": "COMPLETE",
+            "error_message": "",
+            "created_at": "2021-07-27T16:02:08.070557",
+            "updated_at": "2021-07-27T16:22:08.070557",
+            "finished_at": "2021-07-27T16:22:08.070557"
+            "url": "https://studio.datachain.ai/team/TeamName/datasets/jobs/da59df47-d121-4eb6-aa76-dc452755544e"
+        },
+        "timestamp": "2021-07-27T16:22:08.070557",
+    }
+```
+
+# Creating webhooks
+
+You can create a webhook to subscribe to events that occur in a specific team. You should have admin access to that team to create the webhooks in that repository. To create a webhook, go to settings for the team and under the section Webhook, click on Add Webhook. Enter the necessary information to create the webhooks.
+
+- **URL:**  Enter the valid URL where you’d like to receive the webhook payload in
+- **Secret:** A string to use as a secret key. You should choose a random string of text with high entropy. You can use the webhook secret to limit incoming requests to those only originating from Datachain Studio.
+- **Events:** Under events, select the events you would like to trigger the webhook.
+    - **JOB:**
+        - CREATED: When a job is created but not yet scheduled to run
+        - SCHEDULED: Job has been scheduled to run
+        - QUEUED: Job is in the queue waiting to be processed
+        - INIT: Job is initializing (starting up)
+        - RUNNING: When a job starts running
+        - COMPLETE: Job has completed successfully
+        - FAILED: Job failed with error
+        - CANCELED: Job has been cancelled successfully
+        - CANCELING: Job has been scheduled to cancel
+        - TASK: A scheduled task is created.
+
+- SSL Verification: By default, we verify SSL certificates when delivering payloads. SSL verification helps ensure that hook payloads are delivered to your URL endpoint securely, keeping your data away from prying eyes. Disabling this option is **not recommended**..
+- HTTP HEADERS: Some custom extra headers you want to include in the request
+- HTTP Method: By default, we make a post request, but you can specify other http method if necessary.
+- Content Type: Optionally, select the data format you want to receive the webhook payload in
+    - **application/json** will deliver the JSON payload directly as the body of the `POST` request.
+    - **application/x-www-form-urlencoded** will send the JSON payload as a form parameter called `payload`.
+
+# Handling webhook deliveries
+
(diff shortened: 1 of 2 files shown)


docs/studio/webhooks.md
```diff
@@ -0,0 +1,233 @@
...
+- Studio uses HMAC hex digest to compute the hash
+- The hash signature always starts with `sha256=`
+- The hash signature is generated using webhook’s secret token and payload contents.
+- Never use a plain `==` operator. Instead consider using a method like [`secure_compare`](https://www.rubydoc.info/gems/rack/Rack%2FUtils:secure_compare) or [`crypto.timingSafeEqual`](https://nodejs.org/api/crypto.html#cryptotimingsafeequala-b), which performs a "constant time" string comparison to help mitigate certain timing attacks against regular equality operators, or regular loops in JIT-optimized languages.
+
+Updating the example above:
+
+```python
+import hashlib
+import hmac
```
amritghimire, line 216:
```suggestion
import hmac
from flask import abort
```

shcheklein (commented):
excellent! thanks Amrit. Let's also include screenshot also mention how to find them in Studio

docs/studio/webhooks.md
```diff
@@ -0,0 +1,234 @@
...
+else:
+    print("Warning: No signature header found")
+```
+
+# Best practices for using Webhooks
+
+1. You should only subscribe to the webhook events that you need. This will reduce the amount of work your server needs to do.
+2. The webhook secret should be a random string of text with high entropy. You should securely store your webhook secret in a way that your server can access.
+3. You should ensure that your server uses an HTTPS connection. By default, Studio will verify SSL certificates when delivering webhooks. Studio recommends that you leave SSL verification enabled.
+4. You can set up an IP allow list for your server, and add the IP addresses that Studio uses for webhook deliveries. This can block spoofed requests to your server.
```
shcheklein: how can it be done?

docs/studio/webhooks.md
```diff
@@ -0,0 +1,234 @@
...
+5. Your server should respond with a 2XX response within 10 seconds of receiving a webhook delivery. If your server takes longer than that to respond, then Studio terminates the connection and considers the delivery a failure.
```
shcheklein: is 10 second is normal? should we make the timeout smaller?

docs/studio/webhooks.md
```diff
@@ -0,0 +1,234 @@
...
+6. Check the event header and event type before processing the event.
```
shcheklein: Why exactly?

docs/studio/webhooks.md
```diff
@@ -0,0 +1,234 @@
...
+    """
+    if not signature_header:
+        abort(403, "X-datachain-signature-256 is missing!")
+    hash_object = hmac.new(secret_token.encode('utf-8'), msg=payload_body, digestmod=hashlib.sha256)
+    expected_signature = "sha256=" + hash_object.hexdigest()
+    if not hmac.compare_digest(expected_signature, signature_header):
+        abort(403, "Request signatures didn't match!")
+```
+
+Add the following call in between the api receiver.
```
shcheklein: "in the API receiver"

docs/studio/webhooks.md
```diff
@@ -0,0 +1,234 @@
+# Webhooks in Studio
+
+## About webhooks
+
+Webhooks provide a way for the notifications to be delivered to an external web server whenever certain events occur in [Studio](https://studio.datachain.ai).
```
shcheklein: minor: read these paragraphs - seems quite a lot of duplication and can be removed / combined.

docs/studio/webhooks.md
```diff
@@ -0,0 +1,234 @@
...
+
+Webhooks are used to receive data as it happens as opposed to polling an API to see if data is available. With webhooks, you can set a setting once that you want to hear about certain events or activities when you create the webhook.
+
+When you create a webhook, you specify a URL, and necessary information you want us to send to the URL along with the events that you want to listen on Datachain. When the event your webhook is subscribed to occurs, Datachain Studio will send an HTTP request with the data about the event to the URL that you specified. If your server is setup to listen for webhook deliveries at that URL, it can take action when it receives one.
+
+For example, you can subscribe your webhook to events that occur when a job is created, is complete, is failed, is running, and so on. You can then monitor whenever a job is failed through this webhook.
+
+### Alternative
+
+As opposed to webhooks, you can also use CLI command to get the job information
```
shcheklein: link here

docs/studio/webhooks.md
```diff
@@ -0,0 +1,234 @@
...
+
+or some of our available API endpoints but webhook requires less effort than polling an API since it allows near real time updates.
```
shcheklein: combine lines or is it a new paragraph?

docs/studio/webhooks.md
```diff
@@ -0,0 +1,234 @@
...
+
+# Available webhooks
+
+As of now, the webhooks available with us fall under the following categories.
+
+## JOB
```
shcheklein: since it is a single category, just combine it with above. Name it Webhook types or something

docs/studio/webhooks.md
amritghimire: This is basically for the infrastructure part for those implementing it, but they can check the ip address the request comes from and use this. I am assuming we have a constant IP if not, we can remove this from best practice.

docs/studio/webhooks.md
amritghimire: Yes, it is similar to github. Webhook is supposed to pass the data to celery or background task and return as fast as possible.

docs/studio/webhooks.md
shcheklein: if we have IPs (confirm with @0x2b3bfa0 ) we need a link here, otherwise we should remove it

docs/studio/webhooks.md
shcheklein: yeah, I understand ... 10 seconds seems like a lot though (put stress on our servers)... anyways, that's more or less fine

docs/studio/webhooks.md
```diff
@@ -0,0 +1,234 @@
...
+
+Whenever any job is created or any status is changed to the job, you will receive the JOB webhook event. The payload you get with the job webhook is as:
```
0x2b3bfa0, line 19: Nitpick: why use JOB in uppercase, when it's not even the name of the event in the provided example?

docs/studio/webhooks.md
0x2b3bfa0: > _I am assuming we have a constant IP_

Our SaaS IP addresses aren't "constant enough" to recommend users to do this; let's avoid shooting ourselves in the foot prematurely.

HMAC should be enough / preferred for payload validation.

<sub>(Iwould have recommended asymmetric signatures if it wasn't too late) 🤐 </sub>

docs/studio/webhooks.md
amritghimire: So, server has idea which type of event and action has occurred.

docs/studio/webhooks.md
shcheklein: was it addressed?

amritghimire: Please check the updated one. Updated it to add missing PING event as well so, keeping this makes sense now

shcheklein:
The right side menu is wrong, all the main content in under Alternative now as you scroll.

Screenshot 2025-09-11 at 10 29 09 AM

docs/studio/webhooks.md
```diff
@@ -0,0 +1,250 @@
...
+For example, you can subscribe your webhook to events that occur when a job is created, is complete, is failed, is running, and so on. You can then monitor whenever a job is failed through this webhook.
+
+### Alternative
+
+As opposed to webhooks, you can also use [CLI command](../commands/index.md) to get the job information or some of our available [API endpoints](api/index.md) but webhook requires less effort than polling an API since it allows near real time updates.
+
+# Available event type
+As of now, your server can receive two different types of events.
+
+## PING
```
shcheklein: Let's move it after JOB (this is secondary / test item, right?)

docs/studio/webhooks.md
```diff
@@ -0,0 +1,250 @@
...
+# This initializes a new Flask application.
+app = Flask(__name__)
+
+# This defines a POST route at the `/webhook` path. This path matches the path that you specified for the smee.io forwarding.
+#
+# Once you deploy your code to a server and update your webhook URL, you should change this to match the path portion of the URL for your webhook.
+@app.route('/webhook', methods=['POST'])
+def webhook():
+    # Respond to indicate that the delivery was successfully received.
+    # Your server should respond with a 2XX response within 10 seconds of receiving a webhook delivery.
```
shcheklein: Comments are very long and we have scroll because of this

amritghimire:
> Please check some comments (especially right side bar one bothers me), open the draft and read / look at it one more time. Otherwise LGTM

Thanks, updated it.

PR #1343, opened 2025-09-15
Fixing delete namespace

ilongin:
Small fix for session arg

## Summary by Sourcery

Bug Fixes:
- Add default None to the session argument in delete_namespace to match other APIs

### DIFF
--- a/src/datachain/lib/namespaces.py
+++ b/src/datachain/lib/namespaces.py
@@ -77,7 +77,7 @@ def ls(session: Optional[Session] = None) -> list[Namespace]:
     return Session.get(session).catalog.metastore.list_namespaces()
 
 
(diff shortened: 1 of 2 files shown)


ilongin:
> Do we need a test for this?

I don't think so. We don't have it in other places where we accept `session`.

shcheklein:
We didn't have a failure in other places though ... at least yet

ilongin:
> We didn't have a failure in other places though ... at least yet

This was overlooked by me mainly because in tests we always set `session` (that's how all tests are working). Anyway I've added a simple test.

shcheklein (approved):

shcheklein:
I know the reason, yep. (I was adding a few of those that take session). This failure just means that we need to learn and improve a bit from now on. In other places too.

PR #1346, opened 2025-09-16
Added projects base query

ilongin:
Added projects base query to reuse codebase and to make it extendable for custom filters needed for Studio (e.g filter by team id)

### DIFF
--- a/src/datachain/data_storage/metastore.py
+++ b/src/datachain/data_storage/metastore.py
@@ -689,9 +689,6 @@ class AbstractDBMetastore(AbstractMetastore):
             return self._projects.select()
         return select(*columns)
 
-    def _projects_update(self) -> "Update":
-        return self._projects.update()
-
     def _projects_delete(self) -> "Delete":
         return self._projects.delete()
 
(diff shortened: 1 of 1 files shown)


src/datachain/data_storage/metastore.py
```diff
@@ -918,13 +915,8 @@ def list_projects(
...
 
-        query = self._projects_select(
-            *(getattr(n.c, f) for f in self._namespaces_fields),
-            *(getattr(p.c, f) for f in self._projects_fields),
-        )
-        query = query.select_from(n.join(p, n.c.id == p.c.namespace_id))
+        query = self._projects_base_query()
 
         if namespace_id:
             query = query.where(n.c.id == namespace_id)
```
dreadatour: Since we are listing projects would it be more correct to filter by project namespace ID, rather than to filter by joined namespaces table? 🤔 Of course, it should be the same result, just more semantically correct, IMO.

dreadatour (approved):
Looks good to me. Do we have tests for all these methods?

src/datachain/data_storage/metastore.py
ilongin: I personally don't see the difference between those two, but I've changed it as you suggested so it's now using `namespace_id` from projects table to filter.

dreadatour: Yes, no big difference, it is only about semantics. Not critical.
Thank you! 🙏❤️

shcheklein:
We need tests + review the automated review (resolve or fix)

PR #1349, opened 2025-09-18
use pydantic instead of pickle for object configs

shcheklein:
To enforce security we need to migrate from pickle to more structured approach on how we pass metastore, warehouse and other params from one process to another.

The current implementation raises flags across different systems.

## Summary by Sourcery

Migrate metastore and warehouse configuration from pickle-based payloads to structured Pydantic-powered JSON schemas with backward compatibility for legacy serialized data.

New Features:
- Add Pydantic config models (SQLiteDatabaseEngineConfig, SQLiteMetastoreConfig, SQLiteWarehouseConfig) for JSON-based serialization and instance construction.

Enhancements:
- Update get_metastore and get_warehouse to parse JSON config payloads, fallback to legacy base64 pickle, and build instances via new config models.

Tests:
- Refactor unit and end-to-end tests to use JSON serialization through config.from_instance(), adjust deserialization assertions, and clean up DATACHAIN__ROOT environment variable.

### DIFF
--- a/src/datachain/catalog/loader.py
+++ b/src/datachain/catalog/loader.py
@@ -44,6 +44,7 @@ def get_metastore(in_memory: bool = False) -> "AbstractMetastore":
                     APIMetastoreConfig,
                     PostgreSQLMetastoreConfig,
                 )
+
                 cfg_map = {
                     "api_metastore": APIMetastoreConfig,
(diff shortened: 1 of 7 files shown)


shcheklein:
Closing this, too complicated, should be a simpler way to make it safer

PR #1352, opened 2025-09-20
Fixing and refactoring `parse_dataset_uri()`

ilongin:
Fixing function to work with namespace names that start with `@v` which is the same as version separator - this is where the bug was.
Also refactoring function to return namespace name and project name as well. Using regex instead of complicated parsing logic which also simplifies. 
Added missing tests for `parse_dataset_uri()` as well.

## Summary by Sourcery

Refactor dataset URI parsing to extract fully qualified identifiers using regex, fix namespace-vs-version ambiguity, update dependent code paths, and enhance test coverage.

New Features:
- parse_dataset_uri now returns namespace, project, name, and optional version from ds:// URIs

Bug Fixes:
- correctly handle namespaces starting with "@v" without misinterpreting them as version suffix

Enhancements:
- use regex-based parsing and remove redundant parse_dataset_name calls in catalog processing
- update catalog instantiation and node conversion to unpack full URI components

Tests:
- add unit tests for valid and invalid parse_dataset_uri scenarios
- update functional pull_dataset test to expect full semver format

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,13 +24,13 @@ repos:
       - id: trailing-whitespace
         exclude: '^LICENSES/'
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: 'v0.15.4'
+    rev: 'v0.15.5'
     hooks:
       - id: ruff
         args: [--fix, --exit-non-zero-on-fix]
       - id: ruff-format
   - repo: https://github.com/codespell-project/codespell
-    rev: v2.4.1
+    rev: v2.4.2
     hooks:
       - id: codespell
         additional_dependencies: ["tomli"]
--- a/docs/index.md
+++ b/docs/index.md
@@ -42,7 +42,7 @@ database for easy and efficient querying.
     including LLMs.
 2.  **Analytics.** DataChain dataset is a table that combines all the
     information about data objects in one place + it provides
-    dataframe-like API and vectorized engine to do analytics on these
+    DataFrame-like API and vectorized engine to do analytics on these
     tables at scale.
 3.  **Versioning.** DataChain doesn't store, require moving or copying
     data. Perfect use case is a bucket with thousands or millions of
--- a/src/datachain/cache.py
+++ b/src/datachain/cache.py
@@ -84,7 +84,7 @@ class Cache:  # noqa: PLW1641
         size = file.size
         if size < 0:
             size = await client.get_size(from_path, version_id=file.version)
-        from tqdm.auto import tqdm
+        from datachain.progress import tqdm
 
         cb = callback or TqdmCallback(
             tqdm_kwargs={"desc": odb_fs.name(from_path), "bytes": True, "leave": False},
--- a/src/datachain/catalog/catalog.py
+++ b/src/datachain/catalog/catalog.py
@@ -22,7 +22,6 @@ else:
 
 import sqlalchemy as sa
 from sqlalchemy import Column
-from tqdm.auto import tqdm
 
 from datachain.cache import Cache
 from datachain.checkpoint import Checkpoint, CheckpointStatus
@@ -53,6 +52,7 @@ from datachain.error import (
 from datachain.lib.listing import get_listing
 from datachain.node import DirType, Node, NodeWithPath
 from datachain.nodes_thread_pool import NodesThreadPool
+from datachain.progress import tqdm
 from datachain.project import Project
 from datachain.sql.types import DateTime, SQLType
 from datachain.utils import DataChainDir, DatasetIdentifier, interprocess_file_lock
--- a/src/datachain/client/azure.py
+++ b/src/datachain/client/azure.py
@@ -2,9 +2,9 @@ from typing import Any
 from urllib.parse import parse_qs, urlsplit, urlunsplit
 
 from adlfs import AzureBlobFileSystem
-from tqdm.auto import tqdm
 
 from datachain.lib.file import File
+from datachain.progress import tqdm
 
 from .fsspec import DELIMITER, Client, ResultQueue
 
--- a/src/datachain/client/fsspec.py
+++ b/src/datachain/client/fsspec.py
@@ -16,12 +16,12 @@ from urllib.parse import urlparse
 from dvc_objects.fs.system import reflink
 from fsspec.asyn import get_loop, sync
 from fsspec.callbacks import DEFAULT_CALLBACK, Callback
-from tqdm.auto import tqdm
 
 from datachain.cache import Cache
 from datachain.client.fileslice import FileWrapper
 from datachain.nodes_fetcher import NodesFetcher
 from datachain.nodes_thread_pool import NodeChunk
+from datachain.progress import tqdm
 
 if TYPE_CHECKING:
     from fsspec.spec import AbstractFileSystem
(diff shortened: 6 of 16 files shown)


src/datachain/dataset.py
```diff
@@ -43,25 +43,58 @@
...
-    Parse dataser uri to extract name and version out of it (if version is defined)
-    Example:
-        Input: ds://zalando@v3.0.1
-        Output: (zalando, 3.0.1)
+    Parse a dataset URI of the form:
+
+        ds://<namespace>.<project>.<name>[@v<semver>]
+
+    Components:
+    - `ds://`        : required prefix identifying dataset URIs.
```
shcheklein: can we drop the ds:// by now? do you remember why it was needed at all?

src/datachain/dataset.py
```diff
@@ -43,25 +43,58 @@
...
+        r"""
+        ^(?P<namespace>@?\w+)      # namespace, may start with '@'
+        \. (?P<project>\w+)        # project
+        \. (?P<name>\w+)           # dataset name
+        (?:@v                      # optional version prefix must be '@v'
+            (?P<version>\d+\.\d+\.\d+)
+        )?$                        # end of string
+        """,
+        re.VERBOSE,
+    )
```
dreadatour: Instead of using regexp, here we can use `parse_dataset_name` first, to get optional `namespace_name`, optional `project_name`, and `dataset_name` (including version) and then parse `dataset_name` to split it into name and version 🤔

This way we will not split the logic of parsing full dataset name (including namespace and project) between two different methods (`parse_dataset_uri` and `parse_dataset_name`), and will keep it in one place, so next time we will need to change anything related, we will not forget to do so and it will be only one place where changes are required.

src/datachain/dataset.py
```diff
@@ -43,25 +43,58 @@
...
+
+    if not uri.startswith("ds://"):
+        raise ValueError(f"Invalid dataset URI: {uri}")
+
+    body = uri[len("ds://") :]
+
+    pattern = re.compile(
+        r"""
+        ^(?P<namespace>@?\w+)      # namespace, may start with '@'
+        \. (?P<project>\w+)        # project
```
dreadatour: I bet we do allow `@` in project name too. Let's check? 🙏

src/datachain/dataset.py
```diff
@@ -43,25 +43,58 @@
...
+        \. (?P<name>\w+)           # dataset name
+        (?:@v                      # optional version prefix must be '@v'
+            (?P<version>\d+\.\d+\.\d+)
```
dreadatour: If I am not mistaken, we still do support single-digit versions (not semver) 🤔

src/datachain/dataset.py
```diff
@@ -43,25 +43,58 @@
...
-def parse_dataset_uri(uri: str) -> tuple[str, Optional[str]]:
+def parse_dataset_uri(uri: str) -> tuple[str, str, str, str | None]:
     """
-    Parse dataser uri to extract name and version out of it (if version is defined)
-    Example:
-        Input: ds://zalando@v3.0.1
-        Output: (zalando, 3.0.1)
+    Parse a dataset URI of the form:
+
+        ds://<namespace>.<project>.<name>[@v<semver>]
```
dreadatour: Do we always require both namespace and project now? It looks like breaking changes 🤔

src/datachain/dataset.py
dreadatour: Yes, we do: https://github.com/iterative/datachain/blob/main/src/datachain/project.py#L10

src/datachain/dataset.py
```diff
@@ -43,25 +43,58 @@
...
+    if not uri.startswith("ds://"):
+        raise ValueError(f"Invalid dataset URI: {uri}")
+
+    body = uri[len("ds://") :]
+
+    pattern = re.compile(
+        r"""
+        ^(?P<namespace>@?\w+)      # namespace, may start with '@'
+        \. (?P<project>\w+)        # project
+        \. (?P<name>\w+)           # dataset name
```
dreadatour: We also allow `@` in dataset name: https://github.com/iterative/datachain/blob/main/src/datachain/dataset.py#L35

dreadatour (changes requested):
I would prefer to do not use regexp here, but rather use `parse_dataset_name` to get namespace and project name (as described [here](https://github.com/iterative/datachain/pull/1352/files#r2366276966)).

It also looks like it is not only basic fix and refactoring, but also this PR introduce breaking changes: namespace and project are required now. Or am I wrong? If this is ok, then we should have more clear message about these changes in PR description, what do you think?

I would also love to see more cases in added tests, may be ask AI to help with that?

src/datachain/dataset.py
ilongin: `@` is not allowed in project name and dataset name. Also, you cannot create namespace name with it, but we do create automatically namespaces with `@<username>` so it can appear here. Regardless, I think I think you are right to assume it can be present, as it's safer that way.

src/datachain/dataset.py
ilongin: We only allow integer versions in `dc.read_dataset()` to be backward compatible with old user's scripts.  In that method we convert integer version to correct string semver and continue with that.
`parse_dataset_uri()` and other internal methods  expect full valid semver.

src/datachain/dataset.py
ilongin: This method was used in places where it was expected to have both, but you are right to make this more generic and leave it as optional.

src/datachain/dataset.py
ilongin: Refactored to set namespace and project as optional and to use `parse_dataset_name()`.

src/datachain/dataset.py
ilongin: That goes way back ... It was part of the requirements but I don't remember all the reasoning behind it. We can discuss separately as I don't want to remove anything like that in this PR

src/datachain/dataset.py
ilongin: We don't, those are reserved characters that are not allowed

src/datachain/dataset.py
```diff
@@ -43,58 +43,34 @@
...
-        (?:@v                      # optional version prefix must be '@v'
-            (?P<version>\d+\.\d+\.\d+)
-        )?$                        # end of string
-        """,
-        re.VERBOSE,
-    )
-
-    match = pattern.match(body)
+    # Split off optional @v<version>
+    match = re.match(r"^(?P<name>.+?)(?:@v(?P<version>\d+\.\d+\.\d+))?$", body)
```
dreadatour, line 61: We still have quite a lot of cases where this regexp will not works (and we have no tests for all these cases), I would prefer to move parsing dataset name, version, namespace, project, etc to separate module and cover it all with tests, so we can reuse it all over the code, but same time it looks like a separate task to refactor this part and it is good enough for now for me.

tests/unit/test_dataset.py
```diff
@@ -186,29 +186,19 @@ def test_parse_dataset_name_empty_name():
...
-        "ds://result@v1.0.2",
-        "ds://@user@v1.0.5",
-        "ds://@user.dev.result@1.0.5",
-    ],
-)
-def test_parse_dataset_uri_invalid_format(uri):
-    with pytest.raises(ValueError) as excinfo:
-        parse_dataset_uri(uri)
-
-    assert str(excinfo.value) == f"Invalid dataset URI format: {uri}"
```
dreadatour: Not sure why did we remove these tests to catch all the failures and left only "happy path" tests :(

shcheklein:
@ilongin can you wrap it all up?

Issue #1361, opened 2025-09-24
Add ability to utilize checkpoints from local run


ilongin:
User should be able to utilize new checkpoints feature with basic `datachain query` command (or similar).

If we go with `datachain query` here is the implementation plan:
Currently this command creates `Job` instance which has name as the name of the script (without path).
We should change this behavior into:
1. Use full path of the script as `Job` name
2. Try to find `Job` with existing name and without child (this means script is re-run) and if found add it as parent to newly created `Job`
3. Add new flag called `--reset` (maybe there is a better name) that will not use existing checkpoints, but will run the script as it is ran for the first time

Example usage
```python
datachain query run.py   # some error happens
datachain query run.py   # continues where it stopped on the last run using the checkpoints from the last run if they exist
datachain query run.py --reset  # runs the script from the scratch, not utilizing created checkpoints
```

shcheklein:
People don't use query AFAIU, it is not even released at the moment.

Can we somehow do this from a Python itself?

Also we have `datachain job run`

ilongin:
`datachain job run` seems to be only for running jobs in Studio for now .. it has things like `team_id` which are Studio specific.

Regarding running from Python itself the issue is to create job if doesn't exist at beginning and set `Job` as complete / failed on the end of the script. There are some options to solve this (e.g `atexit` hooks), I will try it out.

btw  @dmpetrov mentioned somewhere that he wants  `datachain run my_script.py` (equivalent to `datachain query my_script`) so that's another reason I wanted to bring back that command into the life.

dmpetrov:
We do need two separate actions: run and run-from-scratch.

Python script run should be the run (with checkpoints). We need to introduce run-from-scratch.

Thus can be done as a command: dc run and dc run --reset. Or just a python script with env variable like: DC_NO_CHECKPOINT=True python myscrip.py

I'd introduce both. Python and dc run seems cleaner. Env variable a bit hacky but does not require dc command.

shcheklein:
Let's please avoid first introducing CLI back. Let's focus first on Python alone. Tbh, I don't like the whole datachain run for many many reasons:

- it is not public yet and will need work to make it public and it takes already long time
- it is ugly and bad for local experience - you really just expect Python to work
- it overlaps and will confuse people with datachain job run (which one to use when? why?)

etc , etc

dmpetrov:
do you mean `python myscript.py` and `DC_NO_CHECKPOINT=True python myscript.py` for the two actions?

[...]

PR #1362, opened 2025-09-24
Skip processing chain if checkpoint exists

ilongin:
This PR adds a logic in `DataChain.save()` to create checkpoints where needed and to skip processing chain steps if checkpoint for this chain already exists.

## Summary by Sourcery

Add job-level checkpointing to DataChain.save to reuse prior results and skip redundant processing, introduce parent_job_id tracking, and expand tests to cover checkpoint reset and chain modification scenarios

New Features:
- Skip chain processing in DataChain.save when a matching checkpoint exists for the current job
- Introduce checkpoint creation for each job and support resetting via DATACHAIN_CHECKPOINTS_RESET
- Track parent-child job relationships through a new parent_job_id field in the metastore

Bug Fixes:
- Raise JobNotFoundError when DATACHAIN_JOB_ID does not correspond to an existing job

Enhancements:
- Add _calculate_job_hash method to compute chain hashes by combining previous checkpoint hashes and current chain hash

Tests:
- Add comprehensive unit tests to validate checkpoint behavior across first, subsequent, and multiple runs with and without resets
- Update datachain_job_id fixture to initialize jobs via metastore.create_job

### DIFF
--- a/src/datachain/lib/dc/datachain.py
+++ b/src/datachain/lib/dc/datachain.py
@@ -587,11 +587,7 @@ class DataChain:
         Hash is calculated using previous job checkpoint hash (if exists) and
         adding hash of this chain to produce new hash.
         """
-        last_checkpoint = max(
-            self.session.catalog.metastore.list_checkpoints(job_id),
-            key=lambda obj: obj.created_at,
-            default=None,
-        )
+        last_checkpoint = self.session.catalog.metastore.get_last_checkpoint(job_id)
 
         return hashlib.sha256(
             (bytes.fromhex(last_checkpoint.hash) if last_checkpoint else b"")
--- a/src/datachain/lib/dc/datachain.py
+++ b/src/datachain/lib/dc/datachain.py
@@ -700,7 +700,7 @@ class DataChain:
         metastore = self.session.catalog.metastore
 
         job_id = os.getenv("DATACHAIN_JOB_ID")
-        checkpoints_reset = env2bool("DATACHAIN_CHECKPOINTS_RESET")
+        checkpoints_reset = env2bool("DATACHAIN_CHECKPOINTS_RESET", undefined=True)
 
         if not job_id:
             return None, None, None
--- a/src/datachain/lib/dc/datachain.py
+++ b/src/datachain/lib/dc/datachain.py
@@ -37,6 +37,7 @@ from datachain.error import (
 from datachain.func import literal
 from datachain.func.base import Function
 from datachain.func.func import Func
+from datachain.job import Job
 from datachain.lib.convert.python_to_sql import python_to_sql
 from datachain.lib.data_model import (
     DataModel,
@@ -53,6 +54,7 @@ from datachain.lib.signal_schema import SignalResolvingError, SignalSchema
 from datachain.lib.udf import Aggregator, BatchMapper, Generator, Mapper, UDFBase
 from datachain.lib.udf_signature import UdfSignature
 from datachain.lib.utils import DataChainColumnError, DataChainParamsError
+from datachain.project import Project
 from datachain.query import Session
 from datachain.query.dataset import DatasetQuery, PartitionByType
 from datachain.query.schema import DEFAULT_DELIMITER, Column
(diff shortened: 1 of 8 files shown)


src/datachain/lib/dc/datachain.py
```diff
@@ -578,6 +581,23 @@ def persist(self) -> "Self":
             query=self._query.save(project=project, feature_schema=schema)
         )
 
+    def _calculate_job_hash(self, job_id: str) -> str:
+        """
+        Calculates hash of the job at the place of this chain's save method.
+        Hash is calculated using previous job checkpoint hash (if exists) and
+        adding hash of this chain to produce new hash.
+        """
+        last_checkpoint = max(
```
shcheklein: can we have an API for the last checkpoint to simplify this?

src/datachain/lib/dc/datachain.py
```diff
@@ -578,6 +581,23 @@ def persist(self) -> "Self":
...
+            self.session.catalog.metastore.list_checkpoints(job_id),
+            key=lambda obj: obj.created_at,
+            default=None,
+        )
+
+        return hashlib.sha256(
+            (bytes.fromhex(last_checkpoint.hash) if last_checkpoint else b"")
```
shcheklein, line 595: why do we need the previous hash also?

src/datachain/lib/dc/datachain.py
```diff
@@ -632,6 +659,26 @@ def save(  # type: ignore[override]
             # not being able to create it as creation is not allowed
             raise ProjectCreateNotAllowedError("Creating project is not allowed") from e
 
+        # checking checkpoints and skip re-calculation of the chain if checkpoint exist
+        if job_id:
+            job = metastore.get_job(job_id)  # type: ignore[arg-type]
```
shcheklein: why ignore here? can it be fixed?

src/datachain/lib/dc/datachain.py
```diff
@@ -603,6 +623,13 @@ def save(  # type: ignore[override]
                 Available values: `major`, `minor` or `patch`. Default is `patch`.
         """
         catalog = self.session.catalog
+        metastore = catalog.metastore
+
+        job = None
+        _hash = None
+        job_id = os.getenv("DATACHAIN_JOB_ID")
```
shcheklein: how will users provide this? what is the example of the intefrace?

src/datachain/lib/dc/datachain.py
```diff
@@ -632,6 +659,26 @@ def save(  # type: ignore[override]
...
+            _hash = self._calculate_job_hash(job.id)
+
+            if (
+                job.parent_job_id
+                and not checkpoints_reset
+                and metastore.find_checkpoint(job.parent_job_id, _hash)
+            ):
+                # if we find checkpoint with correct hash, we can skip chain calculation
+                catalog.metastore.create_checkpoint(job.id, _hash)
+                from .datasets import read_dataset
```
shcheklein: do we have to  `from .datasets import read_dataset` in multiple places?

also the whole method is now very large - needs refactoring

src/datachain/lib/dc/datachain.py
```diff
@@ -632,6 +659,26 @@ def save(  # type: ignore[override]
...
+                raise JobNotFoundError(f"Job with id {job_id} not found")
+            _hash = self._calculate_job_hash(job.id)
+
+            if (
+                job.parent_job_id
+                and not checkpoints_reset
+                and metastore.find_checkpoint(job.parent_job_id, _hash)
+            ):
+                # if we find checkpoint with correct hash, we can skip chain calculation
+                catalog.metastore.create_checkpoint(job.id, _hash)
```
shcheklein: not sure I follow the logic tbh .. why we create a new checkpoint here?

src/datachain/lib/dc/datachain.py
```diff
@@ -671,6 +718,8 @@ def save(  # type: ignore[override]
                 # current latest version instead.
                 from .datasets import read_dataset
 
+                if job:
+                    catalog.metastore.create_checkpoint(job.id, _hash)  # type: ignore[arg-type]
```
shcheklein: same, fix linter

src/datachain/lib/dc/datachain.py
```diff
@@ -688,13 +737,8 @@ def save(  # type: ignore[override]
...
         )
 
-        if job_id := os.getenv("DATACHAIN_JOB_ID"):
-            catalog.metastore.create_checkpoint(
-                job_id,
-                _hash=hashlib.sha256(  # TODO this will be replaced with self.hash()
-                    str(uuid4()).encode()
-                ).hexdigest(),
-            )
+        if job:
```
shcheklein: how will it work the first time we run it from Python?

src/datachain/lib/dc/datachain.py
ilongin: Added special metastore api

src/datachain/lib/dc/datachain.py
ilongin: Job can have multiple chains. If something is changed, everything after in the script must be recalculated.
So in this example:
```python
dc.read_values().save("ds1")
dc.read_values().save("ds2")
dc.read_values().save("ds3")
```

If we just calculate hash for chain and not take hashes of previous chains into calculation and for example before re-run user changes first chain we won't catch that change in second and third chains and they could be skipped because we will find checkpoint with their own hash.

On the opposite, If we include hashes of previous chains into calculation of current chain then we don't have this problem. The only issue here might be when I get to the UDF checkpoints as then we have overlapping of calculation but I will see when I get there is it even a problem or not.

src/datachain/lib/dc/datachain.py
ilongin: No need for this, prob it was needed before but now it's a leftover. Removing it.

src/datachain/lib/dc/datachain.py
ilongin: We currently provide this from Studio (and from that `datachain query` command but that one is not that important it seems). Note that this still doesn't touch local CLI / Python checkpoint implementation. That will be handled in https://github.com/iterative/datachain/issues/1361

src/datachain/lib/dc/datachain.py
ilongin: I've put it in the top of the `.save()` now to have it in one place. I will think about refactoring

src/datachain/lib/dc/datachain.py
ilongin: My idea was to have all checkpoints present in current job. So if some checkpoint is already created in parent job we just copy it to current job. This is to avoid need and complication for searching checkpoints in job tree and it's simpler to debug this way etc....

src/datachain/lib/dc/datachain.py
ilongin: This is because linter is not smart enough to realize that `_hash` will 100% be defined if `job` is defined. The only thing I can do here is `assert _hash` before method call

src/datachain/lib/dc/datachain.py
ilongin: Note that this is not implementing for Python... Checkpoints with pure Python run will be implemented in https://github.com/iterative/datachain/issues/1361 . This PR relies on `DATACHAIN_JOB_ID` env to get the job. In Python implementation we will just need to create special function to get the job somehow (either by env var or in different way if run withlocal Python) - TBD.
So either way, job should be fetcher here and code should work. Was that you question? I'm not sure I understood it 100%.

src/datachain/lib/dc/datachain.py
shcheklein: thanks!

shcheklein: update it here?

src/datachain/lib/dc/datachain.py
shcheklein: okay, so what is last_checkpoint.hash - the hash of all the chains of the previous job?

src/datachain/lib/dc/datachain.py
shcheklein: removed or not?

src/datachain/lib/dc/datachain.py
shcheklein: kk, please update the comment then

src/datachain/lib/dc/datachain.py
shcheklein: let's do the refactoring, it is getting out of hand (and we ignore the linter complaint?)

src/datachain/lib/dc/datachain.py
```diff
@@ -632,6 +659,26 @@ def save(  # type: ignore[override]
             # not being able to create it as creation is not allowed
             raise ProjectCreateNotAllowedError("Creating project is not allowed") from e
 
+        # checking checkpoints and skip re-calculation of the chain if checkpoint exist
+        if job_id:
+            job = metastore.get_job(job_id)  # type: ignore[arg-type]
+            if not job:
+                raise JobNotFoundError(f"Job with id {job_id} not found")
```
shcheklein: so, can job below be `None`? (I see `if job:` there) ?

src/datachain/lib/dc/datachain.py
ilongin: So it's always the same, hash of some "chain checkpoint" is `hash(hash(chain) + hash_of_previous_chain_checkpoint)` ... so it could be from previous job, but it can be from current job as well.

src/datachain/lib/dc/datachain.py
ilongin: Updated.

src/datachain/lib/dc/datachain.py
ilongin: Yes, those are not under the same branch. btw I'm working on refactoring this whole method to separate checkpoints, delta, some helper methods etc. I think it will be much cleaner and clearer then

src/datachain/lib/dc/datachain.py
ilongin: New method is now being used here. Is that what you meant?

src/datachain/lib/dc/datachain.py
ilongin: it is removed

src/datachain/lib/dc/datachain.py
ilongin: I've refactored it by calling helper methods for checkpoints, delta etc.

src/datachain/lib/dc/datachain.py
shcheklein: kk, I guess comment "previous job checkpoint" also confused me

shcheklein (approved):

shcheklein:
@ilongin fix tests, make sure everything is green before merging

Issue #1356, opened 2025-09-23
`persist` unexpected behavior after concatenation of big chains


ai-ignatyev:
### Description

```python
import datachain as dc

x = dc.read_records(
    [ {'idx': i} for i in range(10000) ],
    schema={'idx': int}
)
y = dc.read_records(
    [ {'idx': i} for i in range(10000, 20000) ],
    schema={'idx': int}
)

z = x | y
print('before:', x.count(), y.count(), z.count())

z = z.persist()
print('after:', x.count(), y.count(), z.count())

```

Expected output:
```
before: 10000 10000 20000
after: 10000 10000 20000
```

Real output:
```
before: 10000 10000 20000
after: 10000 10000 40000
```

### Version Info

```Text
datachain -V
0.31.2

python -V
Python 3.12.9
```

ai-ignatyev:
It seems like the issue is not with `persist`, but with `union`:

```python
import datachain as dc

x = dc.read_records(
    [ {'idx': i} for i in range(10000) ],
    schema={'idx': int}
)
y = dc.read_records(
    [ {'idx': i} for i in range(10000, 20000) ],
    schema={'idx': int}
)

z = x | y
z = z.merge(x, on='idx')
print(x.count(), y.count(), z.count())
```

Expected output:
```
10000 10000 20000
```

Real output:
```
10000 10000 40000
```

shcheklein:
Thanks @ai-ignatyev for the report. I checked it - seems to be specific to SQLite logic that we are using in open source (specifically copy_table optimizations that is using batching). It works in Studio (where we use Clickhouse). I'll check the exact reason and we will fix open source as well.

shcheklein:
Fix is coming here https://github.com/iterative/datachain/pull/1364

Thanks @ai-ignatyev for the excellent report.

shcheklein:
Should be fixed in the latest version.

Issue #603, opened 2024-11-15
We need a function to split Dataset into train / test / eval


shcheklein:
It's a required by customers and is needed for a meaningful prepare data-> train workflow to work. They way we have been doing this so far:

```python
from datachain import DataChain, C

def train_test_split(file) -> str:
    import random
    labels = ["train", "test", "val"]
    return random.choices(labels, weights = [0.7, 0.2, 0.1])[0]

# Add signal

dc = (
    DataChain.from_dataset("qco-image-catalog")
    .map(split=train_test_split, params=["file"], output=str)
    .save()
)

# Save train, test and val datasets

ds_train = (dc.filter(C.split == 'train').save('qco-train'))
ds_test = (dc.filter(C.split == 'test').save('qco-test'))
ds_val = (dc.filter(C.split == 'val').save('qco-val'))
```

Which is a decent start but has a few issues:

- verbose for such a regular use case
- run UDF (compute) for such a simple use case
- might be not stable (each time different results, even if seed is provided - can we even provide a seed and get some meaningful results if we run it in a distributed fashion)

We need a function, probably not on the `dc.py` side, but somewhere in the `lib` that takes a DC and returns a split.

- we can start with wrapping up an existing implementation to iterate on the interface first
- we need later to see / make it work on the database level (no UDF since it's expensive)

Later (in parallel), let's review the existing HuggingFace functionality for this. They have splits as built-in feature. Q: do we want to have something like this on the DC side? Something that looks and feels like a first-class citizen (vs just a wrapper that manipulates with the existing concepts)

dmpetrov:
@shcheklein thank you for creating the issue. I'm just adding a few clarifications.

1. Randomization. We have `sys.rand` in each dataset for exact this reason. The split should be lightweight operation and run inside DB.
2. Seed can be done by XOR-ing: `sys.rand ^ seed` if seed is provided or without any XOR if not provided.
3. We need a to create a special module and populate it with helper functions like this. How about just `toolkit` for now: `from datachain.toolkit import train_test_split`

dmpetrov:
For exact API, please take a look at hugging face implementation and sklearn. Note, we need only basic functionality for now: `train_size, test_size, seed` and `shuffle` is optional for now.

dreadatour:
Follow-up issue: https://github.com/iterative/datachain/issues/606

PR #1364, opened 2025-09-26
fix(union): remap sys ids value to make them unique

shcheklein:
Fixes https://github.com/iterative/datachain/issues/1356

## Summary by Sourcery

Remap sys__id in union operations to guarantee uniqueness and avoid overflow, and add a test to prevent duplication in large union persists

Bug Fixes:
- Ensure sys__id values are unique across union branches by remapping them to dense row numbers with branch-specific offsets

Enhancements:
- Introduce internal _remap_sys_id_dense function to compute new sys__id values as row_number()*2 + branch_offset

Tests:
- Add functional test to verify no duplicated rows when persisting large unions across batch boundaries

### DIFF
--- a/.github/workflows/tests-studio.yml
+++ b/.github/workflows/tests-studio.yml
@@ -32,7 +32,7 @@ jobs:
           POSTGRES_DB: database
           POSTGRES_HOST_AUTH_METHOD: trust
       clickhouse:
-        image: clickhouse/clickhouse-server:24.8
+        image: clickhouse/clickhouse-server:25.8
         ports:
           - 8123:8123
           - 9010:9000
--- a/src/datachain/query/dataset.py
+++ b/src/datachain/query/dataset.py
@@ -982,21 +982,25 @@ class SQLUnion(Step):
 
         columns1, columns2 = _order_columns(q1.columns, q2.columns)
 
-        columns1 = _remap_sys_id_dense(columns1, branch_offset=0)
-        columns2 = _remap_sys_id_dense(columns2, branch_offset=1)
+        union_select = sqlalchemy.select(*columns1).union_all(
+            sqlalchemy.select(*columns2)
+        )
+        regenerated = self.query1.catalog.warehouse._regenerate_system_columns(
+            union_select
+        )
+        result_columns = tuple(regenerated.selected_columns)
 
         def q(*columns):
-            names = {c.name for c in columns}
-            col1 = [c for c in columns1 if c.name in names]
-            col2 = [c for c in columns2 if c.name in names]
-            res = sqlalchemy.select(*col1).union_all(sqlalchemy.select(*col2))
+            if not columns:
+                return regenerated
 
-            subquery = res.subquery()
-            return sqlalchemy.select(*subquery.c).select_from(subquery)
+            names = {c.name for c in columns}
+            selected = [c for c in result_columns if c.name in names]
+            return regenerated.with_only_columns(*selected)
 
         return step_result(
             q,
-            columns1,
+            result_columns,
             dependencies=self.query1.dependencies | self.query2.dependencies,
         )
 
@@ -1223,28 +1227,6 @@ def _order_columns(
     ]
 
     return [[d[n] for n in column_order] for d in column_dicts]
-
-
-def _remap_sys_id_dense(
-    columns: Iterable[ColumnElement], branch_offset: int
-) -> list[ColumnElement]:
-    """Return a new column list where sys__id is remapped to a dense row number.
-
-    The resulting sys__id is computed as:
-      row_number(order by sys__id) * 2 + branch_offset
-
-    This guarantees uniqueness across branches (even vs odd) and avoids
-    arithmetic overflow from scaling arbitrary pre-existing ids.
-    """
-    remapped: list[ColumnElement] = []
-    for col in columns:
-        if getattr(col, "name", None) == "sys__id":
-            rn = sqlalchemy.func.row_number().over(order_by=col)
-            new_id = (rn * 2) + sqlalchemy.literal(branch_offset)
-            remapped.append(new_id.label("sys__id"))
-        else:
-            remapped.append(col)
-    return remapped
 
 
 @attrs.define
--- a/src/datachain/query/dataset.py
+++ b/src/datachain/query/dataset.py
@@ -982,6 +982,9 @@ class SQLUnion(Step):
 
         columns1, columns2 = _order_columns(q1.columns, q2.columns)
 
+        columns1 = _remap_sys_id_dense(columns1, branch_offset=0)
+        columns2 = _remap_sys_id_dense(columns2, branch_offset=1)
+
         def q(*columns):
             names = {c.name for c in columns}
             col1 = [c for c in columns1 if c.name in names]
@@ -1220,6 +1223,28 @@ def _order_columns(
     ]
 
     return [[d[n] for n in column_order] for d in column_dicts]
+
+
+def _remap_sys_id_dense(
+    columns: Iterable[ColumnElement], branch_offset: int
+) -> list[ColumnElement]:
+    """Return a new column list where sys__id is remapped to a dense row number.
+
+    The resulting sys__id is computed as:
+      row_number(order by sys__id) * 2 + branch_offset
+
+    This guarantees uniqueness across branches (even vs odd) and avoids
+    arithmetic overflow from scaling arbitrary pre-existing ids.
+    """
+    remapped: list[ColumnElement] = []
+    for col in columns:
+        if getattr(col, "name", None) == "sys__id":
+            rn = sqlalchemy.func.row_number().over(order_by=col)
+            new_id = (rn * 2) + sqlalchemy.literal(branch_offset)
+            remapped.append(new_id.label("sys__id"))
+        else:
+            remapped.append(col)
+    return remapped
 
 
 @attrs.define
--- a/src/datachain/query/dataset.py
+++ b/src/datachain/query/dataset.py
@@ -985,8 +985,9 @@ class SQLUnion(Step):
         union_select = sqlalchemy.select(*columns1).union_all(
             sqlalchemy.select(*columns2)
         )
+        union_cte = union_select.cte()
         regenerated = self.query1.catalog.warehouse._regenerate_system_columns(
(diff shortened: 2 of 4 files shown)


src/datachain/query/dataset.py
```diff
@@ -1220,6 +1223,28 @@ def _order_columns(
...
+    The resulting sys__id is computed as:
+      row_number(order by sys__id) * 2 + branch_offset
+
+    This guarantees uniqueness across branches (even vs odd) and avoids
+    arithmetic overflow from scaling arbitrary pre-existing ids.
+    """
+    remapped: list[ColumnElement] = []
+    for col in columns:
+        if getattr(col, "name", None) == "sys__id":
+            rn = sqlalchemy.func.row_number().over(order_by=col)
```
shcheklein: not sure about this one cc @dreadatour  - do we need reproducible order? or is it fine to drop it completely ... means each time we run we can a differnet sequence after union ... which is probably fine 🤔 ?

src/datachain/query/dataset.py
```diff
@@ -982,6 +982,9 @@ def apply(
 
         columns1, columns2 = _order_columns(q1.columns, q2.columns)
 
+        columns1 = _remap_sys_id_dense(columns1, branch_offset=0)
+        columns2 = _remap_sys_id_dense(columns2, branch_offset=1)
```
dreadatour: We should also remap `sys__random` field. After `union` there might (will?) be rows with similar `sys__random` field, and we want (?) `sys__random` to be different across all table.

src/datachain/query/dataset.py
```diff
@@ -1220,6 +1223,28 @@ def _order_columns(
     return [[d[n] for n in column_order] for d in column_dicts]
 
 
+def _remap_sys_id_dense(
+    columns: Iterable[ColumnElement], branch_offset: int
+) -> list[ColumnElement]:
```
dreadatour: I believe this should be part of `warehouse` class implementation to be able to use different default value for different warehouse implementations. E.g. `row_number` for SQLite and `random` for ClickHouse.

src/datachain/query/dataset.py
dreadatour: I believe it is fine to ignore the order.

In ClickHouse order is not guaranteed and it was OK until now. But same time may be (?) someone rely on order as "undocumented feature" and this change might break this behavior.

src/datachain/query/dataset.py
```diff
@@ -1220,6 +1223,28 @@ def _order_columns(
...
+      row_number(order by sys__id) * 2 + branch_offset
+
+    This guarantees uniqueness across branches (even vs odd) and avoids
+    arithmetic overflow from scaling arbitrary pre-existing ids.
+    """
+    remapped: list[ColumnElement] = []
+    for col in columns:
+        if getattr(col, "name", None) == "sys__id":
+            rn = sqlalchemy.func.row_number().over(order_by=col)
+            new_id = (rn * 2) + sqlalchemy.literal(branch_offset)
```
dreadatour: Just wonder, should be have `2` as function argument with default value?

Something like:
```python
def _remap_sys_id_dense(
    columns: Iterable[ColumnElement],
    branch_offset: int,
    branches_total: int = 2,
) -> list[ColumnElement]:
```

to be able to use more than 2 branches in future.

Otherwise we might want to use `bool` as `branch_offset` (like `even`/`odd` branch), since only two branches are supported here.

dreadatour (approved):
Thank you for the fix! Looks good to me, approving to unblock, but please, take a look at comments below 🙏

src/datachain/query/dataset.py
shcheklein: unless it is really needed, why would we want to have to manage two different implementations? is it known that one of them is substantially faster?

src/datachain/query/dataset.py
shcheklein: 🤔 do we guarantee sys__random to exist? where do we use it?

dreadatour: Sorry, it is `sys__rand`, not `sys__random`.

> do we guarantee sys__random to exist?

No 100% guarantee (since this field can be filtered out), but we add both `id` and `rand` sys signals in one place: https://github.com/iterative/datachain/blob/72784953938c23409d0a72d84a6483be0e948929/src/datachain/data_storage/schema.py#L276-L282

> where do we use it?

1. In `train_test_split` here: https://github.com/iterative/datachain/blob/72784953938c23409d0a72d84a6483be0e948929/src/datachain/toolkit/split.py#L72 — we are using pre-baked `random` value from `sys__rand` to get deterministic results.
2. In `apply_steps` here: https://github.com/iterative/datachain/blob/72784953938c23409d0a72d84a6483be0e948929/src/datachain/query/dataset.py#L1397 (but I can not remember any place where we do use `DATACHAIN_QUERY_CHUNK_INDEX` and `DATACHAIN_QUERY_CHUNK_TOTAL`).

src/datachain/query/dataset.py
shcheklein: it's a simple change - can do it later when / if it is needed

src/datachain/query/dataset.py
shcheklein: > but I can not remember any place where we do use DATACHAIN_QUERY_CHUNK_INDEX and DATACHAIN_QUERY_CHUNK_TOTAL)

omg, what is this?

> — we are using pre-baked random value from sys__rand to get deterministic results.

if we use it only there, we should probably just add that column there? it seems heavyhanded to maintain it ... how wound window/ groupby functions like rand() and Clickhouse?

src/datachain/query/dataset.py
dreadatour: This is because ClickHouse is distributed and we can not use autoincremented integers as primary keys.

I am not sure how `row_number()` will works on a large scale in ClickHouse — it should works, I can not see any reasons why not, but same time this may leads to some weird flaky bugs, and it will be really hard to investigate those, especially on a large scale.

Deeper research is needed here IMO, or we can stay on a safe side and use random ints, same as we use it in all other places for `sys__id`.

src/datachain/query/dataset.py
dreadatour: > omg, what is this?

Some leftovers, I guess 🫣

> if we use it only there, we should probably just add that column there?

Then results of `train_test_split` will be not deterministic, if we will add `sys__rand` field there before each run. We can simply use `rand()` SQL function instead in this case. We need to have this `sys__rand` field in each dataset pre-baked, to be able to apply `train_test_split`.

More details in original issue here: https://github.com/iterative/datachain/issues/603

The idea (from very beginning, I guess, I am not sure, we do have this field from the very beginning) was to always have this `sys__rand` field in the DB, just in case we need any random filtering/order/etc

> it seems heavyhanded to maintain it...

It is, yes.

src/datachain/query/dataset.py
shcheklein: I just don't really want to have custom implementation across project unless it is really really needed

shcheklein: > I am not sure how row_number() will works on a large scale in ClickHouse 

do you have any reason / suspicion there?

src/datachain/query/dataset.py
shcheklein: > The idea (from very beginning, I guess, I am not sure, we do have this field from the very beginning) was to always have this sys__rand field in the DB, just in case we need any random filtering/order/etc

omg, this way way too heavy (and probably doesn't work anyways in many cases) :(

shcheklein: okay, I guess I will add it, but it is bad design (too heavy if we just use it for split)

dreadatour: We only use it internally for split, since we already have it, but this field was there from very beginning, also for users to be able to use it if needed. I guess.

src/datachain/query/dataset.py
dreadatour: Me too. Just trying to highlight all possible edge cases.

Let me take a really quick research on row_number, just to be sure?

shcheklein: yep, thanks. Much appreciated

src/datachain/query/dataset.py
shcheklein: discussed with @dreadatour . No apparent reason to believe CH breaks on row_number(), we are waiting a response for a bit from CH team, but then we can merge this.

src/datachain/query/dataset.py
dreadatour: It looks like it should be ok to use `row_number` in ClickHouse 👍

shcheklein: for the record, migrate to the new helper that was used for outer merge as well. It covered sys__rand and it is using backend-specific approaches.

src/datachain/query/dataset.py
shcheklein: done with the new helper

src/datachain/query/dataset.py
```diff
@@ -982,18 +982,26 @@ def apply(
 
         columns1, columns2 = _order_columns(q1.columns, q2.columns)
 
+        union_select = sqlalchemy.select(*columns1).union_all(
+            sqlalchemy.select(*columns2)
+        )
+        union_cte = union_select.cte()
```
shcheklein, line 988: Note: `cte()` added to break a very long queries since union is nesting more and more and some tests fail on older SQLite with stack overflow

.github/workflows/tests-studio.yml
```diff
@@ -32,7 +32,7 @@ jobs:
           POSTGRES_DB: database
           POSTGRES_HOST_AUTH_METHOD: trust
       clickhouse:
-        image: clickhouse/clickhouse-server:24.8
+        image: clickhouse/clickhouse-server:25.8
```
shcheklein, line 35: Note: to support `cte()` below

tests/unit/lib/test_datachain.py
```diff
@@ -3686,7 +3686,10 @@ def to_parent(signal: Signal) -> Parent:
         "cnt": "int",
         "sum": "float",
     }
-    assert chain.to_records() == [
+    assert sorted(
```
shcheklein, line 3689: Note: required on newer CH version

PR #1367, opened 2025-09-29
Implement checkpoints usage in local

ilongin:
Enables checkpoint usage with just running `python my_job.py`.
For that to work couple of changes were required:
* We now create a `Job` instance for every dataset, regardless of if it's created with just `python my_script.py` or from SaaS.  When we create dataset from SaaS then SaaS is taking care of job lifecycle (creating, updating state at the end etc.) but on local run it's managed in `datachain` itself. Job instance in local run has name as the full path to the script, so 2 sequential runs of the same script means the next job created is a child of previous job / run.
* `JobManager` was added to be a single place which we can ask to get a `Job` for current process and which updates job state at the end in local env. If `DATACHAIN_JOB_ID` env is present, it means we manage job's lifecycle from outside and that job is just returned from manager, but otherwise new job is being created.
* `JobManager` is used in `DataChain.save()` and in current checkpoints tests
* New tests were added for `JobManager` and e2e tests for checking if job is present when running local script.

## Summary by Sourcery

Centralize and standardize job lifecycle management in DataChain by adopting JobManager for both SaaS and local runs, integrating job IDs into save and checkpoint workflows, and bolstering coverage with new unit and end-to-end tests.

New Features:
- Introduce JobManager to orchestrate DataChain job creation, parent-child linking, and lifecycle management for local script runs
- Always create and attach a Job instance for each save call, using DATACHAIN_JOB_ID for SaaS or local script path as name and linking sequential runs

Enhancements:
- Integrate JobManager into DataChain.save and checkpoint resolution to propagate job IDs into dataset versions and checkpoints
- Add metastore API get_last_job_by_name and utility get_user_script_source for reading the user script as job query

Tests:
- Add unit tests for JobManager covering job reuse, creation, finalization, parent linking, reset behavior, and exception hook delegation
- Extend checkpoint unit tests to exercise both SaaS and local job modes
- Add tests for get_user_script_source and hash consistency for various read methods
- Introduce end-to-end tests verifying single-job creation per script and failure handling with job status updates

### DIFF
--- a/src/datachain/delta.py
+++ b/src/datachain/delta.py
@@ -1,12 +1,16 @@
+import hashlib
 from collections.abc import Sequence
 from copy import copy
 from functools import wraps
 from typing import TYPE_CHECKING, TypeVar
 
+from attrs import frozen
+
 import datachain
 from datachain.dataset import DatasetDependency, DatasetRecord
 from datachain.error import DatasetNotFoundError
 from datachain.project import Project
+from datachain.query.dataset import Step, step_result
 
 if TYPE_CHECKING:
     from collections.abc import Callable
@@ -14,7 +18,9 @@ if TYPE_CHECKING:
 
     from typing_extensions import ParamSpec
 
+    from datachain.catalog import Catalog
     from datachain.lib.dc import DataChain
+    from datachain.query.dataset import QueryGenerator
 
     P = ParamSpec("P")
 
@@ -43,11 +49,38 @@ def delta_disabled(
     return _inner
 
 
+@frozen
+class _RegenerateSystemColumnsStep(Step):
+    catalog: "Catalog"
+
+    def hash_inputs(self) -> str:
+        return hashlib.sha256(b"regenerate_sys_columns").hexdigest()
+
+    def apply(self, query_generator: "QueryGenerator", temp_tables: list[str]):
+        selectable = query_generator.select()
+        regenerated = self.catalog.warehouse._regenerate_system_columns(
+            selectable,
+            keep_existing_columns=True,
+            regenerate_columns=None,
+        )
+
+        def q(*columns):
+            return regenerated.with_only_columns(*columns)
+
+        return step_result(q, regenerated.selected_columns)
+
+
 def _append_steps(dc: "DataChain", other: "DataChain"):
     """Returns cloned chain with appended steps from other chain.
     Steps are all those modification methods applied like filters, mappers etc.
     """
     dc = dc.clone()
+    dc._query.steps.append(
+        _RegenerateSystemColumnsStep(
+            catalog=dc.session.catalog,
+        )
+    )
+
     dc._query.steps += other._query.steps.copy()
     dc.signals_schema = other.signals_schema
     return dc
--- a/tests/func/test_delta.py
+++ b/tests/func/test_delta.py
@@ -1,4 +1,5 @@
 import os
+import uuid
 
 import pytest
 import regex as re
@@ -222,6 +223,42 @@ def test_delta_update_unsafe(test_session):
         (5, 5),
         (6, 6),
     }
+
+
+def test_delta_replay_regenerates_system_columns(test_session):
+    source_name = f"regen_source_{uuid.uuid4().hex[:8]}"
+    result_name = f"regen_result_{uuid.uuid4().hex[:8]}"
+
+    dc.read_values(
+        measurement_id=[1, 2],
+        err=["", ""],
+        num=[1, 2],
+        session=test_session,
+    ).save(source_name)
+
+    def build_chain(delta: bool):
+        read_kwargs = {"session": test_session}
+        if delta:
+            read_kwargs.update({"delta": True, "delta_on": "measurement_id"})
+        return (
+            dc.read_dataset(source_name, **read_kwargs)
+            .filter(C.err == "")
+            .select_except("err")
+            .map(double=lambda num: num * 2, output=int)
+            .select_except("num")
+        )
+
+    build_chain(delta=False).save(result_name)
+
+    build_chain(delta=True).save(
+        result_name,
+        delta=True,
+        delta_on="measurement_id",
+    )
+
+    assert set(
+        dc.read_dataset(result_name, session=test_session).to_values("measurement_id")
+    ) == {1, 2}
 
 
 def test_delta_update_from_storage(test_session, tmp_dir, tmp_path):


src/datachain/hash_utils.py
```diff
@@ -52,6 +56,57 @@ def serialize_column_element(expr: Union[str, ColumnElement]) -> dict:  # noqa:
             "element": serialize_column_element(expr.element),  # type: ignore[arg-type]
         }
 
+    # Clause lists: AND/OR expressions
```
shcheklein: is it related to this PR? why wasn't it covered in the previous one? let's split it into a separate fix .. how do we ensure other cases are covered? can we simplify all of this? (automate)

tests/unit/test_datachain_hash.py
```diff
@@ -170,4 +218,4 @@ def test_diff(test_session):
             status_col="diff",
         )
         .hash()
-    ) == "aef929f3bf247966703534aa3daffb76fa8802d64660293deb95155ffacd8b77"
+    ) == "a2faf0dd6a8a7ea9daff44b0a7c91b797600de30de37b617694e47425c8f650e"
```
shcheklein: so, here is na example of things we'll have to do often now - how do I know this is correct? 

this test in same way could have just checked that hash looks plausible and that is returns the same value twice - so that we don't have to update them every time

also, we need real tests (where we actually tests that certain things we care about are taken into account)

src/datachain/data_storage/metastore.py
```diff
@@ -1593,6 +1599,18 @@ def list_jobs_by_ids(self, ids: list[str], conn=None) -> Iterator["Job"]:
         query = self._jobs_query().where(self._jobs.c.id.in_(ids))
         yield from self._parse_jobs(self.db.execute(query, conn=conn))
 
+    def get_last_job_by_name(self, name: str, conn=None) -> Optional["Job"]:
+        query = (
+            self._jobs_query()
+            .where(self._jobs.c.name == name)
+            .order_by(self._jobs.c.created_at.desc())
+            .limit(1)
```
shcheklein, line 1697: Q: do we allow same names for jobs?

src/datachain/lib/dc/datachain.py
```diff
@@ -629,6 +628,9 @@ def save(  # type: ignore[override]
         self._validate_version(version)
         self._validate_update_version(update_version)
 
+        # get existing job if running in SaaS, or creating new one if running locally
+        job = job_manager.get_or_create(self.session)
```
shcheklein: we already have metastore / session ,etc - can we avoid another class, another manager for this?

src/datachain/query/dataset.py
```diff
@@ -1009,7 +1009,9 @@ class SQLJoin(Step):
     rname: str
 
     def hash_inputs(self) -> str:
-        predicates = ensure_sequence(self.predicates or [])
+        predicates = (
```
shcheklein, line 994: is it related? move to another PR

src/datachain/job.py
```diff
@@ -58,3 +66,180 @@ def parse(
...
+    Behavior:
+      - If the environment variable ``DATACHAIN_JOB_ID`` is set (SaaS mode),
+        the JobManager attaches to that job and does not manage its lifecycle.
+      - If ``DATACHAIN_JOB_ID`` is not set (local script run), the JobManager
+        will:
+          * Create a new job before any work is done.
+          * Use the script path as the job name.
+          * Store the script source (if available) as the job query.
+          * Link to the most recent job with the same name as its parent, if one exists.
+          * Automatically mark the job as ``COMPLETE`` on normal exit, or ``FAILED`` if
```
shcheklein: do we need three states? (so that we can RESUME the job, and if it is actually FAILED collect finally all the garbage left)

src/datachain/job.py
```diff
@@ -58,3 +66,180 @@ def parse(
...
+            self.job = session.catalog.metastore.get_job(env_job_id)
+            if not self.job:
+                raise JobNotFoundError(
+                    f"Job {env_job_id} from DATACHAIN_JOB_ID env not found"
+                )
+            self.owned = False
+        else:
+            # Local run: create new job
+            script = os.path.abspath(sys.argv[0]) if sys.argv else "interactive"
+            source_code = get_user_script_source()
```
shcheklein: again, seems complicated ... why do we need this? is it always available?

src/datachain/lib/udf.py
```diff
@@ -162,7 +162,7 @@ def hash(self) -> str:
         inputs and outputs.
         """
         parts = [
-            hash_callable(self._func),
+            hash_callable(self._func) if self._func else "",
```
shcheklein: how can it be empty?

tests/func/test_datachain.py
```diff
@@ -1663,21 +1663,30 @@ def test_process_and_open_tar(cloud_test_catalog, cloud_type):
     }
 
 
-def test_datachain_save_with_job(test_session, catalog, datachain_job_id):
+@pytest.mark.parametrize("use_datachain_job_id_env", [True, False])
+def test_job_is_created_after_save(test_session, monkeypatch, use_datachain_job_id_env):
```
shcheklein: let's please move (in a separate PR that we can merge BEFORE this one) all checkpoint related tests out of this generic file (it is huge and we must reducing it, not expanding)

tests/unit/test_datachain_hash.py
```diff
@@ -55,13 +61,55 @@ def mock_get_listing():
...
+    path = tmp_dir / "test.jsonl"
+    dc.read_pandas(pd.DataFrame(DF_DATA), session=test_session).to_jsonl(path)
+    assert (
+        dc.read_json(path.as_uri(), format="jsonl", session=test_session).hash()
+        is not None
     )
-    assert dc.read_values(num=[1, 2, 3]).hash() == ""
+
+
+def test_read_pandas(test_session, tmp_dir):
```
shcheklein, line 95: are these tests related? move them into a separate PR

src/datachain/job.py
```diff
@@ -58,3 +66,180 @@ def parse(
...
+
+                sys.excepthook = lambda et, ev, tb: self.finalize_failure(
+                    session, et, ev, tb
+                )
+                self._hooks_registered = True
+
+        return self.job
+
+    def reset(self):
+        """Reset the JobManager state. Useful for testing."""
```
shcheklein: if it is only for testing, it should be part of the test suite, not part of this code

src/datachain/data_storage/metastore.py
ilongin: Yes we do.

src/datachain/lib/dc/datachain.py
ilongin: I wanted to create new class to not repeate the same mistake we did before of bloating classes with too much responsibility which lead to bad code. But maybe you are right, it is a good idea to put this into `Session` as that one already has hooks and everything so I did it now - `JobManager` is removed and it's code to deal with `Job` instance is moved to `Session`.

src/datachain/job.py
ilongin: We actually handle 3 states here:
1. `COMPLETE`  - all went ok with job
2. `FAILED`  - job failed at some point
3. `CANCELED` - user did `KeyboardInterrupt`

User can re-run the job (there is no resuming in that sense since we always create new job) that is failed or canceled from the start or from the last checkpoint.

src/datachain/job.py
ilongin: `script` (full path to script) is needed - it's used as job's name which is later used to check previous job runs and connecting parent - child jobs with the same name.

`source_code` is not mandatory but since in SaaS we have Job query source code set, I thought why not collect it here as well as it's very easy to do so. It's not available only in interactive shells and Jupyter notebooks (as stated in comments of `get_user_script_source()` function) where this function returns `None` and we have this field empty in DB.

src/datachain/lib/udf.py
ilongin: This happens with `ArrowGenerator` for some reason. That class is used for reading CSV and uses generator but doesn't have generator function and has it's own special logic in `process()` method. Anyway, that code is little bit strange because of that (or maybe I'm just not very familiar with it) but I didn't want to touch it now.

tests/unit/test_datachain_hash.py
ilongin: Some of those were failing (e.g situation from above comment where `self._func` is `None`) so I realized I didn't even had tests for those cases - they all don't work with `hash` (they return inconsistent hash very time) but they should not fail - that is main priority. 
I can move all those fixes and tests into another PR if needed, but I thought to leave it here just to speed things out.

src/datachain/job.py
ilongin: This is now moved to `Session.reset_job_state()` and I would like to keep it there. We already have similar methods and if we move this to tests, then we will couple tests with internals of this class (e.g handling hooks etc) and I don't think that is a good idea. Anyway, this method is generic enough so even if it's used only in tests for now, it doesn't seem like a bad fit to the class.

tests/unit/test_datachain_hash.py
ilongin: I don't think this will be that often. This happened since I fixed some issues in hash calculation itself, otherwise this shouldn't change.

We know it's correct if it's consistent, meaning it doesn't fail - returns always the same value on each run.
If we don't check exact value then we cannot know if 2 different values are returned in SaaS and local and that hash is maybe related to someone's local enviromnent (then maybe each worker could calculate different hash and we wouldn't notice that). 

Currently we have 2 tests:
1. `tests/unit/test_hash_utils.py` -> calculate hashes of column elements, functions etc. which are all inputs to chain steps
2. `test_unit/test_datachain_hash.py` -> testing real chains and their hash. Maybe more tests can be added here, need to think about edge cases etc.

tests/unit/test_datachain_hash.py
shcheklein: let's move please all unrelated things so that we don't spend time reviewing / blocking on them

src/datachain/lib/udf.py
shcheklein: but this seem like a hack? same question - can simplify all of this for now, instead of growing this quick fixes and hacks? (someone else will have to then do this and it will take 2x-3x time)

src/datachain/job.py
shcheklein: >source_code is not mandatory but since in SaaS we have Job query source code set, I thought why not collect it

to keep it simple and less code?

we can do it then if it is clear it is needed

src/datachain/job.py
shcheklein: how do we determine this is the same job or a new one?

tests/unit/test_datachain_hash.py
shcheklein: > If we don't check exact value then we cannot know if 2 different values are returned in SaaS and local

I think we don't care about that ... I just don't see a reason for this for now

src/datachain/job.py
shcheklein: > We already have similar methods and if we move this to tests

it doesn't mean it is a good practice at all, it is a  hack and tech debt in most case and more code to maintain. Is it a big issue to have it tests?

src/datachain/query/dataset.py
ilongin: Moved

src/datachain/job.py
ilongin: Not sure I understand the question. So this is signleton (actually this is all moved to `Session` but that is signleton as well) and it creates a `Job` if it's not yet created. While creating new job, it tries to find it's parent (job with the same name). At the end of the process this newly created job will go into one of those 3 states. On new process (script run) new job is created again with parent connection to that last one from previous run.

src/datachain/job.py
ilongin: Ok, changed it to be empty string always in local env. Also removed util function to get script source code and it's tests.

src/datachain/lib/udf.py
ilongin: When I dig deeper into this issue, it seems like this was my mistake (probably because of lack of knowledge of that `udf.py` codebase). I've moved the correct fix to separate branch and will remove it from here https://github.com/iterative/datachain/pull/1378

src/datachain/job.py
dreadatour: - Is there is a way to run the job from scratch, without linking it to the parent?
- Do we consider env variables or `sys.argv` as part of job `uniqueness` or we only do care about file absolute path?

src/datachain/data_storage/metastore.py
```diff
@@ -1593,6 +1599,18 @@ def list_jobs_by_ids(self, ids: list[str], conn=None) -> Iterator["Job"]:
         query = self._jobs_query().where(self._jobs.c.id.in_(ids))
         yield from self._parse_jobs(self.db.execute(query, conn=conn))
 
+    def get_last_job_by_name(self, name: str, conn=None) -> Optional["Job"]:
```
dreadatour: Let's please move this implementation into `SQLiteMetastore` class?
Just to be on a safe side and be sure this logic will not be run in SaaS?

src/datachain/job.py
ilongin: * All jobs are connected to parent automatically if they have the same name (full python script path), but they can be ran from scratch which means all checkpoints will be ignored - for this user needs to add env var `DATACHAIN_CHECKPOINTS_RESET=True`
* Only absolute script path is being considered

tests/unit/test_datachain_hash.py
ilongin: Removed to another PR which is already merged, marking this as resolved

src/datachain/job.py
ilongin: Ok, I agree but the rest of my response is more important I think - coupling tests with internals etc .. but anyway, it's not a problem to move it there if you think it's still better.

ilongin: Moved to tests. Resolving this to clean PR conversations.

src/datachain/data_storage/metastore.py
ilongin: Moved it to sqlite metastore

tests/unit/test_datachain_hash.py
shcheklein: thanks!

src/datachain/job.py
shcheklein: okay, so I run a job - it creates a lot of stuff and fails. It is in FAILED state and I can rerun it. But how do I FAIL it for real? so that all the resources are cleaned up? Do we have some intermediate state in SaaS for this? how do we solve it there?

tests/func/test_datachain.py
shcheklein: it is still needs to be resolved, also in a separate PR if it is not resolved yet please

src/datachain/data_storage/metastore.py
shcheklein: okay, what is the job name ?

src/datachain/hash_utils.py
shcheklein: was it removed from this PR?

src/datachain/data_storage/metastore.py
```diff
@@ -407,6 +407,12 @@ def create_job(
         Returns the job id.
         """
 
+    @abstractmethod
+    def get_last_job_by_name(self, name: str, conn=None) -> Optional["Job"]:
```
shcheklein: will it work in SaaS? what is the semantics of it - what is the job name?

src/datachain/query/session.py
```diff
@@ -94,6 +105,91 @@ def add_dataset_version(
...
 
+    def get_or_create_job(self) -> "Job":
+        """
+        Get or create a Job for this session.
+
+        Returns:
+            Job: The active Job instance.
+
+        Behavior:
+            - If a job already exists in this session, it is returned.
```
shcheklein: how will it work for the nested sessions? does it mean we can have multiple jobs - different in each session in a script?

src/datachain/query/session.py
```diff
@@ -94,6 +105,91 @@ def add_dataset_version(
...
+            # SaaS run: just fetch existing job
+            self.job = self.catalog.metastore.get_job(env_job_id)
+            if not self.job:
+                raise JobNotFoundError(
+                    f"Job {env_job_id} from DATACHAIN_JOB_ID env not found"
+                )
+            self.owns_job = False
+        else:
+            # Local run: create new job
+            script = os.path.abspath(sys.argv[0]) if sys.argv else "interactive"
```
shcheklein: using the same interactive name is dangerous? it will attaching itself to some very old interactive jobs potentially, right? if we run interactively - should we generate random name per interactive session somehow?

src/datachain/query/session.py
```diff
@@ -173,11 +269,30 @@ def get(
...
     @staticmethod
     def except_hook(exc_type, exc_value, exc_traceback):
-        Session.GLOBAL_SESSION_CTX.__exit__(exc_type, exc_value, exc_traceback)
-        Session._global_cleanup()
-
-        if Session.ORIGINAL_EXCEPT_HOOK:
-            Session.ORIGINAL_EXCEPT_HOOK(exc_type, exc_value, exc_traceback)
+        # Handle KeyboardInterrupt specially - mark as canceled and exit with
+        # signal code
+        if exc_type is KeyboardInterrupt:
```
shcheklein: a lot of code duplication, simplify an refactor it please

src/datachain/hash_utils.py
ilongin: Created issue and will make separate PR today (its mostly done)

src/datachain/data_storage/metastore.py
ilongin: No, this is only for local. In SaaS we create Job with uuid as ID and we track it by it (UI knows which Job is being opened, re-ran etc)

src/datachain/data_storage/metastore.py
ilongin: The job name in local run ,e.g with `python my_script.py` is full path to the script, e.g `/home/ivan/job/scripts/my_script.py`  . We needed to have something to "catch on" to be able to find previous job runs . In SaaS this is not a problem as we have control over jobs and we know which job is rerun by it's ID itself (UI knows and will send this info to the backend). This will be clearly explained in documentation.

src/datachain/job.py
ilongin: What resources exactly? With this new job creation in local env we didn't add any new resources nor we need to clean them. UDFs create some resources that are cleaned afterwards by the UDF code itself, but that hasn't been changed. Maybe we will need to have some cleaner when we get to UDF checkpoints as there we will save UDF tables for future job runs, but that will be handled when we get to that.

src/datachain/query/session.py
ilongin: Current implementation will create new job per new (nested) session and this is not what we want actually - we want one job per process / script run. Thanks for the catch. I updated code to fix this (job is now saved in `Session` class attributes rather then private instance attributes) and added a test to verify it.

src/datachain/query/session.py
ilongin: I didn't know there is a history in those runs ... for some reason I thought they are one time run .. anyway, you are right, I've added random name to it.

src/datachain/query/session.py
ilongin: Done

tests/func/test_datachain.py
ilongin: Moved `test_job_is_created_after_save` to `test_job_management.py` and removed other test as it's covered in `test_checkpoints.py` anyway.

amritghimire:
What will happen in the following case?
- User create a dataset locally. It has a job id with it.
- User pushes the dataset to saas
- Will the job id be the same in saas? will foreign key cause any issue?

src/datachain/job.py
shcheklein: tables by `save` (that supposed to be dropped since job failed), all temporary UDF tables that will are not cleaning but won't be doing this (this can be done when we get to UDFs, but tables / datasets - this is not clear, we are changing semantics here)

ilongin:
> What will happen in the following case?
> 
> * User create a dataset locally. It has a job id with it.
> * User pushes the dataset to saas
> * Will the job id be the same in saas? will foreign key cause any issue?

Good question. We probably shouldn't be pushing `job_id` with it. On the other hand, we can maybe think of syncing the jobs between local and SaaS, but for now I think I will just remove `job_id` from it before sending.

src/datachain/job.py
ilongin: I know we briefly discussed this in a meeting but to repeat a question: What are the cons of just leaving created datasets for good, even if their job failed? As a user, I think I would love to see results of those created ones regardless if some chain after it failed. I think we also spoke about showing not only created datasets but those UDF results which are partially done as well (we can discuss this in details later). IMO showing datasets and then removing them is a bad thing and it is also unnecessary complication, but maybe I'm missing something.

dreadatour:
Getting job name from the script name confuses me a lot.

Also correct be if I am wrong, but we are now using job as required part of query run (many checks in the code where we do assert for `job` to be set or did not check if `job` is defined).

If yes, then it looks like in this PR we are limiting the way how we can run datachain code:
1. Using `python file.py` in CLI.
2. From SaaS.

All other options are not considered or not checked:
1. Using `python -c "<python query>"` — name will be set to `/current/working/directory/-c` 😞. Example:
    ```shell
    ~/iterative/playground$ python -c 'import datachain; datachain.read_storage("~/Downloads/").show()'
    ```
    And job name is:
    ```
    sqlite> select id, name from jobs;
    |                  id                  |                   name                   |
    |--------------------------------------|------------------------------------------|
    | 9656e01b-c7fc-438c-91be-2bd8e369310a | /Users/vlad/iterative/playground/-c |
    sqlite>
    ```
    Does this means all the queries ran this way from the same directory will use the same job checkpoints?
    
2. Jupyter Notebooks? Have we even checked them? What `name` will job have? How checkpoint will works with them? Will all the queries ran from the Notebook inherit checkpoints from the same job?
3. Using datachain as a module from a bigger program. What `name` will be set for the job?
4. I can not think of any other scenarios, but I bet there are many.

Should we keep `job` optional and use it (and checkpoint) only if we are 100% sure what we are doing?

ilongin:
> Getting job name from the script name confuses me a lot.
> 
> Also correct be if I am wrong, but we are now using job as required part of query run (many checks in the code where we do assert for `job` to be set or did not check if `job` is defined).
> 
> If yes, then it looks like in this PR we are limiting the way how we can run datachain code:
> 
> 1. Using `python file.py` in CLI.
> 2. From SaaS.

Not really, we just create `Job` with unique name and that's it.

> 
> All other options are not considered or not checked:
> 
> 1. Using `python -c "<python query>"` — name will be set to `/current/working/directory/-c` 😞. Example:

You are right about this case - I didn't cover it. Until now I covered only intaractive runs (Jupyter) in which case random name Job was created but now I've added more robust function to check this.
     
> 2. Jupyter Notebooks? Have we even checked them? What `name` will job have? How checkpoint will works with them? Will all the queries ran from the Notebook inherit checkpoints from the same job?

This was covered already - random name Job is created and there is no connections between runs


> 3. Using datachain as a module from a bigger program. What `name` will be set for the job?

This is good question. Now we will inherit entrypoint of that bigger program so if someone runs `python main.py` and in that `main.py` datachain is imported and used, `Job` will be created with `/full/path/to/main.py` so after second run new `Job` will be created that points to old job as parent and yes, checkpoints are being reused if they match the hash values (this is the key)

> Should we keep `job` optional and use it (and checkpoint) only if we are 100% sure what we are doing?

I think we should always create a `Job` to be consistent and just try to create unique job name where it doesn't make sense to connect to parent jobs.

src/datachain/job.py
ilongin: btw current behavior is that datasets cleanup is actually related to `Session`, not `Job` itself. For example here `numbers_ok` are not removed even after related job fails.

```python
import datachain as dc
from datachain.query.session import Session

with Session("session1") as session1:
    dc.read_values(num=[1, 2, 3, 4 ,5]).save("numbers_ok")

with Session("session2") as session2:
    dc.read_values(num=[1, 2, 3, 4 ,5]).save("numbers_failed")
    raise Exception("Job failed in session2")
```

Regarding checkpoints I made it in a way to create new job on every re-run since python code can change and it seemed like better design, than to have one single `Job` object updated (we loose history that way) ... this means that we cannot solve this with adding another state (I think) and the issue is that we shouldn't remove datasets (or other resources later on) on job being canceled or failed. We can discuss this on retro or after if there is some time.

src/datachain/query/session.py
```diff
@@ -1,21 +1,38 @@
...
 
 
+def is_script_run() -> bool:
+    """
+    Returns True if this was ran as python script, e.g python my_script.py.
+    Otherwise (if interactive or module run) returns False.
+    """
+    try:
+        argv0 = sys.argv[0]
+    except Exception:  # noqa: BLE001
```
shcheklein: key error here? we can capture it exactly

shcheklein: out of index probably or something?

tests/conftest.py
```diff
@@ -224,6 +224,9 @@ def test_session(catalog):
     with Session("TestSession", catalog=catalog) as session:
         yield session
 
+    # Clean up job-related atexit hooks to prevent errors during pytest shutdown
+    Session.cleanup_for_tests()
```
shcheklein: what is the need do this? can we do some direct manipulations directly and not expose and do a method like "clenup_for_tests" in the object (I'm not a big fan of making code more complicated just bc of tests)

src/datachain/query/session.py
```diff
@@ -94,6 +118,115 @@ def add_dataset_version(
     ) -> None:
         self.dataset_versions.append((dataset, version, listing))
 
+    def get_or_create_job(self) -> "Job":
```
shcheklein, line 113: How does all of this work with parallel=True? have you tested it?

tests/conftest.py
ilongin: `Session.cleanup_for_tests()` method was not added in this PR so I don't want to remove it, but I changed this to use new tests util function `reset_session_job_state()`.
This is to clean all the exit hooks to prevent tests to hang on failure because DB tables are already removed by pytest.

src/datachain/query/session.py
ilongin: Added test for `parallel=True`.

shcheklein: but how does it work in reality? does it pass JOB ID via cloudpickle or something else?

tests/unit/lib/test_checkpoints.py
```diff
@@ -60,6 +83,53 @@ def test_checkpoints(
     assert len(list(catalog.metastore.list_checkpoints(second_job_id))) == 3
 
 
+@pytest.mark.parametrize(
+    "cloud_type,version_aware",
+    [("s3", True)],
+    indirect=True,
+)
+@pytest.mark.xdist_group(name="tmpfile")
```
shcheklein: why do we need this?

tests/unit/lib/test_checkpoints.py
```diff
@@ -60,6 +83,53 @@ def test_checkpoints(
     assert len(list(catalog.metastore.list_checkpoints(second_job_id))) == 3
 
 
+@pytest.mark.parametrize(
+    "cloud_type,version_aware",
```
shcheklein: why do we need (here and check other places) cloud fixtures? (they are heavy, definitely not unit tests)

tests/unit/lib/test_checkpoints.py
```diff
@@ -60,6 +83,53 @@ def test_checkpoints(
...
+    [("s3", True)],
+    indirect=True,
+)
+@pytest.mark.xdist_group(name="tmpfile")
+def test_checkpoints_parallel(cloud_test_catalog_tmpfile, monkeypatch):
```
shcheklein: is it really unit test? looks more like functional?

tests/unit/lib/test_checkpoints.py
```diff
@@ -60,6 +83,53 @@ def test_checkpoints(
...
+        raise Exception("Error")
+
+    test_session = cloud_test_catalog_tmpfile.session
+    catalog = test_session.catalog
+
+    dc.read_values(num=[1, 2, 3], session=test_session).save("nums")
+
+    monkeypatch.setenv("DATACHAIN_CHECKPOINTS_RESET", str(False))
+
+    chain = dc.read_dataset("nums", session=test_session).settings(parallel=True)
```
shcheklein: parallel might not be running (double check) if not enough data is there (might need to adjust thresholds and put some asserts that it is actually running subprocesses)

shcheklein: btw, we have a PR that fixes UDF hanging in some cases - it might happen to you it run a parallel mode

shcheklein: at what point it create a job? at first save? can we remove / or a add a test where save is happening after UDF - probably not a big of a change ... just to make sure ...

does persist btw create a JOB? can I have a job w/o save with a mapper that I would like to restart ?

src/datachain/job.py
ilongin: @shcheklein need your opinion here, although I was thinking of stealing little bit time in retro to discuss live as well

shcheklein: My concern here is that `IMO showing datasets and then removing them is a bad thing and it is also unnecessary complication, but maybe I'm missing something.` was a temporary thing, only because we don't really know or it is too complicated to solve. We can discuss it if enough people join. Otherwise let's do a separate call for this.

Does it really change anything though? There will be some other resources most likely to cleanup anyways? We still need some "paused" state for the job, right?

ilongin: > My concern here is that `IMO showing datasets and then removing them is a bad thing and it is also unnecessary complication, but maybe I'm missing something.` was a temporary thing, only because we don't really know or it is too complicated to solve. We can discuss it if enough people join. Otherwise let's do a separate call for this.

I think there is, or at least can be, a difference between created dataset (final product) and other resources like some UDF tables that are transparent to the user anyway. I'm actually wondering why did we even added this cleaning up of datasets if job fails .. if it was because we don't want user to create N datasets for N re-runs, then this issue will be solved by checkpoints....just thinking out loud.

> Does it really change anything though? There will be some other resources most likely to cleanup anyways? We still need some "paused" state for the job, right?

The thing is that ideally I didn't want to add any new state to Job. Adding "paused" state complicate already complicated `Job` lifecycle.  

My thinking was (not sure if this is the best idea though):
1. Datasets are final products and are never removed / cleaned automatically after created -> this simplifies things like above with shown and then removed datasets and solution to that which is actually complex.
2. Partial UDF tables (and other tables needed for UDF ) are preserved until that UDF is successfully done or some TTL has  passed (this is open to discussion)
3. Successful UDF results - this is a tricky one as Dmitry wants this to save as actual (special kind) dataset (e.g. dataset with `checkpoint` prefix or something), show their results in UI and show them as a dependency in dependency graph. I still need to think about this ones when I start working on UDF checkpoints (soon).

If above is true, I currently don't see the need for special state .. job can go to success, failed or canceled (same as failed from code perspective). If it's in success state there are no resources hanging. If in failed / canceled state then some resources may be active and will be either removed when job goes to success state, or will be removed after some checkpoints TTL (time window for checkpoints to be usable / active), for example. Not sure if I'm missing something.

shcheklein: >  I'm actually wondering why did we even added this cleaning up of datasets if job fails .. 

I think just nobody liked these in-between states, but we can try that I guess. I'm fine to try. Keep in mind - we'll need to change UI somehow to still report errors and show some produced datasets. Also we can introduce an explicit button to drop all of them at once.

> removed after some checkpoints TTL (time window for checkpoints to be usable / active),

Locally it is not clear how we do this. Run some GC like git does on every job or so?

ilongin: I spoke with @dmpetrov and he is more on the side of having job as atomic, meaning either all datasets are created or none. We would still need to save those created ones as checkpoint results and make them usable in the Job itself (if there are inter-dependencies between datasets) but hide them from the users. I will look into how complex would this be to implement.

shcheklein: Let's do this please as a follow up folks cc @dmpetrov  I also hesitant doing this (as you can see from this whole thread - it took a few back and forth), but first - we need to deliver the value, second, I don't see changing semantics would break real scenarios across customers.

src/datachain/query/session.py
shcheklein: please check this @ilongin

src/datachain/query/session.py
shcheklein: also, let's review this ... just to be sure

shcheklein:
@ilongin there are a few other comments, let's please review them

src/datachain/job.py
ilongin: @shcheklein the easiest solution is to just keep datasets alive (remove logic of remove them after job fails) but there is an issue: We will change this behavior back and forth - currently we have somewhat broken removal of datasets (they show and then they disappear), then if we decide here to keep them and not remove this will be a change for the users and then finally in a follow up we will again change behavior to remove them after job fails and keep them active only in the job itself (for dependencies between datasets in a job to work) ... so there is a lot of changes for the users....is this ok?

src/datachain/query/session.py
ilongin: This will be handled when we get to UDF checkpoints but I think even then this won't be needed (passing `job_id`) as parallel execution happens after we already have input defined and checkpoints will be checked before (on input definition itself to skip certain rows etc).

tests/unit/lib/test_checkpoints.py
ilongin: Not needed, removing

tests/unit/lib/test_checkpoints.py
ilongin: It's needed for testing parallel since unlike `test_session` it doesn't use in memory db which is problematic with child processes (they have different memory space from parent process and then they don't see correct tables). This fixture is used in all tests where parallel is tested.

ilongin: Looks like there is already `test_session_tmpfile` fixture that is not in memory and was created for tests with distributed processing so I used that at the end to avoid those heavy cloud types.

tests/unit/lib/test_checkpoints.py
ilongin: Moved to functional test file

tests/unit/lib/test_checkpoints.py
ilongin: It does work as expected, even in my case where I have just 3 rows -> in this case number of processes spawn is the same as number of CPU cores (my case 8) but since I have work for only 3 processes (each receiving one row) other 5 workers are sent STOP_SIGNAL right away.
I've increased number of rows to 1000 just to be more realistic though.

ilongin: > at what point it create a job? at first save? 

Correct.


> can we remove / or a add a test where save is happening after UDF - probably not a big of a change ... just to make sure ...

Not sure I understand. In these tests save is actually happening after UDF. 

> does persist btw create a JOB? 

Persist doesn't create a job. Job is created in actual `DataChain.save()` function before running steps (before persist).

> can I have a job w/o save with a mapper that I would like to restart ?

If there is no `.save()` there is no `Job` created and therefor no checkpoints.

tests/func/test_checkpoints.py
```diff
@@ -12,7 +12,7 @@ def mapper_fail(num) -> int:
     test_session = test_session_tmpfile
     catalog = test_session.catalog
 
-    dc.read_values(num=[1, 2, 3], session=test_session).save("nums")
+    dc.read_values(num=list(range(1000)), session=test_session).save("nums")
```
shcheklein, line 21: is it better to patch the value of the variable that defines the batch size or something?

tests/unit/lib/test_checkpoints.py
shcheklein: yes, exactly we have test_session_tmpfile for that!

tests/unit/lib/test_checkpoints.py
shcheklein: what if we have a script with UDF and show() ? we don't want to save anything at all ... but we also want to reuse results as we iterate.

It feels like quite a limitation and even wrong design tbh that JOB is actually running, but no JOB is created... and JOB is created in save and only if we have save

ilongin: You are right, checpoints / job should be created in that scenario as well but I would do that in follow up issue when implementing UDF checkpoints as here that is not important.

Regarding the place where job should be created - it should probably be before applying steps from DatasetQuery as thats where actual "job" starts.

shcheklein: kk, let's reconsider this in the next step

docs/guide/checkpoints.md
shcheklein, line 1: looks good! how much of this is AI generated?

tests/test_query_e2e.py
```diff
@@ -159,13 +159,14 @@ def run_step(step, catalog):  # noqa: PLR0912
     """Run an end-to-end query test step with a command and expected output."""
     command = step["command"]
     # Note that a process.returncode of -2 is the same as the shell returncode of 130
-    # (canceled by KeyboardInterrupt)
-    interrupt_exit_code = -2
+    # (canceled by KeyboardInterrupt). Also accept 130 explicitly as some Python
+    # scripts may exit with that code directly.
+    interrupt_exit_codes = (-2, 130)
```
shcheklein, line 164: @ilongin why do we have this?

docs/guide/checkpoints.md
ilongin: All. I just did maybe 5-6 prompts to make some things better, but even the first version would be good enough I think...for now I found AI to be most useful in generating docs and helping understand the cause of non trivial bugs.

tests/test_query_e2e.py
shcheklein: @ilongin ping

ilongin: The `130` exit code (128 + SIGINT signal number 2) is the standard Unix convention for "process was interrupted by Ctrl+C".                
                
We need it because our custom `Session.except_hook` in `src/datachain/query/session.py:326` replaces the default `sys.excepthook` to do cleanup (finalize job, close session). Without the explicit `sys.exit(130)` there, after the hook handles `KeyboardInterrupt` and returns, Python would exit with code 0 — making it look like the script succeeded. 
`sys.exit(130)` preserves the correct exit semantics so parent processes/shells can tell the script was interrupted.

The test needs to accept both `-2` (process killed directly by SIGINT before our hook runs) and `130` (our hook caught the interrupt, did cleanup, then exited explicitly).

shcheklein: can't we propagate the same exit code it would have exited with naturally?

PR #1368, opened 2025-09-30
be less strict - pass None on validation errors

shcheklein:
To quickly mitigate broken (e.g. after merge objects) (and may be even to have this as a mode later?) - let's not raise, just pass None and user code can deal with it.

@dmpetrov @dreadatour WDYT?

Other solutions are pretty heavy and clunky and a few users depend on this (running locally SQLite with merge).


dmpetrov:
That's might be a decent workaround but if it works the same in cli and saas. Otherwise, we kind of accumulating tech depth - ok as a short fix but we need to address it very soon.

dreadatour:
Might works, but same time this looks very dangerous to me — this can silently spoil output datasets 😢
For example, if partial validation error will happens, e.g. one field from the whole model will be invalid, the whole model will be zeroed, and no errors shown.

shcheklein:
It works on both the same I guess, I can check.

> Might works, but same time this looks very dangerous to me — this can silently spoil output datasets 😢
For example, if partial validation error will happens, e.g. one field from the whole model will be invalid, the whole model will be zeroed, and no errors shown.

I think this should be only about inputs to UDFs, and people will catch it since they will have None?

dreadatour:
> I think this should be only about inputs to UDFs, and people will catch it since they will have None?

Oh, got it! You're right, for inputs only this should be totally fine! 👍

PR #1371, opened 2025-10-01
fix sys ids in outer merge

shcheklein:
Endless nightmare with system columns.


src/datachain/data_storage/sqlite.py
```diff
@@ -875,8 +875,17 @@ def add_left_rows_filter(exp: BinaryExpression):
                 if isinstance(c, BinaryExpression):
                     right_left_join = add_left_rows_filter(c)
 
-        union = sqlalchemy.union(left_right_join, right_left_join).subquery()
-        return sqlalchemy.select(*union.c).select_from(union)
+        # Use CTE instead of subquery to force SQLite to materialize the result
+        # This breaks deep nesting and prevents parser stack overflow.
+        union_cte = sqlalchemy.union(left_right_join, right_left_join).cte()
```
shcheklein, line 880: C: also adds `cte()` to avoid endless stack trace and exponential SQL growth on outer merges

PR #1374, opened 2025-10-03
Fix none value objects

shcheklein:
Followup https://github.com/iterative/datachain/pull/1368/files

Fixes NaN handling in objects (we covert None to NaN and SQLite since it doesn't distinguish them).

Adds a move comprehensive test to make sure we can consume and return None on SQLite (and it is also handled fine on ClickHouse).

Also moved other UDF tests to a separate file to avoid touching "test_datachain nightmare".

## Summary by Sourcery

Improve null-value handling in SignalSchema by detecting and converting all-None nested inputs to None, add related unit and functional tests, fix NaN handling for SQLite, and reorganize UDF tests into a dedicated file.

New Features:
- Introduce static method to detect all-None values in nested structures
- Catch validation errors when constructing nested Pydantic models with only null inputs and return None instead of failing

Bug Fixes:
- Fix handling of None↔NaN conversions for nested DataModel objects on SQLite (and maintain compatibility with ClickHouse)

Enhancements:
- Add logging in SignalSchema for debugging null-value conversions
- Wrap object creation in row_to_objs and row_to_features to gracefully handle all-null inputs

Tests:
- Add unit tests for row_to_objs behavior with complete and partial None values in nested collections
- Add comprehensive functional test for UDF handling of None nested DataModel after full outer merge
- Move existing UDF tests out of test_datachain.py into new tests/func/test_udf.py to declutter

Chores:
- Remove UDF tests from the monolithic test_datachain.py to simplify the file


tests/func/test_udf.py
```diff
@@ -0,0 +1,754 @@
...
+
+import datachain as dc
+from datachain.func import path as pathfunc
+from datachain.lib.file import AudioFile, AudioFragment, File
+from datachain.lib.udf import Mapper
+from datachain.lib.utils import DataChainError
+from tests.utils import LARGE_TREE, NUM_TREE
+
+
+def test_udf_none_nested_datamodel_after_outer_merge(test_session):
```
shcheklein, line 16: C: this is a new test here, everything else was moved

PR #1379, opened 2025-10-04
Fix parallel workers stuck

dreadatour:
Trying to fix issue with parallel workers stuck using AI.

Draft, experimenting, update and reasoning required.

## Summary by Sourcery

Improve fault handling and cleanup in parallel UDF execution to prevent workers from hanging and avoid queue blocking

Bug Fixes:
- Fail fast on first worker exception by stopping task scheduling and terminating remaining workers to prevent deadlocks

Enhancements:
- Use contextlib.suppress to safely close queues and terminate processes during cleanup without blocking
- Drop non-essential status updates when the task queue is full to avoid producer blocking in put_into_queue

### DIFF
--- a/src/datachain/query/dispatch.py
+++ b/src/datachain/query/dispatch.py
@@ -310,6 +310,9 @@ class UDFDispatcher:
                 elif status == FINISHED_STATUS:
                     n_workers -= 1  # Worker finished
                 else:  # Failed / error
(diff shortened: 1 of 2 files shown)


dreadatour:
Fix in https://github.com/iterative/datachain/pull/1384

PR #1382, opened 2025-10-05
cleanup column utils

shcheklein:
Experimenting for now as a follow up to https://github.com/iterative/datachain/pull/1380

### DIFF
--- a/src/datachain/query/utils.py
+++ b/src/datachain/query/utils.py
@@ -7,11 +7,9 @@ ColT = Union[sa.ColumnClause, sa.Column, sa.ColumnElement, sa.TextClause, sa.Lab
 
 def column_name(col: ColT) -> str:
     """Returns column name from column element."""
-    return (
-        col.name
-        if isinstance(col, (sa.ColumnClause, sa.Column, sa.Label))
-        else str(col)
-    )
+    if isinstance(col, (sa.TextClause)):
+        raise TypeError(f"Got text clause column in select: {col!s}")
+    return col.name
 
 
 def get_query_column(query: sa.Select, name: str) -> Optional[ColT]:
@@ -23,18 +21,3 @@ def get_query_id_column(query: sa.Select) -> Optional[sa.ColumnElement]:
     """Returns ID column element from query or None if column not found."""
     col = get_query_column(query, "sys__id")
     return col if col is not None and isinstance(col, sa.ColumnElement) else None
-
-
-def select_only_columns(query: sa.Select, *names: str) -> sa.Select:
-    """Returns query selecting defined columns only."""
-    if not names:
-        return query
-
-    cols: list[ColT] = []
-    for name in names:
-        col = get_query_column(query, name)
-        if col is None:
-            raise ValueError(f"Column '{name}' not found in query")
-        cols.append(col)
-
-    return query.with_only_columns(*cols)


shcheklein:
Do we have tests besides basic that actually hit this code? 

Probably not that many or none at all unfortunately ...

dreadatour:
> Do we have tests besides basic that actually hit this code?

Unfortunately we don't have direct tests for these methods.

I went even further — completely removed `query/utils.py` — please, take a look?

Also added more tests.

src/datachain/data_storage/warehouse.py
```diff
@@ -228,7 +227,8 @@ def dataset_select_paginated(
             while True:
                 if limit is not None:
                     limit -= num_yielded
-                    if limit == 0:
+                    num_yielded = 0
+                    if limit <= 0:
```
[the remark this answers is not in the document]
dreadatour: This is fix for correct pagination, error was found in tests.

src/datachain/data_storage/warehouse.py
```diff
@@ -228,7 +227,8 @@ def dataset_select_paginated(
             while True:
                 if limit is not None:
                     limit -= num_yielded
-                    if limit == 0:
+                    num_yielded = 0
```
dreadatour, line 230: Found this with tests. Query with limit was broken here.

src/datachain/query/dispatch.py
```diff
@@ -65,7 +64,7 @@ def udf_entrypoint() -> int:
     wh_cls, wh_args, wh_kwargs = udf_info["warehouse_clone_params"]
     warehouse: AbstractWarehouse = wh_cls(*wh_args, **wh_kwargs)
 
-    id_col = get_query_id_column(query)
+    id_col = query.selected_columns.get("sys__id")
```
shcheklein: perfect!

tests/func/test_batching.py
```diff
@@ -240,3 +240,18 @@ def test_partition_ids_only(warehouse, partitioned_animal_dataset_query):
...
+
+
+@pytest.mark.parametrize("cloud_type,version_aware", [("file", False)], indirect=True)
+def test_partition_missing_column_raises(warehouse, animal_dataset):
+    table = warehouse.get_table(
+        warehouse.dataset_table_name(animal_dataset, animal_dataset.latest_version)
+    )
+
+    # Build a query without partition_id to ensure Partition raises
+    query = sa.select(table.c.sys__id, table.c.file__path).order_by(table.c.sys__id)
```
shcheklein: can we btw move sys_id creation after partition to the partition itself? like we do with union and outr merge now?

dreadatour: Please, take a look here: https://github.com/iterative/datachain/pull/1382/commits/57ccd9d39fb3a328d7e0242ce697529848ce8ea0

(I'll add more tests on this if this is ok)

.github/workflows/tests.yml
```diff
@@ -220,6 +219,56 @@ jobs:
...
+          head -20 coverage.xml
+
+      - name: Upload coverage report
+        uses: codecov/codecov-action@v5
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          files: coverage.xml
+          flags: datachain
+          fail_ci_if_error: false
+          verbose: true
```
shcheklein: @dreadatour check this alert. just in case

src/datachain/data_storage/warehouse.py
```diff
@@ -257,18 +257,31 @@ def _regenerate_system_columns(self, selectable):
...
             elif col.name == "sys__rand":
                 expr = self._system_random_expr()
                 expr = sa.cast(expr, system_types["sys__rand"])
                 result_columns.append(expr.label("sys__rand"))
+                rand_col_found = True
             else:
                 result_columns.append(col)
 
+        # Add missing system columns if needed
+        if not id_col_found:
```
shcheklein: it still can be refactored, to much duplication here

tests/func/test_batching.py
```diff
@@ -240,3 +240,18 @@ def test_partition_ids_only(warehouse, partitioned_animal_dataset_query):
 
     assert not db_ids
     assert not partition_ids
+
+
+@pytest.mark.parametrize("cloud_type,version_aware", [("file", False)], indirect=True)
```
shcheklein: do we really have to use cloud fixtures here?

also, looks like a unit test / not functional (that's not exactly clear)

dreadatour: - removed cloud fixtures from here 👍
- move these tests from func to unit 👍

src/datachain/data_storage/warehouse.py
dreadatour: Updated, please, take a look 🙏

shcheklein:
LGTM (check some tests failing though?)

dreadatour:
> check some tests failing though?

Not related to this PR 😥

shcheklein:
> Not related to this PR 😥

we can merge it then I guess then?

src/datachain/query/dataset.py
```diff
@@ -634,21 +634,11 @@ def apply(
...
-                # to hold the query results, so we can join it with the
-                # partition table later.
-                columns = [
-                    c if isinstance(c, Column) else Column(c.name, c.type)
-                    for c in query.subquery().columns
-                ]
-                temp_table = self.catalog.warehouse.create_dataset_rows_table(
-                    self.catalog.warehouse.temp_table_name(),
-                    columns=columns,
+            if "sys__id" not in query.selected_columns:
```
shcheklein, line 637: do we need this `if`? not anymore right?

also we changed the semantics of _regenerate_system_columns - it now generates columns if they are missing (it was not the case before). I'm not 100% sure (yet), but we have cases where we don't select or generate those columns unless they are explicitly needed ...

PR #1383, opened 2025-10-05
Improve column element serialization

ilongin:
Improving `ColumnElement` serialization for hashing. Now this implementation is generic and should work for any subclass of `ColumnElement`, not only the common ones.
New tests are added

TODO: add more tests for `test_query_steps_hash.py`

### DIFF
--- a/src/datachain/hash_utils.py
+++ b/src/datachain/hash_utils.py
@@ -5,99 +5,76 @@ import textwrap
 from collections.abc import Sequence
 from typing import TypeVar, Union
 
-from sqlalchemy.sql.elements import (
-    BinaryExpression,
-    BindParameter,
-    ColumnElement,
-    Label,
-    Over,
-    UnaryExpression,
-)
-from sqlalchemy.sql.functions import Function
+from sqlalchemy.sql.elements import ColumnElement
 
 T = TypeVar("T", bound=ColumnElement)
 ColumnLike = Union[str, T]
 
 
-def serialize_column_element(expr: Union[str, ColumnElement]) -> dict:  # noqa: PLR0911
+def _serialize_value(val):  # noqa: PLR0911
+    """Helper to serialize arbitrary values recursively."""
+    if val is None:
+        return None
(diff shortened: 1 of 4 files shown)


src/datachain/hash_utils.py
```diff
@@ -5,99 +5,71 @@
...
 
-    # Plain columns
-    if hasattr(expr, "name"):
-        return {"type": "column", "name": expr.name}
-
-    # Fallback: stringify unknown nodes
+    # Generic handling for all ColumnElement types using SQLAlchemy's internals
+    if isinstance(expr, ColumnElement):
+        # All standard SQLAlchemy types have _traverse_internals
+        if hasattr(expr, "_traverse_internals"):
```
shcheklein, line 46: I like that it is somewhat simpler now, but how reliable it is? how maintainable is it ? (they will change the API at some point?) and most likely no one will figure out it and will switch to stringify anyways? So, it is really bad idea to just use plain stringify? I think we discussed this before - it was not obvious, but seemed fine

ilongin: `_traverse_internals` is actually part of SQLAlchemy's public visitor/cloning API (https://docs.sqlalchemy.org/en/20/core/visitors.html) and is definetely more stable than pure string generation. It should be reliable as it shouldn't really change unless they change public API. We shouldn't touch this code any more unless we bump on some bugs.

Regarding the string option, even if we ignore the fact that stringifing produces different strings for different dialects which automatically disables using checkpoints between local and SaaS and creates issue with tests, `statement.compile(dialect=dialect, compile_kwargs={"literal_binds": True}))` just doesn't guarantee to return consistent strings even in the same dialects. 
Issues with it are the one I mentioned before in previous PRs with addition of random table names in string results (happens in join) which we would need to deal with somehow,  then it doesn't work with `bytes` if they are not UTF-8 compatible etc etc, .. in general this method was meant to be used only for debugging / logging in development and def not for things like producing consistent hashing.

PR #1384, opened 2025-10-06
fix(dispatcher): simplify and fix termination handling

shcheklein:
Fixes hanging of the UDF in parallel mode. Simple script to reproduce is below. A more advanced script to test this is here:

https://gist.github.com/shcheklein/c988ab49da95628fc7f4b871d4e033b6

```python
# ruff: noqa: INP001

"""Quick local check of running a UDF with parallel execution.

Run this script directly to see the UDF results printed to stdout.
"""
import multiprocess as mp

import datachain as dc

_SEEN_WORKERS: set[str] = set()


def compute_metrics(word: str) -> tuple[str, int]:
    """Return the upper-cased word and its length."""
    proc = mp.current_process()  # type: ignore[attr-defined]
    if proc.name not in _SEEN_WORKERS:
        print(f"running on {proc.name} (pid={proc.pid})")
        _SEEN_WORKERS.add(proc.name)

    if proc.name == "Worker-UDF-1":
        return None.upper(), len(word)
    return word.upper(), len(word)


def main() -> None:
    base_words = ["alpha", "beta", "gamma", "delta", "epsilon"]
    words = base_words * 20_000  # 5 * 20_000 = 100_000 entries

    chain = (
        dc.read_values(word=words)
        .settings(parallel=2)
        .map(
            compute_metrics,
            params=["word"],
            output={"upper": str, "length": int},
        )
        .select("word", "upper", "length")
    )

    total = 0
    for total, (original, upper, length) in enumerate(chain, start=1):
        if total <= 10:
            print(f"{original!r} -> {upper!r}, length={length}")

    print(f"processed {total} rows")


if __name__ == "__main__":
    main()
```

A bit more advanced script to test it in different


src/datachain/query/dispatch.py
```diff
@@ -290,10 +292,20 @@ def run_udf_parallel(  # noqa: C901, PLR0912
...
-                try:
-                    result = get_from_queue(self.done_queue)
-                except KeyboardInterrupt:
-                    break
+                while True:
+                    try:
+                        result = self.done_queue.get_nowait()
+                        break
+                    except Empty:
+                        for p in pool:
```
shcheklein, line 292: Need this in case worker dies abruptly.

src/datachain/query/dispatch.py
```diff
@@ -320,39 +332,23 @@ def run_udf_parallel(  # noqa: C901, PLR0912
...
-                    result = get_from_queue(self.done_queue)
-                    status = result["status"]
-                    if status != OK_STATUS:
-                        n_workers -= 1
-
-                self.done_queue.close()
-                self.done_queue.join_thread()
-
-            # Wait for workers to stop
+            for p in pool:
```
shcheklein: I don't see a reason from the to of my head to complicate it much tbh ... Let's see tests ...

PR #1386, opened 2025-10-06
Add dependency tracking to `hash_callable()`

ilongin:
Currently `hash_callable()` only hashed the function itself. Changes to helper functions in the same script didn't change the hash

Example:

  ```python
  def helper(x):
      return x + 1

  def my_udf(x):
      return helper(x) * 2

  hash1 = hash_callable(my_udf)

  # Change helper
  def helper(x):
      return x + 10

  hash2 = hash_callable(my_udf)  # hash1 == hash2 -> wrong!
```

  Solution

  Now recursively tracks and hashes user-defined helper functions from the same **module**:
  - Inspects `func.__code__.co_names` to find dependencies
  - Only includes functions from same file (ignores imports, built-ins, classes)
  - Handles circular dependencies

  Limitations

  1. Global variables/constants not tracked - e.g. changing THRESHOLD = 100 to 200 (assuming it's used in function) won't change hash
  2. Cross-module imports ignored - from utils import helper changes not tracked (by design)
  3. Closures not tracked - captured values in closures not reflected

## Summary by Sourcery

Add recursive dependency tracking to hash_callable so that changes in user-defined helper functions within the same module affect the hash, with cycle detection to prevent infinite recursion.

New Features:
- Recursively include user-defined helper functions from the same module in hash_callable
- Handle circular dependencies by tracking visited functions

Enhancements:
- Inspect function code names and globals to identify same-module dependencies and ignore imports, builtins, and classes
- Include sorted dependency hashes in the hash payload for determinism

Tests:
- Add unit test to verify that hash_callable changes when a helper function implementation is modified

### DIFF
--- a/src/datachain/hash_utils.py
+++ b/src/datachain/hash_utils.py
@@ -142,11 +142,13 @@ def hash_callable(func, _visited=None):
             lines, _ = inspect.getsourcelines(func)
             payload = textwrap.dedent("".join(lines)).strip()
         except (OSError, TypeError):
-            # Fallback: bytecode if source not available
-            payload = func.__code__.co_code
+            # Fallback: bytecode + constants if source not available
+            code = func.__code__
+            payload = (code.co_code, code.co_consts, code.co_names, code.co_varnames)
     else:
-        # For lambdas, fall back directly to bytecode
-        payload = func.__code__.co_code
+        # For lambdas, use bytecode + constants
+        code = func.__code__
+        payload = (code.co_code, code.co_consts, code.co_names, code.co_varnames)
 
     # Normalize annotations
     annotations = {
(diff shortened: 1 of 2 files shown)


shcheklein:
Let's deprioritize this. Or let me know why this is a priority. Even hashing the function is questionable (we actually want to be able to change it and run the same job again, and it will be very hard to make it robust and cover all cases. But even besides that - we have a way bigger priority - make basic case work e2e. Specifically - single chain with UDF restart. Let's please make it work, then we can discuss improvements like this.

shcheklein:
@ilongin can be closed for now?

shcheklein:
@ilongin is this one still needed?

PR #1387, opened 2025-10-06
Optimize dataset dependency calculation and storage

amritghimire:
Summary:
------
This PR changes improvement in performances to dataset dependency
calculation as well as in storage where we now store the nested
dependency structure directly to optimize the retreival process.

Changes:
------
- Replaced recursive database queries with a single batch query that fetches all dependencies at once
- Reduced Database Round-trips: Dependency calculation now uses get_direct_dataset_dependencies_by_ids() to fetch all required dependencies in one query
- Optimized Memory Usage: New DatasetDependencyMinimal class stores only essential dependency information

Database change:
-------
- Added nested_dependencies Column: New JSON column in datasets_dependencies table to store pre-calculated dependency structures
- Backward Compatibility: Schema migration safely adds the new column to existing databases
- Efficient Storage: Nested dependency trees are stored as JSON, reducing storage overhead

### DIFF
--- a/src/datachain/catalog/catalog.py
+++ b/src/datachain/catalog/catalog.py
@@ -509,6 +509,44 @@ def clone_catalog_with_cache(catalog: "Catalog", cache: "Cache") -> "Catalog":
     return clone
 
 
+def get_all_ids(d):
+    ids = set()
+    for key, value in d.items():
+        ids.add(int(key))
+        if isinstance(value, dict):
+            ids.update(get_all_ids(value))
+    return sorted(ids)
+
+
+def build_nested_dependencies(dataset_deps, dependency_structure, indirect=True):
+    def build_deps(dep_id, deps_dict):
+        if dep_id not in dataset_deps:
+            return []
+
+        deps = []
+        if dep_id in deps_dict:
+            for child_id in deps_dict[dep_id]:
+                if child_id in dataset_deps:
+                    # Recursively build child dependencies if indirect=True
+                    child_deps = (
+                        build_deps(child_id, deps_dict[dep_id]) if indirect else []
+                    )
+                    child_dep = dataset_deps[child_id]
+                    child_dep.dependencies = child_deps
+                    deps.append(child_dep)
+        return deps
+
+    result = []
+    for root_id in dependency_structure:
+        if root_id in dataset_deps:
+            deps = build_deps(root_id, dependency_structure)
+            root_dep = dataset_deps[root_id]
+            root_dep.dependencies = deps
+            result.append(root_dep)
+
+    return result
+
+
 class Catalog:
     def __init__(
         self,
@@ -1198,40 +1236,82 @@ class Catalog:
 
     def get_dataset_dependencies(
         self,
-        name: str,
-        version: str,
+        name: str | None = None,
+        version: str | None = None,
         namespace_name: str | None = None,
         project_name: str | None = None,
         indirect=False,
+        dataset_id: int | None = None,
+        dataset_version_id: int | None = None,
     ) -> list[DatasetDependency | None]:
-        dataset = self.get_dataset(
-            name,
-            namespace_name=namespace_name,
-            project_name=project_name,
-        )
+        if (name and dataset_id) or (version and dataset_version_id):
+            raise ValueError(
+                "Either name and version or"
+                " dataset_id and dataset_version_id must be provided"
+            )
 
-        direct_dependencies = self.metastore.get_direct_dataset_dependencies(
-            dataset, version
-        )
-
-        if not indirect:
-            return direct_dependencies
-
-        for d in direct_dependencies:
-            if not d:
-                # dependency has been removed
-                continue
-            if d.is_dataset:
-                # only datasets can have dependencies
-                d.dependencies = self.get_dataset_dependencies(
-                    d.name,
-                    d.version,
-                    namespace_name=d.namespace,
-                    project_name=d.project,
-                    indirect=indirect,
+        if not dataset_id:
+            if not name or not version:
+                raise ValueError(
+                    "name and version must be provided when dataset_id is not provided"
                 )
+            dataset = self.get_dataset(
+                name,
+                namespace_name=namespace_name,
+                project_name=project_name,
+            )
+            dataset_version = dataset.get_version(version)
+            dataset_id = dataset.id
+            dataset_version_id = dataset_version.id
 
-        return direct_dependencies
+        if dataset_version_id is None:
+            raise ValueError(
+                "dataset_version_id must be provided when dataset_id is provided"
+            )
+
+        dependency_ids = self.get_dataset_dependency_ids(
+            dataset_id,
+            dataset_version_id,
+            indirect=indirect,
+        )
+        all_ids = get_all_ids(dependency_ids)
+
+        dependencies = {
+            str(d.id): d
+            for d in self.metastore.get_direct_dataset_dependencies_by_ids(all_ids)
+        }
+        return build_nested_dependencies(dependencies, dependency_ids, indirect)
+
+    def get_dataset_dependency_ids(
+        self,
+        dataset_id: int,
+        version_id: int,
+        indirect=False,
+    ) -> dict[str, dict]:
+        dependency_tree: dict[str, dict] = {}
+        direct_dependencies = self.metastore.get_dataset_dependency_minimal(
+            dataset_id=dataset_id,
+            version_id=version_id,
+        )
+        for d in direct_dependencies:
+            if not indirect:
+                dependency_tree[str(d.id)] = {}
+                continue
+
+            nested_dependencies = d.nested_dependencies
+            if (
+                nested_dependencies is None
+                and indirect
+                and d.dataset_id
+                and d.dataset_version_id
+            ):
+                nested_dependencies = self.get_dataset_dependency_ids(
+                    d.dataset_id,
+                    d.dataset_version_id,
+                    indirect=True,
+                )
+            dependency_tree[str(d.id)] = nested_dependencies or {}
+        return dependency_tree
 
     def ls_datasets(
         self,
--- a/src/datachain/data_storage/metastore.py
+++ b/src/datachain/data_storage/metastore.py
@@ -31,6 +31,7 @@ from datachain.data_storage import JobQueryType, JobStatus
 from datachain.data_storage.serializer import Serializable
 from datachain.dataset import (
     DatasetDependency,
+    DatasetDependencyMinimal,
     DatasetListRecord,
     DatasetListVersion,
     DatasetRecord,
@@ -78,6 +79,7 @@ class AbstractMetastore(ABC, Serializable):
     dataset_list_class: type[DatasetListRecord] = DatasetListRecord
     dataset_list_version_class: type[DatasetListVersion] = DatasetListVersion
     dependency_class: type[DatasetDependency] = DatasetDependency
+    dependency_minimal_class: type[DatasetDependencyMinimal] = DatasetDependencyMinimal
     job_class: type[Job] = Job
     checkpoint_class: type[Checkpoint] = Checkpoint
 
@@ -347,6 +349,7 @@ class AbstractMetastore(ABC, Serializable):
         source_dataset_version: str,
         dep_dataset: "DatasetRecord",
         dep_dataset_version: str,
+        nested_dependencies: dict | None = None,
     ) -> None:
         """Adds dataset dependency to dataset."""
 
@@ -365,6 +368,18 @@ class AbstractMetastore(ABC, Serializable):
         self, dataset: DatasetRecord, version: str
     ) -> list[DatasetDependency | None]:
         """Gets direct dataset dependencies."""
+
+    @abstractmethod
+    def get_dataset_dependency_minimal(
+        self, dataset_id: int, version_id: int
+    ) -> list[DatasetDependencyMinimal]:
+        """Gets dataset dependency minimal."""
+
+    @abstractmethod
+    def get_direct_dataset_dependencies_by_ids(
+        self, ids: list[int]
+    ) -> list[DatasetDependency]:
+        """Gets dataset dependency minimal by ids."""
 
     @abstractmethod
     def remove_dataset_dependencies(
@@ -639,6 +654,14 @@ class AbstractDBMetastore(AbstractMetastore):
         ]
 
     @classmethod
+    def _datasets_dependency_nested(cls) -> Column:
+        return Column(
+            "nested_dependencies",
+            JSON,
+            nullable=True,
+        )
+
+    @classmethod
     def _datasets_dependencies_columns(cls) -> list["SchemaItem"]:
         """Datasets dependencies table columns."""
         return [
@@ -669,6 +692,7 @@ class AbstractDBMetastore(AbstractMetastore):
                 ForeignKey(f"{cls.DATASET_VERSION_TABLE}.id"),
                 nullable=True,
             ),
+            cls._datasets_dependency_nested(),
         ]
 
     #
@@ -1406,6 +1430,7 @@ class AbstractDBMetastore(AbstractMetastore):
         source_dataset_version: str,
         dep_dataset: "DatasetRecord",
         dep_dataset_version: str,
+        nested_dependencies: dict | None = None,
     ) -> None:
         """Adds dataset dependency to dataset."""
         self.db.execute(
@@ -1416,6 +1441,7 @@ class AbstractDBMetastore(AbstractMetastore):
                 ),
                 dataset_id=dep_dataset.id,
                 dataset_version_id=dep_dataset.get_version(dep_dataset_version).id,
+                nested_dependencies=nested_dependencies,
             )
         )
 
@@ -1482,6 +1508,54 @@ class AbstractDBMetastore(AbstractMetastore):
         )
 
         return [self.dependency_class.parse(*r) for r in self.db.execute(query)]
+
+    def get_direct_dataset_dependencies_by_ids(
+        self, ids: list[int]
+    ) -> list[DatasetDependency]:
+        n = self._namespaces
+        p = self._projects
+        d = self._datasets
+        dd = self._datasets_dependencies
+        dv = self._datasets_versions
+
+        select_cols = self._dataset_dependencies_select_columns()
+
+        query = (
+            self._datasets_dependencies_select(*select_cols)
+            .select_from(
+                dd.join(d, dd.c.dataset_id == d.c.id, isouter=True)
+                .join(dv, dd.c.dataset_version_id == dv.c.id, isouter=True)
+                .join(p, d.c.project_id == p.c.id, isouter=True)
+                .join(n, p.c.namespace_id == n.c.id, isouter=True)
+            )
+            .where(dd.c.id.in_(ids))
+        )
+
+        return [
+            dep
+            for dep in [self.dependency_class.parse(*r) for r in self.db.execute(query)]
+            if dep is not None
+        ]
+
+    def get_dataset_dependency_minimal(
+        self, dataset_id: int, version_id: int
+    ) -> list[DatasetDependencyMinimal]:
+        dd = self._datasets_dependencies
+
+        select_cols = self._datasets_dependencies_columns()
+
+        query = self._datasets_dependencies_select(*select_cols).where(
+            (dd.c.source_dataset_id == dataset_id)
+            & (dd.c.source_dataset_version_id == version_id)
+        )
+
+        return [
+            dep
+            for dep in [
+                self.dependency_minimal_class.parse(*r) for r in self.db.execute(query)
+            ]
+            if dep is not None
+        ]
 
     def remove_dataset_dependencies(
         self, dataset: DatasetRecord, version: str | None = None
--- a/src/datachain/data_storage/sqlite.py
+++ b/src/datachain/data_storage/sqlite.py
@@ -18,7 +18,7 @@ from sqlalchemy import (
     select,
 )
 from sqlalchemy.dialects import sqlite
-from sqlalchemy.schema import CreateIndex, CreateTable, DropTable
+from sqlalchemy.schema import CreateColumn, CreateIndex, CreateTable, DropTable
 from sqlalchemy.sql import func
 from sqlalchemy.sql.elements import BinaryExpression, BooleanClauseList
 from sqlalchemy.sql.expression import bindparam, cast
@@ -327,6 +327,20 @@ class SQLiteDatabaseEngine(DatabaseEngine):
     def create_table(self, table: "Table", if_not_exists: bool = True) -> None:
         self.execute(CreateTable(table, if_not_exists=if_not_exists))
 
+    def create_column(
+        self, table: "Table", column: "Column", if_not_exists: bool = True
+    ) -> None:
+        quoted_table_name = quote_schema(table.name)
+        if if_not_exists:
+            res = self.execute_str(
+                f"SELECT 1 FROM pragma_table_info('{table.name}') "  # noqa: S608
+                f" WHERE name = '{column.name}';"
+            )
+            if res.fetchone():
+                return
+        column_str = str(CreateColumn(column))
+        self.execute_str(f"ALTER TABLE {quoted_table_name} ADD COLUMN {column_str}")
+
     def drop_table(self, table: "Table", if_exists: bool = False) -> None:
         self.execute(DropTable(table, if_exists=if_exists))
 
@@ -460,6 +474,12 @@ class SQLiteMetastore(AbstractDBMetastore):
         self.default_table_names.append(self._datasets_dependencies.name)
         self.db.create_table(self._jobs, if_not_exists=True)
         self.default_table_names.append(self._jobs.name)
+        self.db.create_column(
+            self._datasets_dependencies,
+            self._datasets_dependency_nested(),
+        )
+
+        # ADD  metadata for dependencies
         self.db.create_table(self._checkpoints, if_not_exists=True)
         self.default_table_names.append(self._checkpoints.name)
 
--- a/src/datachain/query/dataset.py
+++ b/src/datachain/query/dataset.py
@@ -1859,13 +1859,20 @@ class DatasetQuery:
                         )
             else:
                 dependencies.add((dep_dataset, dep_dataset_version))
-
         for dep_dataset, dep_dataset_version in dependencies:
+            dataset_version = dep_dataset.get_version(dep_dataset_version)
+
+            nested_deps_values = self.catalog.get_dataset_dependency_ids(
+                dep_dataset.id,
+                dataset_version.id,
+                indirect=True,
+            )
             self.catalog.metastore.add_dataset_dependency(
                 dataset,
                 version,
                 dep_dataset,
                 dep_dataset_version,
+                nested_deps_values,
             )
 
     def exec(self) -> "Self":
--- a/src/datachain/dataset.py
+++ b/src/datachain/dataset.py
@@ -20,6 +20,7 @@ LT = TypeVar("LT", bound="DatasetListRecord")
 V = TypeVar("V", bound="DatasetVersion")
 LV = TypeVar("LV", bound="DatasetListVersion")
 DD = TypeVar("DD", bound="DatasetDependency")
+DDM = TypeVar("DDM", bound="DatasetDependencyMinimal")
 
 DATASET_PREFIX = "ds://"
 QUERY_DATASET_PREFIX = "ds_query_"
@@ -92,6 +93,35 @@ def parse_dataset_name(name: str) -> tuple[str | None, str | None, str]:
 class DatasetDependencyType:
     DATASET = "dataset"
     STORAGE = "storage"
+
+
+@dataclass
+class DatasetDependencyMinimal:
+    id: int
+    dataset_id: int | None
+    dataset_version_id: int | None
+    source_dataset_id: int
+    source_dataset_version_id: int | None
+    nested_dependencies: dict | None
+
+    @classmethod
+    def parse(
+        cls: builtins.type[DDM],
+        id: int,
+        source_dataset_id: int,
+        source_dataset_version_id: int | None,
+        dataset_id: int | None,
+        dataset_version_id: int | None,
+        nested_dependencies: str | None,
+    ) -> "DatasetDependencyMinimal | None":
+        return cls(
+            id,
+            dataset_id,
+            dataset_version_id,
+            source_dataset_id,
+            source_dataset_version_id,
+            json.loads(nested_dependencies) if nested_dependencies else None,
+        )
 
 
 @dataclass
@@ -173,6 +203,12 @@ class DatasetDependency:
 
     def __hash__(self):
         return hash(f"{self.type}_{self.name}_{self.version}")
+
+    def set_dependencies(
+        self, dependencies: list["DatasetDependency | None"]
+    ) -> "DatasetDependency":
+        self.dependencies = dependencies
+        return self
 
 
 class DatasetStatus:


amritghimire:
The code used to compare the performance upgrades are: 
```py
import datachain as dc
from datachain.catalog import get_catalog
import pprint


# Multiple storage sources
gs = dc.read_storage("gs://amrit-datachain-test")
az = dc.read_storage("az://amrit-test-az")

# Level 1: Basic image datasets from each storage
gs_jpgs = gs.filter(dc.C('file.path').glob('*.jpg')).save("gs_jpgs")
gs_pngs = gs.filter(dc.C('file.path').glob('*.png')).save("gs_pngs")
az_jpgs = az.filter(dc.C('file.path').glob('*.jpg')).save("az_jpgs")
az_pngs = az.filter(dc.C('file.path').glob('*.png')).save("az_pngs")

# Level 2: Simple cross-storage unions
all_jpgs = gs_jpgs.union(az_jpgs).save("all_jpgs")
all_pngs = gs_pngs.union(az_pngs).save("all_pngs")

# Level 3: Storage + Dataset unions (simple interconnections)
gs_plus_jpgs = gs.union(all_jpgs).save("gs_plus_jpgs")
gs_plus_pngs = gs.union(all_pngs).save("gs_plus_pngs")

# Level 4: All images combined
all_images = all_jpgs.union(all_pngs).save("all_images")

# Level 5: Storage unions with datasets
gs_gs_union = gs.union(gs).union(all_images).save("gs_gs_with_images")
gs_az_union = gs.union(az).union(all_jpgs).save("gs_az_with_jpgs")

# Level 6: Processed versions
large_jpgs = all_jpgs.filter(dc.C('file.size') > 100000).save("large_jpgs")
large_pngs = all_pngs.filter(dc.C('file.size') > 50000).save("large_pngs")

# Level 7: Processed images combined
processed_images = large_jpgs.union(large_pngs).save("processed_images")

# Level 8: Storage + processed datasets
gs_processed = gs.union(processed_images).save("gs_processed")
gs_processed = gs.union(large_jpgs).save("gs_processed")

# Level 9: Mega unions (combining previous levels)
mega_union_1 = gs_gs_union.union(gs_az_union).save("mega_union_1")
mega_union_2 = gs_processed.union(gs_processed).save("mega_union_2")

# Level 10: Final aggregations
final_images = all_images.union(processed_images).save("final_images")
final_storage = gs.union(gs).union(az).union(final_images).save("final_storage")

# Level 11: Ultimate dataset
ultimate_dataset = final_images.union(final_storage).union(mega_union_1).union(mega_union_2).save("ultimate_dataset")

# Level 12: Final result
result = ultimate_dataset.filter(dc.C("file.path").glob("*2024*")).filter(dc.C('file.size') > 1000).limit(10).save("final_result")


catalog = get_catalog()
dataset = catalog.get_dataset(
    "ultimate_dataset",
    namespace_name="local",
    project_name="local",
)
version = dataset.get_version(dataset.latest_version)

print("Getting dependencies...")
print("_"*100)


dependencies = catalog.get_dataset_dependencies(
    name=dataset.name,
    version=version.version,
    namespace_name="local",
    project_name="local",
    indirect=True,
)

pprint.pprint(dependencies)

breakpoint()


```

With this changes, only following sql queries are run now:
```sql
SELECT id, source_dataset_id, source_dataset_version_id, dataset_id, dataset_version_id, nested_dependencies 
FROM datasets_dependencies 
WHERE datasets_dependencies.source_dataset_id = 24 AND datasets_dependencies.source_dataset_version_id = 113

SELECT namespaces.name, projects.name AS name_1, datasets_dependencies.id, datasets_dependencies.dataset_id, datasets_dependencies.dataset_version_id, datasets.name AS name_2, datasets_versions.version, datasets_versions.created_at 
FROM datasets_dependencies LEFT OUTER JOIN datasets ON datasets_dependencies.dataset_id = datasets.id LEFT OUTER JOIN datasets_versions ON datasets_dependencies.dataset_version_id = datasets_versions.id LEFT OUTER JOIN projects ON datasets.project_id = projects.id LEFT OUTER JOIN namespaces ON projects.namespace_id = namespaces.id 
WHERE datasets_dependencies.id IN (249, 250, 251, 252, 253, 254, 255, 256, 262, 263, 264, 265, 266, 267, 268, 269, 270, 271, 272, 273, 274, 275, 276, 277, 284, 285, 286, 287, 288, 289, 290, 291, 292, 293, 294, 295, 296, 297, 298, 299, 300, 301, 302, 303, 304, 305, 306, 307, 308, 309)
```


Instead of recursive fetch for each dataset dependency in current implementation:
```
....
id, datasets_dependencies.dataset_version_id, datasets.name AS name_2, datasets_versions.version, datasets_versions.created_at 
FROM datasets_dependencies LEFT OUTER JOIN datasets ON datasets_dependencies.dataset_id = datasets.id LEFT OUTER JOIN datasets_versions ON datasets_dependencies.dataset_version_id = datasets_versions.id LEFT OUTER JOIN projects ON datasets.project_id = projects.id LEFT OUTER JOIN namespaces ON projects.namespace_id = namespaces.id 
WHERE datasets_dependencies.source_dataset_id = 39 AND datasets_dependencies.source_dataset_version_id = 128
SELECT EXISTS (SELECT 1 
FROM sqlite_master 
WHERE type = 'table' AND name = 'datasets') AS anon_1
SELECT EXISTS (SELECT 1 
FROM sqlite_master 
WHERE type = 'table' AND name = 'datasets_versions') AS anon_1
SELECT namespaces.id, namespaces.uuid, namespaces.name, namespaces.description, namespaces.created_at, projects.id AS id_1, projects.uuid AS uuid_1, projects.name AS name_1, projects.description AS description_1, projects.created_at AS created_at_1, projects.namespace_id, datasets.id AS id_2, datasets.project_id, datasets.name AS name_2, datasets.description AS description_2, datasets.attrs, datasets.status, datasets.feature_schema, datasets.created_at AS created_at_2, datasets.finished_at, datasets.error_message, datasets.error_stack, datasets.script_output, datasets.sources, datasets.query_script, datasets.schema, datasets_versions.id AS id_3, datasets_versions.uuid AS uuid_2, datasets_versions.dataset_id, datasets_versions.version, datasets_versions.status AS status_1, datasets_versions.feature_schema AS feature_schema_1, datasets_versions.created_at AS created_at_3, datasets_versions.finished_at AS finished_at_1, datasets_versions.error_message AS error_message_1, datasets_versions.error_stack AS error_stack_1, datasets_versions.script_output AS script_output_1, datasets_versions.num_objects, datasets_versions.size, datasets_versions.preview, datasets_versions.sources AS sources_1, datasets_versions.query_script AS query_script_1, datasets_versions.schema AS schema_1, datasets_versions.job_id 
FROM namespaces JOIN projects ON namespaces.id = projects.namespace_id JOIN datasets ON projects.id = datasets.project_id LEFT OUTER JOIN datasets_versions ON datasets.id = datasets_versions.dataset_id 
WHERE datasets.name = 'large_pngs' AND namespaces.name = 'local' AND projects.name = 'local'
SELECT namespaces.name, projects.name AS name_1, datasets_dependencies.id, datasets_dependencies.dataset_id, datasets_dependencies.dataset_version_id, datasets.name AS name_2, datasets_versions.version, datasets_versions.created_at 
FROM datasets_dependencies LEFT OUTER JOIN datasets ON datasets_dependencies.dataset_id = datasets.id LEFT OUTER JOIN datasets_versions ON datasets_dependencies.dataset_version_id = datasets_versions.id LEFT OUTER JOIN projects ON datasets.project_id = projects.id LEFT OUTER JOIN namespaces ON projects.namespace_id = namespaces.id 
WHERE datasets_dependencies.source_dataset_id = 38 AND datasets_dependencies.source_dataset_version_id = 127
SELECT EXISTS (SELECT 1 
FROM sqlite_master 
WHERE type = 'table' AND name = 'datasets') AS anon_1
SELECT EXISTS (SELECT 1 
FROM sqlite_master 
WHERE type = 'table' AND name = 'datasets_versions') AS anon_1
SELECT namespaces.id, namespaces.uuid, namespaces.name, namespaces.description, namespaces.created_at, projects.id AS id_1, projects.uuid AS uuid_1, projects.name AS name_1, projects.description AS description_1, projects.created_at AS created_at_1, projects.namespace_id, datasets.id AS id_2, datasets.project_id, datasets.name AS name_2, datasets.description AS description_2, datasets.attrs, datasets.status, datasets.feature_schema, datasets.created_at AS created_at_2, datasets.finished_at, datasets.error_message, datasets.error_stack, datasets.script_output, datasets.sources, datasets.query_script, datasets.schema, datasets_versions.id AS id_3, datasets_versions.uuid AS uuid_2, datasets_versions.dataset_id, datasets_versions.version, datasets_versions.status AS status_1, datasets_versions.feature_schema AS feature_schema_1, datasets_versions.created_at AS created_at_3, datasets_versions.finished_at AS finished_at_1, datasets_versions.error_message AS error_message_1, datasets_versions.error_stack AS error_stack_1, datasets_versions.script_output AS script_output_1, datasets_versions.num_objects, datasets_versions.size, datasets_versions.preview, datasets_versions.sources AS sources_1, datasets_versions.query_script AS query_script_1, datasets_versions.schema AS schema_1, datasets_versions.job_id 
FROM namespaces JOIN projects ON namespaces.id = projects.namespace_id JOIN datasets ON projects.id = datasets.project_id LEFT OUTER JOIN datasets_versions ON datasets.id = datasets_versions.dataset_id 
WHERE datasets.name = 'all_pngs' AND namespaces.name = 'local' AND projects.name = 'local'
SELECT namespaces.name, projects.name AS name_1, datasets_dependencies.id, datasets_dependencies.dataset_id, datasets_dependencies.dataset_version_id, datasets.name AS name_2, datasets_versions.version, datasets_versions.created_at 
FROM datasets_dependencies LEFT OUTER JOIN datasets ON datasets_dependencies.dataset_id = datasets.id LEFT OUTER JOIN datasets_versions ON datasets_dependencies.dataset_version_id = datasets_versions.id LEFT OUTER JOIN projects ON datasets.project_id = projects.id LEFT OUTER JOIN namespaces ON projects.namespace_id = namespaces.id 
WHERE datasets_dependencies.source_dataset_id = 31 AND datasets_dependencies.source_dataset_version_id = 120
SELECT EXISTS (SELECT 1 
FROM sqlite_master 
WHERE type = 'table' AND name = 'datasets') AS anon_1
SELECT EXISTS (SELECT 1 
FROM sqlite_master 
WHERE type = 'table' AND name = 'datasets_versions') AS anon_1
SELECT namespaces.id, namespaces.uuid, namespaces.name, namespaces.description, namespaces.created_at, projects.id AS id_1, projects.uuid AS uuid_1, projects.name AS name_1, projects.description AS description_1, projects.created_at AS created_at_1, projects.namespace_id, datasets.id AS id_2, datasets.project_id, datasets.name AS name_2, datasets.description AS description_2, datasets.attrs, datasets.status, datasets.feature_schema, datasets.created_at AS created_at_2, datasets.finished_at, datasets.error_message, datasets.error_stack, datasets.script_output, datasets.sources, datasets.query_script, datasets.schema, datasets_versions.id AS id_3, datasets_versions.uuid AS uuid_2, datasets_versions.dataset_id, datasets_versions.version, datasets_versions.status AS status_1, datasets_versions.feature_schema AS feature_schema_1, datasets_versions.created_at AS created_at_3, datasets_versions.finished_at AS finished_at_1, datasets_versions.error_message AS error_message_1, datasets_versions.error_stack AS error_stack_1, datasets_versions.script_output AS script_output_1, datasets_versions.num_objects, datasets_versions.size, datasets_versions.preview, datasets_versions.sources AS sources_1, datasets_versions.query_script AS query_script_1, datasets_versions.schema AS schema_1, datasets_versions.job_id 
FROM namespaces JOIN projects ON namespaces.id = projects.namespace_id JOIN datasets ON projects.id = datasets.project_id LEFT OUTER JOIN datasets_versions ON datasets.id = datasets_versions.dataset_id 
WHERE datasets.name = 'az_pngs' AND namespaces.name = 'local' AND projects.name = 'local'
SELECT namespaces.name, projects.name AS name_1, datasets_dependencies.id, datasets_dependencies.dataset_id, datasets_dependencies.dataset_version_id, datasets.name AS name_2, datasets_versions.version, datasets_versions.created_at 
FROM datasets_dependencies LEFT OUTER JOIN datasets ON datasets_dependencies.dataset_id = datasets.id LEFT OUTER JOIN datasets_versions ON datasets_dependencies.dataset_version_id = datasets_versions.id LEFT OUTER JOIN projects ON datasets.project_id = projects.id LEFT OUTER JOIN namespaces ON projects.namespace_id = namespaces.id 
WHERE datasets_dependencies.source_dataset_id = 29 AND datasets_dependencies.source_dataset_version_id = 118
SELECT EXISTS (SELECT 1 
FROM sqlite_master 
WHERE type = 'table' AND name = 'datasets') AS anon_1
SELECT EXISTS (SELECT 1 
FROM sqlite_master 
WHERE type = 'table' AND name = 'datasets_versions') AS anon_1
SELECT namespaces.id, namespaces.uuid, namespaces.name, namespaces.description, namespaces.created_at, projects.id AS id_1, projects.uuid AS uuid_1, projects.name AS name_1, projects.description AS description_1, projects.created_at AS created_at_1, projects.namespace_id, datasets.id AS id_2, datasets.project_id, datasets.name AS name_2, datasets.description AS description_2, datasets.attrs, datasets.status, datasets.feature_schema, datasets.created_at AS created_at_2, datasets.finished_at, datasets.error_message, datasets.error_stack, datasets.script_output, datasets.sources, datasets.query_script, datasets.schema, datasets_versions.id AS id_3, datasets_versions.uuid AS uuid_2, datasets_versions.dataset_id, datasets_versions.version, datasets_versions.status AS status_1, datasets_versions.feature_schema AS feature_schema_1, datasets_versions.created_at AS created_at_3, datasets_versions.finished_at AS finished_at_1, datasets_versions.error_message AS error_message_1, datasets_versions.error_stack AS error_stack_1, datasets_versions.script_output AS script_output_1, datasets_versions.num_objects, datasets_versions.size, datasets_versions.preview, datasets_versions.sources AS sources_1, datasets_versions.query_script AS query_script_1, datasets_versions.schema AS schema_1, datasets_versions.job_id 
FROM namespaces JOIN projects ON namespaces.id = projects.namespace_id JOIN datasets ON projects.id = datasets.project_id LEFT OUTER JOIN datasets_versions ON datasets.id = datasets_versions.dataset_id 
WHERE datasets.name = 'gs_pngs' AND namespaces.name = 'local' AND projects.name = 'local'
SELECT namespaces.name, projects.name AS name_1, datasets_dependencies.id, datasets_dependencies.dataset_id, datasets_dependencies.dataset_version_id, datasets.name AS name_2, datasets_versions.version, datasets_versions.created_at 
FROM datasets_dependencies LEFT OUTER JOIN datasets ON datasets_dependencies.dataset_id = datasets.id LEFT OUTER JOIN datasets_versions ON datasets_dependencies.dataset_version_id = datasets_versions.id LEFT OUTER JOIN projects ON datasets.project_id = projects.id LEFT OUTER JOIN namespaces ON projects.namespace_id = namespaces.id 
WHERE datasets_dependencies.source_dataset_id = 27 AND datasets_dependencies.source_dataset_version_id = 116
SELECT EXISTS (SELECT 1 
FROM sqlite_master 
WHERE type = 'table' AND name = 'datasets') AS anon_1
SELECT EXISTS (SELECT 1 
FROM sqlite_master 
WHERE type = 'table' AND name = 'datasets_versions') AS anon_1
SELECT namespaces.id, namespaces.uuid, namespaces.name, namespaces.description, namespaces.created_at, projects.id AS id_1, projects.uuid AS uuid_1, projects.name AS name_1, projects.description AS description_1, projects.created_at AS created_at_1, projects.namespace_id, datasets.id AS id_2, datasets.project_id, datasets.name AS name_2, datasets.description AS description_2, datasets.attrs, datasets.status, datasets.feature_schema, datasets.created_at AS created_at_2, datasets.finished_at, datasets.error_message, datasets.error_stack, datasets.script_output, datasets.sources, datasets.query_script, datasets.schema, datasets_versions.id AS id_3, datasets_versions.uuid AS uuid_2, datasets_versions.dataset_id, datasets_versions.version, datasets_versions.status AS status_1, datasets_versions.feature_schema AS feature_schema_1, datasets_versions.created_at AS created_at_3, datasets_versions.finished_at AS finished_at_1, datasets_versions.error_message AS error_message_1, datasets_versions.error_stack AS error_stack_1, datasets_versions.script_output AS script_output_1, datasets_versions.num_objects, datasets_versions.size, datasets_versions.preview, datasets_versions.sources AS sources_1, datasets_versions.query_script AS query_script_1, datasets_versions.schema AS schema_1, datasets_versions.job_id 
FROM namespaces JOIN projects ON namespaces.id = projects.namespace_id JOIN datasets ON projects.id = datasets.project_id LEFT OUTER JOIN datasets_versions ON datasets.id = datasets_versions.dataset_id 
WHERE datasets.name = 'large_jpgs' AND namespaces.name = 'local' AND projects.name = 'local'
SELECT namespaces.name, projects.name AS name_1, datasets_dependencies.id, datasets_dependencies.dataset_id, datasets_dependencies.dataset_version_id, datasets.name AS name_2, datasets_versions.version, datasets_versions.created_at 
FROM datasets_dependencies LEFT OUTER JOIN datasets ON datasets_dependencies.dataset_id = datasets.id LEFT OUTER JOIN datasets_versions ON datasets_dependencies.dataset_version_id = datasets_versions.id LEFT OUTER JOIN projects ON datasets.project_id = projects.id LEFT OUTER JOIN namespaces ON projects.namespace_id = namespaces.id 
WHERE datasets_dependencies.source_dataset_id = 37 AND datasets_dependencies.source_dataset_version_id = 126
SELECT EXISTS (SELECT 1 
FROM sqlite_master 
WHERE type = 'table' AND name = 'datasets') AS anon_1
SELECT EXISTS (SELECT 1 
FROM sqlite_master 
WHERE type = 'table' AND name = 'datasets_versions') AS anon_1
SELECT namespaces.id, namespaces.uuid, namespaces.name, namespaces.description, namespaces.created_at, projects.id AS id_1, projects.uuid AS uuid_1, projects.name AS name_1, projects.description AS description_1, projects.created_at AS created_at_1, projects.namespace_id, datasets.id AS id_2, datasets.project_id, datasets.name AS name_2, datasets.description AS description_2, datasets.attrs, datasets.status, datasets.feature_schema, datasets.created_at AS created_at_2, datasets.finished_at, datasets.error_message, datasets.error_stack, datasets.script_output, datasets.sources, datasets.query_script, datasets.schema, datasets_versions.id AS id_3, datasets_versions.uuid AS uuid_2, datasets_versions.dataset_id, datasets_versions.version, datasets_versions.status AS status_1, datasets_versions.feature_schema AS feature_schema_1, datasets_versions.created_at AS created_at_3, datasets_versions.finished_at AS finished_at_1, datasets_versions.error_message AS error_message_1, datasets_versions.error_stack AS error_stack_1, datasets_versions.script_output AS script_output_1, datasets_versions.num_objects, datasets_versions.size, datasets_versions.preview, datasets_versions.sources AS sources_1, datasets_versions.query_script AS query_script_1, datasets_versions.schema AS schema_1, datasets_versions.job_id 
FROM namespaces JOIN projects ON namespaces.id = projects.namespace_id JOIN datasets ON projects.id = datasets.project_id LEFT OUTER JOIN datasets_versions ON datasets.id = datasets_versions.dataset_id 
WHERE datasets.name = 'all_jpgs' AND namespaces.name = 'local' AND projects.name = 'local'
SELECT namespaces.name, projects.name AS name_1, datasets_dependencies.id, datasets_dependencies.dataset_id, datasets_dependencies.dataset_version_id, datasets.name AS name_2, datasets_versions.version, datasets_versions.created_at 
FROM datasets_dependencies LEFT OUTER JOIN datasets ON datasets_dependencies.dataset_id = datasets.id LEFT OUTER JOIN datasets_versions ON datasets_dependencies.dataset_version_id = datasets_versions.id LEFT OUTER JOIN projects ON datasets.project_id = projects.id LEFT OUTER JOIN namespaces ON projects.namespace_id = namespaces.id 
WHERE datasets_dependencies.source_dataset_id = 30 AND datasets_dependencies.source_dataset_version_id = 119
SELECT EXISTS (SELECT 1 
FROM sqlite_master 
WHERE type = 'table' AND name = 'datasets') AS anon_1
SELECT EXISTS (SELECT 1 
FROM sqlite_master 
WHERE type = 'table' AND name = 'datasets_versions') AS anon_1
SELECT namespaces.id, namespaces.uuid, namespaces.name, namespaces.description, namespaces.created_at, projects.id AS id_1, projects.uuid AS uuid_1, projects.name AS name_1, projects.description AS description_1, projects.created_at AS created_at_1, projects.namespace_id, datasets.id AS id_2, datasets.project_id, datasets.name AS name_2, datasets.description AS description_2, datasets.attrs, datasets.status, datasets.feature_schema, datasets.created_at AS created_at_2, datasets.finished_at, datasets.error_message, datasets.error_stack, datasets.script_output, datasets.sources, datasets.query_script, datasets.schema, datasets_versions.id AS id_3, datasets_versions.uuid AS uuid_2, datasets_versions.dataset_id, datasets_versions.version, datasets_versions.status AS status_1, datasets_versions.feature_schema AS feature_schema_1, datasets_versions.created_at AS created_at_3, datasets_versions.finished_at AS finished_at_1, datasets_versions.error_message AS error_message_1, datasets_versions.error_stack AS error_stack_1, datasets_versions.script_output AS script_output_1, datasets_versions.num_objects, datasets_versions.size, datasets_versions.preview, datasets_versions.sources AS sources_1, datasets_versions.query_script AS query_script_1, datasets_versions.schema AS schema_1, datasets_versions.job_id 
FROM namespaces JOIN projects ON namespaces.id = projects.namespace_id JOIN datasets ON projects.id = datasets.project_id LEFT OUTER JOIN datasets_versions ON datasets.id = datasets_versions.dataset_id 
WHERE datasets.name = 'gs_jpgs' AND namespaces.name = 'local' AND projects.name = 'local'
SELECT namespaces.name, projects.name AS name_1, datasets_dependencies.id, datasets_dependencies.dataset_id, datasets_dependencies.dataset_version_id, datasets.name AS name_2, datasets_versions.version, datasets_versions.created_at 
FROM datasets_dependencies LEFT OUTER JOIN datasets ON datasets_dependencies.dataset_id = datasets.id LEFT OUTER JOIN datasets_versions ON datasets_dependencies.dataset_version_id = datasets_versions.id LEFT OUTER JOIN projects ON datasets.project_id = projects.id LEFT OUTER JOIN namespaces ON projects.namespace_id = namespaces.id 
WHERE datasets_dependencies.source_dataset_id = 26 AND datasets_dependencies.source_dataset_version_id = 115
SELECT EXISTS (SELECT 1 
FROM sqlite_master 
WHERE type = 'table' AND name = 'datasets') AS anon_1
SELECT EXISTS (SELECT 1 
FROM sqlite_master 
WHERE type = 'table' AND name = 'datasets_versions') AS anon_1
SELECT namespaces.id, namespaces.uuid, namespaces.name, namespaces.description, namespaces.created_at, projects.id AS id_1, projects.uuid AS uuid_1, projects.name AS name_1, projects.description AS description_1, projects.created_at AS created_at_1, projects.namespace_id, datasets.id AS id_2, datasets.project_id, datasets.name AS name_2, datasets.description AS description_2, datasets.attrs, datasets.status, datasets.feature_schema, datasets.created_at AS created_at_2, datasets.finished_at, datasets.error_message, datasets.error_stack, datasets.script_output, datasets.sources, datasets.query_script, datasets.schema, datasets_versions.id AS id_3, datasets_versions.uuid AS uuid_2, datasets_versions.dataset_id, datasets_versions.version, datasets_versions.status AS status_1, datasets_versions.feature_schema AS feature_schema_1, datasets_versions.created_at AS created_at_3, datasets_versions.finished_at AS finished_at_1, datasets_versions.error_message AS error_message_1, datasets_versions.error_stack AS error_stack_1, datasets_versions.script_output AS script_output_1, datasets_versions.num_objects, datasets_versions.size, datasets_versions.preview, datasets_versions.sources AS sources_1, datasets_versions.query_script AS query_script_1, datasets_versions.schema AS schema_1, datasets_versions.job_id 
FROM namespaces JOIN projects ON namespaces.id = projects.namespace_id JOIN datasets ON projects.id = datasets.project_id LEFT OUTER JOIN datasets_versions ON datasets.id = datasets_versions.dataset_id 
WHERE datasets.name = 'az_jpgs' AND namespaces.name = 'local' AND projects.name = 'local'
SELECT namespaces.name, projects.name AS name_1, datasets_dependencies.id, datasets_dependencies.dataset_id, datasets_dependencies.dataset_version_id, datasets.name AS name_2, datasets_versions.version, datasets_versions.created_at 
FROM datasets_dependencies LEFT OUTER JOIN datasets ON datasets_dependencies.dataset_id = datasets.id LEFT OUTER JOIN datasets_versions ON datasets_dependencies.dataset_version_id = datasets_versions.id LEFT OUTER JOIN projects ON datasets.project_id = projects.id LEFT OUTER JOIN namespaces ON projects.namespace_id = namespaces.id 
WHERE datasets_dependencies.source_dataset_id = 28 AND datasets_dependencies.source_dataset_version_id = 117
SELECT EXISTS (SELECT 1 
FROM sqlite_master 
WHERE type = 'table' AND name = 'datasets') AS anon_1
SELECT EXISTS (SELECT 1 
FROM sqlite_master 
WHERE type = 'table' AND name = 'datasets_versions') AS anon_1
SELECT namespaces.id, namespaces.uuid, namespaces.name, namespaces.description, namespaces.created_at, projects.id AS id_1, projects.uuid AS uuid_1, projects.name AS name_1, projects.description AS description_1, projects.created_at AS created_at_1, projects.namespace_id, datasets.id AS id_2, datasets.project_id, datasets.name AS name_2, datasets.description AS description_2, datasets.attrs, datasets.status, datasets.feature_schema, datasets.created_at AS created_at_2, datasets.finished_at, datasets.error_message, datasets.error_stack, datasets.script_output, datasets.sources, datasets.query_script, datasets.schema, datasets_versions.id AS id_3, datasets_versions.uuid AS uuid_2, datasets_versions.dataset_id, datasets_versions.version, datasets_versions.status AS status_1, datasets_versions.feature_schema AS feature_schema_1, datasets_versions.created_at AS created_at_3, datasets_versions.finished_at AS finished_at_1, datasets_versions.error_message AS error_message_1, datasets_versions.error_stack AS error_stack_1, datasets_versions.script_output AS script_output_1, datasets_versions.num_objects, datasets_versions.size, datasets_versions.preview, datasets_versions.sources AS sources_1, datasets_versions.query_script AS query_script_1, datasets_versions.schema AS schema_1, datasets_versions.job_id 
FROM namespaces JOIN projects ON namespaces.id = projects.namespace_id JOIN datasets ON projects.id = datasets.project_id LEFT OUTER JOIN datasets_versions ON datasets.id = datasets_versions.dataset_id 
WHERE datasets.name = 'all_pngs' AND namespaces.name = 'local' AND projects.name = 'local'
SELECT namespaces.name, projects.name AS name_1, datasets_dependencies.id, datasets_dependencies.dataset_id, datasets_dependencies.dataset_version_id, datasets.name AS name_2, datasets_versions.version, datasets_versions.created_at 
FROM datasets_dependencies LEFT OUTER JOIN datasets ON datasets_dependencies.dataset_id = datasets.id LEFT OUTER JOIN datasets_versions ON datasets_dependencies.dataset_version_id = datasets_versions.id LEFT OUTER JOIN projects ON datasets.project_id = projects.id LEFT OUTER JOIN namespaces ON projects.namespace_id = namespaces.id 
WHERE datasets_dependencies.source_dataset_id = 31 AND datasets_dependencies.source_dataset_version_id = 120
SELECT EXISTS (SELECT 1 
FROM sqlite_master 
WHERE type = 'table' AND name = 'datasets') AS anon_1
SELECT EXISTS (SELECT 1 
FROM sqlite_master 
WHERE type = 'table' AND name = 'datasets_versions') AS anon_1
SELECT namespaces.id, namespaces.uuid, namespaces.name, namespaces.description, namespaces.created_at, projects.id AS id_1, projects.uuid AS uuid_1, projects.name AS name_1, projects.description AS description_1, projects.created_at AS created_at_1, projects.namespace_id, datasets.id AS id_2, datasets.project_id, datasets.name AS name_2, datasets.description AS description_2, datasets.attrs, datasets.status, datasets.feature_schema, datasets.created_at AS created_at_2, datasets.finished_at, datasets.error_message, datasets.error_stack, datasets.script_output, datasets.sources, datasets.query_script, datasets.schema, datasets_versions.id AS id_3, datasets_versions.uuid AS uuid_2, datasets_versions.dataset_id, datasets_versions.version, datasets_versions.status AS status_1, datasets_versions.feature_schema AS feature_schema_1, datasets_versions.created_at AS created_at_3, datasets_versions.finished_at AS finished_at_1, datasets_versions.error_message AS error_message_1, datasets_versions.error_stack AS error_stack_1, datasets_versions.script_output AS script_output_1, datasets_versions.num_objects, datasets_versions.size, datasets_versions.preview, datasets_versions.sources AS sources_1, datasets_versions.query_script AS query_script_1, datasets_versions.schema AS schema_1, datasets_versions.job_id 
FROM namespaces JOIN projects ON namespaces.id = projects.namespace_id JOIN datasets ON projects.id = datasets.project_id LEFT OUTER JOIN datasets_versions ON datasets.id = datasets_versions.dataset_id 
WHERE datasets.name = 'az_pngs' AND namespaces.name = 'local' AND projects.name = 'local'
SELECT namespaces.name, projects.name AS name_1, datasets_dependencies.id, datasets_dependencies.dataset_id, datasets_dependencies.dataset_version_id, datasets.name AS name_2, datasets_versions.version, datasets_versions.created_at 
FROM datasets_dependencies LEFT OUTER JOIN datasets ON datasets_dependencies.dataset_id = datasets.id LEFT OUTER JOIN datasets_versions ON datasets_dependencies.dataset_version_id = datasets_versions.id LEFT OUTER JOIN projects ON datasets.project_id = projects.id LEFT OUTER JOIN namespaces ON projects.namespace_id = namespaces.id 
WHERE datasets_dependencies.source_dataset_id = 29 AND datasets_dependencies.source_dataset_version_id = 118
SELECT EXISTS (SELECT 1 
FROM sqlite_master 
WHERE type = 'table' AND name = 'datasets') AS anon_1
SELECT EXISTS (SELECT 1 
FROM sqlite_master 
WHERE type = 'table' AND name = 'datasets_versions') AS anon_1
SELECT namespaces.id, namespaces.uuid, namespaces.name, namespaces.description, namespaces.created_at, projects.id AS id_1, projects.uuid AS uuid_1, projects.name AS name_1, projects.description AS description_1, projects.created_at AS created_at_1, projects.namespace_id, datasets.id AS id_2, datasets.project_id, datasets.name AS name_2, datasets.description AS description_2, datasets.attrs, datasets.status, datasets.feature_schema, datasets.created_at AS created_at_2, datasets.finished_at, datasets.error_message, datasets.error_stack, datasets.script_output, datasets.sources, datasets.query_script, datasets.schema, datasets_versions.id AS id_3, datasets_versions.uuid AS uuid_2, datasets_versions.dataset_id, datasets_versions.version, datasets_versions.status AS status_1, datasets_versions.feature_schema AS feature_schema_1, datasets_versions.created_at AS created_at_3, datasets_versions.finished_at AS finished_at_1, datasets_versions.error_message AS error_message_1, datasets_versions.error_stack AS error_stack_1, datasets_versions.script_output AS script_output_1, datasets_versions.num_objects, datasets_versions.size, datasets_versions.preview, datasets_versions.sources AS sources_1, datasets_versions.query_script AS query_script_1, datasets_versions.schema AS schema_1, datasets_versions.job_id 
FROM namespaces JOIN projects ON namespaces.id = projects.namespace_id JOIN datasets ON projects.id = datasets.project_id LEFT OUTER JOIN datasets_versions ON datasets.id = datasets_versions.dataset_id 
WHERE datasets.name = 'gs_pngs' AND namespaces.name = 'local' AND projects.name = 'local'
SELECT namespaces.name, projects.name AS name_1, datasets_dependencies.id, datasets_dependencies.dataset_id, datasets_dependencies.dataset_version_id, datasets.name AS name_2, datasets_versions.version, datasets_versions.created_at 
FROM datasets_dependencies LEFT OUTER JOIN datasets ON datasets_dependencies.dataset_id = datasets.id LEFT OUTER JOIN datasets_versions ON datasets_dependencies.dataset_version_id = datasets_versions.id LEFT OUTER JOIN projects ON datasets.project_id = projects.id LEFT OUTER JOIN namespaces ON projects.namespace_id = namespaces.id 
WHERE datasets_dependencies.source_dataset_id = 27 AND datasets_dependencies.source_dataset_version_id = 116
```

amritghimire:
Also verified, that the dependency calculation works as expected for old database schema in existence.

src/datachain/data_storage/sqlite.py
```diff
@@ -327,6 +327,20 @@ def table_names(self) -> list[str]:
     def create_table(self, table: "Table", if_not_exists: bool = True) -> None:
         self.execute(CreateTable(table, if_not_exists=if_not_exists))
 
+    def create_column(
+        self, table: "Table", column: "Column", if_not_exists: bool = True
+    ) -> None:
+        quoted_table_name = quote_schema(table.name)
+        if if_not_exists:
+            res = self.execute_str(
+                f"SELECT 1 FROM pragma_table_info('{table.name}') "  # noqa: S608
```
shcheklein: let's find a proper way to do this instead of silencing it

src/datachain/data_storage/sqlite.py
```diff
@@ -327,6 +327,20 @@ def table_names(self) -> list[str]:
...
+                f" WHERE name = '{column.name}';"
+            )
+            if res.fetchone():
+                return
+        column_str = str(CreateColumn(column))
+        self.execute_str(f"ALTER TABLE {quoted_table_name} ADD COLUMN {column_str}")
```
shcheklein: same here - is it safe?

amritghimire: Yes, IMO. One reason is we don't pass user input here, we are calling this from our code with known and defined columns during initialization of the table in sqlite only (local database).

shcheklein: but method looks general, let's secure it please

tests/func/test_dataset_query.py
```diff
@@ -1051,7 +1051,7 @@ def test_dataset_dependencies_one_dataset_as_dependency(
 
     catalog.remove_dataset(dogs_dataset.name, force=True)
     # None means dependency was there but was removed in the meantime
-    assert catalog.get_dataset_dependencies(ds_name, "1.0.0") == [None]
+    assert catalog.get_dataset_dependencies(ds_name, "1.0.0") == []
```
shcheklein: why do we change semantics here?

src/datachain/catalog/catalog.py
```diff
@@ -509,6 +509,44 @@ def clone_catalog_with_cache(catalog: "Catalog", cache: "Cache") -> "Catalog":
...
+def get_all_ids(d):
+    ids = set()
+    for key, value in d.items():
+        ids.add(int(key))
+        if isinstance(value, dict):
+            ids.update(get_all_ids(value))
+    return sorted(ids)
+
+
+def build_nested_dependencies(dataset_deps, dependency_structure, indirect=True):
```
shcheklein: what and why does it actually do - put one-two sentence doc string,

src/datachain/catalog/catalog.py
```diff
@@ -1198,40 +1236,77 @@ def get_remote_dataset(
 
     def get_dataset_dependencies(
         self,
-        name: str,
-        version: str,
+        name: str | None = None,
```
shcheklein: why this change? AI?

src/datachain/catalog/catalog.py
```diff
@@ -1198,40 +1236,77 @@ def get_remote_dataset(
...
+        version: str | None = None,
         namespace_name: str | None = None,
         project_name: str | None = None,
         indirect=False,
+        dataset_id: int | None = None,
```
shcheklein: why do we need this API change?

amritghimire: If we already have dataset id with us, we avoid an extra API for Studio (API metastore), one extra database call to get the dataset. 
This is helpful for optimizing in studio.

shcheklein (commented):
It's quite dense PR, please put some PR-comments around to better understand and help to review. Thanks!

src/datachain/catalog/catalog.py
```diff
@@ -509,6 +509,60 @@ def clone_catalog_with_cache(catalog: "Catalog", cache: "Cache") -> "Catalog":
...
 
+def get_all_ids(d: dict[str, Any]) -> list[int]:
+    """
+    Extracts all unique dataset IDs from a nested dictionary structure.
+    """
+    ids = set()
+    for key, value in d.items():
+        ids.add(int(key))
+        if isinstance(value, dict):
+            ids.update(get_all_ids(value))
```
amritghimire: C: This just goes through the dependency dictionary to get the flat list of the ids. See test_dependency for example.

src/datachain/catalog/catalog.py
```diff
@@ -509,6 +509,60 @@ def clone_catalog_with_cache(catalog: "Catalog", cache: "Cache") -> "Catalog":
...
+                if child_id in dataset_deps:
+                    # Recursively build child dependencies if indirect=True
+                    child_deps = (
+                        build_deps(child_id, deps_dict[dep_id]) if indirect else []
+                    )
+                    child_dep = dataset_deps[child_id]
+                    child_dep.dependencies = child_deps
+                    deps.append(child_dep)
+                else:
+                    deps.append(None)
```
amritghimire: C: Case where the dependency is removed from database, it used to exists but not anymore.

src/datachain/catalog/catalog.py
```diff
@@ -509,6 +509,60 @@ def clone_catalog_with_cache(catalog: "Catalog", cache: "Cache") -> "Catalog":
...
+
+    result: list[DatasetDependency | None] = []
+    for root_id in dependency_structure:
+        if root_id in dataset_deps:
+            deps = build_deps(root_id, dependency_structure)
+            root_dep = dataset_deps[root_id]
+            root_dep.dependencies = deps
+            result.append(root_dep)
+        else:
+            result.append(None)
```
amritghimire: C: Case when the root dependency was removed from the database.

src/datachain/catalog/catalog.py
```diff
@@ -509,6 +509,60 @@ def clone_catalog_with_cache(catalog: "Catalog", cache: "Cache") -> "Catalog":
...
+                else:
+                    deps.append(None)
+        return deps
+
+    result: list[DatasetDependency | None] = []
+    for root_id in dependency_structure:
+        if root_id in dataset_deps:
+            deps = build_deps(root_id, dependency_structure)
+            root_dep = dataset_deps[root_id]
+            root_dep.dependencies = deps
```
amritghimire: C: This is the main purpose of this function. This fills the dependencies in nested manner for the dependency structure. See test_dependency.py for example.

src/datachain/catalog/catalog.py
```diff
@@ -509,6 +509,60 @@ def clone_catalog_with_cache(catalog: "Catalog", cache: "Cache") -> "Catalog":
...
+        if dep_id not in dataset_deps:
+            return []
+
+        deps = []
+        if dep_id in deps_dict:
+            for child_id in deps_dict[dep_id]:
+                if child_id in dataset_deps:
+                    # Recursively build child dependencies if indirect=True
+                    child_deps = (
+                        build_deps(child_id, deps_dict[dep_id]) if indirect else []
```
amritghimire: C: Get the dependency from our dependency cache ( dataset_deps ) apply the nested dependency recursively and add it to the list.

src/datachain/catalog/catalog.py
```diff
@@ -1198,40 +1252,77 @@ def get_remote_dataset(
...
+                namespace_name=namespace_name,
+                project_name=project_name,
+            )
+            dataset_version = dataset.get_version(version)
+            dataset_id = dataset.id
+            dataset_version_id = dataset_version.id
+
+        if dataset_version_id is None:
+            raise ValueError(
+                "dataset_version_id must be provided when dataset_id is provided"
```
amritghimire: C: Basic validation for arguments.

src/datachain/catalog/catalog.py
```diff
@@ -1198,40 +1252,77 @@ def get_remote_dataset(
...
+            )
+
+        dependency_ids = self.get_dataset_dependency_ids(
```
amritghimire: C: We first get the dependency ids only from the metadata we store.

src/datachain/catalog/catalog.py
```diff
@@ -1198,40 +1252,77 @@ def get_remote_dataset(
...
+            dataset_id,
+            dataset_version_id,
+            indirect=indirect,
         )
+        all_ids = get_all_ids(dependency_ids)
```
amritghimire: C: Then from the nested structure, we flatten to get the list of ids to fetch in one batch.

src/datachain/catalog/catalog.py
```diff
@@ -1198,40 +1252,77 @@ def get_remote_dataset(
...
 
-        if not indirect:
-            return direct_dependencies
+        dependencies = {
+            str(d.id): d
+            for d in self.metastore.get_direct_dataset_dependencies_by_ids(all_ids)
```
amritghimire, line 1215: C: Before the end, we get all the dependency from the database in one batch using `in` filter.

src/datachain/catalog/catalog.py
```diff
@@ -1198,40 +1252,77 @@ def get_remote_dataset(
...
+        }
+        return build_nested_dependencies(dependencies, dependency_ids, indirect)
```
amritghimire: C: Finally build the dependency in nested manner and return to the user.

src/datachain/catalog/catalog.py
```diff
@@ -1198,40 +1252,77 @@ def get_remote_dataset(
...
-                    d.name,
-                    d.version,
-                    namespace_name=d.namespace,
-                    project_name=d.project,
-                    indirect=indirect,
-                )
 
-        return direct_dependencies
+            nested_dependencies = d.nested_dependencies
+            if nested_dependencies is None and d.dataset_id and d.dataset_version_id:
```
amritghimire, line 1259: C: This is for fallback. If nested_dependencies is None, that means that is the old schema. We will calculate it recursively for such case.

src/datachain/catalog/catalog.py
```diff
@@ -1198,40 +1252,77 @@ def get_remote_dataset(
...
+        return build_nested_dependencies(dependencies, dependency_ids, indirect)
 
+    def get_dataset_dependency_ids(
+        self,
+        dataset_id: int,
+        version_id: int,
+        indirect=False,
+    ) -> dict[str, dict]:
+        dependency_tree: dict[str, dict] = {}
+        direct_dependencies = self.metastore.get_dataset_dependency_minimal(
```
amritghimire: C: This minimal version so that we don't need unnecessary join here.

src/datachain/data_storage/sqlite.py
```diff
@@ -327,6 +327,20 @@ def table_names(self) -> list[str]:
     def create_table(self, table: "Table", if_not_exists: bool = True) -> None:
         self.execute(CreateTable(table, if_not_exists=if_not_exists))
 
+    def create_column(
```
amritghimire: This is for migration purpose where we may need to add column if it doesn't exist.

src/datachain/query/dataset.py
```diff
@@ -1859,13 +1859,20 @@ def _add_dependencies(self, dataset: "DatasetRecord", version: str):
                         )
             else:
                 dependencies.add((dep_dataset, dep_dataset_version))
-
         for dep_dataset, dep_dataset_version in dependencies:
+            dataset_version = dep_dataset.get_version(dep_dataset_version)
+
+            nested_deps_values = self.catalog.get_dataset_dependency_ids(
```
amritghimire: C: Get the nested dependency structure and save it to database.

src/datachain/data_storage/sqlite.py
amritghimire: Updated.

src/datachain/catalog/catalog.py
```diff
@@ -509,6 +509,60 @@ def clone_catalog_with_cache(catalog: "Catalog", cache: "Cache") -> "Catalog":
...
+    """
+    ids = set()
+    for key, value in d.items():
+        ids.add(int(key))
+        if isinstance(value, dict):
+            ids.update(get_all_ids(value))
+    return sorted(ids)
+
+
+def build_nested_dependencies(
```
shcheklein: is it replacing something? how did we do this on the Studio side before/

src/datachain/catalog/catalog.py
```diff
@@ -509,6 +509,60 @@ def clone_catalog_with_cache(catalog: "Catalog", cache: "Cache") -> "Catalog":
...
+    dataset_deps: dict[str, DatasetDependency],
+    dependency_structure: dict[str, dict],
+    indirect: bool = True,
```
shcheklein: do we use it in both modes? why?

src/datachain/catalog/catalog.py
amritghimire: Yes and No, previously, what we did was we recursively call the same function `self.get_dataset_dependencies` recursively, which in turn fetch the direct dependencies first and then go through each dependency to repeat the process. In each call, it used to fetch the dataset id first, then version, then finally the joined sql table. You can imagine how inefficient that was.

What we now do is figure out all the dataset dependencies we need to fetch for a case (using metadata). Then fetch all those dataset dependency in one batch, and build the tree here i.e. just match the id in the tree to the dataset dependency fetched from database. 


Studio side doesn't do anything more, it just calls `get_dataset_dependencies` and return this to user with some changes like fetching user avatars. 
This part is not changed it is as it is.

src/datachain/catalog/catalog.py
amritghimire: This is for convenience, it fetches one level and return if indirect is false. I will add a test to verify this further and make changes if needed.

src/datachain/catalog/catalog.py
shcheklein: can we also check if there are some libs that build graph from a connections matrix (that what is happening here, right?)

amritghimire: I am not sure what you mean by connections matrix tbh. But why do we want to introduce a dependency to do something this simple? 
I mean we are just recursively attaching the dependencies based on the tree we already have. For example if we have: 

```
{"306": {"297": {"255": {"250": {}}, "256": {"252": {}}}}}
```

We will get the dependency we have cached in the variable to form a tree:
`<Dependency (306)> .dependencies = [<Dependency(297)...`. and so on.

This is verified and tested in test_dependency.

src/datachain/catalog/catalog.py
```diff
@@ -509,6 +509,60 @@ def clone_catalog_with_cache(catalog: "Catalog", cache: "Cache") -> "Catalog":
     return clone
 
 
+def get_all_ids(d: dict[str, Any]) -> list[int]:
+    """
+    Extracts all unique dataset IDs from a nested dictionary structure.
+    """
+    ids = set()
+    for key, value in d.items():
+        ids.add(int(key))
```
dreadatour: This confuses me a bit. `dict` keys are strings, and we converting them to integers here: `ids.add(int(key))`.

Below in `get_dataset_dependency_ids` we are getting dataset dependencies from metastore and converting IDs from integers to strings: `dependency_tree[str(d.id)] = nested_dependencies`.

This looks like a bad design to me, to be honest 😢 Will it be better to always treat IDs as integers?

src/datachain/catalog/catalog.py
```diff
@@ -509,6 +509,60 @@ def clone_catalog_with_cache(catalog: "Catalog", cache: "Cache") -> "Catalog":
     return clone
 
 
+def get_all_ids(d: dict[str, Any]) -> list[int]:
```
dreadatour: `catalog.py` already looks overloaded, have you considered to move these methods to other file? Like `dependencies.py` or something like that. Not like in Java-style, but still, to make it a bit cleaner 🤔

src/datachain/catalog/catalog.py
```diff
@@ -509,6 +509,60 @@ def clone_catalog_with_cache(catalog: "Catalog", cache: "Cache") -> "Catalog":
...
+    for key, value in d.items():
+        ids.add(int(key))
+        if isinstance(value, dict):
+            ids.update(get_all_ids(value))
+    return sorted(ids)
+
+
+def build_nested_dependencies(
+    dataset_deps: dict[str, DatasetDependency],
+    dependency_structure: dict[str, dict],
```
dreadatour: I do have the same question, as above, why do we use string types as dataset IDs? 😢

dreadatour (commented):
My comments so far, keep reviewing.

src/datachain/catalog/catalog.py
amritghimire: For some reason, when I tried to save the json in sqlite, it always converted the ids to string. I will try to see if there is possible way to have int itself .

src/datachain/catalog/catalog.py
```diff
@@ -1198,40 +1252,77 @@ def get_remote_dataset(
...
+                "Provide either (name and version),(dataset_id and dataset_version_id)"
+            )
 
-        direct_dependencies = self.metastore.get_direct_dataset_dependencies(
-            dataset, version
+        if not dataset_id:
+            if not name or not version:
+                raise ValueError(
+                    "name and version must be provided when dataset_id is not provided"
+                )
```
dreadatour: Just wonder if it will be a good idea to split this function into two (one call the other), for simpler code, less checks, better function signature and typings? 🤔

src/datachain/catalog/catalog.py
```diff
@@ -1198,40 +1252,77 @@ def get_remote_dataset(
...
-        name: str,
-        version: str,
+        name: str | None = None,
+        version: str | None = None,
         namespace_name: str | None = None,
         project_name: str | None = None,
         indirect=False,
+        dataset_id: int | None = None,
+        dataset_version_id: int | None = None,
     ) -> list[DatasetDependency | None]:
```
dreadatour, line 1205: Not directly related to this PR, but I am still not sure if this is a good return type: `list[DatasetDependency | None]` 😢 Does it make any sense to get Nones in the list of dependencies.

src/datachain/catalog/catalog.py
```diff
@@ -1198,40 +1252,77 @@ def get_remote_dataset(
...
+            for d in self.metastore.get_direct_dataset_dependencies_by_ids(all_ids)
+        }
+        return build_nested_dependencies(dependencies, dependency_ids, indirect)
 
+    def get_dataset_dependency_ids(
+        self,
+        dataset_id: int,
+        version_id: int,
+        indirect=False,
+    ) -> dict[str, dict]:
```
dreadatour: Method name and return type are not correlated. `get_dataset_dependency_ids_tree` or something like that? 🤔

src/datachain/catalog/catalog.py
```diff
@@ -1198,40 +1252,77 @@ def get_remote_dataset(
...
-                )
 
-        return direct_dependencies
+            nested_dependencies = d.nested_dependencies
+            if nested_dependencies is None and d.dataset_id and d.dataset_version_id:
+                nested_dependencies = self.get_dataset_dependency_ids(
+                    d.dataset_id,
+                    d.dataset_version_id,
+                    indirect=True,
+                )
```
dreadatour: Recursive call? Do we have any safeguard against infinite recursion?

src/datachain/data_storage/metastore.py
```diff
@@ -366,6 +369,18 @@ def get_direct_dataset_dependencies(
     ) -> list[DatasetDependency | None]:
         """Gets direct dataset dependencies."""
 
+    @abstractmethod
+    def get_dataset_dependency_minimal(
+        self, dataset_id: int, version_id: int
+    ) -> list[DatasetDependencyMinimal]:
+        """Gets dataset dependency minimal."""
```
dreadatour: What does this means — "minimal"? 🤔

src/datachain/catalog/catalog.py
```diff
@@ -1198,40 +1252,77 @@ def get_remote_dataset(
...
+            dependency_tree[str(d.id)] = nested_dependencies or {}
```
dreadatour, line 1265: Just wonder, should we use `None` (`null` in JSON) as the case for no dependencies instead of empty dict? 🤔

src/datachain/data_storage/sqlite.py
```diff
@@ -460,6 +476,12 @@ def _init_tables(self) -> None:
         self.default_table_names.append(self._datasets_dependencies.name)
         self.db.create_table(self._jobs, if_not_exists=True)
         self.default_table_names.append(self._jobs.name)
+        self.db.create_column(
+            self._datasets_dependencies,
+            self._datasets_dependency_nested(),
+        )
```
dreadatour: Is this a migration?
This will run the query on every run, is this ok for us?

dreadatour (commented):
I would love to avoid calculating and storing data in additional JSON column, we still can fetch all the dependencies in one SQL query from existing columns and tables using recursive CTEs.

dreadatour (approved):
I am approving this to do not block anyone, although me personally I am strongly believe there is no need to bring an additional complexity and code, since we could use recursive SQL query to fetch all the dependencies on the existing dataset dependencies table 👀

Separate thing that bothers me a lot is usage of JSON field, strings to store numerical IDs and multiple type conversions back and forth 😢

src/datachain/catalog/catalog.py
ilongin: This is strange name for a function, I would rename it to make more sense.

src/datachain/catalog/catalog.py
ilongin: This function signature looks very confusing to me. So we have inputs as `dict[str, DatasetDependency]` and then it returns `list[DatasetDependency]` ... even when reading your function comment I don't understand what are the inputs and what are the outputs....maybe it's just me, idk.

src/datachain/data_storage/sqlite.py
ilongin: The same question .. this function was meant to be used only for creating tables, not custom things like this. Why is this even needed in the first place?

ilongin: Now I see this indeed is some kind of migration. Note that we don't add any kind of migration code to datachain ever. We did add some kind of check to see if schema version is up to date and if you are changing schema increase `SCHEMA_VERSION` to +1 ...if user has old schema he will receive error and he needs to then remove `.datachain/db`.

If we want to add real robust DB migrations, that needs to be done with proper tools like alembic.

src/datachain/catalog/catalog.py
```diff
@@ -1198,40 +1252,77 @@ def get_remote_dataset(
 
     def get_dataset_dependencies(
```
ilongin: I don't like that this function now accepts name, version, id, version id  - all optional. Couldn't this be reduced to one of those.. e.g to accept only IDs?
Also note that we probably can now remove `dataset_id` completely from dependencies tables as that one was added before when we could create dataset without dataset versions (shadow datasets or whatever we call them). Now that's not needed so maybe we should've cleaned up that before refactoring this (would've been easier here then)

src/datachain/catalog/catalog.py
ilongin: Note that `get_dataset()` call is probably not needed - we can create query with just name + version without IDs in `metastore.get_dataset_dependencies()`. Also note that getting version in current code is from memory (no DB query).

BTW have you tried to use recursive SQL query (e.g `WITH RECURSIVE all_dependencies AS ....`) as it seems like a perfect fit for the problem? Wondering if that wold be enough to boost performance and it wouldn't require all this complicated code changes and hacks with additional `nested_dependencies` column (I assume this holds `ID`s of all dependencies in a tree below or?)

src/datachain/data_storage/metastore.py
```diff
@@ -638,6 +653,14 @@ def _dataset_list_version_fields(self) -> list[str]:
             in self.dataset_list_version_class.__dataclass_fields__
         ]
 
+    @classmethod
+    def _datasets_dependency_nested(cls) -> Column:
+        return Column(
+            "nested_dependencies",
```
ilongin, line 659: What do we save in this column?

src/datachain/catalog/catalog.py
shcheklein: question is - do we really need it?

src/datachain/catalog/catalog.py
shcheklein: yep, if it is possible to do with a single query - definitely let's try and see if that is enough

src/datachain/data_storage/metastore.py
```diff
@@ -1483,6 +1509,54 @@ def get_direct_dataset_dependencies(
...
+        d = self._datasets
+        dd = self._datasets_dependencies
+        dv = self._datasets_versions
+
+        select_cols = self._dataset_dependencies_select_columns()
+
+        query = (
+            self._datasets_dependencies_select(*select_cols)
+            .select_from(
+                dd.join(d, dd.c.dataset_id == d.c.id, isouter=True)
```
shcheklein, line 1526: why do we do outer joins?

src/datachain/catalog/catalog.py
amritghimire: Tried to use the suggestions for single query today, but it only make it more complicated. Added my findings at https://github.com/iterative/studio/pull/12227#issuecomment-3387010860 . The complication will exist in two place, one to build heirarchy structure and one in query itself if we go through that route and also, the database must relay more data transfer that way.

shcheklein: thanks @amritghimire . Q: where / how do we now get all recursive dependencies before we can build the tree and save it?

shcheklein: By now I mean in this PR ...

src/datachain/catalog/catalog.py
amritghimire: When saving JSON data to SQLite, integers used as keys within a JSON object are typically converted to strings.

shcheklein: because keys must be strings in JSON? should we use a different format to save ids then? should we convert them back on read?

amritghimire: We do convert them back on read, cast this to int. So, should be good IMO.

src/datachain/catalog/catalog.py
amritghimire: Splitted.

src/datachain/catalog/catalog.py
amritghimire: I assume this was by design to let user know that dependency has been removed.

src/datachain/catalog/catalog.py
shcheklein: So, why do we have str then if we convert them back?

src/datachain/catalog/catalog.py
amritghimire: Right now, we can get all recursive dependency tree from the database itself. I mean use the json we have for the next depth to form the new depth and so on.

shcheklein: How / when do we build JSON? do we need the tree to build it in the first place?

src/datachain/catalog/catalog.py
amritghimire: Renamed it.

src/datachain/catalog/catalog.py
amritghimire: Moved it to separate file.

src/datachain/catalog/catalog.py
amritghimire: Updated.

src/datachain/catalog/catalog.py
amritghimire: Added safeguard for infinite recursion and circular recursion.

src/datachain/catalog/catalog.py
amritghimire: Kept none or null as default for non migrated ones where we will use fallback.

src/datachain/catalog/catalog.py
amritghimire: Updated the comments.

src/datachain/catalog/catalog.py
amritghimire: Updated it to two separate functions.

src/datachain/data_storage/metastore.py
amritghimire: We save the nested structure in this column: 
```
{"306": {"297": {"255": {"250": {}}, "256": {"252": {}}}}}
```

src/datachain/data_storage/metastore.py
amritghimire: So that we will be able to show if there are any rows with dataset dependencies removed as None value.

src/datachain/data_storage/sqlite.py
amritghimire: Yes, it simply checks if the column exists similar as what we do when creating table locally in sqlite one only.

src/datachain/catalog/catalog.py
amritghimire: When a dataset is saved with dependencies, the system calls _get_dataset_dependency_ids_tree() to build the nested dependency structure. The _add_dependencies() method in DatasetQuery builds the JSON tree before storing it in the database
The process is as: 
- Get immediate dependencies from the database
- Use the nested dependency field from it to form new json. 
- Save it to the database.
- If there is no dependency it will be saved as `{}`.

So, here the JSON acts as a pre-computed dependency tree.

For example: 
```
{"dep_id": {"child_id": {"grandchild_id": {}}}}
```

when `grandchild_id` is formed, it has no dependencies so it is `{}`
when `child_id` is formed, it sees `grandchild_id` and use it to form new structure: `{"grandchild_id": {}}`
Similarly, dep_id forms the following: 
`{"dep_id": {"child_id": {"grandchild_id": {}}}}`

src/datachain/catalog/catalog.py
amritghimire: Sorry my bad, we cast int to str instead when forming the dependency cache as:
```
        dependencies = {
            str(d.id): d
            for d in self.metastore.get_direct_dataset_dependencies_by_ids(all_ids)
        }
```

So that we compare using same type and it works as expected.

src/datachain/catalog/catalog.py
shcheklein: > the system calls _get_dataset_dependency_ids_tree() to build the nested dependency structure. 

does it mean we still do a lot of call potentially to DB?

can this part be replaced with recursive DB call?

shcheklein: should the method then called differently?

what do we use if for? to expand the tree that was saved before?

amritghimire: No, just one. When saving `grandchild_id`, it just tries to get any dataset dependency , it doesn't find any. 

When saving `child_id` , it gets the `grandchild_id` and uses its json value to form new one. 
When saving `dep_id` , it only gets `child_id` one, uses it to form final one and so on.

shcheklein: > When saving child_id , it gets the grandchild_id and uses its json value to form new one.

Ah, yeah, this is quite handy 👍

amritghimire: We use it to map the dependency. Let me try to clarify once more. We get values like `{"306": {"297": {"255": {"250": {}}, "256": {"252": {}}}}}` from the database metadata we have saved once. 

We then get the list of dataset dependency from the ids before. it will be in flat heirarchy as every sql query returns. 
{"306": <Dataset dependency>, "297": <Dataset dependency>, "255": <Dataset dependency >} and so on.

This function in simple term replace the id with dataset dependency in the tree above. 

i.e. <Dataset dependency (306)> will be the first item. Its dependencies field goes to the list of child and so on. 

I hope it makes it clear.

src/datachain/data_storage/sqlite.py
ilongin: But we should not do this. We decided not to put any kind of DB migration things except pure table creation. If we start to do this, we will end up with bunch of custom migration stuff that is not maintanable. I suggest to remove it and just increase `SCHEMA_VERSION` on top of the file which will indicate user to remove `.datachain/db` completely and re-create it again

amritghimire: I will leave this to product decision, I am okay with either, but with this in place, user don't need to delete any data in process. It provides smooth migration. 
If we are okay with asking user to remove database, then I can remove this part too.

amritghimire: cc. @shcheklein

src/datachain/catalog/catalog.py
ilongin: So to comment your experiment with recursive call:
1. Issue with more result rows is because you used `union_all` instead just `union` which caused for you to visit shared dependency (the one that is dependency to multiple datasets) multiple times (this is causing exponential growth I think). 
2. I quickly used Claude to generate metastore function to get dataset dependencies with recursion and here is the result (note that I didn't put too much effort into refactoring it so it can probably be much cleaner):
<details>
  <summary>Click to expand</summary>
  
  ```python
    
  def get_dataset_dependencies(
        self,
        dataset: DatasetRecord,
        version: str,
        indirect: bool = False,
    ) -> list[DatasetDependency | None]:
        if not indirect:
            return self.get_direct_dataset_dependencies(dataset, version)

        n = self._namespaces
        p = self._projects
        d = self._datasets
        dd = self._datasets_dependencies
        dv = self._datasets_versions

        dataset_version = dataset.get_version(version)

        # Common dependency fields for CTE
        dep_fields = [
            dd.c.id,
            dd.c.source_dataset_id,
            dd.c.source_dataset_version_id,
            dd.c.dataset_id,
            dd.c.dataset_version_id,
        ]

        # Base case: direct dependencies
        base_query = select(
            *dep_fields,
            literal(0).label("depth"),
        ).where(
            (dd.c.source_dataset_id == dataset.id)
            & (dd.c.source_dataset_version_id == dataset_version.id)
        )

        # Recursive CTE
        cte = base_query.cte(name="dependency_tree", recursive=True)

        # Recursive case: dependencies of dependencies
        recursive_query = select(
            *dep_fields,
            (cte.c.depth + 1).label("depth"),
        ).select_from(
            cte.join(
                dd,
                (cte.c.dataset_id == dd.c.source_dataset_id)
                & (cte.c.dataset_version_id == dd.c.source_dataset_version_id),
            )
        )

        cte = cte.union(recursive_query)

        # Fetch all with full details (columns matching dependency_class.parse signature)
        final_query = (
            select(
                n.c.name,
                p.c.name,
                cte.c.id,
                cte.c.dataset_id,
                cte.c.dataset_version_id,
                d.c.name,
                dv.c.version,
                dv.c.created_at,
                cte.c.source_dataset_id,
                cte.c.source_dataset_version_id,
                cte.c.depth,
            )
            .select_from(
                cte.join(d, cte.c.dataset_id == d.c.id, isouter=True)
                .join(dv, cte.c.dataset_version_id == dv.c.id, isouter=True)
                .join(p, d.c.project_id == p.c.id, isouter=True)
                .join(n, p.c.namespace_id == n.c.id, isouter=True)
            )
            .order_by(cte.c.depth, cte.c.id)
        )

        results = list(self.db.execute(final_query))

        # Build hierarchy: map dependency_id -> DatasetDependency
        dep_by_id: dict[int, DatasetDependency | None] = {}
        # Map (source_dataset_id, source_version_id) -> list of dependency_ids
        children_by_source: dict[tuple[int, int], list[int]] = {}

        for row in results:
            *parse_args, source_did, source_vid, depth = row
            dep = self.dependency_class.parse(*parse_args)

            if dep:
                dep_by_id[dep.id] = dep
                parent_key = (source_did, source_vid)
                children_by_source.setdefault(parent_key, []).append(dep.id)

        # Recursively populate dependencies for each node
        def populate_children(dep: DatasetDependency) -> None:
            # Find what dataset this dependency points to
            for row in results:
                if row[2] == dep.id:  # matching dependency id
                    target_did, target_vid = row[3], row[4]
                    if target_did and (target_did, target_vid) in children_by_source:
                        child_ids = children_by_source[(target_did, target_vid)]
                        dep.dependencies = [dep_by_id[cid] for cid in child_ids]
                        for child in dep.dependencies:
                            if child:
                                populate_children(child)
                    break

        # Return direct dependencies (depth 0) with their hierarchies populated
        direct_deps = []
        root_key = (dataset.id, dataset_version.id)
        if root_key in children_by_source:
            for dep_id in children_by_source[root_key]:
                dep = dep_by_id[dep_id]
                if dep:
                    populate_children(dep)
                    direct_deps.append(dep)

        return direct_deps
  ```
</details>

3. I calculated performance using your example in description and compared between main, your solution and this recursive one and here are the results:
* **your** -> ~ 3ms
* **recursive (code above)** -> ~8ms
* **main** -> ~272ms
As we can see, there is no much difference in performance between recursive and your solution when we compare it to `main` so I think recursive solution is better as we don't need to have and maintain additional column in dependencies table that has copy of all dependency tree below. Also code added above is pretty much all that is needed (it can just be called from `Catalog.get_dataset_dependencies()` or even better, just directly)

BTW since your recursive code is pretty similar to mine I assumed you used AI as well? Which one did you use?

amritghimire: As you see, in a simple example as above, recursive code still is around 3 times slower. Also, the query above gives 86 results, still a lot of records imo to be returned from the database. I think in real world scenario, i.e. in prod, that is going to matter, we need to transfer huge chunks of data between database and api. We already had few examples where database was treated as bottleneck. 

 I still am not comfortable to introduce that complexity in the code as you see it has extra complexity in the code without  the performance improvement.

amritghimire: Also, the response or result was way bigger than the actual value as I compared with the code above.

ilongin: I don't think it matter is it 3ms or 8ms compared to 270ms which was before, that was my point. Its like comparing O(n) and O(3n) with O(n^2). 
I don't understand those 86 results, can you explain? It returns all dependencies needed, which is exact number of record to be returned. How can it return less than that, or to ask differently - what is returned but is actually a surplus? Maybe I'm missing something though.

BTW I also don't agree it's more complex compared to solution in this PR since, as mentioned above, we don't need additional column in DB which holds basically duplicated data of dependency tree (at least their IDs) and to maintain that. I actually think it's less complex but maybe we can hear other opinions as well from @shcheklein and @dreadatour .

shcheklein: > I don't think it matter is it 3ms or 8ms compared to 270ms which was before, that was my point. Its like comparing O(n) and O(3n) with O(n^2).

it is hard to get the complexity from these small experiments though. 

> I actually think it's less complex but maybe we can hear other opinions as well from @shcheklein and @dreadatour .

Tbh, I'm not sure why we have an issue with caching dep ids as JSON? Datasets are immutable. We have this information, why not save it and use it? It makes the whole retrieval super efficient and predictable. The idea seems very simple. Can this PR made smaller and cleaner so that it is obvious that it is simple?

I like the idea of running queries against DB, but on the other hand I'm not a big fan of complicated queries and pushing he limits on technologies. I'm not smart enough to fight all of this in production later - I have enough on my plate already - k8s, graphql, celery, name it . These recursive queries on real DB, on a scale - might be fine, or might start pushing DB into some pressure (e.g. when table is large enough and recursion is deep enough). So there are definitely some risks here that no one can predict atm.

So, are there some real, architectural concerns on caching it within a dataset? 

@amritghimire let's also review the PR carefully one more time - it looked indeed intimidating a bit for that simple idea.

src/datachain/catalog/catalog.py
shcheklein: did we remove it? resolve then ... please TAL, cleanup things

src/datachain/catalog/catalog.py
shcheklein: thanks @amritghimire. yes, let's rename it if possible? (it sounds general now and makes it confusing - like we are actually building the tree here). I wonder if there is a way to make it more compact or use something simpler ... also make it probably a protected helper that starts with `_`

src/datachain/data_storage/sqlite.py
shcheklein: I think we can remove the migration and bump the schema, we haven't done this before. We can separately in docs provide a command to run for those who want to do this. I would keep code simpler.

src/datachain/catalog/catalog.py
amritghimire: I have created an alternative PR with recursive one, LMK if you think that one is simpler, I will merge that in that case.  cc. @shcheklein @ilongin @dreadatour 

https://github.com/iterative/datachain/pull/1395

https://github.com/iterative/studio/pull/12236

amritghimire: Updated it.

src/datachain/catalog/catalog.py
amritghimire: Yes, it was needed in Studio. We have indirect as argument for getting dependencies in graphql.

PR #1389, opened 2025-10-06
Adding `reset` argument to `Catalog.query()`

ilongin:
This is needed for settng `DATACHAIN_CHECKPOINTS_RESET` in SaaS.

## Summary by Sourcery

Add a `reset` parameter to `Catalog.query()` to control checkpoint reset behavior via the `DATACHAIN_CHECKPOINTS_RESET` environment variable.

New Features:
- Introduce `reset` argument in `Catalog.query()` defaulting to False

Tests:
- Update unit tests to expect `DATACHAIN_CHECKPOINTS_RESET` in the environment and parametrize `reset` values


shcheklein:
Let's use UI for Studio for this. Or am I missing something?

ilongin:
> Let's use UI for Studio for this. Or am I missing something?

This is needed for SaaS to set `DATACHAIN_CHECKPOINTS_RESET` env var into subprocess.

dreadatour:
> > Let's use UI for Studio for this. Or am I missing something?
> 
> This is needed for SaaS to set `DATACHAIN_CHECKPOINTS_RESET` env var into subprocess.

🤔 Just wonder, if this is needed only for SaaS, why not to set `DATACHAIN_CHECKPOINTS_RESET` env variable in SaaS code before calling the `query` method, since we are passing `env` variables to it anyway? https://github.com/iterative/datachain/blob/main/src/datachain/catalog/catalog.py#L1753

ilongin:
> > > Let's use UI for Studio for this. Or am I missing something?
> > 
> > 
> > This is needed for SaaS to set `DATACHAIN_CHECKPOINTS_RESET` env var into subprocess.
> 
> 🤔 Just wonder, if this is needed only for SaaS, why not to set `DATACHAIN_CHECKPOINTS_RESET` env variable in SaaS code before calling the `query` method, since we are passing `env` variables to it anyway? https://github.com/iterative/datachain/blob/main/src/datachain/catalog/catalog.py#L1753

I would like to have this explicit in a method signature. It's hard to track those env vars and reason with codebase if we do it that way IMO.  Also, reset logic is not SaaS specific (like those other env vars like num of workers, qeues etc.) but also used in local, just passing it is little bit different, so it doesn't really fit in those ..

src/datachain/catalog/catalog.py
```diff
@@ -1765,6 +1766,7 @@ def query(
             {
                 "DATACHAIN_QUERY_PARAMS": json.dumps(params or {}),
                 "DATACHAIN_JOB_ID": job_id or "",
+                "DATACHAIN_CHECKPOINTS_RESET": str(reset),
```
dreadatour, line 1772: Will it be `None` string if `reset` param is not set?
Let's add it to the env variables only if it is `True`?
Or, at least, `str(bool(reset))` or something like `"True" if reset else "False"`

ilongin: `None` will be the same as `False` since we are reading this with [`env_to_bool()`](https://github.com/iterative/datachain/blob/main/src/datachain/utils.py#L492-L499) util ... so that env will be read as `True` only if reset is also `True`, otherwise it will be `False`.

dreadatour: Yes, I agree, but same time having `None` as value in env variable kind of smells 👀😅

dreadatour (approved):
Approving this, but please, let's fix env variables 🙏

src/datachain/catalog/catalog.py
ilongin: I've changed it to be default `False` and check if it's bool (so that we cannot put `None` or some string there)

PR #1395, opened 2025-10-11
Alternative to  #1387

amritghimire:
Alternative to  #1387
Summary:
------
This PR changes improvement in performances to dataset dependency
calculation as well as in storage where we now store the nested
dependency structure directly to optimize the retreival process.

Changes:
------
- Replaced recursive database queries with a single batch query that fetches all dependencies at once
- Reduced Database Round-trips: Dependency calculation now uses get_direct_dataset_dependencies_by_ids() to fetch all required dependencies in one query
- Optimized Memory Usage: New DatasetDependencyMinimal class stores only essential dependency information

Database change:
-------
- Added nested_dependencies Column: New JSON column in datasets_dependencies table to store pre-calculated dependency structures
- Backward Compatibility: Schema migration safely adds the new column to existing databases
- Efficient Storage: Nested dependency trees are stored as JSON, reducing storage overhead

## Summary by Sourcery

Improve dataset dependency retrieval by replacing recursive per-dependency queries with a single batch recursive CTE, storing nested dependency trees directly in the database, and providing utilities to assemble dependency hierarchies in memory.

New Features:
- Introduce DatasetDependencyNode dataclass to represent dependency entries with hierarchy depth
- Add metastore.get_dataset_dependency_node method leveraging a recursive CTE to fetch complete dependency trees
- Implement catalog.get_dataset_dependencies_by_ids method to retrieve and assemble nested dependencies in a single batch
- Provide build_dependency_hierarchy and populate_nested_dependencies utilities for constructing nested dependency trees

Enhancements:
- Replace recursive per-dependency database lookups with a single batched query to reduce round trips
- Add nested_dependencies JSON column to datasets_dependencies table with a migration for efficient storage of precomputed dependency structures

### DIFF
--- a/src/datachain/data_storage/metastore.py
+++ b/src/datachain/data_storage/metastore.py
@@ -27,12 +27,12 @@ from sqlalchemy import (
 )
 from sqlalchemy.sql import func as f
 
+from datachain.catalog.dependency import DatasetDependencyNode
 from datachain.checkpoint import Checkpoint
 from datachain.data_storage import JobQueryType, JobStatus
 from datachain.data_storage.serializer import Serializable
 from datachain.dataset import (
     DatasetDependency,
-    DatasetDependencyNode,
     DatasetListRecord,
     DatasetListVersion,
     DatasetRecord,
@@ -1519,7 +1519,6 @@ class AbstractDBMetastore(AbstractMetastore):
             & (dd.c.source_dataset_version_id == version_id)
         )
 
-        # Recursive CTE
         cte = base_query.cte(name="dependency_tree", recursive=True)
 
         # Recursive case: dependencies of dependencies
@@ -1537,27 +1536,27 @@ class AbstractDBMetastore(AbstractMetastore):
         cte = cte.union(recursive_query)
 
         # Fetch all with full details
-        final_query = (
-            select(
-                n.c.name,
-                p.c.name,
-                cte.c.id,
-                cte.c.dataset_id,
-                cte.c.dataset_version_id,
-                d.c.name,
-                dv.c.version,
-                dv.c.created_at,
-                cte.c.source_dataset_id,
-                cte.c.source_dataset_version_id,
-                cte.c.depth,
-            )
-            .select_from(
-                cte.join(d, cte.c.dataset_id == d.c.id, isouter=True)
-                .join(dv, cte.c.dataset_version_id == dv.c.id, isouter=True)
-                .join(p, d.c.project_id == p.c.id, isouter=True)
-                .join(n, p.c.namespace_id == n.c.id, isouter=True)
-            )
-            .order_by(cte.c.depth, cte.c.id)
+        final_query = select(
+            n.c.name,
+            p.c.name,
+            cte.c.id,
+            cte.c.dataset_id,
+            cte.c.dataset_version_id,
+            d.c.name,
+            dv.c.version,
+            dv.c.created_at,
+            cte.c.source_dataset_id,
+            cte.c.source_dataset_version_id,
+            cte.c.depth,
+        ).select_from(
+            # Use outer joins to handle cases where dependent datasets have been
+            # physically deleted. This allows us to return dependency records with
+            # None values instead of silently omitting them, making broken
+            # dependencies visible to callers.
+            cte.join(d, cte.c.dataset_id == d.c.id, isouter=True)
+            .join(dv, cte.c.dataset_version_id == dv.c.id, isouter=True)
+            .join(p, d.c.project_id == p.c.id, isouter=True)
+            .join(n, p.c.namespace_id == n.c.id, isouter=True)
         )
 
         return [
--- a/src/datachain/dataset.py
+++ b/src/datachain/dataset.py
@@ -20,7 +20,6 @@ LT = TypeVar("LT", bound="DatasetListRecord")
 V = TypeVar("V", bound="DatasetVersion")
 LV = TypeVar("LV", bound="DatasetListVersion")
 DD = TypeVar("DD", bound="DatasetDependency")
-DDN = TypeVar("DDN", bound="DatasetDependencyNode")
 
 DATASET_PREFIX = "ds://"
 QUERY_DATASET_PREFIX = "ds_query_"
@@ -93,62 +92,6 @@ def parse_dataset_name(name: str) -> tuple[str | None, str | None, str]:
 class DatasetDependencyType:
     DATASET = "dataset"
     STORAGE = "storage"
-
-
-@dataclass
-class DatasetDependencyNode:
-    namespace: str
-    project: str
-    id: int
-    dataset_id: int | None
-    dataset_version_id: int | None
-    dataset_name: str | None
-    dataset_version: str | None
-    created_at: datetime
-    source_dataset_id: int
-    source_dataset_version_id: int | None
-    depth: int
-
-    @classmethod
-    def parse(
-        cls: builtins.type[DDN],
-        namespace: str,
-        project: str,
-        id: int,
-        dataset_id: int | None,
-        dataset_version_id: int | None,
-        dataset_name: str | None,
-        dataset_version: str | None,
-        created_at: datetime,
-        source_dataset_id: int,
-        source_dataset_version_id: int | None,
-        depth: int,
-    ) -> "DatasetDependencyNode | None":
-        return cls(
-            namespace,
-            project,
-            id,
-            dataset_id,
-            dataset_version_id,
-            dataset_name,
-            dataset_version,
-            created_at,
-            source_dataset_id,
-            source_dataset_version_id,
-            depth,
-        )
-
-    def to_dependency(self) -> "DatasetDependency | None":
-        return DatasetDependency.parse(
-            namespace_name=self.namespace,
-            project_name=self.project,
-            id=self.id,
-            dataset_id=self.dataset_id,
-            dataset_version_id=self.dataset_version_id,
-            dataset_name=self.dataset_name,
-            dataset_version=self.dataset_version,
-            dataset_version_created_at=self.created_at,
-        )
 
 
 @dataclass
--- a/src/datachain/catalog/dependency.py
+++ b/src/datachain/catalog/dependency.py
@@ -1,4 +1,67 @@
-from datachain.dataset import DatasetDependency, DatasetDependencyNode
+import builtins
+from dataclasses import dataclass
+from datetime import datetime
+from typing import TypeVar
+
+from datachain.dataset import DatasetDependency
+
+DDN = TypeVar("DDN", bound="DatasetDependencyNode")
+
+
+@dataclass
+class DatasetDependencyNode:
+    namespace: str
+    project: str
+    id: int
+    dataset_id: int | None
+    dataset_version_id: int | None
+    dataset_name: str | None
+    dataset_version: str | None
+    created_at: datetime
+    source_dataset_id: int
+    source_dataset_version_id: int | None
+    depth: int
+
+    @classmethod
+    def parse(
+        cls: builtins.type[DDN],
+        namespace: str,
+        project: str,
+        id: int,
+        dataset_id: int | None,
+        dataset_version_id: int | None,
+        dataset_name: str | None,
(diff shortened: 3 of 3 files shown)


src/datachain/dataset.py
```diff
@@ -94,6 +95,62 @@ class DatasetDependencyType:
     STORAGE = "storage"
 
 
+@dataclass
+class DatasetDependencyNode:
```
shcheklein: why do we need this model in this case? (why not regular Dependecy?) What is Node?

src/datachain/data_storage/metastore.py
```diff
@@ -1483,6 +1492,78 @@ def get_direct_dataset_dependencies(
 
         return [self.dependency_class.parse(*r) for r in self.db.execute(query)]
 
+    def get_dataset_dependency_node(
+        self, dataset_id: int, version_id: int
+    ) -> list[DatasetDependencyNode | None]:
+        n = self._namespaces
```
shcheklein: have we tried to run it on PG? with a substantial table size and some deep nesting?

src/datachain/data_storage/metastore.py
```diff
@@ -1483,6 +1492,78 @@ def get_direct_dataset_dependencies(
...
+                cte.c.dataset_version_id,
+                d.c.name,
+                dv.c.version,
+                dv.c.created_at,
+                cte.c.source_dataset_id,
+                cte.c.source_dataset_version_id,
+                cte.c.depth,
+            )
+            .select_from(
+                cte.join(d, cte.c.dataset_id == d.c.id, isouter=True)
```
shcheklein: why outer is needed?

src/datachain/data_storage/metastore.py
```diff
@@ -1483,6 +1492,78 @@ def get_direct_dataset_dependencies(
...
+                .join(dv, cte.c.dataset_version_id == dv.c.id, isouter=True)
+                .join(p, d.c.project_id == p.c.id, isouter=True)
+                .join(n, p.c.namespace_id == n.c.id, isouter=True)
+            )
+            .order_by(cte.c.depth, cte.c.id)
```
shcheklein: do we need order by?

src/datachain/data_storage/metastore.py
```diff
@@ -1483,6 +1492,78 @@ def get_direct_dataset_dependencies(
 
         return [self.dependency_class.parse(*r) for r in self.db.execute(query)]
 
+    def get_dataset_dependency_node(
```
shcheklein: IMPORTANT: check security: team id, etc. Add tests for this if needed

shcheklein: rename - get...nodes ... not a single node

src/datachain/dataset.py
amritghimire: This is to save the information obtained from recursive query. The regular dependency doesn't need some extra field like depth with it.

src/datachain/data_storage/metastore.py
amritghimire: > have we tried to run it on PG? 

Yes, it is in studio PR as well.

src/datachain/data_storage/metastore.py
amritghimire: > So that we will be able to show if there are any rows with dataset dependencies removed as None value.

https://github.com/iterative/datachain/pull/1387#discussion_r2418332547

src/datachain/data_storage/metastore.py
amritghimire: ~This is not for this, but in Studio part. Will add there.~

This uses the dataset_select and so on which is overriden in postgres to add team filter, so everything works as expected.

src/datachain/data_storage/metastore.py
amritghimire: Yes, it helps when forming the dependency heirarchy later.

shcheklein: how? could you elaborate please?

amritghimire: When building the hierarchy using build_dependency_hierarchy, this helps because we go through the depth by depth basis.

src/datachain/dataset.py
shcheklein: Any other fields? I'm just thinking is it really worth it introducing an extra model for this? what are the other fields? can we just use some default value or None when we don't need them?

also, put some comments - it won't be obvious at all later why we have two dependencies and which one to use when

src/datachain/data_storage/metastore.py
shcheklein: how about some real scale in PG?

src/datachain/data_storage/metastore.py
shcheklein: so, we keep ids, but we remove them physically ... please add comment here on this

src/datachain/data_storage/metastore.py
shcheklein: still, can you elaborate pls?

src/datachain/data_storage/metastore.py
shcheklein: can you point me to both places pls? 

can we add a test for this?

amritghimire: https://github.com/iterative/studio/blob/14a8465ba463626284d54b2c6355783858cb5b65/backend/datachain_server/src/datachain_server/postgresql.py#L646 and https://github.com/iterative/studio/blob/14a8465ba463626284d54b2c6355783858cb5b65/backend/datachain_server/src/datachain_server/postgresql.py#L674

src/datachain/data_storage/metastore.py
amritghimire: By ordering by depth first, we ensure that the direct dependencies appear before their own dependencies. That makes the result much more easier to work with, you see the immediate dependencies (depth=0) . then second level dependencies (depth=1) and so on. 
Also, having a consistent sort order makes the api predicatable as well. Which is why I always prefer backend to return a sorted output. It also provides a stable testsing that don't have flaky ordering issue.

src/datachain/data_storage/metastore.py
shcheklein: how can we be sure that some logic doesn't change and it won't be using those directly anymore? we need a test for this ..

src/datachain/data_storage/metastore.py
```diff
@@ -1483,6 +1492,78 @@ def get_direct_dataset_dependencies(
...
+        # Base case: direct dependencies
+        base_query = select(
+            *dep_fields,
+            literal(0).label("depth"),
+        ).where(
+            (dd.c.source_dataset_id == dataset_id)
+            & (dd.c.source_dataset_version_id == version_id)
+        )
+
+        # Recursive CTE
```
shcheklein: remove obvious comments

src/datachain/data_storage/metastore.py
amritghimire: Whenever we delete the dataset or dataset version, we remove the entries (cascade) only in case of source dataset or source dataset id, we don't remove the entries of dependencies. i.e. `dataset` and `dataset_version` . So, we set null to those values. In such case, the dataset_id will be null. Outer join will make sure the dataset dependencies in the query will return those with null values too.

src/datachain/data_storage/metastore.py
shcheklein: since we also do some outer joins - can we end up getting some objects that do not belong to the team

e.g. team_id is not provided, datasets() query return empty, but since it is outer it will still fetch some stuff from other tables?

how do we check all of this situations?

src/datachain/data_storage/metastore.py
shcheklein: okay it if it for aesthetics, let's not make it part of the API though if it not really needed (e.g. we can't promise this to FE, I would also not use it in tests)

src/datachain/data_storage/metastore.py
shcheklein: okay, add comments here why outer is really needed

src/datachain/data_storage/metastore.py
amritghimire: WDYM by for aesthetics only? 
> By ordering by depth first, we ensure that the direct dependencies appear before their own dependencies. That makes the result much more easier to work with, you see the immediate dependencies (depth=0) . then second level dependencies (depth=1) and so on.

It is so we have depth 0 dependencies first, then second level dependencies second so it is used :
> When building the hierarchy using build_dependency_hierarchy,

I am not sure I understand the scrutiny over a simple order_by here.

shcheklein: okay, then, I'm missing the function (non-aesthetics) part here? are we relying on this somewhere?

shcheklein: > I am not sure I understand the scrutiny over a simple order_by here.

performance, it is additional operation

+ basic understanding (when I read code I ask myself if I really understand why and what is happening here and what is important what is not) - it is harder then to refactor later for example - you have to have answers on why each line exists pretty much

src/datachain/data_storage/metastore.py
amritghimire: I am not sure I understand tbh. How will outer fetch stuff from other tables, if the dataset id is null? 

If you mean what if team_id is not provided in postgresql metastore, then it is the fundamental part of the metastore. With studio, it will always be there.

If that failed to be there, then we would have  been returning everything everywhere in Studio by that logic.

shcheklein: I don't know, I didn't check the details. This is to double check.

Hypothetical situation: we are getting some ID from the API (not from the select). E.g. dataset id, etc. We then run select from `datasets` with this id and get empty result (since team doesn't match) - but we keep using the id itself in outer merge to merge with deps table based on this id alone. We'll get a bunch of deps results + nulls for dataset columns bc of outer merge.

This what I want to double check ^^. It is hard for me to be precise / exact here. Asking to double and triple check and may be add some tests here.

(overall, we must add a wrapper around all APIs I think to check every single object that it returns)

amritghimire: Lets assume that happened, even in such case, the dataset id whose information we were supposed to return will not be available  in outer join right, then we will just return None value in dependency tree. With this outer join and this code implementation alone, you can see that we will not have any information returned to user if the user don't have the access to the team id. The whole information, project id, and dataset id depends on the same query dataset_id. 

> (overall, we must add a wrapper around all APIs I think to check every single object that it returns)

I am not even sure how we can achieve this, basically you are asking is capture the response for each api, check for the access in such case. Not every object will have team_id or even similar structure. How about the items in nested structure?

src/datachain/data_storage/metastore.py
amritghimire: Removed.,

src/datachain/data_storage/metastore.py
amritghimire: Removed it.

src/datachain/data_storage/metastore.py
amritghimire: Added

src/datachain/dataset.py
amritghimire: Yes, it has source_dataset_version_id, source_dataset_id, dataset_version_id and dataset_id and so on. Basically this is for internal purpose. Moving it to different file.

shcheklein: kk (just want to be 1000% sure we need an extra model)

src/datachain/data_storage/metastore.py
shcheklein: > Lets assume that happened, even in such case, the dataset id whose information we were supposed to return will not be available in outer join right,

it really depends on how exactly we do it. Again, I didn't check exact queries. I'm asking to double and triple check and think about all possible cases.

Feel free to resolve if you think it is safe.

shcheklein: (But we need to think how to automate these checks)

src/datachain/data_storage/metastore.py
amritghimire: Yes locally with possible reasonable scale,  also added some indexes in studio.

PR #1396, opened 2025-10-11
Fix checkpoints hash tests

dreadatour:
After sqlalchemy update to `2.0.44`.

## Summary by Sourcery

Tests:
- Update expected hash digests in hash utility, query steps, and datachain tests to align with SQLAlchemy 2.0.44

### DIFF
--- a/tests/unit/test_datachain_hash.py
+++ b/tests/unit/test_datachain_hash.py
@@ -195,7 +195,7 @@ def test_all_possible_steps(test_session):
             right_on=["player.name"],
         )
         .hash()
-    ) == "1e2caf5d0bcda7ea75e33bb4db43cc7321191306b47a77ae541b61bbbfcea32b"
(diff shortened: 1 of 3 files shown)


dreadatour:
> Does this change have any implications for end users?

In current implementation this should not affects end users.

ilongin:
> > Does this change have any implications for end users?
> 
> In current implementation this should not affects end users.

I still wanted to have hash not rely on SQLAlchemy version changes too much at least. I will investigste this as a followup

dreadatour:
> I still wanted to have hash not rely on SQLAlchemy version changes too much at least. I will investigste this as a followup

Would be great to check if we can exclude internal "protected" sqlalchemy variables from hash calculation. E.g. add check for `_` prefix in `attr_name` here: https://github.com/iterative/datachain/blob/main/src/datachain/hash_utils.py#L51

PR #1397, opened 2025-10-12
remove duplicate tests

shcheklein:
Got introduced probably with bad conflict resolve here https://github.com/iterative/datachain/pull/1365/files

## Summary by Sourcery

Tests:
- Remove duplicate functional tests for UDF mapping in serial, parallel, and distributed modes


tests/func/test_udf.py
```diff
@@ -131,6 +131,11 @@ def extract_paths(
     assert rows[2][4].source == "file://"
 
 
+@pytest.mark.parametrize(
```
shcheklein, line 134: restores decorators were also removed by mistake I think

PR #1398, opened 2025-10-13
Fix union breaking schema order

ilongin:
Union was breaking schema order mixing signals from multiple objects (e.g `File`). This PR fixes this issue.
Related Studio issue: https://github.com/iterative/studio/issues/12188

### DIFF
--- a/src/datachain/query/dataset.py
+++ b/src/datachain/query/dataset.py
@@ -1238,33 +1238,27 @@ class UnionSchemaMismatchError(ValueError):
         return cls(f"Cannot perform union. {'. '.join(parts)}")
 
 
-def _validate_columns(
-    left_columns: Iterable[ColumnElement], right_columns: Iterable[ColumnElement]
-) -> list[str]:
-    left_names = [c.name for c in left_columns]
-    right_names = [c.name for c in right_columns]
-
-    if sorted(left_names) == sorted(right_names):
-        return left_names
-
-    left_names_set = set(left_names)
-    right_names_set = set(right_names)
-
-    raise UnionSchemaMismatchError.from_column_sets(
-        left_names_set - right_names_set,
-        right_names_set - left_names_set,
-    )
-
-
 def _order_columns(
     left_columns: Iterable[ColumnElement], right_columns: Iterable[ColumnElement]
 ) -> list[list[ColumnElement]]:
-    column_order = _validate_columns(left_columns, right_columns)
+    left_names = [c.name for c in left_columns]
+    right_names = [c.name for c in right_columns]
+
+    # validate
+    if sorted(left_names) != sorted(right_names):
+        left_names_set = set(left_names)
+        right_names_set = set(right_names)
+        raise UnionSchemaMismatchError.from_column_sets(
+            left_names_set - right_names_set,
+            right_names_set - left_names_set,
+        )
+
+    # Order columns to match left_names order
     column_dicts = [
         {c.name: c for c in columns} for columns in [left_columns, right_columns]
     ]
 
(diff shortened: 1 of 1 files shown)


src/datachain/query/dataset.py
```diff
@@ -1170,15 +1170,18 @@ def apply_sql_clause(self, query) -> Select:
...
     left_columns: Iterable[ColumnElement], right_columns: Iterable[ColumnElement]
-) -> set[str]:
-    left_names = {c.name for c in left_columns}
-    right_names = {c.name for c in right_columns}
+) -> list[str]:
+    left_names = [c.name for c in left_columns]
+    right_names = [c.name for c in right_columns]
 
-    if left_names == right_names:
+    if sorted(left_names) == sorted(right_names):
```
dmpetrov: It suppose to be much more strict - you cannot union if **any** mismatch.

If we try to be smart here, it opens a can of worms when someone will be always not happy with the results.

More details:


1. Number of columns. Must have.
2. Types of columns (some exception: it's ok to make it nullable or convert int to float).
3. Names of columns.

In SQL, they require only (1) and (2) while ignoring (3).

We might have issues with (2) - sqlalchemy is not good at types. So, for us it's better to use (3) in addition to (1) and ignore types (2). It should be == without any sorting.

ilongin:
> Why does it fix the issue, could you explain it please?
> 
> Specifically, I don't quite understand why would columns we have in `select` / `subquery` affect the signal schema that we have attached to the chain /query. Is it schema that defines the order / signals, etc? Or is it done in some other different way?

Schema is derived directly from selected columns from built SQLAlchemy query. Note that this is flatten schema (e.g it has keys like `file__path`, `file__size` etc.). We also have `feature_schema` which has higher level objects defined like `file: File` etc but that one is not important in this issue.

At some point in `SQLUnion` logic we were validating and constructing columns for union and in the process of validating we were using `set` which broke the original order.

shcheklein:
> Schema is derived directly from selected columns from built SQLAlchemy query. 

could you point me to it please?

src/datachain/query/dataset.py
ilongin: So you are saying that we should not sort in order to not mix columns that are named the same but actually have different types? If it's only a matter of removing sorting then I will add it here, but if it's more complex I would add another PR / issue for this since it's not actually related to this PR (before my change we were comparing sets of column names which is pretty much the same)

ilongin:
> > Schema is derived directly from selected columns from built SQLAlchemy query.
> 
> could you point me to it please?

1. Columns created out of query and sent to `create_dataset()` -> https://github.com/iterative/datachain/blob/main/src/datachain/query/dataset.py#L1908-L1928
2. Columns used to create schema in `create_dataset()` -> https://github.com/iterative/datachain/blob/main/src/datachain/catalog/catalog.py#L834-L836

src/datachain/query/dataset.py
ilongin: A lot of tests failing after removing sorting  ... let's do this in a separate issue as it's not trivial it seems ,and as mention above, it's not really related to this PR anyway.

shcheklein:
that's really weird, why aren't we using signal schema? it feels it can be tricky to preserve and guarantee order of columns in all these subqueries and selects ...

ilongin:
> that's really weird, why aren't we using signal schema? it feels it can be tricky to preserve and guarantee order of columns in all these subqueries and selects ...

I've added separate PR where I'm experimenting with using actual signals schema for calculating this, as it's not super simple it seems (need more time fixing that and making sure it's not breaking) https://github.com/iterative/datachain/pull/1404

PR #1400, opened 2025-10-13
fix(steps): drop sys columns when it is not safe to keep them

shcheklein:
Continuation of 

https://github.com/iterative/datachain/pull/1371
https://github.com/iterative/datachain/pull/1371

We need to drop sys columns since it is not safe to carry them and regeneration is expensive and is not actually needed it seems. We should be operating in a mode where sys columns don't exist after each step and generated when they are actually needed.

## TODO:

- [x] Check apply `_regenerate_system_columns` used in partition by
- [x] Check if need to drop `sys` from the schema
- [x] Check tests
- [x] Check Studio part
- [x] Optimize `_regenerate_system_columns` to not wrap into subqueries if it is not needed (we kept existing columns)

## Summary by Sourcery

Drop system columns in dataset query unions to avoid unsafe propagation and expensive regeneration

Bug Fixes:
- Prevent carrying unsafe sys__ columns across steps by filtering them out before union operations

Enhancements:
- Add helper _drop_system_columns to filter out sys__-prefixed columns with fallback
- Refactor apply() in dataset to use filtered columns for unions and simplify selection logic via subqueries

## Summary by Sourcery

Drop system columns by default in query operations, selectively regenerate them when necessary, and refactor dataset and warehouse logic to avoid unsafe propagation of sys__ columns.

Bug Fixes:
- Prevent unsafe propagation of system columns across union and full outer join steps
- Avoid null sys__id values after merge operations by dropping system columns before union

Enhancements:
- Introduce helper _drop_system_columns to filter out sys__-prefixed columns
- Extend _regenerate_system_columns to selectively regenerate system columns and reuse existing ones
- Refactor dataset apply, clone, union, and merge logic to drop system columns by default and regenerate only when needed

Tests:
- Update tests to assert ds.schema and absence of system columns instead of signals_schema
- Adjust function and hash tests to reflect changes in default system column handling

### DIFF
--- a/src/datachain/data_storage/sqlite.py
+++ b/src/datachain/data_storage/sqlite.py
@@ -868,11 +868,8 @@ class SQLiteWarehouse(AbstractWarehouse):
                 if isinstance(c, BinaryExpression):
                     right_left_join = add_left_rows_filter(c)
 
-        # Use CTE instead of subquery to force SQLite to materialize the result
-        # This breaks deep nesting and prevents parser stack overflow.
         union_cte = sqlalchemy.union(left_right_join, right_left_join).cte()
-
-        return self._regenerate_system_columns(union_cte)
+        return sqlalchemy.select(*union_cte.c).select_from(union_cte)
 
     def _system_row_number_expr(self):
         return func.row_number().over()
@@ -884,11 +881,7 @@ class SQLiteWarehouse(AbstractWarehouse):
         """
         Create a temporary table from a query for use in a UDF.
         """
-        columns = [
-            sqlalchemy.Column(c.name, c.type)
-            for c in query.selected_columns
-            if c.name != "sys__id"
-        ]
+        columns = [sqlalchemy.Column(c.name, c.type) for c in query.selected_columns]
         table = self.create_udf_table(columns)
 
         with tqdm(desc="Preparing", unit=" rows", leave=False) as pbar:
--- a/src/datachain/data_storage/warehouse.py
+++ b/src/datachain/data_storage/warehouse.py
@@ -5,7 +5,7 @@ import random
 import string
 from abc import ABC, abstractmethod
 from collections.abc import Callable, Generator, Iterable, Iterator, Sequence
-from typing import TYPE_CHECKING, Any, Union
+from typing import TYPE_CHECKING, Any, Union, cast
 from urllib.parse import urlparse
 
 import attrs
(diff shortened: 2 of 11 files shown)


src/datachain/query/dataset.py
```diff
@@ -438,12 +438,13 @@ def create_result_query(
         """
 
     def populate_udf_table(self, udf_table: "Table", query: Select) -> None:
-        if "sys__id" not in query.selected_columns:
```
shcheklein, line 441: @dreadatour need you review here specifically 

(and if you know we need to change something else in some other place overall(

src/datachain/data_storage/sqlite.py
```diff
@@ -868,11 +868,8 @@ def add_left_rows_filter(exp: BinaryExpression):
                 if isinstance(c, BinaryExpression):
                     right_left_join = add_left_rows_filter(c)
 
-        # Use CTE instead of subquery to force SQLite to materialize the result
-        # This breaks deep nesting and prevents parser stack overflow.
-        union_cte = sqlalchemy.union(left_right_join, right_left_join).cte()
-
-        return self._regenerate_system_columns(union_cte)
+        union = sqlalchemy.union(left_right_join, right_left_join).subquery()
```
shcheklein: C: hopefully this just reverts it to where it was before introducing `_regenerate_system_columns`

src/datachain/data_storage/warehouse.py
```diff
@@ -248,45 +248,57 @@ def dataset_select_paginated(
...
+            present.add(col.name)
+            regen = col.name in regenerate and not keep_existing_columns
+            columns.append(build(col.name) if regen else col)
+            changed |= regen
+
+        for name in regenerate - present:
+            columns.append(build(name))
+            changed = True
+
+        if not changed:
```
shcheklein, line 299: C: optimization was missing - we don't now wrap into subqueries if there are not changes

src/datachain/diff/__init__.py
```diff
@@ -151,11 +151,7 @@ def _to_list(obj: str | Sequence[str] | None) -> list[str] | None:
     if status_col:
         cols_select.append(diff_col)
 
-    if not dc_diff._sys:
```
shcheklein, line 154: C: @ilongin - that was no clear why it is needed

src/datachain/lib/dc/datachain.py
```diff
@@ -894,7 +896,7 @@ def gen(
                 udf_obj.to_udf_wrapper(self._settings.batch_size),
                 **self._settings.to_dict(),
             ),
-            signal_schema=udf_obj.output,
+            signal_schema=SignalSchema({"sys": Sys}) | udf_obj.output,
```
shcheklein, line 899: C: I assume chain.singnal_schema always reflects actual state, including sys, etc. So, AFAIU (cc @dreadatour ) our UDFs will include sys columns (since we create UDF table with them). Is it true?

@dreadatour need an extra pair of eyes on this

PR #1404, opened 2025-10-14
Using signals schema instead of columns to calculate dataset schema

### LINKED ISSUE #1403: Use `SignalSchema` to calculate final dataset version schema
Currently to calculate final dataset version `schema` column we use sqlalchemy select statement columns, but since we do have `SignalSchema` instance calculated in chain we should try to use that one instead.

ilongin:
WIP

## Summary by Sourcery

Use the signals schema to compute and pass both hierarchical and flat dataset schemas across save and dataset creation APIs, eliminating manual column-based schema logic

New Features:
- Add clone_with_sys_signals method to SignalSchema for merging system and user signals

Enhancements:
- Generate a flat_schema from signals_schema (including system signals) for dataset storage
- Propagate flat_schema through dataset save and create_dataset calls to replace ad-hoc column-based schema building
- Remove inline column-based schema computation in catalog.create_dataset and pass flat_schema to metastore.create_dataset


shcheklein:
@ilongin ping

PR #1410, opened 2025-10-18
Don't duplicate temp tables on reusing queries

shcheklein:
We need a proper fix for the whole temp table tracking. I'll take a look hopefully. We should be using some kind of "run context" instead of reusing one global temp_tables from the initial chain. 

For now this makes the manipulations with temp tables more precise.

## Summary by Sourcery

Fix temporary table tracking by slicing appended names to avoid duplicates and add tests to validate correct behavior

Bug Fixes:
- Prevent duplicate temp table entries by appending only newly added names in various query methods
- Log an error if duplicate temp tables are detected during cleanup

Tests:
- Add functional tests to verify no duplicate temp tables in nested merge, union, and subtract operations


0x2b3bfa0 (approved):
🐱 🤖

PR #1412, opened 2025-10-19
fix delta expecting sys columns in apply steps

shcheklein:
One more sys_id fix.

We are applying steps in delta directly on top of newly generated step. Generally speaking this is not safe (we are not rebuilding steps properly, we are applying cached versions). Now as union drops columns we are reintroducing them here. 

Ideally, we should be generating steps in the first place where sys__ids don't exist.

## Summary by Sourcery

Regenerate system columns before applying delta steps to ensure cached steps have required sys columns

New Features:
- Add _RegenerateSystemColumnsStep to re-add system columns via warehouse._regenerate_system_columns
- Modify _append_steps to insert the regeneration step before merging delta steps

Bug Fixes:
- Fix missing system columns in delta apply steps by regenerating them

Tests:
- Add test verifying that delta replay regenerates system columns and preserves measurement_id

### DIFF
--- a/src/datachain/delta.py
+++ b/src/datachain/delta.py
@@ -1,12 +1,16 @@
+import hashlib
 from collections.abc import Sequence
 from copy import copy
 from functools import wraps
 from typing import TYPE_CHECKING, TypeVar
 
+from attrs import frozen
+
 import datachain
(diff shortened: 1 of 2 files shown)


dmpetrov (approved):
LG - haven't look deep

PR #1421, opened 2025-10-21
fix(diff): edge case when there no columns except provided in on

shcheklein:
Fixes an edge case in `diff` (see test). When only column(s) that dataset has are the ones provided in `on` we treated records as different. It fixes to treat those as same. (It affects delta among other things).

## Summary by Sourcery

Fix diff edge case where datasets with only 'on' columns were incorrectly marked as different by treating an empty compare list as no modifications and add a test for this scenario

Bug Fixes:
- Handle empty compare list in diff to avoid marking rows as modified when only 'on' columns are present

Tests:
- Add test to verify diff treats records as same when only 'on' columns exist


dreadatour (approved):
👍

PR #1424, opened 2025-10-25
always drop sys id in merge

shcheklein:
Continuation of my favorite thing - `sys__id` and `sys__rand` management.

It is not really safe to keep in merge / join in general since there could be duplicates after merge AFAIU.

It is better to drop it and don't advertise as part of the schema outside.

## Summary by Sourcery

Enforce dropping of system columns (sys signals) during merge and join operations to prevent duplicate or exposed internal schema fields and verify correct behavior through updated and new tests.

Enhancements:
- Always drop sys signals when merging or joining datasets in both client and query implementations.

Tests:
- Remove sys key assertions from existing merge tests and add a new test to confirm system columns are excluded after merge and correctly restored on save/load.

### DIFF
--- a/src/datachain/delta.py
+++ b/src/datachain/delta.py
@@ -150,7 +150,9 @@ def _get_retry_chain(
         error_records = result_dataset.filter(C(delta_retry) != "")
         error_source_records = source_dc.merge(
             error_records, on=on, right_on=right_on, inner=True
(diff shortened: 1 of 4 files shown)


dmpetrov (approved):
LG - I haven't look deep

PR #1430, opened 2025-10-27
fix sys ids in agg, add more tests

shcheklein:
Cleanup tests a bit, add more tests. Primarily for sys is gen stuff.

## Summary by Sourcery

Clean up the functional test suite and add comprehensive tests for UDF generation and aggregation behavior after union and merge operations

Tests:
- Add tests verifying UDF generation works correctly after union operations
- Add tests verifying UDF generation works correctly after merge operations in both full and non-full modes
- Add tests validating aggregation with functional partitions after union and merge operations
- Remove redundant parallel generation test in test_union.py

### DIFF
--- a/src/datachain/query/dataset.py
+++ b/src/datachain/query/dataset.py
@@ -630,23 +630,18 @@ class UDFStep(Step, ABC):
     def apply(
         self, query_generator: QueryGenerator, temp_tables: list[str]
     ) -> "StepResult":
-        _query = query = query_generator.select()
+        query, tables = self.process_input_query(query_generator.select())
+        _query = query
 
         # Apply partitioning if needed.
         if self.partition_by is not None:
-            _query = query = self.catalog.warehouse._regenerate_system_columns(
-                query_generator.select(),
-                keep_existing_columns=True,
-                regenerate_columns=["sys__id"],
-            )
             partition_tbl = self.create_partitions_table(query)
-            temp_tables.append(partition_tbl.name)
             query = query.outerjoin(
                 partition_tbl,
                 partition_tbl.c.sys__id == query.selected_columns.sys__id,
             ).add_columns(*partition_columns())
+            tables = [*tables, partition_tbl]
 
-        query, tables = self.process_input_query(query)
         temp_tables.extend(t.name for t in tables)
         udf_table = self.create_udf_table(_query)
         temp_tables.append(udf_table.name)


src/datachain/query/dataset.py
```diff
@@ -630,23 +630,18 @@ def clone(self, partition_by: PartitionByType | None = None) -> "Self":
...
-            )
             partition_tbl = self.create_partitions_table(query)
-            temp_tables.append(partition_tbl.name)
             query = query.outerjoin(
                 partition_tbl,
                 partition_tbl.c.sys__id == query.selected_columns.sys__id,
             ).add_columns(*partition_columns())
+            tables = [*tables, partition_tbl]
 
-        query, tables = self.process_input_query(query)
```
dreadatour, line 649: We are now passing query with `JOIN` between two tables (`pre_udf_table` and `partition_tbl`) into `populate_udf_table` instead of passing basic `SELECT ... FROM ...`.

shcheklein: hmm, seems the same as before, no?

dreadatour: I've checked, just in case.

Running `test_agg_works_after_union` from this PR with breakpoint on next line (`temp_tables.extend(t.name for t in tables)`):

```
In[1]: str(self.catalog.metastore.db.db_file)
Out[1]: '/private/var/folders/0f/9dz34qjs1gv62hv6rf5p6lkh0000gn/T/pytest-of-vlad/pytest-174/test_agg_works_after_union0/test.db'
In[2]: [t.name for t in tables]
Out[2]: ['udf_tPFbhM', 'udf_oVPRE2']
In[3]: print(query)
SELECT "udf_tPFbhM".sys__id, "udf_tPFbhM".sys__rand, "udf_tPFbhM".path, partition_id 
FROM "udf_tPFbhM" LEFT OUTER JOIN "udf_oVPRE2" ON "udf_oVPRE2".sys__id = "udf_tPFbhM".sys__id
```

Now in CLI:

```
$ sqlite3 /private/var/folders/0f/9dz34qjs1gv62hv6rf5p6lkh0000gn/T/pytest-of-vlad/pytest-174/test_agg_works_after_union0/test.db
sqlite> SELECT * FROM udf_tPFbhM LIMIT 5;
| sys__id |      sys__rand      |      path      |
|---------|---------------------|----------------|
| 1       | 8231059254174362419 | group-0/item-0 |
| 2       | 5550085067197592710 | group-1/item-1 |
| 3       | 7259957605758204303 | group-2/item-2 |
| 4       | 5854051897282815088 | group-3/item-3 |
| 5       | 7787358695868082018 | group-4/item-4 |
sqlite> SELECT * FROM udf_oVPRE2 LIMIT 5;
| sys__id |      sys__rand      | partition_id |
|---------|---------------------|--------------|
| 1       | 6241631001278032599 | 1            |
| 2       | 690157723433635761  | 2            |
| 3       | 2259346411529837530 | 3            |
| 4       | 674255673837620915  | 4            |
| 5       | 1150541152001007498 | 5            |
sqlite>
```

We do have 2 tables now with JOIN between them. In addition `partition_tbl` now have `sys__rand` column which is redundant for this table.

shcheklein: yep, so not issues with this I assume

Issue #1437, opened 2025-10-29
Wrong chain after `mutate` on `list[dc.DataModel]`


ai-ignatyev:
### Description

Let's consider the following code.

```python
import datachain as dc

class A(dc.DataModel):
    x: int

class B(dc.DataModel):
    x: list[A]

chain = dc.read_values(
    x=[
        B(x=[A(x=0)]),
        B(x=[A(x=1), A(x=2)]),
    ]
)

chain = chain.mutate(y=dc.C('x.x'))
chain.print_schema()
```

Real output:
```bash
x: B@v1
  x: list[A@v1]
    * list of: A@v1
      x: int
y: list[A@v1]
  * list of: A@v1
    x: int
```

Expected output:
```bash
x: B@v1
  x: list[A@v1]
    * list of: A@v1
      x: int
y: list[dict]
```

### Version Info

```Text
datachain -V
0.37.6

python -V
Python 3.12.9
```

shcheklein:
@ai-ignatyev hey, thanks for the report and a simple script to reproduce. If I'm not missing anything it should be other way around - we should be restoring it back to the list of models when we read it. So, mutate is probably correct, we just don't handle list of models (?). I'll double check and fix it asap.

shcheklein:
@ai-ignatyev please give a try again.

PR #1439, opened 2025-10-29
fix(models): support list, dicts of models

shcheklein:
Fixes https://github.com/iterative/datachain/issues/1437 and adds more support across different scenarios for lists / dicts / unions.

## Summary by Sourcery

Enable SignalSchema to handle list and dict annotations of Pydantic models (including optional and union types) during feature conversion, update downstream serialization in warehouse and values_to_tuples, and extend chain type detection to include dict types.

New Features:
- Support conversion of list[DataModel] and dict[key, DataModel] annotations in SignalSchema.row_to_features
- Allow optional and union types for collections when converting feature values

Enhancements:
- Introduce _convert_feature_value to recursively convert raw values into Pydantic models, lists, and dicts based on type hints
- Update warehouse.convert_type to serialize Pydantic models inside dicts and lists to JSON
- Enhance values_to_tuples to infer types for empty lists and dicts during tuple conversion
- Extend is_chain_type to recognize dict types as chain operations

Tests:
- Add unit tests for list/dict of models and optional/union collection scenarios in row_to_features
- Remove obsolete tests for iterable_chain, test_to_list_single_item, and test_collect_deprecated

### DIFF
--- a/src/datachain/data_storage/warehouse.py
+++ b/src/datachain/data_storage/warehouse.py
@@ -18,6 +18,7 @@ from datachain.data_storage.schema import convert_rows_custom_column_types
 from datachain.data_storage.serializer import Serializable
 from datachain.dataset import DatasetRecord, StorageURI
 from datachain.lib.file import File
+from datachain.lib.model_store import ModelStore
 from datachain.lib.signal_schema import SignalSchema
 from datachain.node import DirType, DirTypeGroup, Node, NodeWithPath, get_path
 from datachain.query.batch import RowsOutput
@@ -76,6 +77,29 @@ class AbstractWarehouse(ABC, Serializable):
     def cleanup_for_tests(self):
         """Cleanup for tests."""
 
+    def _to_jsonable(self, obj: Any) -> Any:
+        """Recursively convert Python/Pydantic structures into JSON-serializable
+        objects.
+        """
+
+        if ModelStore.is_pydantic(type(obj)):
+            return obj.model_dump()
+
+        if isinstance(obj, dict):
+            out: dict[str, Any] = {}
+            for k, v in obj.items():
+                if not isinstance(k, str):
+                    key_str = json.dumps(self._to_jsonable(k), ensure_ascii=False)
+                else:
+                    key_str = k
+                out[key_str] = self._to_jsonable(v)
+            return out
+
+        if isinstance(obj, (list, tuple, set)):
+            return [self._to_jsonable(i) for i in obj]
+
+        return obj
+
     def convert_type(  # noqa: PLR0911
         self,
         val: Any,
@@ -122,11 +146,13 @@ class AbstractWarehouse(ABC, Serializable):
             if col_python_type is dict or col_type_name == "JSON":
                 if value_type is str:
                     return val
-                if value_type in (dict, list):
-                    return json.dumps(val, ensure_ascii=False)
-                raise ValueError(
-                    f"Cannot convert value {val!r} with type {value_type} to JSON"
-                )
+                try:
+                    json_ready = self._to_jsonable(val)
+                    return json.dumps(json_ready, ensure_ascii=False)
+                except Exception as e:
+                    raise ValueError(
+                        f"Cannot convert value {val!r} with type {value_type} to JSON"
+                    ) from e
 
             if isinstance(val, col_python_type):
                 return val
--- a/src/datachain/lib/convert/values_to_tuples.py
+++ b/src/datachain/lib/convert/values_to_tuples.py
@@ -13,40 +13,152 @@ class ValuesToTupleError(DataChainParamsError):
         super().__init__(f"Cannot convert signals for dataset{ds_name}: {msg}")
 
 
-def values_to_tuples(  # noqa: C901, PLR0912
+def _find_first_non_none(sequence: Sequence[Any]) -> Any | None:
+    """Find the first non-None element in a sequence."""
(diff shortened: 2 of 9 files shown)


src/datachain/data_storage/warehouse.py
```diff
@@ -76,6 +77,52 @@ def __exit__(self, exc_type, exc_value, traceback) -> None:
...
+        # Collections
+        if isinstance(obj, dict):
+            out: dict[str, Any] = {}
+            for k, v in obj.items():
+                # Convert key to a JSON-friendly string
+                if isinstance(k, str):
+                    key_str = k
+                elif ModelStore.is_pydantic(type(k)) or isinstance(
+                    k, (dict, list, tuple, set)
+                ):
```
dreadatour: In what cases key can be a pydantic model or `(dict, list, tuple, set)`?

src/datachain/data_storage/warehouse.py
```diff
@@ -76,6 +77,52 @@ def __exit__(self, exc_type, exc_value, traceback) -> None:
...
+                    try:
+                        key_str = json.dumps(self._to_jsonable(k), ensure_ascii=False)
+                    except Exception:  # noqa: BLE001
+                        key_str = str(k)
+                elif isinstance(k, (int, float, bool)) or k is None:
+                    key_str = str(k)
+                else:
+                    key_str = str(k)
```
dreadatour: This can be simplified IMO:
```suggestion
                else:
                    key_str = str(k)
```

src/datachain/lib/convert/values_to_tuples.py
```diff
@@ -81,7 +81,18 @@ def values_to_tuples(  # noqa: C901, PLR0912
                         f" Please use DataModel types: {DataTypeNames}",
                     )
                 if isinstance(first_not_none_element, list):
-                    typ = list[type(first_not_none_element[0])]  # type: ignore[assignment, misc]
+                    if len(first_not_none_element) > 0:
+                        typ = list[type(first_not_none_element[0])]  # type: ignore[assignment, misc]
```
dreadatour: Can `first_not_none_element[0]` be `None`? 🤔

src/datachain/lib/convert/values_to_tuples.py
```diff
@@ -81,7 +81,18 @@ def values_to_tuples(  # noqa: C901, PLR0912
...
+                    else:
+                        typ = list  # type: ignore[assignment]
+                elif isinstance(first_not_none_element, dict):
+                    if len(first_not_none_element) > 0:
+                        # Get the first key-value pair to infer types
+                        first_key = next(iter(first_not_none_element.keys()))
+                        first_value = first_not_none_element[first_key]
```
dreadatour: Can `first_value` be `None`?

src/datachain/lib/signal_schema.py
```diff
@@ -585,6 +585,69 @@ def row_to_features(
...
+
+        if ModelStore.is_pydantic(annotation):
+            if isinstance(value, annotation):
+                obj = value
+            elif isinstance(value, Mapping):
+                obj = annotation(**value)
+            else:
+                return result
+            assert isinstance(obj, BaseModel)
+            SignalSchema._set_file_stream(obj, catalog, cache)
```
dreadatour, line 620: Not crucial for this PR, but this line confuses be a bit. Just something my eyes get stuck on and I can not understand without going deeper into the code and logic. May be this is because `set_file_stream` is not a best name for this method or it deserves a comment here.

src/datachain/lib/signal_schema.py
```diff
@@ -898,13 +961,13 @@ def _type_to_str(type_: type | None, subtypes: list | None = None) -> str:  # no
             args = get_args(type_)
             type_str = SignalSchema._type_to_str(args[0], subtypes)
             return f"Optional[{type_str}]"
-        if origin in (list, List):  # noqa: UP006
+        if origin is list:
```
dreadatour, line 967: If I am not mistaken, I've on purpose left this as is while deprecating Python 3.9 just in case someone still have models with `List` as a type (even in Python 3.10+ codebase).

It would be great to check if this case is still works with this change. Or we do agree we do not support `List` (and `Dict` below) types anymore.

shcheklein: I think it still works, we have tests for this / or they were added as part of this PR

tests/unit/lib/test_signal_schema.py
```diff
@@ -946,6 +1022,32 @@ def test_resolve_types():
             assert SignalSchema._resolve_type(s, {}) == t
 
 
+def test_type_to_str_typing_module_vs_builtin_generics():
```
shcheklein, line 1025: @dreadatour here is the test for List vs list

src/datachain/lib/signal_schema.py
shcheklein: yes, we have quite a few places around like this ... this is always needed when we deserialize Files 

the method name is not good, but it is wide-spread atm

src/datachain/lib/convert/values_to_tuples.py
shcheklein: It can be, needs to be fixed

src/datachain/lib/convert/values_to_tuples.py
shcheklein: same, probably can be

PR #1444, opened 2025-11-03
unify json usage, make sure datetime is handled

shcheklein:
A step towards better support of complex types in UDFs.

We can't now serialize list of Files since datetime is not supported. This unifies our json library (to use ujson) and adds support for datetime.

## Summary by Sourcery

Unify JSON handling across DataChain by introducing a central JSON utility based on ujson with extended type support, remove the old JSONSerialize encoder, and add tests to verify datetime serialization.

New Features:
- Add datachain/json module wrapping ujson with built-in support for datetime, date, time, bytes, memoryview, and UUID serialization

Bug Fixes:
- Ensure datetime objects are serialized correctly in SQL JSON conversions

Enhancements:
- Replace legacy JSONSerialize class with unified datachain.json across the codebase
- Update JSON dumps/loads calls to use datachain.json and drop cls=JSONSerialize arguments

Tests:
- Add unit test for to_json on storage Files to verify path, size, and last_modified serialization
- Update functional tests to remove JSONSerialize usage in preview data

### DIFF
--- a/src/datachain/json.py
+++ b/src/datachain/json.py
@@ -8,6 +8,7 @@ All code inside DataChain should import this module instead of using
 """
 
 import datetime as _dt
+import json as _json
 import uuid as _uuid
 from collections.abc import Callable
 from typing import Any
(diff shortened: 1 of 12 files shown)


src/datachain/json.py
```diff
@@ -0,0 +1,99 @@
...
+
+def _coerce(value: Any) -> Any:
+    """Return a JSON-serializable representation for supported extra types."""
+
+    if isinstance(value, (_dt.datetime, _dt.date, _dt.time)):
+        # ``datetime`` family classes expose ``isoformat`` with timezone support
+        # when available, so no additional handling is required here.
+        return value.isoformat()
+    if isinstance(value, (bytes, bytearray)):
+        return list(bytes(value)[:_MAX_PREVIEW_BYTES])
```
dreadatour: If I am not mistaken, this may affects storing bytes data in warehouse since it is limiting it to `_MAX_PREVIEW_BYTES`.

I'd check tests for `bytes` type stored in warehouse (if we have some) and in that tests try to store data with `_MAX_PREVIEW_BYTES + 1` bytes, for example.

shcheklein: this is only limits bytes serilized to JSON AFAIU ... or am I missing something else?

Issue #1450, opened 2025-11-06
Inconsistent `map` behavior depending on the input type


ai-ignatyev:
### Description

Quick heads-up: I am not trying to flood you with issues. These reports come from regular use, and I hope they will be helpful to you. Thank you all for your work!

Now, let's look at the following two code snippets. If we run the first snippet, we don't get any error.
```python
import datachain as dc

class X(dc.DataModel):
    x: int

chain = dc.read_values(x=[X(x=0), X(x=1)])
chain = chain.map(
    lambda x: x.model_dump(),
    params=['x'],
    output={'x': dict},
)
chain.show()
```

But if we ran the second snippet, it fails.
```python
import datachain as dc

chain = dc.read_values(x=[0, 1])
chain = chain.map(
    lambda x: {'x': x},
    params=['x'],
    output={'x': dict},
)
chain.show()
```

```bash
Traceback (most recent call last): 
  File "/home/aignatyev/main.py", line 9, in <module>
    chain.show()
  File "/home/aignatyev/.venv/lib/python3.12/site-packages/datachain/lib/dc/datachain.py", line 2013, in show
    df = dc.to_pandas(
         ^^^^^^^^^^^^^
  File "/home/aignatyev/.venv/lib/python3.12/site-packages/datachain/lib/dc/datachain.py", line 1986, in to_pandas
    results = self.results(include_hidden=include_hidden)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/aignatyev/.venv/lib/python3.12/site-packages/datachain/lib/dc/datachain.py", line 1468, in results
    return list(self._leaf_values(include_hidden=include_hidden))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/aignatyev/.venv/lib/python3.12/site-packages/datachain/lib/dc/datachain.py", line 1419, in _leaf_values
    with self._query.ordered_select(*db_signals).as_iterable() as rows:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.12/contextlib.py", line 137, in __enter__
    return next(self.gen)
           ^^^^^^^^^^^^^^
[...]

PR #1460, opened 2025-11-13
fix(udf): don't allow silent wrong schema merges

shcheklein:
Fixes https://github.com/iterative/datachain/issues/1450

Makes things more strict overall for schema merges after UDFs (and in other places where we do `schema1 | schema2`).

## Summary by Sourcery

Enforce strict schema merging to catch type mismatches and root conflicts in SignalSchema unions and dataset column additions, preventing silent schema errors in UDF pipelines

Bug Fixes:
- Prevent silent wrong schema merges by raising errors on type or root conflicts in schema unions and dataset column additions

Enhancements:
- Implement strict conflict detection in SignalSchema.__or__, raising errors for differing types or root duplicates while allowing identical and system signals
- Introduce root-level conflict checks and sorted overlap error messages when adding columns in dataset queries
- Refactor UDF map operation to compose the sys schema before merging with existing and output schemas

Tests:
- Add unit tests for SignalSchema union behavior (type mismatches, root conflicts, system signal rules) and functional tests for UDF schema override rejection

### DIFF
--- a/src/datachain/lib/dc/datachain.py
+++ b/src/datachain/lib/dc/datachain.py
@@ -850,14 +850,13 @@ class DataChain:
         if (prefetch := self._settings.prefetch) is not None:
             udf_obj.prefetch = prefetch
 
+        sys_schema = SignalSchema({"sys": Sys})
         return self._evolve(
             query=self._query.add_signals(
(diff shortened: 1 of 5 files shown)


dmpetrov (approved):
LG

PR #1463, opened 2025-11-16
convert booleans back on read

shcheklein:
Attempt to fix https://github.com/datachain-ai/datachain/issues/1357

### DIFF
--- a/src/datachain/sql/types.py
+++ b/src/datachain/sql/types.py
@@ -12,6 +12,7 @@ for sqlite we can use `sqlite.register_converter`
 ( https://docs.python.org/3/library/sqlite3.html#sqlite3.register_converter )
 """
 
+import numbers
 from datetime import datetime
 from types import MappingProxyType
 from typing import Any, Union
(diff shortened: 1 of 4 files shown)


src/datachain/sql/types.py
```diff
@@ -445,6 +446,18 @@ def string(self, value):
...
 
     def boolean(self, value):
+        if value is None or isinstance(value, bool):
+            return value
+
+        if isinstance(value, numbers.Integral):
+            return bool(value)
+        if isinstance(value, str):
+            normalized = value.strip().lower()
+            if normalized in {"true", "t", "yes", "y", "1"}:
```
amritghimire, line 456: What if sometimes user really wanted a value named yes in string format? Wouldn't that be unexpected for user if some rows with value yes was converted to boolean under assumption?

ilongin: Type read converter for boolean should be called only for Boolean column types from schema so I think this should be ok

amritghimire: I was just concerned from the test:

https://github.com/datachain-ai/datachain/pull/1463/files#diff-aa7469c93b1ceea63d0bd80dc827a3c1aedc85557a3780e0e7e10b1c40195459R13

What if there was a string with value yes for some row?

shcheklein: `read_csv` is using Arrow and automatically derives the type. It has separate `column_types` parameter to override that:

> Dictionary of column names and their corresponding types. It is passed to CSV reader and for each column specified type auto inference is disabled.

PR #1464, opened 2025-11-17
UDF error handling improvements

dreadatour:
Improve UDF error handling: return raw error from UDF to handle it properly in the parent process.


src/datachain/lib/udf.py
```diff
@@ -42,8 +40,44 @@
...
+
+    def __str__(self) -> str:
+        return f"{self.__class__.__name__!s}: {self.message!s}"
+
+    def __reduce__(self):
+        """Custom reduce method for pickling."""
+        return self.__class__, (self.message,)
+
+
+class UdfRunError(Exception):
```
dreadatour, line 57: A new exception type that is caught by the parent process, capturing both the traceback and the UDF name for display in the UI.

src/datachain/lib/udf.py
```diff
@@ -312,20 +346,6 @@ def _prepare_row_and_id(self, row, udf_fields, catalog, cache, download_cb):
...
-        except Exception as e:  # noqa: BLE001
-            msg = f"============== Error in user code: '{self.name}' =============="
-            print(msg)
-            exc_type, exc_value, exc_traceback = sys.exc_info()
-            traceback.print_exception(exc_type, exc_value, exc_traceback.tb_next)
-            print("=" * len(msg))
-            raise DataChainError(
-                f"Error in user code in class '{self.name}': {e!s}"
-            ) from None
-        return result_objs
```
dreadatour, line 327: No need to catch and log this here — the parent process now catches the exception and handles logging and user-facing output.

src/datachain/query/dispatch.py
```diff
@@ -159,9 +160,16 @@ def _run_worker(self) -> None:
...
+                # We put the exception into the done queue so the main process
+                # can handle it appropriately. We include the stacktrace to propagate
+                # it to the main process and show it to the user.
                 put_into_queue(
                     self.done_queue,
-                    {"status": FAILED_STATUS, "exception": e},
+                    {
+                        "status": FAILED_STATUS,
+                        "exception": e,
+                        "stacktrace": traceback.format_exc(),
```
dreadatour, line 171: Send also the traceback.

src/datachain/query/dispatch.py
```diff
@@ -316,7 +324,9 @@ def run_udf_parallel(  # noqa: C901, PLR0912
                 else:  # Failed / error
                     n_workers -= 1
                     if exc := result.get("exception"):
-                        raise exc
+                        if isinstance(exc, KeyboardInterrupt):
+                            raise exc
+                        raise UdfRunError(exc, stacktrace=result.get("stacktrace"))
```
dreadatour, line 329: `UdfRunError` wrapper for `exc` exception is needed to preserve the traceback.

tests/func/test_to_database.py
```diff
@@ -898,7 +898,7 @@ def process_item(id_val):
     ).map(process_item, params=["id"], output={"data": str})
 
     # The export should fail during processing, after table creation
-    with pytest.raises(dc.DataChainError, match="Processing failed on third item"):
+    with pytest.raises(ValueError, match="Processing failed on third item"):
```
dreadatour, line 901: Direct exception is now raised.

tests/unit/lib/test_checkpoints.py
```diff
@@ -2,12 +2,15 @@
...
 
 
-def mapper_fail(num) -> int:
-    raise Exception("Error")
+class CustomMapperError(Exception):
+    pass
+
+
+def mapper_fail(num: int) -> int:
+    raise CustomMapperError("Error")
```
dreadatour, line 13: This to check if we are raising and catching any exceptions as is.

.github/workflows/tests-studio.yml
```diff
@@ -69,6 +69,14 @@ jobs:
           ref: ${{ env.STUDIO_BRANCH }}
           token: ${{ secrets.ITERATIVE_STUDIO_READ_ACCESS_TOKEN }}
 
+      - uses: astral-sh/setup-uv@v7
```
shcheklein: why this change?

dreadatour: We are using this trick to test datachain and studio related changes from the git branch in case both repos have branches with the same name.

- Here, in datachain, we are already checking out changes from the desired branch.
- Next, we do check if Studio have the branch with the same name — we will checkout the same branch from Studio.
- But Studio itself have datachain as a dependency, and we need to make sure it will use the same datachain branch as in this PR.

Before Studio tests in datachain repo was broken sometimes because of this — rarely, only if API changes, but still. Same happens in this PR.

This change is not related directly to this PR, let me separate it 👍

dreadatour: For the records: same changes were applied here: https://github.com/datachain-ai/datachain/pull/1444/commits/4a0c6a1a52fc5793c252f21b150882d8e8eb8aa5

PR #1467, opened 2025-11-19
try to use latest torch deps

shcheklein:
Trying to see if it works now.


shcheklein:
> It looks like it works 👍

yep, 🤞  ... they might break it again, I guess ... for now it works. quite complicated set of packages.

PR #1470, opened 2025-11-20
UDF checkpoints for aggregator

ilongin:
Implementing UDF checkpoints for aggregator.


src/datachain/lib/udf.py
```diff
@@ -583,9 +583,22 @@ def run(
         download_cb: Callback = DEFAULT_CALLBACK,
         processed_cb: Callback = DEFAULT_CALLBACK,
     ) -> Iterator[Iterable[UDFResult]]:
+        from datachain.data_storage.schema import PARTITION_COLUMN_ID
+
         self.setup()
 
+        # Check if partition_id is available (when partition_by is used)
+        partition_id_idx = None
```
dreadatour, line 620:
```suggestion
        partition_id_idx: int | None = None
```

src/datachain/lib/udf.py
```diff
@@ -598,9 +611,12 @@ def run(
             ]
             result_objs = self.process_safe(udf_args)
             udf_outputs = (self._flatten_row(row) for row in result_objs)
-            output = (
-                dict(zip(self.signal_names, row, strict=False)) for row in udf_outputs
-            )
+            # Include sys__input_id to track which partition produced each output
+            output = [
+                {"sys__input_id": input_id}
```
dreadatour, line 645: Do we want to add `sys__input_id` as constant (same way we added `from datachain.data_storage.schema import PARTITION_COLUMN_ID`)?

src/datachain/query/dataset.py
```diff
@@ -640,9 +640,13 @@ def populate_udf_output_table(self, udf_table: "Table", query: Select) -> None:
...
 
-    def create_partitions_table(self, query: Select) -> "Table":
+    def create_partitions_table(self, query: Select, table_name: str) -> "Table":
         """
-        Create temporary table with group by partitions.
+        Create table with partition mappings (sys__id -> partition_id).
+
+        Args:
+            query: Input query with sys__id column
+            table_name: Name for the partition table.
```
dreadatour, line 779: ❤️

PR #1474, opened 2025-11-24
cleanup warnings, fail on any new warnings

shcheklein:
Fix resources cleanup. Mostly to avoid tons of warnings in tests, but also probably avoid some leaks.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -167,12 +167,6 @@ markers = [
 ]
 filterwarnings = [
   "error",  # Treat all warnings as errors by default
-  # Specific warnings we want to be strict about (redundant with error default, but explicit)
-  "error::pandas.errors.PerformanceWarning",
-  "error::pydantic.warnings.PydanticDeprecatedSince20",
-  "error::pytest_mock.PytestMockWarning",
-  "error::pytest.PytestCollectionWarning",
-  "error::sqlalchemy.exc.SADeprecationWarning",
   # Third-party deprecation warnings we can't fix
   "ignore::DeprecationWarning:timm.*",
   "ignore::DeprecationWarning:botocore.auth",


src/datachain/catalog/catalog.py
```diff
@@ -485,13 +497,36 @@ def get_init_params(self) -> dict[str, Any]:
         }
 
     def copy(self, cache=True, db=True):
+        """
+        Create a shallow copy of this catalog.
```
ilongin, line 501: Why is this kind of copy even needed? Why we just cannot copy the connections (metastore/warehoue objects) as well to avoid extra complexity and potential sources for strange / hard to debug issues.

src/datachain/data_storage/metastore.py
```diff
@@ -118,6 +119,16 @@ def close_on_exit(self) -> None:
         differently."""
         self.close()
 
+    @contextmanager
+    def _init_guard(self):
```
ilongin, line 123: little bit weird name IMO but cannot think of alternatives though. Maybe `_init_cleanup()` or `_on_failure_cleanup()`

ilongin: btw why did we need this as a special method? Maybe it would be easier to just move this code where it's used, since it's used only in one place....it will be clear what it does and saves reader for one jump / search for this method.
So something like:
```python
  try:
      self._init_meta_table()
      self._init_meta_schema_value()
      self._check_schema_version()
      self._init_tables()
      self._init_namespaces_projects()
  except Exception:
      with suppress(Exception):
          self.close_on_exit()
      raise
```

src/datachain/data_storage/sqlite.py
```diff
@@ -291,6 +291,8 @@ def cursor(self, factory=None):
         return self.db.cursor(factory)
 
     def close(self) -> None:
+        if self.is_closed:
```
ilongin, line 294: I think `close()` is idempotent so we can call it multiple times.

src/datachain/lib/pytorch.py
```diff
@@ -93,6 +94,10 @@ def __init__(
             self._cache = self._prefetch_cache
             weakref.finalize(self, self._prefetch_cache.destroy)
 
+        # Close the catalog if we created it - we only needed it for clone params
+        if owns_catalog:
```
ilongin, line 98: I don't like that we need to worry about every specific class to close / clean resources. It will be hard to manage all this which will lead to bugs. I think we should only work with `Session` as main object and we should only worry about it's cleaning / closing which would then close all the other specific resources like catalog / metastore / warehouse etc but those would be transparent to the user. This could be done in followup.

src/datachain/query/session.py
```diff
@@ -57,6 +57,7 @@ class Session:
 
     GLOBAL_SESSION_CTX: "Session | None" = None
     SESSION_CONTEXTS: ClassVar[list["Session"]] = []
+    _ALL_SESSIONS: ClassVar[WeakSet["Session"]] = WeakSet()
```
ilongin, line 60: What is the difference between `_ALL_SESSIONS` and `SESSION_CONTEXTS`  ? Aren't they redundant or I'm missing something?

src/datachain/catalog/catalog.py
shcheklein: it is indeed look redundant (there is some logic on CH side where we share connections, but this probably even not related directly). Proper / careful research is needed to remove the whole copy probably. Goes a bit beyond the intention of this PR

(close is still needed, context manager methods like __exit_, etc are also good to have to my mind since we have closeable resources anyways)

src/datachain/data_storage/metastore.py
shcheklein: I think, intention was to also use it in the CH warehouse (it requires a separate PR to cleanup resource leaks)

src/datachain/data_storage/sqlite.py
shcheklein: yep, but I think it fine to not abuse it if the call is not needed

src/datachain/lib/pytorch.py
shcheklein: agreed, session should be the way to go in most cases. Sometimes (tests) may be we will have to create catalog

shcheklein: resolving this, it goes way beyond the point of this PR (cleanup and make it strict)

src/datachain/query/session.py
shcheklein: SESSION_CONTEXTS is managed by `with Session ...` (enter / exit mechanism) to keep the current session at the top of the stack, etc. If someone directly creates a session w/o enter / exit - we sill want to clean it up at exit at least. In certain cases it might be hard to wrap all code into `with` (e.g. we pass it down and we don't control anymore its lifecycle, etc). It is also not 100% clear that we have to require from user always using `with`.

Some sessions might be created automatically (e.g. with some custom config) for users.

Even if we decide to make it very strict - it is also a bigger change that goes beyond this PR's intention to my mind.

pyproject.toml
```diff
@@ -166,16 +166,17 @@ markers = [
   "incremental_processing: Delta and retry examples"
 ]
 filterwarnings = [
-  "error::pandas.errors.PerformanceWarning",
-  "error::pydantic.warnings.PydanticDeprecatedSince20",
-  "error::pytest_mock.PytestMockWarning",
-  "error::pytest.PytestCollectionWarning",
-  "error::sqlalchemy.exc.SADeprecationWarning",
+  "error",  # Treat all warnings as errors by default
```
dreadatour, line 169: This is awesome, thank you! 🔥

src/datachain/catalog/catalog.py
```diff
@@ -545,41 +580,49 @@ def _remove_dataset_rows_and_warehouse_info(
...
-                nodes = [Node.from_file(client.get_file_info(file_path))]
-                dir_only = False
-            else:
-                nodes = listing.expand_path(file_path)
-                dir_only = file_path.endswith("/")
-            dsrc_all.extend(
-                DataSource(listing, client, node, dir_only) for node in nodes
-            )
-        return dsrc_all
+    ) -> Iterator[list["DataSource"] | None]:
```
dreadatour, line 591: I think we might want to improve this in a follow-up. Iterator returning a list or None looks a bit odd to me. I may be wrong, anyway, it looks like it is a bit out of the scope of this PR.

src/datachain/query/session.py
ilongin: Ok, make sense since we shouldn't enforce usage of context manager.

dreadatour (approved):
Looks good to me overall. A lot of changes, hard to review them all in one go. I am totally OK with merging this and move forward. Added one tiny non-related comment.

PR #1476, opened 2025-11-25
Set dataset version default status to 'CREATED'

dreadatour:
Default dataset status is `CREATED` while default dataset version status is `PENDING`. I am not sure what this `PENDING` status even means — no usage for this status was found in the codebase.

Setting default dataset version status to `CREATED` for the consistency.

Also renamed `create_new_dataset_version` to `create_dataset_version`, since "create new" looks a bit odd.

See also related Studio PR.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -18,7 +18,7 @@ jobs:
   run:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
       - name: Set up Python 3.13
         uses: actions/setup-python@v6
         with:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,7 +17,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Check out the repository
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
         with:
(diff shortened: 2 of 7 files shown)


shcheklein:
I think in case of Datasets CREATED sounds like COMPLETED - thus we had pending (?) ... not sure we need to change that, but not strong opinion either

dreadatour:
> I think in case of Datasets CREATED sounds like COMPLETED - thus we had pending (?) ... not sure we need to change that, but not strong opinion either

For datasets we also have `COMPLETE` status and we do have a lot of checks for this in the code. By default dataset is creating with `CREATED` status and after all it became `COMPLETE`.

My guess is `PENDING` was used for the storage indexing, I can go deeper in git history to find where it was used.

For now `DatasetStatus.PENDING` is not used anywhere, — only as a default status during dataset version creation, and it is updating to `COMPLETE` in the end.

I am going to go deeper with these statuses, like for example, why do we need `finished_at` based on statuses for dataset versions ([here](https://github.com/datachain-ai/datachain/blob/main/src/datachain/data_storage/metastore.py#L1054-L1057))? But a bit later.

shcheklein:
@dreadatour good to go?

dreadatour:
> @dreadatour good to go?

Yes! I was waiting for a review. I am going to merge it now, thank you! 🙏

PR #1480, opened 2025-11-26
Add many-to-many relationship between dataset versions and jobs

ilongin:
When a dataset is created in a job that fails, re-running the script creates a new job but the dataset still references the old (failed) job. This makes the dataset invisible in subsequent job runs in Studio UI.
This PR implements a many-to-many relationship between dataset versions and jobs using a dataset_version_jobs junction table with an is_creator flag. It also fixes current issue in checkpoint code - when checkpoint is found what dataset version to return? Currently we return latest version but that's wrong as it might not be the one that is created in this job chain (it may be created by some other random job) so in this PR we try to find the correct version created in this specif job chain / hierarchy.
  
Changes:
  - Added `dataset_version_jobs` table to metastore schemas
  - Added `link_dataset_version_to_job()`, `get_ancestor_job_ids()`, and `get_dataset_version_for_job_ancestry()` metastore methods
  - Updated `DataChain._resolve_checkpoint()` to use job ancestry for finding dataset versions
  - Updated `DatasetQuery.save()` to link datasets to jobs on creation

The existing `job_id` column in `dataset_versions` table remains unchanged for backward compatibility and can be deprecated in a future release.
Jobs that create datasets are linked with `is_creator=True`, while jobs that reuse datasets via checkpoints are linked with `is_creator=False`.

### DIFF
--- a/src/datachain/lib/dc/datachain.py
+++ b/src/datachain/lib/dc/datachain.py
@@ -647,6 +647,9 @@ class DataChain:
 
         # Checkpoint handling
         _hash, result = self._resolve_checkpoint(name, project, job, kwargs)
+        if bool(result):
+            # Checkpoint was found and reused
+            print(f"Checkpoint found for dataset '{name}', skipping creation")
 
         # Schema preparation
         schema = self.signals_schema.clone_without_sys_signals().serialize()
@@ -669,9 +672,6 @@ class DataChain:
                     **kwargs,
                 )
             )
-        else:
-            # Checkpoint was found and reused
-            print(f"Checkpoint found for dataset '{name}', skipping creation")
 
         catalog.metastore.create_checkpoint(job.id, _hash)  # type: ignore[arg-type]
         return result
--- a/tests/func/test_delta.py
+++ b/tests/func/test_delta.py
@@ -594,7 +594,10 @@ def test_delta_update_from_storage(test_session, tmp_dir, tmp_path):
     ).to_values("file.etag")[0] > etags[6]
 
 
-def test_delta_update_check_num_calls(test_session, tmp_dir, tmp_path, capsys):
+def test_delta_update_check_num_calls(
+    test_session, tmp_dir, tmp_path, capsys, monkeypatch
+):
+    monkeypatch.setenv("DATACHAIN_CHECKPOINTS_RESET", "True")
     ds_name = "delta_ds"
     path = tmp_dir.as_uri()
     tmp_dir = tmp_dir / "images"


src/datachain/data_storage/metastore.py
```diff
@@ -484,6 +488,38 @@ def create_checkpoint(
...
+        self,
+        dataset_version_id: int,
+        job_id: str,
+        is_creator: bool = False,
+        conn=None,
+    ) -> None:
+        """Link dataset version to job (many-to-many)."""
+
+    @abstractmethod
+    def get_ancestor_job_ids(self, job_id: str, conn=None) -> list[str]:
```
amritghimire, line 525: Will this be sorted always? A helper to get the latest ancestor job id would be helpful/

src/datachain/data_storage/metastore.py
```diff
@@ -484,6 +488,38 @@ def create_checkpoint(
...
+
+    @abstractmethod
+    def link_dataset_version_to_job(
+        self,
+        dataset_version_id: int,
+        job_id: str,
+        is_creator: bool = False,
+        conn=None,
+    ) -> None:
+        """Link dataset version to job (many-to-many)."""
```
amritghimire: May be redundant.

src/datachain/data_storage/metastore.py
```diff
@@ -1850,6 +1885,38 @@ def _checkpoints(self) -> "Table":
...
+    @cached_property
+    def _dataset_version_jobs(self) -> "Table":
+        return Table(
+            self.DATASET_VERSION_JOBS_TABLE,
+            self.db.metadata,
+            *self._dataset_version_jobs_columns(),
+        )
+
+    @abstractmethod
+    def _dataset_version_jobs_insert(self) -> "Insert": ...
```
amritghimire, line 1939: Lets add _dataset_version_jobs_select, delete and other functions too.

src/datachain/data_storage/metastore.py
```diff
@@ -1850,6 +1885,38 @@ def _checkpoints(self) -> "Table":
...
+        """Junction table for dataset versions and jobs many-to-many relationship."""
+        return [
+            Column("id", Integer, primary_key=True),
+            Column(
+                "dataset_version_id",
+                Integer,
+                nullable=False,
+            ),
+            Column("job_id", Text, nullable=False),
+            Column("is_creator", Boolean, nullable=False, default=False),
```
amritghimire, line 1920: What exactly is this for?

src/datachain/data_storage/metastore.py
```diff
@@ -1928,3 +1995,117 @@ def get_last_checkpoint(self, job_id: str, conn=None) -> Checkpoint | None:
...
+                self._jobs.join(
+                    ancestors_cte,
+                    self._jobs.c.id
+                    == cast(ancestors_cte.c.parent_job_id, self._jobs.c.id.type),
+                )
+            )
+        )
+
+        # Select all ancestor IDs except the starting job itself
+        query = select(ancestors_recursive.c.id).where(
```
amritghimire: Lets add a recursive limit in case we go into circular dependency somehow.

src/datachain/data_storage/metastore.py
```diff
@@ -1928,3 +1995,117 @@ def get_last_checkpoint(self, job_id: str, conn=None) -> Checkpoint | None:
...
+
+    def _get_dataset_version_for_job_ancestry_query(
+        self,
+        dataset_name: str,
+        namespace_name: str,
+        project_name: str,
+        job_ancestry: list[str],
+    ) -> "Select":
+        """Build query to find dataset version created by job ancestry."""
+        return (
```
amritghimire, line 2132: Maybe split this into variables. looks too complex otherwise

src/datachain/data_storage/metastore.py
ilongin: To distinguish if this particular job actually created new dataset version or just reused it (because of checkpoints). It's not that important now but it maybe might be important in future so I added it right away to not loose info.

src/datachain/data_storage/metastore.py
ilongin: I actually like it better like this. Adding variables would just complicate things IMO as now I would need to think about them and their names - it would complicate my mental model. Now, it's formatted in a way that you can right away see all the joins and what query does.

src/datachain/data_storage/metastore.py
ilongin: added

src/datachain/data_storage/metastore.py
ilongin: added

src/datachain/data_storage/metastore.py
ilongin: It's not but ATM it doesn't need to be. If we will need to have it sorted in future we can easily add this as well

src/datachain/data_storage/metastore.py
```diff
@@ -1928,3 +2009,126 @@ def get_last_checkpoint(self, job_id: str, conn=None) -> Checkpoint | None:
...
+                index_elements=["dataset_version_id", "job_id"]
+            )
+        self.db.execute(query, conn=conn)
+
+    def get_ancestor_job_ids(self, job_id: str, conn=None) -> list[str]:
+        # Use recursive CTE to walk up the parent chain
+        # Format: WITH RECURSIVE ancestors(id, parent_job_id, depth) AS (...)
+        # Include depth tracking to prevent infinite recursion in case of
+        # circular dependencies
+        max_depth = 1000
```
amritghimire: I think 100 is more appropriate, 1000 looks too much.

src/datachain/catalog/catalog.py
```diff
@@ -1581,7 +1582,7 @@ def _instantiate(ds_uri: str) -> None:
             leave=False,
         )
 
-        schema = DatasetRecord.parse_schema(remote_ds_version.schema)
```
shcheklein, line 1629: unrelated change, cleanup, remove

ilongin: It's not unrelated. Because that `parse_schem`a was inside `DatasetRecord` I needed to state each single row column when parsing results from database in `get_dataset_version_for_job_ancestry` instead of just passing all columns as one variable. Anyway, it's better to have it as a separate method - no reason to be under DatasetRecprd

src/datachain/data_storage/metastore.py
```diff
@@ -1928,3 +2009,126 @@ def get_last_checkpoint(self, job_id: str, conn=None) -> Checkpoint | None:
...
         return self.checkpoint_class.parse(*rows[0])
+
+    def link_dataset_version_to_job(
+        self,
+        dataset_version_id: int,
+        job_id: str,
+        is_creator: bool = False,
+        conn=None,
+    ) -> None:
+        query = self._dataset_version_jobs_insert().values(
```
shcheklein: do we check that both objects are actually related and belong to the same team?

src/datachain/data_storage/metastore.py
```diff
@@ -1928,3 +2009,126 @@ def get_last_checkpoint(self, job_id: str, conn=None) -> Checkpoint | None:
...
+                self._jobs.c.parent_job_id.label("parent_job_id"),
+                (ancestors_cte.c.depth + 1).label("depth"),
+            ).select_from(
+                self._jobs.join(
+                    ancestors_cte,
+                    (
+                        self._jobs.c.id
+                        == cast(ancestors_cte.c.parent_job_id, self._jobs.c.id.type)
+                    )
+                    & (ancestors_cte.c.depth < max_depth),
```
shcheklein: what happens if depth is exceeded? do we silently pretty much fail / don't behave as expected?

src/datachain/data_storage/metastore.py
```diff
@@ -1928,3 +2009,126 @@ def get_last_checkpoint(self, job_id: str, conn=None) -> Checkpoint | None:
...
+
+    def get_ancestor_job_ids(self, job_id: str, conn=None) -> list[str]:
+        # Use recursive CTE to walk up the parent chain
+        # Format: WITH RECURSIVE ancestors(id, parent_job_id, depth) AS (...)
+        # Include depth tracking to prevent infinite recursion in case of
+        # circular dependencies
+        max_depth = 100
+
+        ancestors_cte = (
+            self._jobs_select(
```
shcheklein, line 2068: does _jobs_select filter by team? do we have a test for this?

src/datachain/data_storage/metastore.py
```diff
@@ -1928,3 +2009,126 @@ def get_last_checkpoint(self, job_id: str, conn=None) -> Checkpoint | None:
...
+            is_creator=is_creator,
+            created_at=datetime.now(timezone.utc),
+        )
+        if hasattr(query, "on_conflict_do_nothing"):
+            query = query.on_conflict_do_nothing(
+                index_elements=["dataset_version_id", "job_id"]
+            )
+        self.db.execute(query, conn=conn)
+
+    def get_ancestor_job_ids(self, job_id: str, conn=None) -> list[str]:
```
shcheklein, line 2062: what is the perf of such query ? did you test it on a large DB?

src/datachain/data_storage/metastore.py
```diff
@@ -1928,3 +2009,126 @@ def get_last_checkpoint(self, job_id: str, conn=None) -> Checkpoint | None:
...
+        return [str(row[0]) for row in results]
+
+    def _get_dataset_version_for_job_ancestry_query(
+        self,
+        dataset_name: str,
+        namespace_name: str,
+        project_name: str,
+        job_ancestry: list[str],
+    ) -> "Select":
+        """Build query to find dataset version created by job ancestry."""
```
shcheklein: for consistency - review here in other places docstrings ... is there some logic for them to be missing or exist and in different formats

src/datachain/data_storage/metastore.py
```diff
@@ -1928,3 +2009,126 @@ def get_last_checkpoint(self, job_id: str, conn=None) -> Checkpoint | None:
...
+        return (
+            self._datasets_versions_select()
```
shcheklein, line 2133: do both selects filter by team id already? do we have tests for this?

can we actually make all those helpers raise if metastore doesn't implement them?

src/datachain/data_storage/metastore.py
shcheklein: it's not informative (duplicates name) ... what does it actually do? it is not clear

src/datachain/data_storage/metastore.py
```diff
@@ -1928,3 +1995,117 @@ def get_last_checkpoint(self, job_id: str, conn=None) -> Checkpoint | None:
...
+                    self._namespaces,
+                    self._projects.c.namespace_id == self._namespaces.c.id,
+                )
+            )
+            .where(
+                self._datasets.c.name == dataset_name,
+                self._namespaces.c.name == namespace_name,
+                self._projects.c.name == project_name,
+                self._dataset_version_jobs.c.job_id.in_(job_ancestry),
+                self._dataset_version_jobs.c.is_creator == True,  # noqa: E712
```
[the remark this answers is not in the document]
shcheklein: @ilongin ping

src/datachain/data_storage/metastore.py
```diff
@@ -1928,3 +2009,126 @@ def get_last_checkpoint(self, job_id: str, conn=None) -> Checkpoint | None:
...
+
+    def get_dataset_version_for_job_ancestry(
+        self,
+        dataset_name: str,
+        namespace_name: str,
+        project_name: str,
+        job_id: str,
+        conn=None,
+    ) -> DatasetVersion | None:
+        # Get job ancestry (current job + all ancestors)
```
shcheklein, line 2172: one more comment that look different from everything above

src/datachain/data_storage/metastore.py
```diff
@@ -1928,3 +2009,126 @@ def get_last_checkpoint(self, job_id: str, conn=None) -> Checkpoint | None:
...
+
+        query = self._get_dataset_version_for_job_ancestry_query(
+            dataset_name, namespace_name, project_name, job_ancestry
+        )
+
+        results = list(self.db.execute(query, conn=conn))
+        if not results:
+            return None
+
+        return self.dataset_version_class.parse(*results[0])
```
shcheklein, line 2188: raise an exception if there are > 1

src/datachain/lib/dc/datachain.py
```diff
@@ -712,10 +712,41 @@ def _resolve_checkpoint(
             and not checkpoints_reset
             and metastore.find_checkpoint(job.parent_job_id, _hash)
         ):
-            # checkpoint found → reuse dataset
+            # checkpoint found → find which dataset version to reuse
+
+            # Find dataset version that was created by any ancestor job
+            dataset_version = metastore.get_dataset_version_for_job_ancestry(
```
shcheklein: can we add inside those implementations some logs to actually trace it then (if job was found or not), if found - print its id, etc

src/datachain/lib/dc/datachain.py
```diff
@@ -712,10 +712,41 @@ def _resolve_checkpoint(
...
+                name,
+                project.namespace.name,
+                project.name,
+                job.id,
+            )
+
+            if not dataset_version:
```
shcheklein, line 738: do we have a test for this?

src/datachain/lib/dc/datachain.py
```diff
@@ -712,10 +712,41 @@ def _resolve_checkpoint(
...
+                version=dataset_version.version,
+                **kwargs,
+            )
+
+            # Link current job to this dataset version (not creator)
+            metastore.link_dataset_version_to_job(
+                dataset_version.id, job.id, is_creator=False
             )
+
+            # Update dataset_version.job_id to point to the latest job
```
shcheklein: why do we need this pointer? 

we need then think about transaction here (link above and this update)

src/datachain/query/dataset.py
```diff
@@ -1989,6 +1991,13 @@ def save(
             )
             self.catalog.update_dataset_version_with_warehouse_info(dataset, version)
 
+            # Link this dataset version to the job that created it
+            if job_id and dataset.has_version(version):
```
shcheklein: how can it be that we don't have job_id?

how can it be we don't have this version?

tests/func/test_metastore.py
```diff
@@ -907,3 +907,39 @@ def test_get_job_status(metastore):
...
+            workers=1,
+            parent_job_id=parent_id,
+        )
+        job_ids.append(job_id)
+        parent_id = job_id
+
+    # The last job is the leaf (youngest)
+    leaf_job_id = job_ids[-1]
+
+    ancestors = metastore.get_ancestor_job_ids(leaf_job_id)
```
shcheklein, line 939: what happens if one of the jobs is not found? a job in the middle

src/datachain/data_storage/metastore.py
ilongin: Yes, it does filter by team_id. Safest thing is always use these helper methods for any kind of start of the query as it's then overriden in Studio to include team_id filter. We generally don't have explicit tests for these methods .. we should think of what is the best way to add them as they are not really for datachain repo.

src/datachain/data_storage/metastore.py
ilongin: For some reason I thought exception will be raised but it was silently failing. Now I've added whole error handling of this including test.

src/datachain/data_storage/metastore.py
ilongin: Excellent performance. ~1ms for lookup in job chain of depth 99 (our maximum atm) and with table of 100k jobs inside.

src/datachain/data_storage/metastore.py
ilongin: we usually put docstrings in abstract methods. This one is helper method so I've added simple doc to explain.

src/datachain/data_storage/metastore.py
ilongin: `DatasetVersion` table is one of those old tables which doesn't have copied `team_id` but has connection to Dataset which does have `team_id`. In this case, whole `_get_dataset_version_for_job_ancestry_query` method is overriden in Studio by adding `team_id` filter since we couldn't do the same with `_datasets_versions_select`

ilongin: Regarding other question: Since this repeats for all models (we have the same problem) I think we need separate task to:
1. Think of how to automate this or raise exception or something as you mentioned
2. Write tests for it

src/datachain/data_storage/metastore.py
ilongin: Added better comment

src/datachain/data_storage/metastore.py
ilongin: This is not pydoc but comment for specific line below. Pydocs are added to abstract methods

src/datachain/data_storage/metastore.py
ilongin: Query returns only one record i.e. has limit so that check would be redundant. Even if we don't have limit it's ok for query to return multiple results - that would mean we created multiple versions of the same dataset and we are picking the last one created.

src/datachain/lib/dc/datachain.py
ilongin: Added  couple of debugs

src/datachain/lib/dc/datachain.py
ilongin: No, added it

src/datachain/lib/dc/datachain.py
ilongin: I left this since it will be used in pipelines (this was the main issue for that task) and to be backward compatible but yes, this is now redundant and should be deprecated or even maybe removed .I will see if pipelines can just use the same `get_dataset_version_for_job_ancestry` method

src/datachain/query/dataset.py
ilongin: When we create temp datasets with .persist we don't track that with a job so job_id is not defined.
Second check for dataset version is redundant - removing it.

src/datachain/data_storage/metastore.py
shcheklein: resolving this, since it is stylistic and we can let author decide

src/datachain/data_storage/metastore.py
shcheklein: * we should be raising an exception by default in each of those helpers (do we do this?) in abstract metastore
* we can have test on the Studio side - we already have plenty of tests

src/datachain/data_storage/metastore.py
shcheklein: I know, it is just some extra defense mechanism

src/datachain/query/dataset.py
shcheklein: hmm, a lot of questions ...

* why do we do things like `self.catalog.update_dataset_version_with_warehouse_info(dataset, version)` for persist()?
* if this is a difference between different types - it is better to have some explicit flag / name. It is weird that I have to remember not to pass job_id in persist. Or what we decide to refactor and always get it here instead of providing it?

src/datachain/lib/dc/datachain.py
shcheklein: let's remove it then for now, and add separately? otherwise we have a risk on leaving as is and fighting then with some half-broken state (I saw this happening already)

src/datachain/lib/dc/datachain.py
shcheklein: I means Studio primarily btw, not datachain (or datachain if we know that it will be visible on backend in Studio)

tests/func/test_metastore.py
ilongin: it would return partial list of jobs -to the point job is removed. In checkpoints, this could result in dataset not being found in ancestors which would result in re-calculating it from the scratch. I think that's ok.

src/datachain/data_storage/metastore.py
ilongin: Both objects are guaranteed to belong to the same team because they're both created/fetched through the same metastore instance which is scoped to a `team_id`. The method is only called from internal business logic within the same session/team context, not from any external API that could pass arbitrary IDs.  Therefor I don't think we need to add this kind of checks everywhere.... what is enough is that those CRUD database functions e.g `_dataset_version_jobs_insert` are enforced to be used and cannot be bypassed but this is way out of the scope of this PR .. I think we need another issue with adding those and related Studio tests and anything else that's needed.

src/datachain/data_storage/metastore.py
ilongin: As mentioned in related comment https://github.com/datachain-ai/datachain/pull/1480#discussion_r2584485205 we need separate task for this as in every PR we have similar questions / issues.

src/datachain/data_storage/metastore.py
ilongin: added

src/datachain/lib/dc/datachain.py
ilongin: I was working on this for the past day but changes on Studio are so complex and affect performance of couple of queries there (e.g fetching datasets, dependencies etc.) that I suggest to not to this. Yes, we will have one redundant field but we will most likely need this field again when we do proper solution of having datasets marked as "pending" or whatever until job is successfully done. I opened datachain and Studio draft PRs just for the reference but I think we should close them and finally merge this PR

src/datachain/query/dataset.py
ilongin: Actually the issue here was that I wasn't connecting temp dataset created with .persist() with a job which is actually the case now in main branch - there was no reason to change this so I'm returning the logic to connect it to a job .. t his then solves the related issue off even need for this condition i.e. it's not needed any more so it's all ok now I think.

src/datachain/lib/dc/datachain.py
shcheklein: what happens if we fail in between of those calls?

src/datachain/data_storage/metastore.py
shcheklein: is it resolved? added tests? etc?

src/datachain/data_storage/metastore.py
shcheklein: hmm ... I think in this case it might be going a bit beyond basic API call issue / us using or not using CRUD helpers. Is it possible for a user from a random team call this with some values from a different team (can be guessed)? AFAIU it was not and is not possible in all other API calls (e.g. when we select we check team id, so we never select some random external value, when we create objects we also set team_id, not sure if we have create or select but probably there it also checks for team_id). Here is the first time I see a bit different beast - we just put into DB directly any provided values. It can potentially lead to cascade removals, or us exposing some objects, etc. Again, not sure we had anything like this before. And this seems like another important security consideration.

src/datachain/data_storage/metastore.py
shcheklein: > in every PR we have similar questions / issues.

or we can just start adding them in each PR? or do you think it is possible to have a single test that covers everything. Having a separate task doesn't contradict anything in this case. Just less tech debt to catch up with, unless I'm missing something. Same with raising - we can do this for helpers that are in scope of this PR, right?

src/datachain/lib/dc/datachain.py
ilongin: ~~I've added transaction on top of those, but in general our whole code is kind of without proper transactions - we had task for this in old repo but never got to work on it. Before I think we had some issues of transaction objects not being exposed to outer scope outside metastore but now it should be easier to use it in catalog, datachain and in other places. I will create separate issue to go through codebase and add transactions everywhere where it's missing~~

EDIT: both of those calls are now in single metastore method under transaction.

src/datachain/data_storage/metastore.py
ilongin: Added couple of tests testing team isolation in companion Studio branch.

src/datachain/data_storage/metastore.py
ilongin: I've added couple of tests in Studio

src/datachain/data_storage/metastore.py
ilongin: I think we need to separate two things here:
1. Business logic validation - create defensive code to avoid bug in a system that could lead to what you described. I will add additional validation, even though it's probably not needed.
2. Someone tries to hack the system -  there is not defense here. If we add validation in this or any other function, it can still be bypassed by user running query directly on DB.  When we were speaking about DB isolation to prevent this kind of issues at beginning of datachain main argument was that we will only have on-prem users where we don't have this issues in the first place since every customer will have it's own database etc., but it seems like it's not the case right now. I think we need to brainstorm how to avoid this kind of things, if they can be avoided in the first place.

shcheklein: >it can still be bypassed by user running query directly on DB.

How? 

I assume users can access PG DB (metastore) only via API from a cluster. They can't access it directly. Am I missing something here?

src/datachain/data_storage/metastore.py
shcheklein: thanks!

src/datachain/data_storage/metastore.py
shcheklein: can we also raise exception in those helpers that we added within this PR? (if there is not implementation provided we don't just pass silently?

src/datachain/lib/dc/datachain.py
shcheklein: can we check that we have enough logs and details on the Studio side please one more time

src/datachain/lib/dc/datachain.py
```diff
@@ -712,10 +715,66 @@ def _resolve_checkpoint(
...
+                name,
+                namespace=project.namespace.name,
+                project=project.name,
+                version=dataset_version.version,
+                **kwargs,
             )
+
+            # Link current job to this dataset version (not creator) and update
+            # dataset_version.job_id atomically
+            with metastore.db.transaction() as conn:  # type: ignore[attr-defined]
```
shcheklein: how would this transaction work in Studio? (the main concern) ... I'm not sure I understand this

src/datachain/data_storage/metastore.py
ilongin: Maybe I'm missing something... @dreadatour is `APIMetastore` really helping to avoid these kind of issues in all cases?

shcheklein: No, it doesn't and bites us if you consumer code is not taking lack of transactions into account. 

Is there a way to avoid two (or more) calls? Is  there a way to design them in a way that it doesn't matter if it fails in the middle? Otherwise is there a way to make it a single call?

src/datachain/data_storage/metastore.py
```diff
@@ -1928,3 +2007,150 @@ def get_last_checkpoint(self, job_id: str, conn=None) -> Checkpoint | None:
...
+        the dataset where is_creator=True. Returns newest by created_at, or
+        None if no version was created by any job in the ancestry chain.
+
+        Used for checkpoint resolution to find which version to reuse when
+        continuing from a parent job.
+        """
+        return (
+            self._datasets_versions_select()
+            .select_from(
+                self._dataset_version_jobs.join(
```
amritghimire, line 2135: Does this pass on the team id filter or jobs too?

src/datachain/data_storage/metastore.py
```diff
@@ -1928,3 +2007,150 @@ def get_last_checkpoint(self, job_id: str, conn=None) -> Checkpoint | None:
...
+                    self._datasets_versions,
+                    self._dataset_version_jobs.c.dataset_version_id
+                    == self._datasets_versions.c.id,
+                )
+                .join(
+                    self._datasets,
+                    self._datasets_versions.c.dataset_id == self._datasets.c.id,
+                )
+                .join(
+                    self._projects,
```
amritghimire, line 2145: Same for projects, namespaces too

src/datachain/data_storage/sqlite.py
```diff
@@ -471,6 +471,8 @@ def _init_tables(self) -> None:
         self.default_table_names.append(self._jobs.name)
         self.db.create_table(self._checkpoints, if_not_exists=True)
         self.default_table_names.append(self._checkpoints.name)
+        self.db.create_table(self._dataset_version_jobs, if_not_exists=True)
```
amritghimire, line 477: What is the chance that we would end up with large table locally for dataset versions jobs if we don't cleanup properly? I mean every run will have separate job id locally and we tend to run many scripts faster locally.

src/datachain/data_storage/metastore.py
ilongin: Yes, this whole method actually is extended in Studio metastore to add `team_id` filter on that table.

src/datachain/data_storage/metastore.py
ilongin: We don't need to filter by `team_id` every element of a join. Usually we validate only the starting point, e.g here we validate dataset version `team_id` and all other's are related objects which belong to the same team.

src/datachain/data_storage/sqlite.py
ilongin: I don't see this as an issue atm. this table is small (just many to many link with one more field) and in future we will implement proper solution of "draft" datasets that won't become actual one until job is successful

src/datachain/data_storage/metastore.py
ilongin: I updated this method to do both linking dataset version to a job and update dataset job_id in single transaction.

src/datachain/lib/dc/datachain.py
ilongin: You are right, I forgot that we use APIMetastore everywhere, not just in UDFs. So the only solution here is to make `link_dataset_version_to_job` do both of those things and wrap in transaction. This way we avoud multiple API calls and not being able to do it atomically.

src/datachain/data_storage/metastore.py
ilongin: I have some ideas but I would like to leave it for different / followup PR since this one is already too big and solution for that should be generic  / applied to all methods we have

src/datachain/lib/dc/datachain.py
shcheklein: is it done? just to confirm ...

src/datachain/lib/dc/datachain.py
shcheklein: resolved?

src/datachain/lib/dc/datachain.py
ilongin: Yep

PR #1483, opened 2025-11-29
Remove 'datachain query' command

dreadatour:
Remove `datachain query` command since it is not used anymore. It was required before because of "last statement", but now instead of running `datachain query script.py` one can run `python script.py`.

Also removed dead unused code, except for `process_output` — it was moved to Studio repo, since it is used in Studio (see related PR).

---

One downside. `datachain query` supports `--param` argument to set query params, see here: https://github.com/iterative/dvcx/pull/1362, but we don't use it anyway.

May be we need to think about other way to set query params, like env variables or query args.

### DIFF
--- a/src/datachain/catalog/__init__.py
+++ b/src/datachain/catalog/__init__.py
@@ -1,7 +1,6 @@
 from .catalog import (
     QUERY_DATASET_PREFIX,
     QUERY_SCRIPT_CANCELED_EXIT_CODE,
-    QUERY_SCRIPT_INVALID_LAST_STATEMENT_EXIT_CODE,
     Catalog,
     is_namespace_local,
 )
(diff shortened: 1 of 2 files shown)


tests/func/test_query.py
```diff
@@ -1,19 +1,10 @@
-import os.path
```
amritghimire, line 1: Can this file be removed? Does it have tests anymore?

shcheklein (approved):

tests/func/test_query.py
dreadatour: Nice catch! 🙏

dreadatour: Removed it, thank you for noticing that.

PR #1490, opened 2025-12-04
fix(file): exclude catalog before serializing

shcheklein:
Make File cloudpick-able.

We need this to be able to do `read_values(file=[File, File ....])` since internally it is using UDF atm and hits cloudpickle.

### DIFF
--- a/src/datachain/lib/file.py
+++ b/src/datachain/lib/file.py
@@ -297,6 +297,16 @@ class File(DataModel):
         super().__init__(**kwargs)
         self._catalog = None
         self._caching_enabled: bool = False
+        self._download_cb: Callback = DEFAULT_CALLBACK
+
+    def __getstate__(self):
+        state = super().__getstate__()
+        # Exclude _catalog from pickling - it contains SQLAlchemy engine and other
+        # non-picklable objects. The catalog will be re-set by _set_stream() on the
+        # worker side when needed.
+        state["__dict__"] = state["__dict__"].copy()
+        state["__dict__"]["_catalog"] = None
+        return state
 
     def as_text_file(self) -> "TextFile":
         """Convert the file to a `TextFile` object."""
--- a/tests/func/test_file.py
+++ b/tests/func/test_file.py
@@ -8,6 +8,51 @@ from datachain.query import C
 from datachain.utils import TIME_ZERO
 
 
+def test_file_pickle_with_catalog(tmp_dir, test_session_tmpfile):
+    """Test that File objects with catalog can be pickled for parallel processing.
+
+    File objects created with File.at() have a _catalog attached which contains
+    SQLAlchemy connections, weakrefs, etc that cannot be pickled. The __getstate__
+    method must exclude _catalog to allow pickling.
+    """
+    import cloudpickle
+
+    # Create a real file using File.at + open (this attaches _catalog)
+    file = File.at(tmp_dir / "test.txt", session=test_session_tmpfile)
+    with file.open("w") as f:
+        f.write("hello world")
+
+    assert file._catalog is not None, "File should have catalog attached"
+
+    # This would fail with "cannot pickle 'sqlite3.Connection' object"
+    # if __getstate__ doesn't exclude _catalog
+    data = cloudpickle.dumps(file)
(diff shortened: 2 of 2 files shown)


src/datachain/lib/file.py
```diff
@@ -298,6 +298,15 @@ def __init__(self, **kwargs):
         self._catalog = None
         self._caching_enabled: bool = False
 
+    def __getstate__(self):
+        state = super().__getstate__()
+        # Exclude _catalog from pickling - it contains SQLAlchemy engine and other
+        # non-picklable objects. The catalog will be re-set by _set_stream() on the
+        # worker side when needed.
+        state["__dict__"] = state["__dict__"].copy()
```
amritghimire, line 307: Will this actually update the state of the class in place removing catalog from it if it is used further ?

src/datachain/lib/file.py
```diff
@@ -298,6 +298,15 @@ def __init__(self, **kwargs):
         self._catalog = None
         self._caching_enabled: bool = False
 
+    def __getstate__(self):
```
dreadatour, line 302: We can also use [\_\_reduce\_\_](https://docs.python.org/3/library/pickle.html#object.__reduce__) to make it pickable.

dreadatour: Example: https://github.com/datachain-ai/datachain/blob/bd49c21807d4dbb2c5f6efb1d347f7bb0e201c59/src/datachain/lib/file.py#L95-L96

shcheklein: then we would have to use it in all child classes and manage all fields explicitly, no?

dreadatour: Hm, good point! You're right.

dreadatour: Same time we can define `__reduce__` in `Catalog`, what do you think?

shcheklein: okay, but then we will potentially get some weird state in the catalog? I would err on the side of just removing catalog altogether (and setting it up back later).

src/datachain/lib/file.py
shcheklein: no, I don't think so, we copy that dict since dict indeed contains references to the actual values 🤔

src/datachain/lib/file.py
dreadatour: > then we will potentially get some weird state in the catalog

From my understanding it will be reinitialized from scratch if using `__reduce__`. We can get weird state if something is wrong in `Catalog` but from what I remember it should be ok, since we are already initializing it from scratch in UDFs in distributed workers.

dreadatour: Anyway, I am absolutely ok to move forward with this implementation, I just think it might be more clean to use `__reduce__` since it was designed for this. And fix all the errors after with "weird state" or something else.

dreadatour (approved):
Should we also have tests for this?

src/datachain/lib/file.py
```diff
@@ -298,6 +298,15 @@ def __init__(self, **kwargs):
...
+        state = super().__getstate__()
+        # Exclude _catalog from pickling - it contains SQLAlchemy engine and other
+        # non-picklable objects. The catalog will be re-set by _set_stream() on the
+        # worker side when needed.
+        state["__dict__"] = state["__dict__"].copy()
+        state["__dict__"]["_catalog"] = None
+        return state
```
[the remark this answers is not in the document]
shcheklein: added a test

src/datachain/lib/file.py
shcheklein: but what exactly would you be preserving in Catalog?

shcheklein: also, not sure Pydantic model can even restore this way ... (it has some internal attributes)

dreadatour: I am not sure we need to preserve state, so everything from the constructor may be? https://github.com/datachain-ai/datachain/blob/main/src/datachain/catalog/catalog.py#L455-L465

shcheklein:
> Should we also have tests for this?

added

amritghimire (approved):

PR #1492, opened 2025-12-05
is_local()

dmpetrov:
more descriptive than is_studio()

### DIFF
--- a/docs/references/datachain.md
+++ b/docs/references/datachain.md
@@ -7,25 +7,18 @@ for examples of how to create a chain.
 
 ::: datachain.query.schema.C
 
-::: datachain.query.schema.Column
+::: datachain.lib.dc.database.ConnectionType
+
+
+::: datachain.lib.dc.listings.listings
 
 ::: datachain.lib.dc.csv.read_csv
 
 ::: datachain.lib.dc.datasets.read_dataset
 
-::: datachain.lib.dc.datasets.datasets
-
-::: datachain.lib.dc.datasets.delete_dataset
-
-::: datachain.lib.dc.datasets.move_dataset
-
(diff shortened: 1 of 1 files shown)


src/datachain/lib/dc/utils.py
```diff
@@ -28,6 +28,10 @@ def is_studio() -> bool:
     return getenv_bool("DATACHAIN_IS_STUDIO", default=False)
 
 
+def is_local() -> bool:
```
shcheklein, line 32: do we need docs for this?

shcheklein: also a test (if we have something for is_studio)

docs/references/datachain.md
```diff
@@ -43,6 +43,10 @@ for examples of how to create a chain.
 
 ::: datachain.lib.dc.DataChain
 
+::: datachain.lib.dc.utils.is_studio
+
+::: datachain.lib.dc.utils.is_local
```
dmpetrov, line 46: @yathomasi please check if it's a right way of doing docs

dmpetrov (commented):
@yathomasi your review is needed. please take a look at the doc changes

docs/references/datachain.md
yathomasi: @dmpetrov yes, it's correct.

I think we can arrange this better i.e right now it's between the classes and we could move up.

Screenshot 2025-12-08 at 08 58 37

Let me know if it's okay I can do that and merge the PR.

yathomasi: Reorganized them.

Screenshot 2025-12-08 at 11 06 03

PR #1493, opened 2025-12-06
to_json/to_csv - return file written

shcheklein:
Make `to_json` and `to_csv` return `File` instance and use File `at` / `open` to write files.

Need actual written File returned for a customer scenario.

### DIFF
--- a/src/datachain/lib/dc/datachain.py
+++ b/src/datachain/lib/dc/datachain.py
@@ -41,7 +41,7 @@ from datachain.lib.data_model import (
     StandardType,
     dict_to_data_model,
 )
-from datachain.lib.file import EXPORT_FILES_MAX_THREADS, ArrowRow, FileExporter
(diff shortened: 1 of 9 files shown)


src/datachain/lib/dc/datachain.py
```diff
@@ -2318,123 +2318,102 @@ def to_csv(
...
+        with target.open("wb", client_config=fs_kwargs) as f:
+            self._write_json_stream(f, headers, include_outer_list)
+        return target
 
+    def _write_json_stream(
+        self,
+        f: IO[bytes],
+        headers: list[list[str]],
+        include_outer_list: bool,
+    ) -> None:
```
amritghimire, line 2382: Probably a comment that the file should be open and the burden of closing file lies in the caller would be helpful

shcheklein: it's an internal protected helper, and code is more or less clear (I hope) ... not sure it is that meaningful in this particular case (and it is a usual convention that if something is passed outside it is a responsibility of the caller to manage it)

PR #1497, opened 2025-12-08
cli: New CLI command datachain pipeline create

amritghimire:
#1497 (this, create) > #1498 (status) > #1500 (list) > #1504 (pause/resume) > #1505 (remove-job)
------------

## Summary of Changes

Added a new CLI command `datachain pipeline create` to trigger pipelines for dataset dependencies in Studio.

**What it does:**
- Creates a pipeline for a dataset dependency
- Supports review mode (pauses the pipeline for review before running)
- Handles dataset names in fully qualified or short form
- Resolves the latest version if none is specified

**Implementation details:**

1. **CLI parser** (`pipeline.py`): Adds argument parsing for the `pipeline create` command, including dataset name, version, review flag, namespace, project, and team.

2. **Studio client updates** (`remote/studio.py`):
   - Adds `create_dataset_dependency_update()` to call the api with the dataset identifier and review flag

3. **Main handler** (`studio.py`): Adds `trigger_pipeline()` that:
   - Resolves the full dataset name (namespace, project, name)
   - Fetches the latest version if not provided
   - Creates a `DatasetIdentifier` and calls the Studio client
   - Prints the pipeline name and trigger source on success

5. **DatasetIdentifier class** (`utils.py`): A small class to represent a dataset with namespace, project, name, and version. Includes hashing, equality, and string formatting.

6. **CLI wiring** (`cli/__init__.py` and `parser/__init__.py`): Hooks the new command into the CLI so `datachain pipeline trigger` works.

7. **Documentation** (`docs/commands/pipeline/create.md`): Adds docs with examples and usage.

**How it works:**
When you run `datachain pipeline create "dataset_name"`, it resolves the dataset identifier, calls Studio’s API to create the pipeline, and prints the result. If you pass `--review`, the pipeline starts in a paused state so you can review it in Studio before resuming.

### DIFF
--- a/docs/commands/pipeline/create.md
+++ b/docs/commands/pipeline/create.md
@@ -15,9 +15,9 @@ usage: datachain pipeline create [-h] [-v] [-q]
 
 This command creates a pipeline in Studio that will update the specified dataset. The pipeline automatically includes all necessary jobs to update the dataset based on its dependencies. If no version is specified, the latest version of the dataset is used.
 
-The pipeline is created in paused state for review. You can open Studio to review the pipeline configuration and resume it when ready.
+The pipeline is created in paused state. Use `datachain pipeline resume` to start pipeline execution.
 
-The dataset name can be provided in fully qualified format (e.g., `@namespace.project.name`) or as a short name. If using a short name, default project and namespace in Studio will be used.
+The dataset name can be provided in fully qualified format (e.g., `@namespace.project.name`) or as a short name. If using a short name, Studio uses the default project and namespace.
 
 ## Arguments
 
@@ -47,8 +47,3 @@ datachain pipeline create "final_result" --version "1.0.9"
 ```bash
 datachain pipeline create "@amritghimire.default.final_result"
 ```
-
-## Notes
-
-* The pipeline is always created in a paused state in Studio. You can open Studio to review the pipeline configuration and resume it when ready.
-* The pipeline automatically includes all jobs needed to update the dataset based on its dependencies.
--- a/src/datachain/cli/parser/pipeline.py
+++ b/src/datachain/cli/parser/pipeline.py
@@ -22,11 +22,11 @@ def add_pipeline_parser(subparsers, parent_parser) -> None:
         " dataset. The pipeline automatically includes all necessary jobs to update"
         " the dataset based on its dependencies. "
         "If no version is specified, the latest version of the dataset is used.\n\n"
-        "The pipeline is created in paused state for review. You can open Studio to"
-        " review the pipeline configuration and resume it when ready.\n\n"
+        "The pipeline is created in paused state. Use `datachain pipeline resume`"
+        " to start pipeline execution.\n\n"
         "The dataset name can be provided in fully qualified format "
         "(e.g., @namespace.project.name) or as a short name. "
-        "If using a short name, default project and namespace in Studio will be used."
+        "If using a short name, Studio uses the default project and namespace."
     )
     pipeline_create_parser = pipeline_subparser.add_parser(
         "create",


src/datachain/utils.py
```diff
@@ -108,6 +110,54 @@ def find(cls, create: bool = True) -> "Self":
...
+    @classmethod
+    def to_string(cls, dataset_identifier: "DatasetIdentifier") -> str:
+        """Format DatasetIdentifier as a string: namespace.project.name@version"""
+        return (
+            f"{dataset_identifier.namespace}.{dataset_identifier.project}."
+            f"{dataset_identifier.name}@{dataset_identifier.version}"
+        )
+
+    def __str__(self) -> str:
+        return self.to_string(self)
```
dreadatour: This looks odd 👀

shcheklein:
>datachain graph trigger

let's revisit this. This really should be something like `dc pipeline create` most likely. graph trigger is too technical, low level, etc

docs/commands/graph/trigger.md
```diff
@@ -0,0 +1,53 @@
...
+* `-t TEAM, --team TEAM` - Team to run job for (default: from config)
+* `-h`, `--help` - Show the help message and exit.
+* `-v`, `--verbose` - Be verbose.
+* `-q`, `--quiet` - Be quiet.
+
+## Examples
+
+1. Run a trigger with fully qualified dataset name
+```bash
+datachain graph trigger "@amritghimire.default.final_result" --version "1.0.9"
```
shcheklein: do we accept short names? (and resolve according to rules)

shcheklein: think about multiple targets - will we have to change API here?

amritghimire: > do we accept short names? (and resolve according to rules)

What short names? 

> think about multiple targets - will we have to change API here?

I am not sure what you mean by multiple targets here.

shcheklein: > short name

final_result

> multiple targets

datachain graph trigger final_result1  final_result2

(it will be required pretty much soon most likely)

amritghimire: > > short name
> 
> final_result


Yes, it uses the same code read_dataset does to identify and attach namespace and projects. 

> > multiple targets
> 
> datachain graph trigger final_result1 final_result2

> (it will be required pretty much soon most likely)

How is it different than calling the command twice?

shcheklein: >Yes, it uses the same code read_dataset does to identify and attach namespace and projects.

Let's I guess in the example use also a short name then (if we have examples)

shcheklein: >How is it different than calling the command twice?

there will be two different pipelines? 

(a few targets could easily share a lot of jobs between them)

docs/commands/pipeline/create.md
```diff
@@ -0,0 +1,53 @@
+# pipeline create
+
+Create an update for dataset dependency in Studio
```
shcheklein: seems wrong / bad descritption

please read carefully docs, check style, cleanup . etc. Think about phrases - do they make sense, do they use consistent terms, would you understand them w/o your knowledge of the details

Ask some powerful AI also to review with a similar prompt

docs/commands/pipeline/create.md
```diff
@@ -0,0 +1,53 @@
...
+This command creates an pipeline for a dataset dependency. The pipeline is determined based on the dependency for the dataset in Studio and is created accordingly. The dataset name can be a fully qualified name including the namespace and project, or a regular name, in which case the explicitly defined namespace and project will be used if they are set; otherwise, default values will be applied.
+
+## Arguments
+
+* `dataset` - Name of the dataset (can be a fully qualified name including the namespace and project or regular name)
+
+## Options
+
+* `-V VERSION, --version VERSION` - Version of the dataset (default: latest)
+* `-r, --review` - Review the pipeline before creating. (Opens in paused state)
```
shcheklein: as I mentioned in Slack - let's get rid of this complexity. Create always in "review" mode.

docs/commands/pipeline/create.md
```diff
@@ -0,0 +1,53 @@
...
+* `-n NAMESPACE, --namespace NAMESPACE` - Namespace of the dataset
```
shcheklein: let's not complicate - no need to allow namespace and project separately

src/datachain/cli/parser/pipeline.py
```diff
@@ -0,0 +1,79 @@
...
+    )
+    pipeline_subparser = pipeline_parser.add_subparsers(
+        dest="cmd",
+        help="Use `datachain pipeline CMD --help` to display command-specific help",
+    )
+
+    pipeline_create_help = "Create an update for dataset dependency in Studio"
+    pipeline_create_description = "Create an pipeline for a dataset dependency.\n"
+    pipeline_create_description += (
+        "The pipeline will be determined based on the dependency for the"
```
shcheklein: based on the dependecies probably

src/datachain/cli/parser/pipeline.py
```diff
@@ -0,0 +1,79 @@
...
+        help=pipeline_helper,
+        formatter_class=CustomHelpFormatter,
+    )
+    pipeline_subparser = pipeline_parser.add_subparsers(
+        dest="cmd",
+        help="Use `datachain pipeline CMD --help` to display command-specific help",
+    )
+
+    pipeline_create_help = "Create an update for dataset dependency in Studio"
+    pipeline_create_description = "Create an pipeline for a dataset dependency.\n"
```
shcheklein: Create a pipeline

Please again, review carefully docs with and w/o AI. Let me know when it is done

src/datachain/remote/studio.py
```diff
@@ -508,3 +508,22 @@ def cancel_job(
 
     def get_clusters(self) -> Response[ClusterListData]:
         return self._send_request("datachain/clusters/", {}, method="GET")
+
+    # Pipeline API
+    def create_dataset_dependency_update(
```
shcheklein: why is called dataset_dependency_update?

docs/commands/pipeline/create.md
amritghimire: Creating in review mode always means it will not continue automatically, which may not what user want all the time. With CLI, user may want the pipeline to start directly probably part of some periodic CI or something similar. I believe it is good to provide options to user.

docs/commands/pipeline/create.md
amritghimire: Why not? Its not required and it is in line with existing behavior of read_dataset

docs/commands/pipeline/create.md
shcheklein: let's keep it simple initialy, we can add "auto start" (instead of review) later if needed

docs/commands/pipeline/create.md
shcheklein: since there is an obvious way of doing this already and syntax is shorter ... again, can be added if needed, but harder to remove (keep it simple initially, don't bloat options, etc)

docs/commands/pipeline/create.md
```diff
@@ -0,0 +1,54 @@
...
+```
+
+3. Create a pipeline for the latest version of a dataset:
+```bash
+datachain pipeline create "@amritghimire.default.final_result"
+```
+
+## Notes
+
+* The pipeline is always created in a paused state in Studio. You can open Studio to review the pipeline configuration and resume it when ready.
```
shcheklein: We don't need these two clarifications (overall - rule of thumb - if you need to Note or clarify something - it smells - worth paying extra attention since it can be design issue). Rather mention: Use "pipeline run" (or whatever he command is) to start pipeline execution

src/datachain/cli/parser/pipeline.py
```diff
@@ -0,0 +1,61 @@
...
+        help="Use `datachain pipeline CMD --help` to display command-specific help",
+    )
+
+    pipeline_create_help = "Create a pipeline to update a dataset in Studio"
+    pipeline_create_description = (
+        "This command creates a pipeline in Studio that will update the specified"
+        " dataset. The pipeline automatically includes all necessary jobs to update"
+        " the dataset based on its dependencies. "
+        "If no version is specified, the latest version of the dataset is used.\n\n"
+        "The pipeline is created in paused state for review. You can open Studio to"
```
shcheklein: do I have to open Studio? is it he only way to review? Also too much text and clarifications (see comment above)

shcheklein (approved):

PR #1501, opened 2025-12-09
add proper tuple type support in models

shcheklein:
Fixes support for models like this:

```python
class ImageGeometry(BaseModel):
    image_position: Optional[Tuple[float, float, float]]
    image_orientation: Optional[Tuple[float, float, float, float, float, float]]
    pixel_spacing: Optional[Tuple[float, float]]
    slice_thickness: Optional[float]
```

Also, fixes set and tuple support on SQLite.

### DIFF
--- a/src/datachain/data_storage/warehouse.py
+++ b/src/datachain/data_storage/warehouse.py
@@ -135,7 +135,7 @@ class AbstractWarehouse(ABC, Serializable):
                     if isinstance(val[0], item_python_type):
                         # SQLite ARRAY storage expects a list; tuples/sets must be
                         # converted to lists even when element types already match.
-                        return val if value_type is list else list(val)
+                        return list(val)
                     if item_python_type is float and isinstance(val[0], int):
                         return [float(i) for i in val]
 


src/datachain/data_storage/warehouse.py
```diff
@@ -128,12 +128,17 @@ def convert_type(  # noqa: PLR0911
...
                     return []
+
                 item_python_type = self.python_type(col_type.item_type)
+
                 if item_python_type is not list:
                     if isinstance(val[0], item_python_type):
-                        return val
+                        # SQLite ARRAY storage expects a list; tuples/sets must be
+                        # converted to lists even when element types already match.
+                        return val if value_type is list else list(val)
```
amritghimire: Why not simple list(val)? 
```suggestion
                        return list(val)
```

shcheklein: avoids extra copy I think ... not hugely important though

amritghimire: list will do a shallow copy though.  The part `return val if value_type is list else list(val)` returns two type of return to my mind, one is mutable if it is list, another is shallow copy hence doesn't matter if passed list is edited. I would like to have uniformity in return behavior. Also, simple list is cleaner, more idiomatic, and immediately clear what the intent is (ensure output is a list).

shcheklein: shallow copy is still a copy O(n) allocation time + space to pointer (most likely) + time it takes to do the operation (penalizing a bit the most common part), Again, not a big deal in this particular case, fee free to merge the change if you feel strongly about it.

amritghimire (approved):

PR #1503, opened 2025-12-10
Fix nested column flattening in group_by to use underscores

nablabits:
This is an early attempt to fix https://github.com/datachain-ai/datachain/issues/1329

I'm not very familiar with the code but with my knowledge and the help of some LLM I managed to put together this solution. I'm aware that this could be quite disruptive for users as it's changing the output of some well established function such as `to_list` (see the example in [this test](https://github.com/nablabits/datachain/blob/2149857c3c5661564cc8b532bd8abc99b9489c3b/tests/unit/lib/test_partition_by.py#L306)). People more knowledgeable of the use cases can judge this. 

If we want to preserve the `nested.level1.name` structure, I feel that the options go through:
1. how the signals are extracted [here](https://github.com/nablabits/datachain/blob/2149857c3c5661564cc8b532bd8abc99b9489c3b/src/datachain/lib/dc/datachain.py#L1215)
2. the `to_pandas` method and more precisely this `get_headers_with_length` folk ([source](https://github.com/datachain-ai/datachain/blob/2149857c3c5661564cc8b532bd8abc99b9489c3b/src/datachain/lib/dc/datachain.py#L1986)) which is responsible of getting the columns that then become a multi-index.

Let me know what you think 

(There are a couple of lint errors that we can address once a consensus on the approach is reached)

### DIFF
--- a/src/datachain/lib/dc/datachain.py
+++ b/src/datachain/lib/dc/datachain.py
@@ -1234,10 +1234,11 @@ class DataChain:
                         schema_partition_by.append(col)
                     else:
                         # BaseModel or other - add flattened columns directly
+                        # Use underscores to flatten the column name
+                        # to avoid MultiIndex in pandas output
                         for column in cast("list[Column]", columns):
-                            col_type = self.signals_schema.get_column_type(column.name)
-                            schema_fields[column.name] = col_type
-                        schema_partition_by.append(col)
+                            # Use the column's db name which already uses underscores
+                            signal_columns.append(column)
                 else:
                     # simple signal - but we need to check if it's a complex signal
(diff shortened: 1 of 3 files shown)


nablabits:
@shcheklein Yep, I have rebased the branch on top of #1496 and yes, it is working as expected, see: https://github.com/nablabits/datachain/commit/1570d50f2b1f0f9f971f61759c47b7c81c02be19

I will close this now, but feel free to reopen if something else comes up :slightly_smiling_face: 

Edit: wait, it's working because I didn't remove the `flatten=True`  :facepalm:

nablabits:
It seems that it was fixed at some point :slightly_smiling_face:

PR #1506, opened 2025-12-11
Remove non complete datasets from failed job

ilongin:
### Summary
  Implements dataset status management to hide incomplete datasets (CREATED/FAILED) from public APIs and adds automatic garbage collection for failed dataset versions.

  ### Problem
  When jobs fails, dataset versions remain visible in incomplete states, confusing users and wasting disk space with no cleanup mechanism.

  ### Solution

  **1. Status Management**
  - Move `COMPLETE` status to end of `save()` - only set after all operations succeed
  - Mark dataset versions as `FAILED` when their job fails
  - Added `include_incomplete` parameter throughout stack (metastore → catalog → public API)

  **2. Filtering**
  - Public APIs (`read_dataset`, `delete_dataset`, `move_dataset`, `datasets`) only show COMPLETE datasets
  - Internal methods retain full access via `include_incomplete=True`

  **3. Garbage Collection**
  - New `cleanup_failed_dataset_versions()` removes FAILED/CREATED dataset versions
  - **Only cleans versions from jobs in final status** (FAILED, COMPLETE, CANCELED) - never running jobs
  - Configurable retention period via `--retention-days` parameter
  - Updated `datachain gc` CLI command

### DIFF
--- a/src/datachain/catalog/catalog.py
+++ b/src/datachain/catalog/catalog.py
@@ -31,6 +31,7 @@ from datachain.dataset import (
     create_dataset_uri,
     parse_dataset_name,
     parse_dataset_uri,
+    parse_dataset_with_version,
     parse_schema,
 )
 from datachain.error import (
@@ -46,7 +47,7 @@ from datachain.node import DirType, Node, NodeWithPath
 from datachain.nodes_thread_pool import NodesThreadPool
 from datachain.project import Project
 from datachain.sql.types import DateTime, SQLType
-from datachain.utils import DataChainDir
+from datachain.utils import DataChainDir, DatasetIdentifier
 
 from .datasource import DataSource
 from .dependency import build_dependency_hierarchy, populate_nested_dependencies
@@ -89,6 +90,7 @@ class DatasetRowsFetcher(NodesThreadPool):
         warehouse: "AbstractWarehouse",
         remote_ds: DatasetRecord,
         remote_ds_version: str,
+        export_id: int,
         local_ds: DatasetRecord,
         local_ds_version: str,
         schema: dict[str, SQLType | type[SQLType]],
@@ -103,12 +105,16 @@ class DatasetRowsFetcher(NodesThreadPool):
         self.warehouse = warehouse
         self.remote_ds = remote_ds
         self.remote_ds_version = remote_ds_version
+        self.export_id = export_id
         self.local_ds = local_ds
         self.local_ds_version = local_ds_version
         self.schema = schema
         self.last_status_check: float | None = None
         self.studio_client = StudioClient()
         self.progress_bar = progress_bar
+        self._last_export_status: str | None = None
+        self._last_export_files_done: int | None = None
+        self._last_export_num_files: int | None = None
 
     def done_task(self, done):
         for task in done:
@@ -139,12 +145,36 @@ class DatasetRowsFetcher(NodesThreadPool):
         Checks are done every PULL_DATASET_CHECK_STATUS_INTERVAL seconds
         """
         export_status_response = self.studio_client.dataset_export_status(
-            self.remote_ds, self.remote_ds_version
+            self.export_id
         )
         if not export_status_response.ok:
             raise DataChainError(export_status_response.message)
 
-        export_status = export_status_response.data["status"]  # type: ignore [index]
+        data = export_status_response.data
+        export_status = data["status"]
+
+        # Surface Studio-side export progress (if available) while we're pulling.
+        files_done = data.get("files_done")
+        num_files = data.get("num_files")
+
+        if (
+            files_done is not None
+            and num_files is not None
+            and (
+                export_status != self._last_export_status
+                or files_done != self._last_export_files_done
+                or num_files != self._last_export_num_files
+            )
+        ):
+            self._last_export_status = export_status
+            self._last_export_files_done = files_done
+            self._last_export_num_files = num_files
+
+            if self.progress_bar is not None:
+                # Keep the main bar semantics (rows) and just add a postfix.
+                self.progress_bar.set_postfix_str(
+                    f"studio_export={files_done}/{num_files} ({export_status})"
+                )
 
         if export_status == "failed":
             raise DataChainError("Dataset export failed in Studio")
@@ -1098,6 +1128,29 @@ class Catalog:
 
         return namespace_name, project_name, name
 
+    def parse_dataset_name(
+        self,
+        dataset_name: str,
+        namespace_name: str | None = None,
+        project_name: str | None = None,
+        version: str | None = None,
+    ) -> DatasetIdentifier:
+        if not version:
+            dataset_name, version = parse_dataset_with_version(dataset_name)
+
+        namespace, project, name = self.get_full_dataset_name(
+            dataset_name,
+            namespace_name=namespace_name,
+            project_name=project_name,
+        )
+
+        return DatasetIdentifier(
+            namespace=namespace,
+            project=project,
+            name=name,
+            version=version,
+        )
+
     def get_dataset(
         self,
         name: str,
@@ -1418,31 +1471,29 @@ class Catalog:
 
     def export_dataset_table(
         self,
-        bucket_uri: str,
+        bucket: str,
         name: str,
         version: str,
         project: Project | None = None,
+        *,
+        file_format: str | None = None,
+        base_file_name: str,
         client_config=None,
-    ) -> list[str]:
+    ) -> None:
         dataset = self.get_dataset(
             name,
             namespace_name=project.namespace.name if project else None,
             project_name=project.name if project else None,
         )
 
-        return self.warehouse.export_dataset_table(
-            bucket_uri, dataset, version, client_config
+        self.warehouse.export_dataset_table(
+            bucket,
+            dataset,
+            version,
+            file_format=file_format,
+            base_file_name=base_file_name,
+            client_config=client_config,
         )
-
-    def dataset_table_export_file_names(
-        self, name: str, version: str, project: Project | None = None
-    ) -> list[str]:
-        dataset = self.get_dataset(
-            name,
-            namespace_name=project.namespace.name if project else None,
-            project_name=project.name if project else None,
-        )
-        return self.warehouse.dataset_table_export_file_names(dataset, version)
 
     def remove_dataset(
         self,
@@ -1696,7 +1747,9 @@ class Catalog:
         if not export_response.ok:
             raise DataChainError(export_response.message)
 
-        signed_urls = export_response.data
+        export_data = export_response.data
+        export_id = export_data["export_id"]
+        signed_urls: list[str] = export_data["signed_urls"]
 
         if signed_urls:
             with (
@@ -1713,7 +1766,7 @@ class Catalog:
                     """
                     res = [[] for i in range(PULL_DATASET_MAX_THREADS)]
                     current_worker = 0
-                    for url in signed_urls:
+                    for url in urls:
                         res[current_worker].append(url)
                         current_worker = (current_worker + 1) % PULL_DATASET_MAX_THREADS
 
@@ -1724,11 +1777,13 @@ class Catalog:
                     warehouse,
                     remote_ds,
                     remote_ds_version.version,
+                    export_id,
                     local_ds,
                     local_ds_version,
                     schema,
                     progress_bar=dataset_save_progress_bar,
                 )
+
                 try:
                     rows_fetcher.run(
                         iter(batch(signed_urls)), dataset_save_progress_bar
--- a/.github/workflows/tests-studio.yml
+++ b/.github/workflows/tests-studio.yml
@@ -31,6 +31,8 @@ jobs:
           POSTGRES_USER: test
           POSTGRES_DB: database
           POSTGRES_HOST_AUTH_METHOD: trust
+        options: >-
+          --add-host=host.docker.internal:host-gateway
       clickhouse:
         image: clickhouse/clickhouse-server:25.8
         ports:
@@ -41,10 +43,14 @@ jobs:
           CLICKHOUSE_USER: studio_local
           CLICKHOUSE_PASSWORD: ch123456789!
           CLICKHOUSE_DEFAULT_ACCESS_MANAGEMENT: 1
+        options: >-
+          --add-host=host.docker.internal:host-gateway
       redis:
         image: redis:7.2.5
         ports:
           - 6379:6379
+        options: >-
+          --add-host=host.docker.internal:host-gateway
     steps:
       - name: Studio branch name
         env:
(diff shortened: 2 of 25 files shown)


src/datachain/data_storage/metastore.py
```diff
@@ -317,9 +318,34 @@ def remove_dataset_version(
...
+    @abstractmethod
+    def get_failed_dataset_versions_to_clean(
+        self, retention_days: int | None = None
+    ) -> list[tuple[DatasetRecord, str]]:
+        """
+        Get failed/incomplete dataset versions that are safe to clean up.
+
+        Returns dataset versions that:
+        - Have status CREATED or FAILED (incomplete/failed)
+        - Were created more than retention_days ago (if retention_days specified)
```
shcheklein: why do we want retention?

src/datachain/data_storage/metastore.py
```diff
@@ -317,9 +318,34 @@ def remove_dataset_version(
...
+        - Belong to jobs that are not running (COMPLETE, FAILED, CANCELED)
```
shcheklein, line 331: can we drop FAILED even if job is still running?

src/datachain/data_storage/metastore.py
ilongin: I thought it might be useful in future if user wants to debug failed datasets or something similar. Currently it's not used so maybe I can even remove it.

shcheklein: yes, if we need to debug - it is better to save it as some separate object (it is not a good idea to keep failed datasets around). Also I hope we are removing them automatically right away on a failed job? We should not be relying fully on GC to my mind (GC should be the last resort only)

ilongin: Yes, they are being removed on failed job automatically, actually on Session `__exit__` where old cleanup logic was being triggering since we don't have any "job hooks" to catch on. GC is, as you mentioned, used if whole process (e.g pod) crashes so cleanup is not being called.

src/datachain/data_storage/metastore.py
ilongin: we could but it's cleaner IMO to do removing / cleaning only when job is done running.

src/datachain/data_storage/metastore.py
shcheklein: kk, but since dataset is staying in this FAILED form for a while, what happens to the next job run? Is that version visible to some internal logic? E.g. what will be the version on the next successful run?

src/datachain/catalog/catalog.py
```diff
@@ -960,6 +960,52 @@ def cleanup_tables(self, names: Iterable[str]) -> None:
...
+        - Are older than retention_days (if specified)
+        - Belong to completed/failed/canceled jobs (not running)
+
+        Args:
+            retention_days: Days to retain failed versions. If None, cleans all.
+
+        Returns:
+            List of cleaned version IDs (for logging/metrics)
+        """
+        import logging
```
shcheklein: why do we do logging in such a weird way? AI?

src/datachain/catalog/catalog.py
```diff
@@ -960,6 +960,52 @@ def cleanup_tables(self, names: Iterable[str]) -> None:
...
+
+        versions_to_clean = self.metastore.get_failed_dataset_versions_to_clean(
+            retention_days=retention_days, job_id=job_id
+        )
+
+        cleaned_ids = []
+        for dataset, version in versions_to_clean:
+            try:
+                # Track version ID before deletion
+                version_obj = dataset.get_version(version)
```
shcheklein: do we need really this tracking?

why? can it be all simplified?

shcheklein: can we do this in some simpler way at least? return version itself, etc?

shcheklein: way to much hassle to then just to len() on it to print a report to user

src/datachain/data_storage/metastore.py
```diff
@@ -358,6 +388,18 @@ def update_dataset_status(
     ) -> DatasetRecord:
         """Updates dataset status and appropriate fields related to status."""
 
+    @abstractmethod
+    def mark_job_dataset_versions_as_failed(self, job_id: str) -> None:
+        """
+        Mark all non-COMPLETE dataset versions created by a job as FAILED.
```
shcheklein, line 389: So, do we mark them as failed or do we drop them? Why would we want to keep them around?

src/datachain/data_storage/metastore.py
```diff
@@ -1317,37 +1361,50 @@ def _get_dataset_query(
...
             n.join(p, n.c.id == p.c.namespace_id)
             .join(d, p.c.id == d.c.project_id)
-            .join(dv, d.c.id == dv.c.dataset_id, isouter=isouter)
+            .join(dv, join_condition, isouter=isouter)
         )
         return query.select_from(j)
 
-    def _base_dataset_query(self) -> "Select":
+    def _base_dataset_query(self, include_incomplete: bool = True) -> "Select":
+        # When filtering by status, use inner join so datasets without COMPLETE
```
shcheklein, line 1377: what does it practically mean? dataset exists but not visible in UI?

src/datachain/data_storage/metastore.py
```diff
@@ -1465,6 +1592,30 @@ def update_dataset_status(
 
         return dataset
 
+    def mark_job_dataset_versions_as_failed(self, job_id: str) -> None:
```
shcheklein, line 1573: is it exposed as a API metastore call? do we check that user has access to that particular job_id, that job belongs to the team that we are accessing this API from?

src/datachain/data_storage/metastore.py
```diff
@@ -1415,24 +1477,89 @@ def remove_dataset_version(
...
+                self.db.execute(self._datasets_delete().where(d.c.id == dataset.id))
 
         dataset.remove_version(version)
         return dataset
 
+    def get_failed_dataset_versions_to_clean(
+        self, retention_days: int | None = None, job_id: str | None = None
+    ) -> list[tuple[DatasetRecord, str]]:
+        """
+        Get failed/incomplete dataset versions that are safe to clean up.
```
shcheklein: do we need to duplicate really the whole docstring?

src/datachain/data_storage/metastore.py
```diff
@@ -1415,24 +1477,89 @@ def remove_dataset_version(
...
+
+        Returns dataset versions that:
+        - Have status CREATED or FAILED (incomplete/failed)
+        - Were created more than retention_days ago (if retention_days
+          specified)
+        - Belong to jobs that are not running (COMPLETE, FAILED, CANCELED)
+        - Optionally returning them only for specific job_id
```
shcheklein: is it needed? why?

src/datachain/data_storage/metastore.py
shcheklein: also, still not clear why complicating this with retention ...

src/datachain/data_storage/metastore.py
```diff
@@ -1415,24 +1477,89 @@ def remove_dataset_version(
...
-            # had only one version, fully deleting dataset
-            self.db.execute(self._datasets_delete().where(d.c.id == dataset.id))
+            if dataset.versions and len(dataset.versions) == 1:
+                # had only one version, fully deleting dataset
+                self.db.execute(self._datasets_delete().where(d.c.id == dataset.id))
 
         dataset.remove_version(version)
         return dataset
 
+    def get_failed_dataset_versions_to_clean(
```
shcheklein: `to_clean` - remove it? can it be just general method?

can we reuse some existing stuff? like get all dataset versions with FAILED state?

src/datachain/lib/dc/datasets.py
```diff
@@ -147,11 +147,30 @@ def read_dataset(
...
+    if not is_listing or version is not None:
+        # Always validate non-listing datasets
+        # For listing datasets, only validate if specific version is requested
         dataset = session.catalog.get_dataset_with_remote_fallback(
-            name, namespace_name, project_name, update=update
+            name,
+            namespace_name,
+            project_name,
+            update=update,
+            include_incomplete=is_listing,  # Allow incomplete for listing datasets
```
shcheklein: why?

src/datachain/lib/dc/datasets.py
```diff
@@ -147,11 +147,30 @@ def read_dataset(
         namespace_name=namespace,
     )
 
-    if version is not None:
+    # Validate dataset exists and is COMPLETE (filter non-COMPLETE datasets from
```
shcheklein: not sure what is happening here .. code and comments look a bit messy . Could you clarify all of this? (and may be cleanup)

src/datachain/lib/dc/records.py
```diff
@@ -95,4 +95,12 @@ def read_records(
     records = (adjust_outputs(warehouse, record, col_types) for record in to_insert)
     warehouse.insert_rows(table, records, batch_size=READ_RECORDS_BATCH_SIZE)
     warehouse.insert_rows_done(table)
+
+    # Mark dataset as COMPLETE since it's fully created
+    from datachain.dataset import DatasetStatus
```
shcheklein: do we really want this import like this? (looks messy, ad hoc, including the comment)

src/datachain/query/session.py
```diff
@@ -220,6 +220,14 @@ def _finalize_job_as_failed(self, exc_type, exc_value, tb):
                 error_message=str(exc_value),
                 error_stack=error_stack,
             )
+
+            # Mark any incomplete dataset versions created by this job as FAILED
+            self.catalog.metastore.mark_job_dataset_versions_as_failed(
```
shcheklein, line 225: can we avoid extra call and just remove non complete datasets right away (we know they are incomplete w/o this call)

src/datachain/catalog/catalog.py
ilongin: I thought it's good practice to return what was actually deleted, but maybe you are right since we don't actually use that info len of removed should be sufficient.

src/datachain/data_storage/metastore.py
ilongin: Dataset versions with FAILED statuses are not visible to the any kind of public API (big part of this PR was to ensure that) but version itself is taken e.g if we have failed dataset in version `1.0.4` and someone tries to create new version it will be `1.0.5` even though `1.0.4` is marked for removal.
Regarding retention, I removed all code related to that and we can introduce it back when needed.

src/datachain/catalog/catalog.py
ilongin: Yes, fixed.

src/datachain/data_storage/metastore.py
ilongin: Dataset exists but it's not visible to any public API (UI, CLI, `dc.read_dataset()` etc.).

src/datachain/data_storage/metastore.py
ilongin: We mark them as FAILED and then removing them right away in `Session._finalize_job_as_failed()`. Question is why is this fist step of marking them as FAILED even needed since cleaning up cleans up versions in **all** non-final states if job is in final states? Reasons:
  1. It's a cheap operation (just status update)
  2. Cleaning up is heavy operation that might fail (it cleans warehouse table, related objects etc) so we will have datasets in correct state even if that happens - looks better in DB, easier to debug etc.
  3. Semantically correct (job failed → mark outputs failed)

So it's not technically needed but it's better IMO since we will have versions in cleaner state (FAILED) even if cleaning up fails afterwards.

src/datachain/data_storage/metastore.py
ilongin: No need, removing it

src/datachain/data_storage/metastore.py
ilongin: I renamed whole method to `get_incomplete_dataset_versions` to make it more correct and generic. We cannot reuse existing methods since this one creates specific query i.e. fetches versions in all incomplete statuses and related job must be in complete status.

src/datachain/data_storage/metastore.py
ilongin: In PG I've added whole new method that checks `team_id` connection to all. We couldn't just use `_datasets_versions_update()` (one of those CRUD methods that does team filtering automatically) since we don't have `team_id` column in dataset versions so we need to join with parent object (dataset). 

We should think about introducing that connection to team to every object even though it means we won't have fully normalized schema, but code will look much better (removing duplicate functions and not needed joins in PG metastore). Added issue for this https://github.com/datachain-ai/studio/issues/12438

src/datachain/data_storage/metastore.py
ilongin: When doing cleaning directly in Session `__exit__` we have specific `job_id` for which this session is made so cleaning versions only for that one.
When doing GC cleaning we clean for all the jobs.

src/datachain/data_storage/metastore.py
shcheklein: kk, let's keep it ... better for visibility indeed

src/datachain/data_storage/metastore.py
shcheklein: >since we don't have team_id column in dataset versions so we need to join with parent object (dataset).

can we add / do things like this quickly? I mean add it and deploy? what prevents us from doing this?

shcheklein: >In PG I've added whole new method that checks team_id connection to all. We couldn't just use _datasets_versions_update()

aren't keep building up tech debt with this?

src/datachain/data_storage/metastore.py
shcheklein: just to double check - do we filter by team_id here?

ilongin: Yes

src/datachain/lib/dc/datasets.py
ilongin: Refactored, this change at the end was not needed. It's always False as it should be, no special case for listings.

src/datachain/lib/dc/datasets.py
ilongin: Same as above comment, this part is returned to it's original codebase, except just adding flag to return only complete datasets.

src/datachain/query/session.py
ilongin: Explained in above comment https://github.com/datachain-ai/datachain/pull/1506#discussion_r2622591941

src/datachain/data_storage/metastore.py
ilongin: The problem is that adding `team_id` to dataset versions is probably not a few lines of code. We need to have migrations, make sure all current ones are connected to `team_id`, refactor metastore methods etc etc. Not sure, maybe it wouldn't be that hard but still this PR is already way too big, specially if we count Studio companion one. Therefor I would leave it for separate PR, but if you have strong opinion about it I can do it here as well.

PR #1513, opened 2025-12-15
Allow updating multiple targets to create pipeline

amritghimire:
## Summary

Updates the `pipeline create` command to support multiple datasets and inline version specification. Removes the `--version` flag in favor of a `@version` suffix on dataset names.

### Command interface

- Accepts multiple datasets: changed from a single `dataset` argument to `dataset [dataset ...]` (using `nargs="+"`).
- Inline version syntax: versions are specified with a `@version` suffix (e.g., `dataset@1.0.9`). If omitted, the latest version is used.
- Removed `-V/--version` flag: no longer needed.

### Implementation

Added `_parse_dataset_with_version()` to parse dataset names with optional version suffixes:
- Splits on the last `@` using `rsplit("@", 1)`.
- Validates the version part with semver.
- Returns a dict with `dataset_name` and `version` (empty string if no version).

Updated `create_pipeline()`:
- Accepts a list of dataset names.
- Parses each name to extract dataset and version.
- Passes a list of dataset dicts to the Studio API.

Updated `StudioClient.create_pipeline()`:
- Accepts `datasets` (list of dicts) instead of separate `dataset_name` and `dataset_version`.
- Each dict contains `dataset_name` and `version`.

### Documentation

- Updated `docs/commands/pipeline/create.md` with the new syntax and examples.
- Added examples showing multiple datasets and inline version specification.

## Example usage

```bash
# Single dataset with version
datachain pipeline create "@amritghimire.default.final_result@1.0.9"

# Multiple datasets with mixed version specifications
datachain pipeline create "@amritghimire.default.final_result@1.0.9" "final_result_new" "final_result_updated@1.0.9"
```

This aligns the CLI with the Studio API that accepts multiple datasets in a single request.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -39,13 +39,13 @@ jobs:
         run: uv pip install nox --system
 
       - name: Cache mypy
-        uses: actions/cache@v4
+        uses: actions/cache@v5
         with:
           path: .mypy_cache
           key: mypy-${{ runner.os }}-${{ env.pythonLocation }}-${{ hashFiles('pyproject.toml') }}
 
       - name: Cache pre-commit hooks
-        uses: actions/cache@v4
+        uses: actions/cache@v5
         with:
           path: ~/.cache/pre-commit
           key: pre-commit-3|${{ env.pythonLocation }}|${{ hashFiles('.pre-commit-config.yaml') }}
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,7 +24,7 @@ repos:
       - id: trailing-whitespace
         exclude: '^LICENSES/'
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: 'v0.14.8'
+    rev: 'v0.14.9'
     hooks:
       - id: ruff
         args: [--fix, --exit-non-zero-on-fix]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -118,7 +118,7 @@ tests = [
 ]
 dev = [
   "datachain[docs,tests]",
-  "mypy==1.19.0",
+  "mypy==1.19.1",
   "types-python-dateutil",
   "types-dateparser",
   "types-pytz",


src/datachain/studio.py
```diff
@@ -561,14 +561,32 @@ def list_clusters(team_name: str | None):
     print(tabulate.tabulate(rows, headers="keys", tablefmt="grid"))
 
 
+def _parse_dataset_with_version(dataset_input: str) -> dict[str, str]:
```
ilongin: we should move this method to `src/datachain/dataset.py`

amritghimire: This is internal function to this feature. Do we need this elsewhere?

shcheklein: we should already have similar helpers and validators? let's check this please. If not for the whole logic here, but probably for extracting semver, etc

also, in UI we still use `v1.0.0` suffix, btw ... let's research a bit though out the code all the conventions and align / cleanup this

src/datachain/utils.py
```diff
@@ -144,7 +144,12 @@ def from_dataset_dependency(
 
     def __str__(self) -> str:
         """Format DatasetIdentifier as a string: namespace.project.name@version"""
-        return f"{self.namespace}.{self.project}.{self.name}@{self.version}"
+        dataset_name = ".".join(
+            v for v in [self.namespace, self.project, self.name] if v
+        )
+        if self.version:
+            dataset_name += f"@v{self.version}"
```
shcheklein: do we really want to keep using v? I think not, since other our APIs don't accept v and it confuses everyone ...

src/datachain/studio.py
amritghimire: I couldn't find a better way to pass multiple values for dataset and versions in cli command hence the need to attach the version in the dataset name itself. 
Regarding the helper, we don't have one yet. We only have `get_full_dataset_name` which extracts the namespace, project and dataset name from the fully qualified name. But we can't use this, because it attaches the default namespace, project for local.  I couldn't find one helper where we extract semver through it. 
I found one to validate the semver, it is used here. 


> also, in UI we still use v1.0.0 suffix, btw ... let's research a bit though out the code all the conventions and align / cleanup this

Regarding the convention, I am open to any ideas on how to interpret the version here.

src/datachain/utils.py
amritghimire: This is for UI to show in the list of pipeline and v felt more natural to me. But removed v now.

src/datachain/studio.py
ilongin: My point was that this is really domain of dataset - how to parse name from version, and it's better to have it in dataset file IMO

amritghimire: Pushed there with similar utils. PTAL @ilongin

PR #1537, opened 2026-01-05
Added simple local db migration

ilongin:
Simple DB automatic migration for SQLite to add new columns to existing DB.
Note that this code doesn't cover:
1. Altering existing columns
2. Removing columns

Migration runs on every SQLite DB initialization but it is fast enough not to create overhead (there is a test added that measures performance)


pyproject.toml
```diff
@@ -173,6 +173,8 @@ filterwarnings = [
   "ignore::DeprecationWarning:datasets.utils._dill",
   "ignore::DeprecationWarning:librosa.core.intervals",
   "ignore::FutureWarning:google.api_core._python_version_support",  # Python 3.10 EOL warning
+  "ignore:The default formatters.*will be replaced by ruff:FutureWarning:datamodel_code_generator.*",  # Formatter migration warning
```
shcheklein: let's not do this please, properly fix this

src/datachain/data_storage/sqlite.py
```diff
@@ -261,6 +261,40 @@ def execute_str(self, sql: str, parameters=None) -> sqlite3.Cursor:
             return self.db.execute(sql)
         return self.db.execute(sql, parameters)
 
+    def add_column(self, table_name: str, column: Column) -> None:
```
shcheklein, line 265: is it compatible if one decides to install a previous version for example?

shcheklein: means - we have an extra column already, but it is not used

ilongin: Yes, it will work. Extra columns will be ignored. I added a test for this.

pyproject.toml
```diff
@@ -44,7 +44,7 @@ dependencies = [
   "cloudpickle",
   "pydantic",
   "jmespath>=1.0",
-  "datamodel-code-generator>=0.25",
+  "datamodel-code-generator>=0.27",
```
shcheklein: let's do these changes and Pillow as a separate PR

src/datachain/data_storage/sqlite.py
shcheklein: 👍

pyproject.toml
ilongin: fixed in another PR

pyproject.toml
ilongin: fixed and merged

src/datachain/data_storage/sqlite.py
```diff
@@ -261,6 +261,40 @@ def execute_str(self, sql: str, parameters=None) -> sqlite3.Cursor:
...
+            default_val = column.default.arg
+            if isinstance(default_val, str):
+                parts.append(f"DEFAULT '{default_val}'")
+            elif isinstance(default_val, bool):
+                parts.append(f"DEFAULT {int(default_val)}")
+            else:
+                parts.append(f"DEFAULT {default_val}")
+
+        column_def = " ".join(parts)
+        alter_query = f"ALTER TABLE {table_name} ADD COLUMN {column_def}"
```
dreadatour, line 287: I think we can use sqlalchemy here and not build raw SQL, which is fragile/not safe.

Brief example:
```python
from sqlalchemy.schema import AddColumn

table.append_column(col)
self.execute(AddColumn(table, col))
```

Haven't checked it but it looks more robust to me, what do you think?

ilongin: There is no `AddColumn` any more, SQLAlchemy stopped supporting it a while ago since it wasn't stable / portable I guess. Only thing left is to create raw query.

PR #1541, opened 2026-01-07
Fix CI warnings

ilongin:
Fix warnings from dependencies:
1. Suppresing pillow warning since we need version up to 13
2. Limit `datamodel-code-generator` since newer versions need special formatters to be specified but they are not compatible with pydantic

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -45,8 +45,8 @@ database for easy and efficient querying.
     dataframe-like API and vectorized engine to do analytics on these
     tables at scale.
 3.  **Versioning.** DataChain doesn't store, require moving or copying
-    data (unlike DVC). Perfect use case is a bucket with thousands or
-    millions of images, videos, audio, PDFs.
+    data. Perfect use case is a bucket with thousands or millions of
+    images, videos, audio, PDFs.
 
 ## Key Features
 


pyproject.toml
```diff
@@ -44,7 +44,7 @@ dependencies = [
   "cloudpickle",
   "pydantic",
   "jmespath>=1.0",
-  "datamodel-code-generator>=0.25",
+  "datamodel-code-generator>=0.25,<0.27",
```
shcheklein, line 47: @ilongin can do a fix in the code to avoid this?

pyproject.toml
```diff
@@ -107,6 +107,7 @@ tests = [
   "pytest-servers[all]>=0.5.9",
   "pytest-benchmark[histogram]",
   "pytest-xdist>=3.3.1",
+  "azure-storage-blob<12.28",  # Pin to avoid API version 2026-02-06 not supported by Azurite emulator
```
shcheklein, line 110: let's create a ticket and a reminder to update this

pyproject.toml
```diff
@@ -173,6 +174,7 @@ filterwarnings = [
   "ignore::DeprecationWarning:datasets.utils._dill",
   "ignore::DeprecationWarning:librosa.core.intervals",
   "ignore::FutureWarning:google.api_core._python_version_support",  # Python 3.10 EOL warning
+  "ignore:Image.Image.getdata is deprecated:DeprecationWarning",  # Pillow 14 deprecation (we use <13)
```
shcheklein, line 177: why do we use < 13? 

let's not ignore this ... what is the proper fix for this?

pyproject.toml
shcheklein: 0.27 is also a very old version 

https://github.com/koxudaxi/datamodel-code-generator/releases

PR #1545, opened 2026-01-08
Added logs when updating dataset version stats and preview

ilongin:
Adding logs since we had issues / inconsistency with this part of the codebase.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,7 +24,7 @@ repos:
       - id: trailing-whitespace
         exclude: '^LICENSES/'
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: 'v0.14.10'
+    rev: 'v0.14.11'
     hooks:
       - id: ruff
         args: [--fix, --exit-non-zero-on-fix]
@@ -35,7 +35,7 @@ repos:
       - id: codespell
         additional_dependencies: ["tomli"]
   - repo: https://github.com/macisamuele/language-formatters-pre-commit-hooks
-    rev: v2.15.0
+    rev: v2.16.0
     hooks:
       - id: pretty-format-toml
         args: [--autofix, --no-sort]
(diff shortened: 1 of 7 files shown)


shcheklein (changes requested):
Will it be surfaced in Studio?

Should all these logs be debug?

can we instead do this on the CH side and log some unusual responses for example? (no to clutter open source)


let's think this thotugh please ... make it clean

src/datachain/catalog/catalog.py
```diff
@@ -926,14 +926,11 @@ def update_dataset_version_with_warehouse_info(
...
 
-        logger.info(
-            "update_dataset_version_with_warehouse_info: %s@%s, "
-            "initial_num_objects=%s, initial_size=%s, has_preview=%s",
-            dataset.name,
-            version,
-            dataset_version.num_objects,
-            dataset_version.size,
-            bool(dataset_version.preview),
+        print(
```
shcheklein: does it mean it becomes pretty much always visible?

ilongin: I've added missing if clause. It will be visible only if inconsistency happens. 
Also looks like print is not needed since logger.warning is also printed in UI so I changed it to use logger. Debug and info werent printed so I though it catches only default prints for some reason but it wasn't the case.

ilongin:
> Will it be surfaced in Studio?
> 
> Should all these logs be debug?
> 
> can we instead do this on the CH side and log some unusual responses for example? (no to clutter open source)
> 
> let's think this thotugh please ... make it clean

I added some debug (changed from info to debug) in metastore. This will be visible in our Kibana.
I didn't log directly into CH warehouse but in catalog method that does actual update and calls both warehouse methods. This is where discrepancy can be seen (diff between count and preview). This log / warning will be seen in UI. It will not clutter open source since it will be logged only if issue happens.

shcheklein (approved):

shcheklein (commented):
@ilongin it was approved a while ago. Are we waiting for something with this PR?

PR #1549, opened 2026-01-12
Fix for local db migrations

ilongin:
Fixing index creation for local db migrations.
Also, added backward compatibility for old local jobs that are already connected to parents but are missing new run_group_id field.

### DIFF
--- a/src/datachain/data_storage/sqlite.py
+++ b/src/datachain/data_storage/sqlite.py
@@ -241,9 +241,6 @@ class SQLiteDatabaseEngine(DatabaseEngine):
             result = conn.execute(*self.compile_to_args(query))
         else:
             result = self.db.execute(*self.compile_to_args(query))
-        if isinstance(query, CreateTable) and query.element.indexes:
-            for index in query.element.indexes:
-                self.execute(CreateIndex(index, if_not_exists=True), cursor=cursor)
         return result
 
     @retry_sqlite_locks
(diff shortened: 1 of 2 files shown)


dreadatour (commented):
It looks like two separate fixes to me.

src/datachain/data_storage/sqlite.py
```diff
@@ -241,9 +241,6 @@ def execute(
             result = conn.execute(*self.compile_to_args(query))
         else:
             result = self.db.execute(*self.compile_to_args(query))
-        if isinstance(query, CreateTable) and query.element.indexes:
-            for index in query.element.indexes:
-                self.execute(CreateIndex(index, if_not_exists=True), cursor=cursor)
```
dreadatour, line 246: Hm... I am not sure how is this "fixing index creation for local db migrations". Could you provide more information, please? 🙏
Does this also means we will not have indexes in SQLite tables anymore?

ilongin: Indexes are still created in SQLite, just in the different place which is specific for index creation. This is the flow now:

  1. `create_table(if_not_exists=True)` → creates table structure
  2. _migrate_table_schema() → adds missing columns like `run_group_id`
  3. `_create_table_indexes()` → creates indexes AFTER columns exist

Having some implicit index creation in execute method, which shouldn't actually worry about those kind of things in the first place btw, was creating issues with missing fields where we were trying to create indexes for fields that are not even yet added to the table - whole flow was broken.

PR #1554, opened 2026-01-14
Ability to rerun a job with checkpoints in Studio using `datachain job run` CLI command

ilongin:
This PR fixes job connection (correctly setting `rerun_job_id` field) when re-running same script multiple times from local to Studio using `datachain job run my_script.py`

This was fixed by:
* adding new field `is_studio_copy` boolean field to job.
* copying newly created job in Studio to local DB (`is_studio_copy` is set to True). Job name is set to full script path (similar as we do when running jobs locally via python). Then on re-running we check for jobs that have same full script path as name and if it's copy from studio job and if we find job we send `rerun_from_job_id` field as well to Studio. Then Studio connects new job to previous one correctly and can use checkpoints etc.

### DIFF
--- a/src/datachain/studio.py
+++ b/src/datachain/studio.py
@@ -522,6 +522,8 @@ def create_job(  # noqa: PLR0913
         job_id=str(job_id),  # Use Studio's job ID
     )
 
+    catalog.close()
+
     if parsed_start_time or cron:
         print(f"Job {job_id} is scheduled as a task in Studio.")
         return 0


src/datachain/data_storage/metastore.py
```diff
@@ -468,6 +468,8 @@ def create_job(
         parent_job_id: str | None = None,
         rerun_from_job_id: str | None = None,
         run_group_id: str | None = None,
+        is_studio_copy: bool = False,
```
dreadatour: Are there any better options for this name?
- `is_remote_job`
- `is_saas_job`
- etc

ilongin: I was thinking a lot about naming and the reason I chose `is_studio_copy` is that it's by default `False` in both Studio and CLI .. it's only `True` in rare cases in CLI when we copy job from Studio. It is also very expressive / verbose and those 2 will give us the least amount of mental overhead (and less migration as well).
`is_remote_job` and `is_saas_job` will both be `True` in Studio and `False` in CLI by default which is little bit of minus IMO.. otherwise they are all pretty much similar.

amritghimire: Does is_studio_copy mean it is completely only in Studio or `has_saas_copy` or even `has_studio_copy` makes more sense?

dreadatour: But we do not _copy_ job from Studio, we are running it remotely in Studio, this is why I am agains `copy` in the name.

amritghimire: Yeah, that makes sense too. How about, `has_studio_run` or `is_in_saas`?

ilongin: @dreadatour yes, but job is still created and ran in Studio. Locally we just save link / reference / copy to it.

Couple of other ideas:
* `is_studio_reference`
* `is_executed_remotely`

Maybe the second one since it's similar to your suggestion `is_remote_job` but it doesn't have "job" word in it since it's not needed

dreadatour: You're right! `job` is redundant. What do you think about simple `is_remote` flag? (`executed` is also not needed IMO)

Anyway, these options looks much better to me personally 👍🙏

src/datachain/studio.py
```diff
@@ -424,6 +427,15 @@ def create_job(
         with open(req_file) as f:
             requirements = f.read() + "\n" + requirements
 
+    script_path = os.path.abspath(query_file)
+
+    rerun_from_job_id = None
+    rerun_from_job = catalog.metastore.get_last_job_by_name(
```
shcheklein, line 466: are we fetching it from Studio here?

tests/test_cli_studio.py
```diff
@@ -498,14 +504,147 @@ def test_studio_run_task(capsys, mocker, tmp_dir, studio_token):
     assert request_json["cron_expression"] == "0 0 * * *"
 
 
+@skip_if_not_sqlite
+def test_studio_run_connect_to_previous_job(capsys, mocker, tmp_dir, studio_token):
+    """Test that job run sends rerun_from_job_id from local Studio job copy."""
```
shcheklein: description is not clear (AI generated) / cleanup, make test name clear enough, remove comment, cleanup other comments and noise in the test

src/datachain/studio.py
shcheklein: or is it a local metastore / catalog in this case?

shcheklein (commented):
We need docs update

src/datachain/data_storage/metastore.py
ilongin: The problem with `is_remote` is that it is little bit ambiguous. Before we were calling Studio as **remote** and then the question is what is that value going to be in for example scenario where Studio creates job in Studio ... It should be `False` since job is executed locally in Studio context. Anyway, I think `is_executed_remotely` or better `is_remote_execution` is more verbose but less ambiguous IMO.

ilongin: Changed it to is_remote_execution

src/datachain/studio.py
ilongin: We are fetching from local SQLite db to see if we have cached Job with same script path .. if yes we are setting rerun_from_job_id to ensure checkpoint connection when job will be ran on Studio

tests/test_cli_studio.py
ilongin: done

ilongin:
> We need docs update

Studio docs part is updated automatically I think (I will check)

shcheklein:
> Studio docs part is updated automatically I think (I will check)

I mean review where job run is mentioned, update it probably (if needed)

shcheklein:
@ilongin tests fail (rebase?)

ilongin:
> @ilongin tests fail (rebase?)

It seems like all new PRs started to fail with some non related errors  .. I speculate it has to do with maybe new lib versions that are being released. I will investigate first thing tomorrow

shcheklein:
To repeat myself, do we need to update docs (e.g. for `datachain job run`)?

ilongin:
> To repeat myself, do we need to update docs (e.g. for `datachain job run`)?

added all needed docs.
I also added new flag --ignore-checkpoints, e.g `job run --ignore-checkpoints my-script.py` .. with it user can ignore checkpoints and run from scratch

shcheklein:
@ilongin tests, GitHub copliot review pending ... otherwise LTGM

PR #1559, opened 2026-01-16
Show logs from archives in job when logs are processed

amritghimire:
Associated PR for https://github.com/datachain-ai/studio/pull/12516

### DIFF
--- a/src/datachain/studio.py
+++ b/src/datachain/studio.py
@@ -355,16 +355,13 @@ async def _fetch_log_blob(blob_url: str, token: str, timeout: float) -> str:
 
 
 async def _show_log_blobs(log_blobs: list[str], client):
-    print("\n>>>> Fetching logs from archive files...")
-    for i, blob_url in enumerate(log_blobs, 1):
+    for blob_url in log_blobs:
         try:
             log_content = await _fetch_log_blob(blob_url, client.token, client.timeout)
             if log_content:
-                if len(log_blobs) > 1:
-                    print(f"\n--- Log file {i} ---")
                 print(log_content, end="")
-        except (requests.RequestException, OSError) as e:
-            print(f"\n>>>> Warning: Failed to fetch log blob {i}: {e}")
+        except (requests.RequestException, OSError):
+            print("\n>>>> Warning: Failed to fetch logs from studio")
 
 
 def show_logs_from_client(client, job_id):


src/datachain/studio.py
```diff
@@ -340,15 +341,47 @@ def parse_start_time(start_time_str: str | None) -> str | None:
...
+
+
+async def _show_log_blobs(log_blobs: list[str], client):
+    print("\n>>>> Fetching logs from archive files...")
+    for i, blob_url in enumerate(log_blobs, 1):
+        try:
+            log_content = await _fetch_log_blob(blob_url, client.token, client.timeout)
+            if log_content:
+                if len(log_blobs) > 1:
+                    print(f"\n--- Log file {i} ---")
```
dreadatour: Do we need this?

src/datachain/studio.py
```diff
@@ -340,15 +341,47 @@ def parse_start_time(start_time_str: str | None) -> str | None:
...
+        headers = {"Authorization": f"token {token}"}
+        response = requests.get(blob_url, headers=headers, timeout=timeout)
+        response.raise_for_status()
+        return response.text
+
+    return await asyncio.to_thread(_fetch)
+
+
+async def _show_log_blobs(log_blobs: list[str], client):
+    print("\n>>>> Fetching logs from archive files...")
```
shcheklein: people should not care about this - they don't care archive or not. It is just logs for them.

src/datachain/studio.py
```diff
@@ -340,15 +341,47 @@ def parse_start_time(start_time_str: str | None) -> str | None:
...
+    for i, blob_url in enumerate(log_blobs, 1):
+        try:
+            log_content = await _fetch_log_blob(blob_url, client.token, client.timeout)
+            if log_content:
+                if len(log_blobs) > 1:
+                    print(f"\n--- Log file {i} ---")
+                print(log_content, end="")
+        except (requests.RequestException, OSError) as e:
+            print(f"\n>>>> Warning: Failed to fetch log blob {i}: {e}")
```
shcheklein: `log blob` is not clear for users / they don't care about this . You can add a debug log if we need this, but users should not see things like "blob"

src/datachain/studio.py
shcheklein: yes, 100%, we don't need this

src/datachain/studio.py
```diff
@@ -340,15 +341,47 @@ def parse_start_time(start_time_str: str | None) -> str | None:
...
+
+    return await asyncio.to_thread(_fetch)
+
+
+async def _show_log_blobs(log_blobs: list[str], client):
+    print("\n>>>> Fetching logs from archive files...")
+    for i, blob_url in enumerate(log_blobs, 1):
+        try:
+            log_content = await _fetch_log_blob(blob_url, client.token, client.timeout)
+            if log_content:
```
shcheklein, line 361: can it be empty? when / why?

amritghimire: It is a file for the logs, it is possible it can be empty sometime if there are no logs produced.

PR #1565, opened 2026-01-22
Fixing pandas test

ilongin:
1. pandas 3.0.0: Fixed `test_read_csv_column_types` - pandas 3.0 changed default string `dtype` behavior, breaking a buggy comparison (`dtype == pd.StringDtype` class instead of instance). Now uses `pd.api.types.is_string_dtype()`.                                                       
  2. PyTorch 2.10: Added `torch<2.10.0` constraint to hf extra - torchcodec 0.9.1 only supports up to PyTorch 2.9, causing `test_hf_audio` to fail.

### DIFF
--- a/tests/unit/lib/test_datachain.py
+++ b/tests/unit/lib/test_datachain.py
@@ -1328,11 +1328,20 @@ def test_parse_nested_json(tmp_dir, test_session):
     assert sorted(df1["na_me"]["first_select"].to_list()) == sorted(
         d["first-SELECT"] for d in df["nA-mE"].to_list()
     )
+
+    # Normalize nan/None to None for comparison (pandas 3.0 returns nan instead of None)
+    def normalize_null(x):
+        return None if pd.isna(x) else x
+
     assert sorted(
-        df1["na_me"]["l_as_t"].to_list(), key=lambda x: (x is None, x)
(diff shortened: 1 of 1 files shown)


tests/unit/lib/test_datachain.py
```diff
@@ -1594,7 +1594,7 @@ def test_read_csv_column_types(tmp_dir, test_session):
         path.as_uri(), column_types={"age": "str"}, session=test_session
     )
     df1 = chain.select("first_name", "age", "city").to_pandas()
-    assert df1["age"].dtype == pd.StringDtype
+    assert pd.api.types.is_string_dtype(df1["age"])
```
dreadatour, line 1606: Will it works in both Pandas 2.x and 3.x?

dreadatour (commented):
Looks good to me, but tests are failing 😢

tests/unit/lib/test_datachain.py
dreadatour: It should works.

ilongin: Yes, it should

ilongin:
> Looks good to me, but tests are failing 😢

Additional unrelated issue with torch and torchcodec dependencies .. added fix for that as well here.

PR #1567, opened 2026-01-22
improve --env handling in CLI job run

shcheklein:
Now we can pass:

```
--env something=value \
--env something_else=value1
```

to `datachain job run` - it confused me, it confused AI, etc (it was silently just using the last entry)


amritghimire:
Tests seems to be failing, can you please take a look?

shcheklein:
@amritghimire unrelated, tocrchcodec, but yes, need to be addressed before merging ... I thought @ilongin fixed it tbh today, but it seems it something different.

PR #1568, opened 2026-01-23
Added `InsertBuffer` with flush_interval and using it in sqlite.py

ilongin:
Added a new `InsertBuffer` (moved from clickhouse.py) class in `data_storage/buffer.py` that provides unified batch insert handling with time-based flushing support. This reduces data loss on UDF failures by automatically flushing results to the database at regular intervals (default: **60** seconds), regardless of batch size.                                                                                           
                                                                                                                                 
Introduced a new `flush_interval` setting that can be configured via `DataChain.settings(flush_interval=...)` and which is used by `InsertBuffer `                           
                                                                                                                                 
Refactored SQLite's `insert_rows` to use the new `InsertBuffer`


src/datachain/lib/settings.py
```diff
@@ -22,8 +22,9 @@ class Settings:
     _project: str | None
     _min_task_size: int | None
     _batch_size: int | None
+    _flush_interval: float | None
```
dreadatour: If I am not mistaken, we do have tests for settings, it would be great to update tests with `flush_interval`.

ilongin: Added

src/datachain/data_storage/buffer.py
```diff
@@ -0,0 +1,98 @@
+"""Buffer utilities for batch database operations."""
```
shcheklein: AI-generaged / cleanup

Review everything else, line by line

src/datachain/data_storage/buffer.py
```diff
@@ -0,0 +1,98 @@
...
+    ) -> None:
+        """
+        Initialize the insert buffer.
+
+        Args:
+            table: SQLAlchemy table to insert into
+            execute_callback: Callback function to execute inserts.
+                Signature: (table, entries, final=False, cursor=None) -> None
+            buffer_size: Number of entries to accumulate before flushing
+            flush_interval: Optional time in seconds between flushes.
```
shcheklein, line 36: AFAIU there is no guarantee though that is it flushed at that interval, right? (it is still triggered by insert call or something). Can we use an actual timer that would trigger an event

src/datachain/data_storage/warehouse.py
```diff
@@ -489,20 +493,47 @@ def dataset_stats(
...
-    @abstractmethod
     def insert_rows(
         self,
         table: sa.Table,
         rows: Iterable[dict[str, Any]],
         batch_size: int | None = None,
+        flush_interval: float | None = None,
     ) -> None:
-        """Does batch inserts of any kind of rows into table"""
+        """Inserts rows in batches using an InsertBuffer.
```
shcheklein: do we care at this level about specific implementation? previous comment was better ...

src/datachain/lib/dc/datachain.py
```diff
@@ -387,6 +388,9 @@ def settings(
             batch_size: Number of rows per insert by UDF to fine tune and balance speed
                 and memory usage. This might be useful when processing large rows
                 or when running into memory issues. Defaults to 2000.
+            flush_interval: Time interval in seconds for flushing insert buffer.
```
shcheklein: let's not expose this to settings, this is way too low level, hopefully not needed

src/datachain/query/dataset.py
```diff
@@ -813,6 +822,7 @@ class RowGenerator(UDFStep):
     workers: bool | int = False
     min_task_size: int | None = None
     batch_size: int | None = None
+    flush_interval: float | None = None
```
shcheklein: can we avoid propargating this across all the settings please, I really hope this is not needed ... let's not complicate

src/datachain/data_storage/buffer.py
ilongin: You're right, it's only checked on insert/insert_many calls, not at the exact interval. A real timer would need a background thread with locking around buffer access and lifecycle management. Since inserts flow continuously in our pipeline, I think current implementation is good enough and I don't think we gain much with actual timer thread.

src/datachain/lib/dc/datachain.py
ilongin: Removed it

src/datachain/query/dataset.py
ilongin: Removed from settings

src/datachain/data_storage/buffer.py
shcheklein: kk, let's leave it simple for now

shcheklein:
@ilongin tests fail, make sure to fix them before merging

PR #1569, opened 2026-01-23
fix(subtract): make it work according docs

shcheklein:
Fixes `subtract`.

We had an issue where were not properly comparing items to subtract (not only `on` columns, but on all columns in reality). E.g. `sys__rand` was included. It means if source query had a few matching rows but with a different sys__rand we were never really subtracting it. It was leading to random / unpredictable results in production.

## Key Changes

- Replaced `EXCEPT` with `NOT EXISTS` correlated subquery for ClickHouse
- Replaced `LEFT JOIN ... WHERE IS NULL` for SQLite (base implementation)
- Used CTEs (`__dc_src_cte_*`, `__dc_tgt_cte_*`) to avoid repeated subqueries
- Added table aliasing (`__ds_t_*`) to reduce SQL verbosity

## Before / after comparison

Note: OLD results are even worse in reality, perf script is monkey patching only certain parts to the existing implementation

Execution Time (15,000 rows dataset):

| Backend | Old (EXCEPT) | New (Anti-Join) | Speedup |
|---------|-------------|-----------------|---------|
| **ClickHouse** | 3.129s | 1.860s | **1.68x faster** |
| **SQLite** | 0.968s | 0.684s | **1.42x faster** |

## Generated SQL Size (diff query)

| Backend | Old (EXCEPT) | New (Anti-Join) | Reduction |
|---------|-------------|-----------------|-----------|
| **ClickHouse** | 1,033 chars | 797 chars | **23% smaller** |
| **SQLite** | 1,043 chars | 911 chars | **13% smaller** |


tests/unit/lib/test_datachain.py
```diff
@@ -2266,75 +2266,6 @@ def test_union_different_column_order(test_session):
     ]
 
 
-def test_subtract(test_session):
```
shcheklein, line 2269: C: these three tests moved as is

src/datachain/query/dataset.py
```diff
@@ -185,9 +185,14 @@ def q(*columns):
             return sqlalchemy.select(*columns)
 
         dr = self.catalog.warehouse.dataset_rows(self.dataset, self.dataset_version)
+        # Use a short alias with dataset ID suffix for uniqueness and SQL brevity
+        ds_id = dr.table.name.rsplit("_", 1)[-1]
```
shcheklein, line 189: C: this hopefully should reduce all queries length a lot

src/datachain/query/dataset.py
```diff
@@ -264,31 +274,29 @@ def q(*columns):
...
-                    )
-                )
-                for col_name in self.on
-            ]
+        key_pairs: list[tuple[str, str]] = []
+        for col in self.on:
+            if isinstance(col, tuple):
+                key_pairs.append(col)
+            else:
+                key_pairs.append((col, col))
```
dreadatour: Duplicate (almost) with code from `hash_inputs` above:

```
normalized = [(col, col) if isinstance(col, str) else col for col in self.on]
```

Move into separate helper/method? 🤔

shcheklein: yep, abstracted it out

PR #1570, opened 2026-01-23
Lock torchcodec dependency to version <0.10.0

dreadatour:
`torchcodec` version `0.10.0` was released Jan 22, 2026. SaaS tests are now failing. Lock dependency until fix.


shcheklein:
Alternative https://github.com/datachain-ai/datachain/pull/1571 

Ideally it should be this way ^^. Have you tried it?

dreadatour:
> Alternative #1571
> 
> Ideally it should be this way ^^. Have you tried it?

Hm... Trying 👍

PR #1573, opened 2026-01-31
Make `pull` and `read_dataset` from Studio atomic

shcheklein:
Downloading a remote dataset (via `datachain pull` or `read_dataset` with a remote source) wrote directly to the final table. A crash mid-download left corrupt partial data with no cleanup path.

**Fix:** downloads now stage into a temporary table, then atomically swap it into place. Failures drop the temp table — the final table is never left in a partial state.

- File lock per dataset version prevents concurrent downloads from corrupting each other
- GC now detects and cleans up orphaned partial downloads (with a 1h grace period for in-flight ones)
- SQLite retry logic narrowed to only retry on actual lock contention

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,7 +24,7 @@ repos:
       - id: trailing-whitespace
         exclude: '^LICENSES/'
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: 'v0.15.0'
+    rev: 'v0.15.1'
     hooks:
       - id: ruff
         args: [--fix, --exit-non-zero-on-fix]


tests/func/test_read_dataset_remote.py
```diff
@@ -244,13 +224,92 @@
...
 
 
+@pytest.fixture
+def mock_studio_server(
+    remote_dataset_single_version, compressed_parquet_data, dog_entries
+):
+    parquet_data = compressed_parquet_data(dog_entries("1.0.0"))
+
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+        s.bind(("", 0))
```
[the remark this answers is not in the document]
shcheklein: this is just a pytest, with some random data, no risk here

src/datachain/catalog/catalog.py
```diff
@@ -1604,7 +1601,7 @@ def ls(
             for source in data_sources:
                 yield source, source.ls(fields)
 
-    def pull_dataset(  # noqa: C901, PLR0915
+    def pull_dataset(  # noqa: C901, PLR0915, PLR0912
```
shcheklein, line 1620: Def need a refactoring, comes as a next PR to this ... including some cleanups to progress bars

src/datachain/data_storage/warehouse.py
```diff
@@ -390,6 +390,41 @@ def create_dataset_rows_table(
     ) -> sa.Table:
         """Creates a dataset rows table for the given dataset name and columns"""
 
+    def create_temp_dataset_table(
```
shcheklein: @ilongin will some of these clash with the checkpoints PR?

src/datachain/data_storage/metastore.py
```diff
@@ -1549,29 +1575,50 @@ def get_incomplete_dataset_versions(
         dv = self._datasets_versions
         j = self._jobs
 
-        # Query dataset + version info for failed versions from non-running jobs
+        select_cols = (
```
shcheklein, line 1642: C: `get_incomplete_dataset_versions` changed to properly handle job_id == None (can be the case when dataset is created by `pull`)

shcheklein: (job can be deleted also)

src/datachain/data_storage/sqlite.py
```diff
@@ -90,6 +94,16 @@ def wrapper(*args, **kwargs):
             try:
                 return func(*args, **kwargs)
             except sqlite3.OperationalError as operror:
+                code = getattr(operror, "sqlite_errorcode", None)
```
shcheklein: optimization: w/o this some tests run 10 seconds since they go into retries on each OperationalError - some of those errors (e.g. rename table collides) should not be retried

it saves a lot of time

src/datachain/utils.py
```diff
@@ -111,6 +111,82 @@ def find(cls, create: bool = True) -> "Self":
         return instance
 
 
+@contextmanager
+def interprocess_file_lock(
```
shcheklein, line 126: C: this is to prevent multiple processes pulling the same dataset. One will be waiting ...

src/datachain/utils.py
```diff
@@ -111,6 +111,82 @@ def find(cls, create: bool = True) -> "Self":
...
+        else:
+            lock.acquire(timeout=timeout)
+
+        _write_pid()
+        yield
+    finally:
+        try:
+            os.remove(pid_path)
+        except FileNotFoundError:
+            pass
```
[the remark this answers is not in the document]
shcheklein: good catch, resolved

src/datachain/catalog/catalog.py
```diff
@@ -1691,155 +1694,192 @@ def _instantiate(ds_uri: str) -> None:
...
+                    local_ds_name,
+                    namespace_name=namespace.name,
+                    project_name=project.name,
+                    include_incomplete=False,
+                )
+                if local_dataset.has_version(local_ds_version):
+                    raise DataChainError(
+                        f"Local dataset {local_ds_uri} already exists with different"
+                        " uuid, please choose different local dataset name or version"
+                    )
```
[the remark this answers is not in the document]
shcheklein: good catch, resolved

src/datachain/catalog/catalog.py
```diff
@@ -1691,155 +1694,203 @@ def _instantiate(ds_uri: str) -> None:
...
 
-        export_data = export_response.data
-        export_id = export_data["export_id"]
-        signed_urls: list[str] = export_data["signed_urls"]
+                        try:
+                            batches = _round_robin_batch(
+                                signed_urls, PULL_DATASET_MAX_THREADS
+                            )
+                            rows_fetcher.run(iter(batches), dataset_save_progress_bar)
+                        except:
```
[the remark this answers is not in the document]
shcheklein: @copilot I think that is fine as is since we are raising it again right next line, wdyt?

src/datachain/utils.py
```diff
@@ -111,6 +111,89 @@ def find(cls, create: bool = True) -> "Self":
...
+    finally:
+        if acquired:
+            try:
+                os.remove(pid_path)
+            except FileNotFoundError:
+                logger.debug(
+                    "PID file %s was already missing during lock cleanup",
+                    pid_path,
+                )
+            lock.release()
```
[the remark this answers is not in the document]
shcheklein: addressed

src/datachain/data_storage/sqlite.py
```diff
@@ -154,14 +168,22 @@ def _connect(
...
-            # Enable Write-Ahead Log Journaling
-            db.execute("PRAGMA journal_mode = WAL")
+            # Switching to WAL requires an exclusive lock, so retry briefly
+            # in case another process is initializing the same DB file.
+            for _ in range(5):
+                try:
+                    db.execute("PRAGMA journal_mode = WAL")
+                    break
+                except sqlite3.OperationalError:
+                    sleep(1)
```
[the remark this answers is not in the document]
shcheklein: addressed

src/datachain/utils.py
```diff
@@ -122,6 +122,95 @@ def find(cls, create: bool = True) -> "Self":
...
+                    "Failed to remove PID file %s during lock cleanup",
+                    pid_path,
+                    exc_info=True,
+                )
+            finally:
+                lock.release()
+                try:
+                    os.remove(lock_path)
+                except OSError:
+                    pass
```
amritghimire: Lock file is deleted after release, which can break inter-process exclusion. Removing the lock file after release() creates a race where another process can acquire a lock on the old inode while a third process recreates the path and acquires a second lock, breaking mutual exclusion.

src/datachain/data_storage/metastore.py
```diff
@@ -373,6 +377,14 @@ def list_datasets_by_prefix(
         projects.
         """
 
+    def get_dataset_by_version_uuid(
```
amritghimire, line 380: Do we need this to be abstract method?

src/datachain/data_storage/warehouse.py
ilongin: I think it won't clash. I do think that maybe this wrapper is not even needed since you actually just call 2 methods in it and I think it's used only in one place, but it's a minor

src/datachain/data_storage/metastore.py
shcheklein: no, it raise NotImplemented (on other metastores except the local one that we need to use for pull)

src/datachain/utils.py
shcheklein: good catch!, removed os.remove

src/datachain/data_storage/warehouse.py
shcheklein: inlined it, yep

src/datachain/catalog/catalog.py
```diff
@@ -1691,155 +1692,204 @@ def _instantiate(ds_uri: str) -> None:
...
+                if signed_urls:
+                    with self.warehouse.clone() as warehouse:
+                        rows_fetcher = DatasetRowsFetcher(
+                            warehouse=warehouse,
+                            temp_table_name=temp_table_name,
+                            export_id=export_id,
+                            schema=schema,
+                            studio_client=studio_client,
+                            progress_bar=dataset_save_progress_bar,
+                        )
```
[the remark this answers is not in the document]
shcheklein: resolved

src/datachain/catalog/catalog.py
```diff
@@ -1691,155 +1692,204 @@ def _instantiate(ds_uri: str) -> None:
...
-        if signed_urls:
-            with (
-                self.metastore.clone() as metastore,
-                self.warehouse.clone() as warehouse,
-            ):
+                    final_table_name = self.warehouse.dataset_table_name(
+                        local_ds, local_ds_version
+                    )
+                    temp_table = self.warehouse.get_table(temp_table_name)
+                    self.warehouse.rename_table(temp_table, final_table_name)
```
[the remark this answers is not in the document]
shcheklein: @ilongin take a look at this pattern, rename table might need a fix (there is also copy_columns or something)

PR #1576, opened 2026-02-05
UDF checkpoint visibility

ilongin:
Companion PR of https://github.com/datachain-ai/studio/pull/12594

### DIFF
--- a/src/datachain/query/dataset.py
+++ b/src/datachain/query/dataset.py
@@ -598,9 +598,12 @@ class UDFStep(Step, ABC):
         query: Select,
         continued: bool = False,
         rows_reused: int = 0,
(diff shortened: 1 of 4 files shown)


src/datachain/query/dataset.py
```diff
@@ -1069,6 +1078,18 @@ def _skip_udf(
             rows_reused=rows_reused,
         )
 
+        # Register skipped UDF in the registry (no-op for local metastores)
+        self.metastore.add_udf(
```
ilongin, line 1088: This is needed to have `UdfRegistry` entry for skipped UDFs as well

shcheklein:
@ilongin tests failing

PR #1577, opened 2026-02-05
job run: add --no-follow and fix behavior when websocket closes early

amritghimire:
Add --no-follow so CI can wait for job completion without streaming logs to the console. When --no-follow is set we still consume the log stream and only skip printing log lines and log blobs.

When the log stream websocket closes before a final status we now fetch job status via REST and only show dataset versions if the job actually finished. Otherwise we print "Lost connection" and exit 1. Also fix the status check to use JobStatus.finished() and break on unknown status to avoid an infinite loop.

### DIFF
--- a/tests/test_cli_studio.py
+++ b/tests/test_cli_studio.py
@@ -855,7 +855,7 @@ def test_studio_run_tail_job_logs_filters_ping_and_no_follow(
             msg = messages[call_index]
             call_index += 1
             return msg
-        raise websockets.exceptions.ConnectionClosed(None, None)
+        raise websockets.exceptions.ConnectionClosed("Connection closed")
 
     captured_url = {}
 


shcheklein:
> Otherwise we print "Lost connection" and exit 1

Does it fix the issue we discussed? Can we actually retry there?

The use case is - we need to wait on CI for a job to complete ... may be we can do via a different command - like job status in a loop? But to be honest sporadic failures on connection loss is also not good .. I think it should be trying to reestablish the connection ....

docs/commands/job/run.md
```diff
@@ -43,6 +43,7 @@ This command runs a job in Studio using the specified query file. You can config
 * `--start-time START_TIME` - Time to schedule the task in YYYY-MM-DDTHH:mm format or natural language.
 * `--cron CRON` - Cron expression for the cron task.
 * `--no-wait` - Do not wait for the job to finish.
+*  `--no-follow` - Do not print the job logs to the console
```
dreadatour: Extra space?

```suggestion
* `--no-follow` - Do not print the job logs to the console
```

src/datachain/studio.py
```diff
@@ -366,21 +367,31 @@ async def _show_log_blobs(log_blobs: list[str], client):
             print("\n>>>> Warning: Failed to fetch logs from studio")
 
 
-def show_logs_from_client(client, job_id):
+def _get_job_status(client, job_id):
```
dreadatour: Lets please use types 🙏

src/datachain/studio.py
```diff
@@ -390,20 +401,35 @@ async def _run():
...
+            rest_status = _get_job_status(client, job_id)
+            if rest_status:
+                latest_status = rest_status
+
             try:
-                if retry_count > RETRY_MAX_TIMES or (
-                    latest_status and JobStatus[latest_status].finished()
-                ):
+                if latest_status and JobStatus[latest_status] in JobStatus.finished():
+                    break
```
dreadatour, line 419: Just wonder, do we want to log something here?

src/datachain/studio.py
```diff
@@ -390,20 +401,35 @@ async def _run():
...
+                if retry_count > RETRY_MAX_TIMES:
                     break
```
dreadatour, line 422: Also here? Any logs, errors? 🤔

src/datachain/studio.py
```diff
@@ -390,20 +401,35 @@ async def _run():
...
 
     final_status = asyncio.run(_run())
 
+    try:
+        job_finished = final_status and JobStatus[final_status] in JobStatus.finished()
+    except KeyError:
+        job_finished = False
+
+    if not job_finished:
+        print(f"\n>>>> Lost connection. Job status: {final_status or 'unknown'}.")
```
dreadatour, line 440: Hm... It looks like it will be a "Lost connection" for all kind of errors.

src/datachain/studio.py
```diff
@@ -417,11 +443,13 @@ async def _run():
...
-        "FAILED": 1,
-        "CANCELED": 2,
-    }
-    return exit_code_by_status.get(final_status.upper(), 0) if final_status else 0
+    if final_status.upper() == "COMPLETE":
+        return 0
+    if final_status.upper() == "FAILED":
+        return 1
+    if final_status.upper() == "CANCELED":
+        return 2
```
dreadatour, line 462: I am not saying we should use it, but we do have [QUERY_SCRIPT_CANCELED_EXIT_CODE](https://github.com/datachain-ai/datachain/blob/main/src/datachain/catalog/catalog.py#L77) exit status.

src/datachain/studio.py
```diff
@@ -366,21 +367,31 @@ async def _show_log_blobs(log_blobs: list[str], client):
...
+                if "log_blobs" in message and not no_follow:
                     log_blobs = message.get("log_blobs", [])
                     if log_blobs and not log_blobs_processed:
                         log_blobs_processed = True
                         await _show_log_blobs(log_blobs, client)
 
-                elif "logs" in message:
+                elif "logs" in message and not no_follow:
                     for log in message["logs"]:
                         print(log["message"], end="")
```
dreadatour: I am not sure about the logic here.

- We are using `tail_job_logs` method
- We will utilize the connection and receive all the logs even if we are ignoring them all

What is the reason of tailing all the logs and not showing them? 🤔

src/datachain/studio.py
amritghimire: It will just pollute in midst of actual logs. We handle the case later.

src/datachain/studio.py
amritghimire: IMO It will just pollute in midst of actual logs. We handle the case later.

src/datachain/studio.py
amritghimire: It will be lost connection in case we are unable to fetch the job information i.e websocket connection closes, we exhausted retry times, and final status is still not among finished state.

src/datachain/studio.py
amritghimire: It was according to request from @shcheklein at https://datachain-ai.slack.com/archives/C04A9RWEZBN/p1770254719556309

amritghimire:
> > Otherwise we print "Lost connection" and exit 1
> 
> Does it fix the issue we discussed? Can we actually retry there?

The issue was that even if the latest status was running or not completed, and the connection was lost, it returned success exit code. This fixes that. It also fixes retry and we do have retry, we retry the connection for 10 times with 1 sec sleep and if it fails, then we now fetches the status from the rest api and adjust the return code accordingly.

src/datachain/studio.py
dreadatour: Not sure if I am following. Does this means we will retry again in other place? Why do we retry here then?

amritghimire: I mean we retry here. If the retry count exceeds and job is not completed, we print lost connection later.

src/datachain/studio.py
dreadatour: Right above it is:

```python
    try:
        job_finished = final_status and JobStatus[final_status] in JobStatus.finished()
    except KeyError:
        job_finished = False
```

So it is not only "lost connection", it is all kind of errors?

src/datachain/studio.py
dreadatour: Yes, right, but again, if we are not printing the logs, why do we need to fetch from from server then? Let's just only fetch (or even poll?) job status change, this is much simpler, faster and more robust?

Otherwise there is no sense in `--no-follow` param, we can basically redirect output to `/dev/null`.

src/datachain/studio.py
dreadatour: Got it. Does this means we are "polluting in midst of actual logs" anyway, but from parent method with one common error message for all cases?

Again, about logger — what do you think of using it with, for example, info or debug level, so we can find all these issues if needed? Also we have this CLI option:

> `-v`, `--verbose` - Be verbose.

Why not to use it?

src/datachain/studio.py
amritghimire: I actually like this, and it solves the use case too. LMK what you think ? cc. @shcheklein

src/datachain/studio.py
amritghimire: No, because the `tail_job_logs` uses the folllowing: 
```py
               except websockets.exceptions.ConnectionClosed:
                    break
```

In case of other exceptions, we log the error using `logger.error`.

src/datachain/studio.py
amritghimire: We log in case of other exceptions only connection closed doesn't produce log. In other case we have logger.error .

src/datachain/studio.py
shcheklein: no, it doesn't fully solve the use case - you want to see in CI some summary, but not 1GB of logs (it breaks it), so /dev/null is not good there ... or I misunderstood the suggestion

we can definitely optimize and don't fetch / pull logs if they are not needed

shcheklein:
> . It also fixes retry and we do have retry, we retry the connection for 10 times with 1 sec sleep and if it fails, then we now fetches the status from the rest api and adjust the return code accordingly.

what if job is still running? what will be the return code?

are we retrying only websockets?

should we fail if 10 times is not enough?

also, are we going to receive messages we missed while disconnected?

src/datachain/studio.py
amritghimire: Then yeah, current implementation helps here. We are just printing status changes (summary). Logs are discarded from the websocket.

amritghimire:
> > . It also fixes retry and we do have retry, we retry the connection for 10 times with 1 sec sleep and if it fails, then we now fetches the status from the rest api and adjust the return code accordingly.
> 
> what if job is still running? what will be the return code?
> 
> are we retrying only websockets?
> 
> should we fail if 10 times is not enough?

If job is still running, it returns 1 exit code with lost connection error. But only after all the 10 retries exhausts.

src/datachain/studio.py
shcheklein: Let's not use verbose please (`-v` is to control the app logs itself - e.g. websocket reconnects, etc, not user logs)

dreadatour: > `-v` is to control the app logs itself - e.g. websocket reconnects, etc, not user logs

This is exactly what I was talking about — to add more trace or info level logs and to show them if verbose mode is on.

src/datachain/studio.py
dreadatour: > I actually like this

It was an example of bad design 👀😢

> Then yeah, current implementation helps here. We are just printing status changes (summary). Logs are discarded from the websocket.

My point about current implementation is what we are getting all the logs from server and just hiding them. I am suggesting to do not fetch them if `--no-follow` option is on.

dreadatour: As an option it might be 2 different threads — one to fetch job status and stats and other one to fetch job logs (if needed, if `--no-follow` is not set). Then we can retry second one as many times as needed, if job is still running.

src/datachain/studio.py
dreadatour: Wait. Here is the code:

```python
    try:
        job_finished = final_status and JobStatus[final_status] in JobStatus.finished()
    except KeyError:
        job_finished = False

    if not job_finished:
        print(f"\n>>>> Lost connection. Job status: {final_status or 'unknown'}.")
```

If it will be a `KeyError` exception, `job_finished` will be set to `False` and it will end up with "Lost connection" message.

Is that correct?

src/datachain/studio.py
amritghimire: Yeah, let me add a param no follow in websocket connection.

src/datachain/studio.py
amritghimire: So, the check above is to see if the job is finished or not after all the retry is exhausted. Key error here usually mean , we got new status, that was introduced in studio, but not propagated in datachain. In either case, it is not among the finished status. ` return cls.COMPLETE, cls.FAILED, cls.CANCELED, cls.TASK` . So, it still means we lost the connection while job was not completed yet. Hope it make sense.

src/datachain/studio.py
amritghimire: Added a param to no follow, when it passed, we will not get logs.

src/datachain/studio.py
amritghimire: Updated, PTAL.

shcheklein:
@amritghimire tests fail

@dreadatour review one more time please, we need to release this

shcheklein:
@amritghimire codecov fails, please review

@dreadatour please review your comments. We need this released pretty much today.

amritghimire:
> @amritghimire codecov fails, please review

Added more tests, but wasn't aware codecov patch was also required check.

shcheklein:
> Added more tests, but wasn't aware codecov patch was also required check.

I guess we have it set to fail for a reason? To keep coverage growing at least ...

src/datachain/studio.py
```diff
@@ -390,20 +404,43 @@ async def _run():
...
+            rest_status = _get_job_status(client, job_id)
+            if rest_status:
+                latest_status = rest_status
+
             try:
-                if retry_count > RETRY_MAX_TIMES or (
-                    latest_status and JobStatus[latest_status].finished()
-                ):
+                if latest_status and JobStatus[latest_status] in JobStatus.finished():
+                    if verbose:
```
shcheklein: can it be done with logs (and just ignore DEBUG log level if -v is not provided)

it looks weird to have all this manual manipulations and prints

src/datachain/studio.py
```diff
@@ -366,21 +368,33 @@ async def _show_log_blobs(log_blobs: list[str], client):
...
 
 
-def show_logs_from_client(client, job_id):
+def _get_job_status(client, job_id: str):
+    try:
+        response = client.get_jobs(job_id=job_id)
+        if response.ok and response.data and len(response.data) > 0:
+            return response.data[0].get("status")
+    except (requests.RequestException, OSError, KeyError):
+        pass
```
shcheklein: why is it safe to pass here?

amritghimire:
> > Added more tests, but wasn't aware codecov patch was also required check.
> 
> 
> 
> I guess we have it set to fail for a reason? To keep coverage growing at least ... 

Is it set? I think merge is blocked due to review required.

shcheklein:
I mean it fails. It is not a blocker since there are lot of cases where code was moved that didn't have coverage before and some other false positives. It fails since we agreed to pay reasonable attention to it, we agree to fail if actual new code is not covered (reasonably, not religiously just to get numbers bumped). I just notices recently that we all stopped paying attention (reviewing if new code in some edge cases is not covered). This is expected (us losing attentions) - since it is not a blocker. So, it was just a reminder (I aksed @ilongin also to review UDFs).

Screenshot 2026-02-08 at 12 11 52 PM

src/datachain/studio.py
amritghimire: We handle it graciously later, with message network error if we cannot get status.

shcheklein: let's add types

let's add logger 

do we really need all three exception types?

do we have test for this situation?

tests/test_cli_studio.py
```diff
@@ -694,3 +702,205 @@ async def mock_tail_job_logs(jid):
         ">>>> Dataset versions created during the job:\n"
         "    - dataset_name@v1.0.0"
     )
+
+
+def test_studio_run_websocket_disconnect_fetches_status_via_rest(
```
shcheklein, line 710: will it continue waiting? or will it just exit now when socket is disconnected (I guess how reliable default mode is?)

amritghimire: It waits for 10 retry until the job is completed.

shcheklein: > It waits for 10 retry until the job is completed. 

what does it exactly mean? how long is retry?

amritghimire: It means it retries for 10 times with 1s sleep, when the connection is disconnected while trying the websocket.

src/datachain/studio.py
```diff
@@ -390,20 +406,41 @@ async def _run():
                     processed_statuses.add(latest_status)
                     print(f"\n>>>> Job is now in {latest_status} status.")
 
+            # After websocket closes, check actual job status via REST
+            rest_status = _get_job_status(client, job_id)
+            if rest_status and rest_status != latest_status:
+                print(f"\n>>>> Job is now in {rest_status} status.")
+            if rest_status:
```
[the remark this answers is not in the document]
amritghimire: Should be okay, it is in case websockets are disconnected.

PR #1578, opened 2026-02-06
Avoid  error if orphaned dataset table exist

ilongin:
Orphaned dataset version table can appear if error happens between removing dataset version metadata from metastore and removing it's data table from warehouse. In this case, user cannot create new dataset version any more since it hits the issue of existing dataset table every time - user is blocked here.

This PR unblocks user by removing all existing rows in destination table in `copy_table()` method so if orphaned table exist, it will be cleaned before actually doing copy.

Original issue: https://github.com/datachain-ai/studio/issues/12572


ilongin:
> * Why are we getting these orphaned tables?

In general the issue is that we cannot have transaction having both metastore (PG) and warehouse (CH) tables included. 
Example: if something fails in `DatasetQuery.save()` after we created new dataset version the metastore cleanup happens in `Session.__exit__` via `cleanup_failed_dataset_versions`. The orphan occurs because `remove_dataset_version` does metastore removal first, then warehouse table drop - and if the drop fails (or the process dies between the two), you get an orphaned warehouse table.
Similar can happen on delete dataset version I think as well - we first remove metastore and then if process fails we never get to remove warehouse table.
We could have some kind of cron job checking for orphan tables and cleaning them up.

> * Can we focus and get checkpoints done first (some tests are still failing in some PR, etc, etc)- keep cleaning up, etc

Sure, PRs are ready for merge.

> * What happens if this table is created in a parallel still running process?

Having concurrent processes creating the same dataset version is an invalid state by itself, that shouldn't happen at all.


BTW I moved fix in other place, not in copy_table - that was mistake to have it there since it's strange to have implicit constraint for destination table to be empty. Fix is now in place where we actually create new dataset version - there we do expect not having orphan dataset version tables and we now clean it if they occur.

shcheklein:
> In general the issue is that we cannot have transaction having both metastore (PG) and warehouse (CH) tables included.

so, should we then do something like:

on create:

- first do CREATED metastore entry
- then do warehouse table (probably with rename to make it as quick and atomic as possible)
- then do status update to COMPLETE

on cleanup:

- delete warehouse first
- delete entry then

on reentry to create:

- if CREATED exists - wait a bit it can be a parallel process creating, we can get it and continue or create a new one then
- run cleanup otherwise (see above) - proceed to create

> Having concurrent processes creating the same dataset version is an invalid state by itself, that shouldn't happen at all.

is it actually prevented now? do we calculate next version by also including incomplete versions?

ilongin:
> > In general the issue is that we cannot have transaction having both metastore (PG) and warehouse (CH) tables included.
> 
> so, should we then do something like:
> 
> on create:
> 
> * first do CREATED metastore entry
> * then do warehouse table (probably with rename to make it as quick and atomic as possible)
> * then do status update to COMPLETE
> 
> on cleanup:
> 
> * delete warehouse first
> * delete entry then
> 

Yes, we can do that as well. But I would still leave this cleanup before creating just in case...it doesn't hurt to have it there.
EDIT: I added reverse order on cleanup. Creation is already handled the way you described.

> is it actually prevented now? do we calculate next version by also including incomplete versions?

Yes, incomplete versions are included in the calculation. This means two concurrent processes won't compute the same version number - the incomplete version bumps the counter. There's also a unique constraint on (dataset_id, version).

shcheklein:
@ilongin 

> https://github.com/datachain-ai/studio/issues/12572

Let's double check, I don't think this is about orphaned tables. In that example two jobs were doing listing simultaneously.

> it doesn't hurt to have it there.

I think it is better to fail fast then? to detect an actual issue somewhere upstream?

ilongin:
> @ilongin
> 
> > [datachain-ai/studio#12572](https://github.com/datachain-ai/studio/issues/12572)
> 
> Let's double check, I don't think this is about orphaned tables. In that example two jobs were doing listing simultaneously.

No, it was about orphaned tables. I checked it in details, including lookup into CH database. This happened for Vlad in dev for 2 older datasets

> I think it is better to fail fast then? to detect an actual issue somewhere upstream?

If we fail fast this means user won't be able to create new version of dataset if orphan table somehow exists in the system, and he will be blocked there, therefor I think solution in this PR is ok.

shcheklein:
> No, it was about orphaned tables. I checked it in details, including lookup into CH database. This happened for Vlad in dev for 2 older datasets

share the exact details please

ilongin:
Closing in favor of https://github.com/datachain-ai/datachain/pull/1675

PR #1579, opened 2026-02-08
fix(udf): fix errors pickling / unpicking to show proper messages

shcheklein:
Fixes error handing in `setup()` and in some other case. 

We don't really show error message now if setup fails in distributed / parallel mode. We show some ugly pickle error since it can't restore the error properly.

### DIFF
--- a/tests/unit/lib/test_datachain_bootstrap.py
+++ b/tests/unit/lib/test_datachain_bootstrap.py
@@ -1,11 +1,15 @@
+import os
+import pickle
+
 import pytest
 
 import datachain as dc
 from datachain.lib.dc import DatasetPrepareError
+from datachain.lib.signal_schema import SetupError
 from datachain.lib.udf import Mapper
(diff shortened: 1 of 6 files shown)


tests/unit/lib/test_datachain_bootstrap.py
```diff
@@ -1,11 +1,14 @@
...
 import pytest
 
 import datachain as dc
 from datachain.lib.dc import DatasetPrepareError
+from datachain.lib.signal_schema import SetupError
 from datachain.lib.udf import Mapper
 
 
-def test_udf(monkeypatch):
+def test_udf(test_session, monkeypatch):
```
shcheklein, line 12: @dreadatour why distributed is ignored here?

also, is it fine to use test_session? does it change anything in terms of distributed or not, etc?

dmpetrov (approved):
LG

PR #1590, opened 2026-02-16
UDF Checkpoints cleanup 

ilongin:
Implements checkpoint cleanup with TTL-based garbage collection:
- cleanup_checkpoints() method in Catalog to remove old checkpoints
- _remove_checkpoint() helper to clean checkpoint and its UDF tables
- list_checkpoints() enhancement to support time-based filtering
- get_descendant_job_ids() to find child jobs for dependency checking
- remove_checkpoint() in metastore to delete checkpoint records
- CLI integration in garbage_collect command
- Comprehensive tests for cleanup scenarios

### DIFF
--- a/src/datachain/data_storage/metastore.py
+++ b/src/datachain/data_storage/metastore.py
@@ -54,7 +54,6 @@ from datachain.dataset import (
     DatasetRecord,
     DatasetStatus,
     DatasetVersion,
-    StorageURI,
     parse_schema,
 )
 from datachain.error import (
@@ -96,8 +95,6 @@ class AbstractMetastore(ABC, Serializable):
     This manages the storing, searching, and retrieval of indexed metadata.
     """
 
-    uri: StorageURI
-
     schema: "schema.Schema"
     namespace_class: type[Namespace] = Namespace
     project_class: type[Project] = Project
@@ -111,12 +108,6 @@ class AbstractMetastore(ABC, Serializable):
     checkpoint_class: type[Checkpoint] = Checkpoint
     checkpoint_event_class: type[CheckpointEvent] = CheckpointEvent
 
-    def __init__(
-        self,
-        uri: StorageURI | None = None,
-    ):
-        self.uri = uri or StorageURI("")
-
     def __enter__(self) -> Self:
         """Returns self upon entering context manager."""
         return self
@@ -127,10 +118,9 @@ class AbstractMetastore(ABC, Serializable):
     @abstractmethod
     def clone(
         self,
-        uri: StorageURI | None = None,
         use_new_connection: bool = False,
     ) -> "AbstractMetastore":
-        """Clones AbstractMetastore implementation for some Storage input.
+        """Clones AbstractMetastore implementation.
         Setting use_new_connection will always use a new database connection.
         New connections should only be used if needed due to errors with
         closed connections."""
@@ -705,10 +695,6 @@ class AbstractDBMetastore(AbstractMetastore):
     CHECKPOINT_EVENTS_TABLE = "checkpoint_events"
 
     db: "DatabaseEngine"
-
-    def __init__(self, uri: StorageURI | None = None):
-        uri = uri or StorageURI("")
-        super().__init__(uri)
 
     def close(self) -> None:
         """Closes any active database connections."""
--- a/src/datachain/data_storage/sqlite.py
+++ b/src/datachain/data_storage/sqlite.py
@@ -448,14 +448,11 @@ class SQLiteMetastore(AbstractDBMetastore):
 
     def __init__(
         self,
-        uri: StorageURI | None = None,
         db: SQLiteDatabaseEngine | None = None,
         db_file: str | None = None,
         in_memory: bool = False,
     ):
-        uri = uri or StorageURI("")
         self.schema: DefaultSchema = DefaultSchema()
-        super().__init__(uri)
 
         # needed for dropping tables in correct order for tests because of
         # foreign keys
@@ -478,14 +475,9 @@ class SQLiteMetastore(AbstractDBMetastore):
 
     def clone(
         self,
-        uri: StorageURI | None = None,
         use_new_connection: bool = False,
     ) -> "SQLiteMetastore":
-        uri = uri or StorageURI("")
-        if not uri and self.uri:
-            uri = self.uri
-
-        return SQLiteMetastore(uri=uri, db=self.db.clone())
+        return SQLiteMetastore(db=self.db.clone())
 
     def clone_params(self) -> tuple[Callable[..., Any], list[Any], dict[str, Any]]:
         """
@@ -496,7 +488,6 @@ class SQLiteMetastore(AbstractDBMetastore):
             SQLiteMetastore.init_after_clone,
             [],
             {
-                "uri": self.uri,
                 "db_clone_params": self.db.clone_params(),
             },
         )
@@ -509,11 +500,10 @@ class SQLiteMetastore(AbstractDBMetastore):
     def init_after_clone(
         cls,
         *,
-        uri: StorageURI,
         db_clone_params: tuple[Callable, list, dict[str, Any]],
     ) -> "SQLiteMetastore":
         (db_class, db_args, db_kwargs) = db_clone_params
-        return cls(uri=uri, db=db_class(*db_args, **db_kwargs))
+        return cls(db=db_class(*db_args, **db_kwargs))
 
     @cached_property
     def _meta(self) -> Table:
--- a/tests/unit/test_catalog_loader.py
+++ b/tests/unit/test_catalog_loader.py
@@ -14,7 +14,6 @@ from datachain.data_storage.sqlite import (
     SQLiteMetastore,
     SQLiteWarehouse,
 )
-from datachain.dataset import StorageURI
 
 
 class DistributedClass:
@@ -23,11 +22,8 @@ class DistributedClass:
 
 
 def test_get_metastore(sqlite_db):
-    uri = StorageURI("s3://bucket")
-
-    metastore = SQLiteMetastore(uri, sqlite_db.clone())
+    metastore = SQLiteMetastore(db=sqlite_db.clone())
     try:
-        assert metastore.uri == uri
         assert metastore.db.db_file == sqlite_db.db_file
 
         with patch.dict(os.environ, {"DATACHAIN__METASTORE": metastore.serialize()}):
@@ -35,7 +31,6 @@ def test_get_metastore(sqlite_db):
             try:
                 assert metastore2
                 assert isinstance(metastore2, SQLiteMetastore)
-                assert metastore2.uri == uri
                 assert metastore2.db.db_file == sqlite_db.db_file
                 assert metastore2.clone_params() == metastore.clone_params()
             finally:
@@ -146,8 +141,7 @@ def test_get_distributed_class(monkeypatch):
 
 
 def test_get_catalog(sqlite_db):
-    uri = StorageURI("s3://bucket")
-    metastore = SQLiteMetastore(uri, sqlite_db.clone())
+    metastore = SQLiteMetastore(db=sqlite_db.clone())
     warehouse = SQLiteWarehouse(sqlite_db.clone())
     env = {
         "DATACHAIN__METASTORE": metastore.serialize(),
@@ -161,7 +155,6 @@ def test_get_catalog(sqlite_db):
 
             assert catalog.metastore
             assert isinstance(catalog.metastore, SQLiteMetastore)
-            assert catalog.metastore.uri == uri
             assert catalog.metastore.db.db_file == sqlite_db.db_file
             assert catalog.metastore.clone_params() == metastore.clone_params()
 
--- a/tests/unit/test_metastore.py
+++ b/tests/unit/test_metastore.py
@@ -8,23 +8,18 @@ from datachain.checkpoint import CheckpointStatus
 from datachain.data_storage.job import JobQueryType, JobStatus
 from datachain.data_storage.serializer import deserialize
 from datachain.data_storage.sqlite import SCHEMA_VERSION, SQLiteMetastore
-from datachain.dataset import StorageURI
 from datachain.error import OutdatedDatabaseSchemaError
 from tests.conftest import cleanup_sqlite_db
 
 
 def test_sqlite_metastore(sqlite_db):
-    uri = StorageURI("s3://bucket")
-
-    obj = SQLiteMetastore(uri, sqlite_db)
-    assert obj.uri == uri
+    obj = SQLiteMetastore(db=sqlite_db)
     assert obj.db == sqlite_db
 
     # Test clone
     obj2 = obj.clone()
     try:
         assert isinstance(obj2, SQLiteMetastore)
-        assert obj2.uri == uri
         assert obj2.db.db_file == sqlite_db.db_file
         assert obj2.clone_params() == obj.clone_params()
 
@@ -35,7 +30,6 @@ def test_sqlite_metastore(sqlite_db):
         data = json.loads(raw.decode())
         assert data["callable"] == "sqlite.metastore.init_after_clone"
         assert data["args"] == []
-        assert data["kwargs"]["uri"] == uri
         nested = data["kwargs"]["db_clone_params"]
         assert nested["callable"] == "sqlite.from_db_file"
         assert nested["args"] == [":memory:"]
@@ -44,7 +38,6 @@ def test_sqlite_metastore(sqlite_db):
         obj3 = deserialize(serialized)
         try:
             assert isinstance(obj3, SQLiteMetastore)
-            assert obj3.uri == uri
             assert obj3.db.db_file == sqlite_db.db_file
             assert obj3.clone_params() == obj.clone_params()
         finally:


src/datachain/catalog/catalog.py
```diff
@@ -2042,3 +2044,75 @@ def index(
...
             pass
+
+    def _remove_checkpoint(self, checkpoint: Checkpoint) -> None:
+        """Remove a checkpoint and its associated job-specific UDF tables."""
+        self.metastore.remove_checkpoint(checkpoint.id)
+
+        # Remove job-specific tables for this checkpoint
+        # Table patterns: udf_{job_id}_{hash}_{suffix}
+        # where suffix can be: input, output, output_partial, processed
+        job_id_sanitized = checkpoint.job_id.replace("-", "")
```
shcheklein: is there some helper that gets these names reliably?

shcheklein: do we test that some other tables are not removed along the way by some mistake?

src/datachain/data_storage/metastore.py
```diff
@@ -533,8 +544,24 @@ def get_last_job_by_name(
...
-        """Returns all checkpoints related to some job"""
+    def list_checkpoints(
+        self,
+        job_id: str | None = None,
+        created_after: datetime | None = None,
+        created_before: datetime | None = None,
+        conn=None,
+    ) -> Iterator[Checkpoint]:
+        """
+        List checkpoints by job id, or all checkpoints if job_id is None.
```
shcheklein: probably wrong comments

review and cleanup AI generated stuff

src/datachain/cli/commands/misc.py
```diff
@@ -13,18 +13,24 @@ def clear_cache(catalog: "Catalog"):
...
+        return
 
-        if num_versions_removed:
-            print(f"Cleaned {num_versions_removed} failed/incomplete dataset versions.")
+    # Print and perform cleanup for each category that has items
+    if temp_tables:
+        print(f"Garbage collecting {len(temp_tables)} temporary tables.")
+        catalog.cleanup_tables(temp_tables)
+
+    if num_versions_removed:
```
shcheklein: Should we print after each action (better for seeing some progress).

Also we need something similar to what we had for temp tables ... Collecting ... Done style (two messages)

src/datachain/catalog/catalog.py
```diff
@@ -2042,3 +2044,75 @@ def index(
...
+        if ttl_seconds is None:
+            ttl_seconds = int(os.environ.get("CHECKPOINT_TTL", str(TTL_INT)))
+
+        ttl_threshold = datetime.now(timezone.utc) - timedelta(seconds=ttl_seconds)
+
+        outdated_checkpoints = list(
+            self.metastore.list_checkpoints(created_before=ttl_threshold)
+        )
+
+        if not outdated_checkpoints:
```
shcheklein: let's add more logs here - how many we got to clean, which one we are cleaning now, how many tables (and may be even table names in a single message)

src/datachain/catalog/catalog.py
shcheklein: let's Checkpoint make aware which tables belong to it?

ilongin: If you think about saving this tables to checkpoint model, I don't agree with this. Checkpoint related UDF tables are detereministic and we 100% know which one belong to specific checkpoint by job_id and hash combo. this would be just another redundant  thing to maintain and risk of being non consistent / in bad state.

ilongin: I updated cleaning logic a little bit, therre was some bug there as it was relying on logic before some updates made in udf checkpoints branch with run_group_id etc.

ilongin: added test to check that non related tables are not cleaned up

src/datachain/cli/commands/misc.py
ilongin: For checkpoints and failed dataset versions we don't have a way to count before so it will be:
"Collecting outdated checkpoints..."
"Removed 5 outdated checkpoints."

I did the same for temp tables to be consistent in messaging

src/datachain/catalog/catalog.py
ilongin: added

ilongin:
> Do we need to update docs?
> What is the expected user workflow for this? Run manually? Is `gc` even documented? Is it even available or still hidden?

it's to run manually as in general with `gc`. It's not documented anywhere ATM, just visible on `--help` from CLI  

Anyway, I added docs for GC here in this PR, take a look.

docs/commands/gc.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Options
+
+* `-h`, `--help` - Show the help message and exit.
+* `-v`, `--verbose` - Be verbose.
+* `-q`, `--quiet` - Be quiet.
+
+## Environment Variables
+
+* `CHECKPOINT_TTL` - Time-to-live for checkpoints in seconds. Checkpoints older than this value are considered outdated and eligible for cleanup. Defaults to 4 hours (14400 seconds).
```
shcheklein: it should have prefix DATACHAIN

also, do we really need yet another variable for this? 

it is better to use a direct proper CLI argument

src/datachain/catalog/catalog.py
```diff
@@ -2092,3 +2093,103 @@ def index(
...
+            )
+            self.metastore.remove_checkpoint(ch.id)
+            cleaned_job_ids.add(ch.job_id)
+            cleaned_run_group_ids.add(run_group_id)
+            removed_count += 1
+
+        # Clean up all UDF tables for cleaned jobs and run groups.
+        # Prefix scan catches output, partial, and input tables regardless
+        # of which hash variant was used to name them.
+        for job_id in cleaned_job_ids:
```
shcheklein: Are we removing warehouse tables after metadata? It should be other way around to my mind and be resilient for missing tables

src/datachain/catalog/catalog.py
```diff
@@ -2092,3 +2093,103 @@ def index(
...
+        # Track cleaned job_ids and run_group_ids for table cleanup
+        cleaned_job_ids: set[str] = set()
+        cleaned_run_group_ids: set[str] = set()
+
+        removed_count = 0
+        for ch in outdated_checkpoints:
+            job = jobs.get(ch.job_id)
+            run_group_id = job.run_group_id if job else None
+
+            if not run_group_id:
```
shcheklein: how can we have checkpoint w/o a run group?

src/datachain/catalog/catalog.py
```diff
@@ -2092,3 +2093,103 @@ def index(
...
+        for a job/group, cleans up all associated UDF tables (output,
+        partial, and input) by prefix scan.
+        """
+        if ttl_seconds is None:
+            ttl_seconds = int(os.environ.get("CHECKPOINT_TTL", str(TTL_INT)))
+
+        ttl_threshold = datetime.now(timezone.utc) - timedelta(seconds=ttl_seconds)
+
+        outdated_checkpoints = list(
+            self.metastore.list_checkpoints(created_before=ttl_threshold)
```
shcheklein: can we redesign it to get run_groups that are 100% inactive ... then get checkpoint that belong to those run_groups

btw, should we actually check for a job, not a run_group .. run_group can be very long potentially (thus we won't be removing checkpoints for a long time)

tests/func/checkpoints/test_checkpoint_cleanup.py
```diff
@@ -0,0 +1,164 @@
...
+    catalog.cleanup_checkpoints()
+
+    # Expired job's checkpoints and tables are gone
+    assert len(list(metastore.list_checkpoints(job_id))) == 0
+    assert len(warehouse.db.list_tables(prefix=f"udf_{job_id}_")) == 0
+
+    # Unrelated table survives
+    assert warehouse.db.has_table(fake_table_name)
+
+    # Clean up
```
shcheklein: denoise, cleanup AI generated stuff 

please pay attention to this

src/datachain/catalog/catalog.py
ilongin: Legacy ones, although probably none of those are in use any more.

docs/commands/gc.md
ilongin: I've added CLI arg and removed ENV var

src/datachain/catalog/catalog.py
ilongin: I refactored to not wait for all run group checkpoints to be outdated to clean them to clean those that are outdated and that's it.
Function is still complex as it has couple of important constraints, but since this is only going to run every now and then I don't think it's important.
Constraints are:
1. Clean all job checkpoints or none (I think it would be weird if some checkpoints are removed and some not in single job)
2. Clean input table only if whole run group checkpoints are going to be cleaned (because input tables are shared in run group)

docs/commands/gc.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Description
+
+This command cleans up internal DataChain storage by removing:
+
+- **Temporary tables** created during query execution that were not properly cleaned up (e.g., due to crashes or interrupted operations).
+- **Failed dataset versions** that were left in an incomplete or failed state.
+- **Outdated checkpoints** and their associated UDF tables that have exceeded the time-to-live (TTL) threshold. See [Checkpoints](../guide/checkpoints.md) for more details.
+
+Each category is collected and cleaned independently. Progress is reported for each step.
```
shcheklein: this is not needed

src/datachain/catalog/catalog.py
```diff
@@ -2092,3 +2094,95 @@ def index(
...
+    def cleanup_checkpoints(self, ttl_seconds: int | None = None) -> int:
+        """Clean up outdated checkpoints and their associated UDF tables.
+
+        For each job whose checkpoints are ALL outdated (older than TTL),
+        removes per-job output/partial tables and checkpoint metadata.
+        Input tables (shared across a run group) are only removed when
+        the entire run group has no active checkpoints.
+        """
+        if ttl_seconds is None:
+            ttl_seconds = TTL_INT
```
shcheklein: what is the semantics of TTL_INT  - is it okay to reuse it here / use it multiple places?

src/datachain/catalog/catalog.py
shcheklein: what is TTL on SaaS?

let's simplify it still

get old jobs + get old run groups -> then just get all checkpoints we want to clean that belong to them (depending on group being outdated or not)

src/datachain/catalog/catalog.py
```diff
@@ -2092,3 +2094,95 @@ def index(
...
+
+        # Per-job output/partial tables (exclude input tables which are
+        # shared across the run group via run_group_id prefix)
+        for job_id in jobs_to_clean:
+            self._cleanup_udf_tables(f"udf_{job_id}_", suffix="_output")
+            self._cleanup_udf_tables(f"udf_{job_id}_", suffix="_output_partial")
+
+        # Shared input tables: only remove when entire run group is inactive
+        for run_group_id in run_groups_to_clean:
+            if not self.metastore.has_active_checkpoints_in_run_group(
```
shcheklein: just make run group active or not (but its timestamp)?

docs/commands/gc.md
ilongin: Removed

src/datachain/catalog/catalog.py
ilongin: It was some default TTL used on multiple places but now it seems that it's used only for checkpoints. Renamed to `CHECKPOINTS_TTL`

src/datachain/catalog/catalog.py
ilongin: We don't explicitly set TTL in SaaS .. it uses default one. I can add that one to be from env var in backend.

Regarding logic, for that I need to add new methods to metastore which I wanted to avoid but agree that it would make code simpler a little bit.

ilongin: Implemented

src/datachain/catalog/catalog.py
shcheklein: > if you think about saving this tables to checkpoint model

not, not at all - not saving (only may be for debugging purposes - sometimes can be handy to cross check things)

more like having a method on Checkpoint instance that can "generate" all the needed names back. 

Catalog should not be ideally aware about "Table patterns: udf_{job_id}_{hash}_{suffix}" things. Checkpoint model can do that.

Does it make sense?

src/datachain/catalog/catalog.py
shcheklein: removed? (if yes - resolve it)

src/datachain/catalog/catalog.py
```diff
@@ -2092,3 +2093,67 @@ def index(
...
+        )
+
+        # Step 2: Remove warehouse tables first (before metadata) so that if
+        # we crash midway, metadata still exists to retry on next GC run.
+        for job in inactive_jobs:
+            self._cleanup_udf_tables(f"udf_{job.id}_", suffix="_output")
+            self._cleanup_udf_tables(f"udf_{job.id}_", suffix="_output_partial")
+
+        # Shared input tables — only when entire run group is inactive
+        for group_id in run_group_ids:
```
shcheklein: do these tables belong to checkpoints?

if we are below deleting checkpoint objects - then we'll never come back to these tables?

shcheklein: or if we don't depend on checkpoint object presence - then we'll be returning again and again all inactive groups here?

shcheklein: I think this "smells" a bit - if this table is considered a part of the checkpoints workflow then we should not be deleting the checkpoint metadata until we delete it. So, checkpoint object essentially has different states ... complicated ... is it all really needed?

src/datachain/catalog/catalog.py
```diff
@@ -2092,3 +2093,67 @@ def index(
...
+    def _cleanup_udf_tables(self, prefix: str, suffix: str = "") -> None:
+        """Remove all UDF tables matching a prefix and optional suffix."""
+        tables = self.warehouse.db.list_tables(prefix=prefix)
+        if suffix:
+            tables = [t for t in tables if t.endswith(suffix)]
+        if tables:
+            logger.info("Removing %d UDF tables: %s", len(tables), tables)
+            self.warehouse.cleanup_tables(tables)
+
+    def cleanup_checkpoints(self, ttl_seconds: int | None = None) -> int:
```
shcheklein, line 2100: looking more into this ... it seems we need to split it into three jobs?

1. delete non-shared tables for inactive checkpoints 
2. delete shared tables for inactive run-groups for all checkpoint there
3. delete checkpoints metadata (btw, do we even need this? we can keep metadata - just set status DELETED - like we do for exports - it is better for audit). We do this for active checkpoints when we know there are no tables left in warehouse and all conditions are satisfied.

we should think how to make it in a way that catalog doesn't directly manipulate with things like `f"udf_{job.id}_", suffix="_output_partial"`

src/datachain/catalog/catalog.py
ilongin: The current order is safe — we always delete tables before metadata:                                                     
                                                                                                                     
1. Delete per-job output/partial tables for inactive jobs
2. Delete shared input tables only if entire run group is inactive
3. Delete checkpoint metadata last

If GC crashes between steps, metadata still exists to retry next run.

Shared input tables don't belong to any specific checkpoint — they belong to the run group as a whole (named with `run_group_id`, shared across all jobs in the group). So we don't need checkpoint metadata to track them, just the `run_group_id` from any remaining job.

The "partial run group" case also works: let's say run group G has jobs J1, J2, J3.
- On GC run 1, J1/J2 are inactive but J3 is still active — we clean J1/J2 output tables and metadata, but skip G's input tables. 
- On GC run 2, J3 expires — we find it via `get_jobs_with_expired_checkpoints()`, it carries run_group_id=G, now the whole group is inactive → input tables get cleaned.

So checkpoint metadata doesn't need intermediate states — the "tables first, metadata last" ordering is sufficient.

src/datachain/catalog/catalog.py
ilongin: On the 3 steps — that's what we already do, in exactly that order: (1) per-job output/partial tables, (2) shared input  tables for fully inactive run groups, (3) checkpoint metadata. It's one method but the steps are clearly separated.      
            
On soft-delete — I think that's a separate concern, we can add status=DELETED later if we need audit. For now hard-delete keeps checkpoint queries simple (no need to filter out deleted ones everywhere).

On moving table name patterns out of catalog — the cleaner intentionally uses prefix+suffix matching to find all tables for a job regardless of hash. That's a fundamentally different operation from table creation (which uses exact names via
helpers in dataset.py). The cleaner needs to know what to clean — moving prefix patterns to the model would just relocate the same knowledge without real benefit IMO.

src/datachain/catalog/catalog.py
ilongin: We already have a single source of truth for table naming — free functions in `dataset.py` (`udf_output_table_name()`,  `udf_input_table_name()`, etc.). Moving them to Checkpoint methods would be incomplete: output/partial table names could live there (Checkpoint has `job_id` + `hash`), but input table names need `run_group_id` which is a Job concept, not Checkpoint. So we'd end up with a split — some names on Checkpoint, some elsewhere.                                                    
            
Also, callers don't always have a Checkpoint object — e.g. `_run_from_scratch()` creates tables before any checkpoint exists. And the cleanup code uses prefix matching, not exact names, so Checkpoint methods wouldn't help there either.

I think the current free functions are the right level of abstraction for this.

src/datachain/catalog/catalog.py
shcheklein: >  So we don't need checkpoint metadata to track them, just the run_group_id from any remaining job.

do we need checkpoints to find those Jobs though? or do we traverse ALL jobs?

how do you define active / inactive job?

src/datachain/catalog/catalog.py
shcheklein: > On soft-delete — I think that's a separate concern, we can add status=DELETED later if we need audit. For now hard-delete keeps checkpoint queries simple (no need to filter out deleted ones everywhere).

it is by no means more complicated, and it is safer

> moving prefix patterns to the model would just relocate the same knowledge without real benefit IMO.

it would isolate it to an appropriate place - should be less error prone if it is done correctly ... but also overall this current implementation raises questions bc of this - something feels wrong. Many ad-hoc manipulations with internals in the high lever business logic.

src/datachain/catalog/catalog.py
ilongin: We don't traverse all jobs. `get_jobs_with_expired_checkpoints()` is a single SQL query that finds jobs where ALL checkpoints are older than TTL. It only looks at jobs that have checkpoints.
- "Active job" -> has at least one checkpoint newer than TTL threshold. 
- "Inactive job" -> all checkpoints older than TTL.

src/datachain/catalog/catalog.py
ilongin: > it is by no means more complicated, and it is safer

I've added soft-delete for checkpoints (`remove_checkpoint` now does `UPDATE SET deleted=True` instead of `DELETE`).

> it would isolate it to an appropriate place - should be less error prone if it is done correctly ... but also overall this current implementation raises questions bc of this - something feels wrong. Many ad-hoc manipulations with internals in the high lever business logic.

I see `cleanup_checkpoints()` as infrastructure code that lives in catalog for convenience (whole Catalog is legacy btw but that's separate issue), not as high-level business logic. The cleaner needs to know what to clean — it matches table prefixes/suffixes to identify output, partial, and input tables. This is part to the cleanup responsibility and I don't think moving it elsewhere would reduce that. ...  I just don't see any benefits but if you have strong opinion about it, I can try to move it...

src/datachain/catalog/catalog.py
shcheklein: so, does it require checkpoints to exist though? does it mean that the number of inactive jobs will keep growing also infinitely?

src/datachain/catalog/catalog.py
shcheklein: > I've added soft-delete for checkpoints (remove_checkpoint now does UPDATE SET deleted=True instead of DELETE).

let's add a status field (like we do for all other models). Also, let's probably show them in the admin panel, etc.

> I see cleanup_checkpoints() as infrastructure code that lives in catalog for convenience

I'm not sure I understand the argument. To me catalog defines high level functions (e.g. remove dataset). 

To make it clear - having `cleanup_checkpoints` in the catalog (as it is now) it probably fine, my concern is what is inside the cleanup_checkpoints. It manipulates with a lot of internal details from different places (e.g. table names constructed ad-hoc), it requires 2-3 custom API calls to operate. Logic is complicated (e.g. you need to think if removing metadata affects next attempts).

All of this makes me think hard how to split it or rewrite it in a way that it is 100% clear what is happening + reduce exposure of internal details inside high-order code.

> I just don't see any benefits but if you have strong opinion about it, I can try to move it...

Catalog is already like a trash bin (2000+? lines of random code and classes together). And we keep making it worse. Let's not grow tech debt.

src/datachain/catalog/catalog.py
```diff
@@ -2094,3 +2095,62 @@ def index(
...
+        3. For run groups where all member jobs are inactive: also remove
+           shared input tables
+        """
+        if ttl_seconds is None:
+            ttl_seconds = CHECKPOINTS_TTL
+
+        ttl_threshold = datetime.now(timezone.utc) - timedelta(seconds=ttl_seconds)
+
+        inactive_jobs = list(
+            self.metastore.get_jobs_with_expired_checkpoints(ttl_threshold)
```
shcheklein: Q: 

can it be all something like:

checkpoints = self.metastore.get_checkpoints() # returns all active checkpoints (status != DELETED or something) + returns in some way information about run group and job (it can be done on the backend with a single query most likely)

```
for ch in checkpoints:
    ch.cleanup()
```

cleanup() in turn incapsulates logic to pick the right tables to cleanup, delete them, set proper status for the checkpoint object, etc.

src/datachain/catalog/catalog.py
ilongin: I'm not sure I understand the question, can you give more context? Jobs are separate from checkpoints. Job can have checkpoints,  some of them can be "active", some "expired". This query finds jobs that have all expired checkpoints, where some of them are not deleted and then it soft deletes them. That's it.

src/datachain/catalog/catalog.py
ilongin: **Status field**: Done — replaced `deleted` boolean with a `status` field (`ACTIVE`/`DELETED`), added to admin panel already.            
                                                                                                                               
**On Catalog tech debt**: Agree that Catalog is overloaded and we should stop growing it. But refactoring it (extracting domain-specific services that have access to both metastore and warehouse) is a separate effort, for which we had a special issue opened long time ago (I cannot find it any more, probably dropped due repo migration). Currently it's the only class that bridges both (metastore + warehouse) and is used everywhere, including Studio so it's hard to just stop using it - first we need to refactor / restructure IMO.
But I don't think we should block this PR on that restructuring. The current `cleanup_checkpoints` is 3 clear steps  and the complexity is inherent to the domain (job tables → run group tables → checkpoint metadata).

If you want we can brainstorm / discuss about this in a call.

src/datachain/catalog/catalog.py
ilongin: This is I think continuation / duplication of above comment / discussion. 
Point is:
1. We need to pull domain specific logic out of Catalog and eventually deprecate Catalog.
2. We need to decide how to do it - should it be on a level of dataclasses (as you suggested), or on a level of just plain modules (I currently lean on that approach more) .. this is up to discussion / brainstorm since it's important decision.

src/datachain/catalog/catalog.py
shcheklein: at the previous implementation we were not soft deleting, but physically deleting them. Then it means there was no chance to pick the job on the next run as inactive (e.g. to delete shared input tables)? Is it correct? This to make sure I understand this right and make sure that soft delete (or status field) is not nice-to-have, but actually needed. 

Now, is there a status for checkpoint objects now that we consider fully deleted (means input tables also gone). This is needed to stop picking up some jobs as inactive AFAIU (otherwise list will keep growing, no?)

happy to discuss this is Slack if I missing something

src/datachain/catalog/catalog.py
shcheklein: to make it clear (as I mentioned before) - this is not about catalog restructuring, this discussion / suggestions apply only to this new functionality, which is a bit too complicated and sometimes exposes a bit too low level details

src/datachain/catalog/catalog.py
shcheklein: > We need to pull domain specific logic out of Catalog and eventually deprecate Catalog.

I'm not sure why this discussion keeps leading to Catalog refactoring tbh. 

This is purely about this function to me. 

I see a few API calls, a bit of exposed internals, questions re reentry or growing list of ids - this makes me think hard how to make it properly (with proper isolation, resuability of code - e.g. table names), statuses (to make sure it is reentrant), less APIs to simplify.

>We need to decide how to do it - should it be on a level of dataclasses

I'm fine I think on both - module or dataclass - this is less important probably.

docs/commands/gc.md
```diff
@@ -0,0 +1,46 @@
...
+- **Temporary tables** created during query execution that were not properly cleaned up (e.g., due to crashes or interrupted operations).
+- **Failed dataset versions** that were left in an incomplete or failed state.
+- **Outdated checkpoints** and their associated UDF tables that have exceeded the time-to-live (TTL) threshold. See [Checkpoints](../guide/checkpoints.md) for more details.
+
+## Options
+
+* `-h`, `--help` - Show the help message and exit.
+* `-v`, `--verbose` - Be verbose.
+* `-q`, `--quiet` - Be quiet.
+* `--checkpoint-ttl` - Time-to-live for checkpoints in seconds. Checkpoints older than this value are considered outdated and eligible for cleanup. Defaults to 4 hours (14400 seconds).
```
dreadatour, line 24: Just wonder if we need dry-run option.

src/datachain/catalog/catalog.py
```diff
@@ -2094,3 +2095,62 @@ def index(
...
         ):
             pass
+
+    def _cleanup_udf_tables(self, prefix: str, suffix: str) -> None:
+        """Remove all UDF tables matching a prefix and optional suffix."""
+        tables = self.warehouse.db.list_tables(prefix=prefix)
+        tables = [t for t in tables if t.endswith(suffix)]
+        if tables:
+            logger.info("Removing %d UDF tables: %s", len(tables), tables)
+            self.warehouse.cleanup_tables(tables)
```
dreadatour: Do we want to use TTL here too or any other guard? It may be the case when other query is running (e.g. in separate terminal) and we will break it by removing its temporary tables.

src/datachain/catalog/catalog.py
```diff
@@ -2094,3 +2095,62 @@ def index(
...
+
+    def cleanup_checkpoints(self, ttl_seconds: int | None = None) -> int:
+        """Clean up outdated checkpoints and their associated UDF tables.
+
+        Algorithm:
+        1. Find inactive jobs (all checkpoints older than TTL) — single query
```
dreadatour: Is it safe to reply on TTL only? It might be a long-running active job (running longer than TTL), e.g. in another process, and we might break this job by removing it's checkpoints.

src/datachain/catalog/catalog.py
```diff
@@ -2094,3 +2095,62 @@ def index(
...
+            self.metastore.get_jobs_with_expired_checkpoints(ttl_threshold)
+        )
+        if not inactive_jobs:
+            return 0
+
+        inactive_job_ids = [job.id for job in inactive_jobs]
+        run_group_ids = {job.run_group_id for job in inactive_jobs if job.run_group_id}
+
+        logger.info(
+            "Cleaning %d inactive jobs across %d run groups",
```
dreadatour: Tbh it is not clear what `run groups` are.

src/datachain/catalog/catalog.py
```diff
@@ -2094,3 +2095,62 @@ def index(
             only_index=True,
         ):
             pass
+
+    def _cleanup_udf_tables(self, prefix: str, suffix: str) -> None:
+        """Remove all UDF tables matching a prefix and optional suffix."""
+        tables = self.warehouse.db.list_tables(prefix=prefix)
+        tables = [t for t in tables if t.endswith(suffix)]
```
dreadatour: Can we update `list_tables` so it will also take `suffix` argument instead of bringing this logic here? Or may be even `mask` or something like that.

This call below: `self._cleanup_udf_tables(f"udf_{job.id}_", "_output")` looks a bit odd IMO.

src/datachain/cli/commands/misc.py
```diff
@@ -10,21 +10,28 @@ def clear_cache(catalog: "Catalog"):
...
 
-def garbage_collect(catalog: "Catalog"):
+def garbage_collect(catalog: "Catalog", checkpoint_ttl: int | None = None):
+    print("Collecting temporary tables...")
     temp_tables = catalog.get_temp_table_names()
-    num_versions_removed = catalog.cleanup_failed_dataset_versions()
-
-    total_cleaned = len(temp_tables) + num_versions_removed
+    if temp_tables:
+        catalog.cleanup_tables(temp_tables)
```
dreadatour, line 17: Do we want to use TTL here too or any other guard? It may be the case when other query is running (e.g. in separate terminal) and we will break it by removing its temporary tables while it is running.

src/datachain/data_storage/metastore.py
```diff
@@ -545,8 +556,14 @@ def get_last_job_by_name(
     #
 
     @abstractmethod
-    def list_checkpoints(self, job_id: str, conn=None) -> Iterator[Checkpoint]:
-        """Returns all checkpoints related to some job"""
+    def list_checkpoints(
+        self,
+        job_id: "str | list[str] | None" = None,
```
dreadatour: Let's please do not mix `str` and `list[str]` types in one argument? E.g. we can easily use `list[str]` here and pass `[job_id]` for a single job.

src/datachain/data_storage/metastore.py
```diff
@@ -577,6 +594,12 @@ def get_or_create_checkpoint(
     def remove_checkpoint(self, checkpoint_id: str, conn: Any | None = None) -> None:
         """Removes a checkpoint by ID"""
 
+    @abstractmethod
+    def get_jobs_with_expired_checkpoints(
+        self, ttl_threshold: datetime, conn=None
```
dreadatour: In `list_checkpoints` method above we are using `created_before` param, here it is `ttl_threshold`. Should these names be consistent? What do you think?

src/datachain/data_storage/metastore.py
```diff
@@ -2326,8 +2379,12 @@ def _dataset_version_jobs_delete(self) -> "Delete":
...
     def _checkpoints_select(self, *columns) -> "Select":
         if not columns:
-            return self._checkpoints.select()
-        return select(*columns)
+            return self._checkpoints.select().where(
+                self._checkpoints.c.status != CheckpointStatus.DELETED,
+            )
+        return select(*columns).where(
+            self._checkpoints.c.status != CheckpointStatus.DELETED,
+        )
```
dreadatour: This can me simplified imo.

src/datachain/data_storage/metastore.py
```diff
@@ -2326,8 +2379,12 @@ def _dataset_version_jobs_delete(self) -> "Delete":
...
 
     def _checkpoints_delete(self) -> "Delete":
         return self._checkpoints.delete()
```
dreadatour: Do we still need this method?

src/datachain/data_storage/metastore.py
dreadatour: Also let's use consistent type hints, e.g. it is `job_id: str | list[str] | None = None` below (without quotes).

src/datachain/data_storage/metastore.py
```diff
@@ -2371,9 +2429,23 @@ def get_or_create_checkpoint(
...
             return checkpoint
 
-    def list_checkpoints(self, job_id: str, conn=None) -> Iterator[Checkpoint]:
-        """List checkpoints by job id."""
-        query = self._checkpoints_query().where(self._checkpoints.c.job_id == job_id)
+    def list_checkpoints(
+        self,
+        job_id: str | list[str] | None = None,
+        created_after: datetime | None = None,
+        created_before: datetime | None = None,
```
dreadatour: Do we use these params (`created_after`/`created_before`) somewhere in the code? 🤔

src/datachain/data_storage/metastore.py
```diff
@@ -2673,6 +2745,34 @@ def get_dataset_version_for_job_ancestry(
...
+
+    def get_jobs_with_expired_checkpoints(
+        self, ttl_threshold: datetime, conn=None
+    ) -> Iterator["Job"]:
+        """Return jobs where ALL checkpoints are older than ttl_threshold."""
+        ch = self._checkpoints
+        jobs = self._jobs
+
+        active = ch.c.status != CheckpointStatus.DELETED
+        job_id_cast = cast(ch.c.job_id, jobs.c.id.type)
```
dreadatour, line 2723: In `has_active_checkpoints_in_run_group` above we don't use helpers like this. It would be great if code will be consistent and we use same patterns everywhere.

src/datachain/query/dataset.py
```diff
@@ -101,6 +101,18 @@
...
+def udf_input_table_name(run_group_id: str, _hash: str) -> str:
+    return f"udf_{run_group_id}_{_hash}_input"
+
+
+def udf_output_table_name(job_id: str, _hash: str) -> str:
+    return f"udf_{job_id}_{_hash}_output"
+
+
+def udf_partial_output_table_name(job_id: str, _hash: str) -> str:
+    return f"udf_{job_id}_{_hash}_output_partial"
```
dreadatour: Not sure, but would be great to have docstrings here with some information about these tables usage and mission.

src/datachain/catalog/catalog.py
```diff
@@ -2094,3 +2095,62 @@ def index(
...
+        # Shared input tables — only when entire run group is inactive
+        for group_id in run_group_ids:
+            if not self.metastore.has_active_checkpoints_in_run_group(
+                group_id, ttl_threshold
+            ):
+                self._cleanup_udf_tables(f"udf_{group_id}_", "_input")
+
+        checkpoints = list(self.metastore.list_checkpoints(job_id=inactive_job_ids))
+        for ch in checkpoints:
+            self.metastore.remove_checkpoint(ch.id)
```
dreadatour: What do you think about adding `remove_checkpoints` method to metastore, so we can pass `job_id` (or multiple `job_ids`) and mark all checkpoints at once as deleted? This will be one single atomic call instead of multiple requests (would be useful for SaaS).

src/datachain/catalog/catalog.py
dreadatour: Will it be safer to mark checkpoints as `deleted` first and then to delete actual data? I am asking because code may breaks at any moment and we may ends up with inconsistent state, we can not rely on a single complex function with 3 steps.

We have had this issues before when no tables in warehouse exists, but we do have records in metastore and it fail queries completely without an option to bypass it.

In this case we may want to add one more check to garbage collector: to list all tables in warehouse and to clean up orphan tables. IMO it looks like a good idea in any case — to implement this check too.

dreadatour (commented):
Main question from me is: what if we have job running in parallel terminals/processes. May this garbage collector implementation make them fail?

src/datachain/catalog/catalog.py
ilongin: Hard-delete actually also worked correctly here. The key is that `has_active_checkpoints_in_run_group` checks `status != DELETED AND created_at >=  ttl_threshold` — so expired-but-not-yet-deleted checkpoints are treated as "not active." Since we always delete tables before metadata, the checkpoints are still physically present when we evaluate the run group, they just don't pass the TTL check.
So soft-delete is nice-to-have (auditability, safety net), not strictly required for correctness.

On the growing list — once all checkpoints for a job are soft-deleted, `get_jobs_with_expired_checkpoints` won't return that job anymore (it filters on `status != DELETED`). So the list doesn't grow.

Does this explain or maybe I'm missing something?

src/datachain/catalog/catalog.py
ilongin: Make sense, added

src/datachain/catalog/catalog.py
ilongin: Ok, what I did:

  1. Moved table name generation into `Checkpoint` as static methods (`output_table_name`, `partial_output_table_name`, `input_table_name`,
  `input_table_pattern`). Catalog no longer knows about `udf_{id}_{hash}_{suffix}` patterns — it just calls `ch.table_name` for output tables and
  `Checkpoint.input_table_pattern(group_id)` for input tables.
  2. Refactored `remove_checkpoint` → `remove_checkpoints` accepting a list of IDs — single bulk update instead of a loop.

Not sure if there's anything else to do

src/datachain/catalog/catalog.py
ilongin: Done, added `pattern` instead of `prefix` in `list_tables`

src/datachain/query/dataset.py
ilongin: Added, btw now those methods are in Checkpoint dataclass

docs/commands/gc.md
ilongin: I don't think this is needed ATM....gc only cleans expired/temp data .. we can always add it in future if we see the real need.

src/datachain/catalog/catalog.py
ilongin: Yes, this is a valid concern. I added additional filter for active jobs when fetching the ones that have expired checkpoints not to remove checkpoints for them. I think this simple addition fixes this.

src/datachain/catalog/catalog.py
ilongin: Added small comment above `run_group_ids = ...`

src/datachain/cli/commands/misc.py
ilongin: Yes, it make sense, but since it's not related to the domain of this PR I would create separate issue for it

src/datachain/catalog/catalog.py
shcheklein: Do we (at did we needed) checkpoints object to exits to eventually remove shared tables? (e.g. to even find inactive jobs next time it runs)? Meanwhile we run `self.metastore.remove_checkpoints([ch.id for ch in checkpoints])` even if some run groups are still active, right? It means next time it runs it won't find inactive jobs, won't delete inactive input tables.

src/datachain/catalog/catalog.py
shcheklein: reduce amount of API calls, still need to review the logic ... and because of this multilayered setup I'm not sure if is correct / fully scalable. I'll comment separately to continue.

src/datachain/data_storage/metastore.py
ilongin: Updated to be list of strings always

src/datachain/data_storage/metastore.py
ilongin: `list_checkpoints` uses `created_before`/`created_after` as general-purpose query filters, while `get_jobs_with_expired_checkpoints` and `has_active_checkpoints_in_run_group` use `ttl_threshold` — which is domain-specific to GC. I think the naming is correct as-is since they serve different purposes, even though the underlying type is the same.

src/datachain/data_storage/metastore.py
ilongin: No need any more since we are doing soft delete, removing it

src/datachain/data_storage/metastore.py
ilongin: No, we don't anymore after refactoring, removing it as well

src/datachain/data_storage/metastore.py
ilongin: Extracted in job_id_cast var above as well, I think that's cleaner

src/datachain/catalog/catalog.py
shcheklein: so, if I'm not mistaken there is still a bug here

1. We take `get_jobs_with_expired_checkpoints`. Let's say results is [id1, id2, id3]
2. We remove tables all tables for id1 and id3, but we keep id2 since run groups is still active
3. We set DELETE for _ALL_ [id1, id2, id3] (mind some input tables are not deleted yet)
4. Any next time we run it again - we'll never get `id2` via `get_jobs_with_expired_checkpoints` AFAIU.
5. Shared input table is orphaned now, no matter when group becomes inactive.

Is it the case? Do we have a test for this?

Does it mean we should not be deleting checkpoints (metadata) until even input tables are gone? But then there will be other problems - e.g. checkpoint exists but some tables are removed (sounds dangerous - btw are we filtering DELETED checkpoints in jobs now? we must add a filter).

So, most likely you we must set status to DELETE, but since we have potentially other things to remove, we need multiple statuses. Check export - it has REMOVING, REMOVED or something like this.

Also, still this whole logic should be simplified.

* one API to get all checkpoints metadata (including active / non-active / and REMOVING status, including run-group information)
* iterate and set REMOVING status via another API call - update_checkpoint_status
* iterate and remove tables
* if all tables removed - set REMOVED status

so, we need only checkpoint APIs - get_checkpoints and update_checkpoint_status (which makes sense to me)

we also need to check that we have proper filters throughout the code of checkpoints (to avoid fetching REMOVED, REMOVING statuses or similar)

src/datachain/catalog/catalog.py
ilongin: Yes, I think you are right. I will return to this order again - metastore, then warehouse...that is much safer in general.

ilongin: It seems like that removal of orphan tables is not very easy .. a lot of heavy DB queries and code is complex. 
Alternative is to add another status - EXPIRED and do atomic update of all checkpoints which are created before TTL time to EXPIRED. Then, after that, gc would continue as is now, just removing expired to removed at it's own pace. . if fails in the middle it will continue later on the next run and no need for additional orphan tables cleanup.
So statuses of checkpoint would be:
1. ACTIVE
2. EXPIRED
3. REMOVED

When finding checkpoints in actual business logic we would just search for ACTIVE ones.

src/datachain/catalog/catalog.py
ilongin: I don't think there's a bug in the described scenario. Input table cleanup is driven by `run_group_id`, not by specific checkpoint  metadata. Here's why it works:                                                                                                             
                                                                                                                                           
1. `get_jobs_with_expired_checkpoints` → [id1, id2, id3]
2. Output tables removed for all three                                                                                                     
3. id2's run group G still has active job id4 → skip G's input tables
4. Soft-delete checkpoints for [id1, id2, id3]
5. Later, id4 expires → GC finds id4 → `id4.run_group_id = G` → `has_active_checkpoints_in_run_group(G, ttl)` returns false (id2's checkpoints are DELETED, id4's are expired) → G's input tables get cleaned

So the input tables aren't orphaned — they're cleaned when the **last** job in the group expires. Any expired job in the group carries the same `run_group_id` and can trigger the cleanup.

But the broader concern about safety is valid. I'm adding an `EXPIRED` status to make the GC flow more robust (please look at [com](https://github.com/datachain-ai/datachain/pull/1590#discussion_r2861551719):

- `ACTIVE` → usable by running jobs (`find_checkpoint` only returns ACTIVE)
- `EXPIRED` → marked atomically at GC start, ignored by `find_checkpoint`, tables pending cleanup
- `DELETED` → fully cleaned up

GC flow becomes:
1. Find expired checkpoints → mark them `EXPIRED` (atomic, prevents any running job from using them)
2. Remove output/input tables
3. Mark `DELETED`

This way even if GC crashes mid-cleanup, `find_checkpoint` won't return expired checkpoints, and next GC run can pick up where it left off.

shcheklein: Sounds good.

Re this `Later, id4 expires → GC finds id4 → id4.run_group_id = G` - sounds even more fragile and unobvious (thus I'm asking to simplify it all - so that we don't depend on such workflows). Also, I have no idea if it by itself has some hidden issues. E.g. is there always guarantee that id4 exits, that it won't disappear also? (I don't know the answer, my point is you need to know quite a lot here to make sure it works)

ilongin: You're right — if the last remaining job in a run group gets manually deleted (with its checkpoints), no one carries that `run_group_id` anymore and the input tables become orphaned. This is very rare edge case but it could happen.                                                                                                
 
This is a constraint of the system — input tables are shared across a run group and can only be removed when there are no jobs with active checkpoints in that group.

I don't think there is a solution in terms of refactoring or simplifying logic here  — the only fix is to add separate method that cleans orphaned input UDF tables - that's what I now added.

src/datachain/data_storage/metastore.py
ilongin: Simplified — extracted the common .where() clause.

src/datachain/catalog/catalog.py
shcheklein: resolved? cc @ilongin @dreadatour ?

src/datachain/catalog/catalog.py
shcheklein: resolved cc @dreadatour @ilongin ?

src/datachain/catalog/catalog.py
```diff
@@ -2094,3 +2096,98 @@ def index(
...
+        5. Mark checkpoints as DELETED
+        """
+        if ttl_seconds is None:
+            ttl_seconds = CHECKPOINTS_TTL
+
+        ttl_threshold = datetime.now(timezone.utc) - timedelta(seconds=ttl_seconds)
+
+        # Atomically mark expired checkpoints — after this, find_checkpoint
+        # won't return them, so running jobs are safe
+        self.metastore.expire_checkpoints(ttl_threshold)
```
shcheklein: can it happen that job is starting and run find checkpoint while we are expiring it?

shcheklein: let's use the same pattern we use everywhere - update_status / update_state - we don't need one call per each action

src/datachain/catalog/catalog.py
```diff
@@ -2094,3 +2096,98 @@ def index(
...
+
+        expired_jobs = list(self.metastore.get_expired_jobs())
```
shcheklein: let's just use and expand regular list_jobs - again, way too many extra calls that are very specific for this. Seriously, 20-30 lines of code and single function required introducing 4-5 APIs? this is wrong to my mind in most cases ...

src/datachain/catalog/catalog.py
```diff
@@ -2094,3 +2096,98 @@ def index(
...
+        if not expired_jobs:
+            return 0
+
+        expired_job_ids = [job.id for job in expired_jobs]
+        # Run group = chain of rerun jobs (parent → child) sharing input tables
+        run_group_ids = {job.run_group_id for job in expired_jobs if job.run_group_id}
+
+        checkpoints = list(self.metastore.list_checkpoints(job_ids=expired_job_ids))
```
shcheklein: let's make the flow more conventional and simpler:

- list_checkpoint(ttl) - should be returning also information about run groups (active / non active status)
- update_status (expired)
- remove tables (doesn't require any extra calls)
- update_status (deleted) - only those which we already removed ALL tables

this should fix a few possible edge cases in this implementation + removes a lot of custom / specific API calls

src/datachain/catalog/catalog.py
ilongin: I think it should be, resolving

src/datachain/catalog/catalog.py
ilongin: `expire_checkpoints` is already a single **atomic** SQL UPDATE that marks all expired checkpoints at once — it's not one call per checkpoint. 
The method  has domain-specific logic (TTL threshold filtering, finished-job check, GROUP BY/HAVING to ensure all checkpoints in a job are expired) that  wouldn't fit a generic `update_status(id, new_status)` pattern. Same for `remove_checkpoints` — single UPDATE for a batch of IDs.

src/datachain/catalog/catalog.py
ilongin: Consolidated `remove_checkpoints` into a generic `update_checkpoint(checkpoint_ids, status=...)`

Regarding the API count — here's what each method does and why it exists:

1. **`expire_checkpoints(ttl_threshold)`** — atomic batch `UPDATE` with complex filtering (TTL + finished jobs + GROUP BY/HAVING). Can't be genericized into
a simple status setter.
2. **`get_expired_jobs()`** — joins checkpoints with jobs to find jobs that have EXPIRED checkpoints. This is a checkpoint-specific query that doesn't fit
into generic `list_jobs` without leaking checkpoint into the jobs API.
3. **`get_active_run_group_ids()`** — joins checkpoints with jobs to find run groups with ACTIVE checkpoints. Used to protect shared input tables during
cleanup and orphan detection.
4. **`update_checkpoint(checkpoint_ids, status=...)`** — generic field updater (replaces `remove_checkpoints`). Follows the `update_job` pattern.
5. **`has_active_checkpoints_in_run_group(group_id, ttl_threshold)`** — targeted check needed during cleanup to decide if shared input tables can be removed.

Methods 1, 2, 3, 5 encapsulate domain-specific join queries — pushing them into generic list endpoints would couple unrelated models.

src/datachain/catalog/catalog.py
ilongin: The proposed `list_checkpoints(ttl)` would need to replicate the same complex logic that currently lives in dedicated methods:                               
                                                                                                                                                             
1. Find finished jobs where **all** active checkpoints are older than TTL (`GROUP BY job_id / HAVING max(created_at) < ttl`) — this is what                  
`expire_checkpoints` does today
2. For each such checkpoint's run group, determine if the **entire** run group is inactive (i.e. no other jobs in the group have ACTIVE checkpoints) — this  
is what `has_active_checkpoints_in_run_group` does today

So the single `list_checkpoints(ttl)` call would actually be two complex queries behind the scenes, returning a richer object (checkpoint + run_group_active
flag). It doesn't reduce complexity — it moves it into one method.

More importantly, this would actually **introduce** edge cases rather than fix them. Currently `expire_checkpoints` does find + mark EXPIRED in a single atomic UPDATE — there's no window between "deciding a checkpoint is expired" and "making it invisible to running jobs via `find_checkpoint`". 
With the proposed list-then-update approach, there's a gap between step 1 (list expired) and step 2 (mark EXPIRED) where a running job could still pick up a
checkpoint that GC is about to expire and remove tables for. The whole reason `expire_checkpoints` exists as an atomic operation is to prevent this race condition.

Current approach, which is better IMO: each method does one clear thing, `expire_checkpoints` atomically marks + hides in a single query, no race window.

src/datachain/catalog/catalog.py
ilongin: Good catch. Even though `expire_checkpoints()` is atomic update, there was still a race condition. A rerun job (Job B) calls `find_checkpoint(job_A_id, hash)` to look up checkpoints from its finished parent (Job A). So there's a window where Job A is finished, its checkpoints are old, but Job B is actively using them.                                                                                
                                                                                                                                                             
Fixed by adding a run group guard to `expire_checkpoints` — it now excludes jobs whose run group has any non-finished (running) job. This way, if Job B is running in the same run group as Job A, Job A's checkpoints won't be expired until Job B finishes. Once all jobs in the group are finished, each job's  checkpoints expire independently based on TTL.
                                                                                                                                                             
This is safe because the check is part of the same atomic UPDATE query.

src/datachain/catalog/catalog.py
shcheklein: > The proposed list_checkpoints(ttl) would need to replicate the same complex logic that currently lives in dedicated methods:

that's the point. One API call. An existing API call. It might be even able to run it as a single query.

>  It doesn't reduce complexity — it moves it into one method.

even by reducing the amount of custom APIs is already a good improvement. Also, importantly this functionality will become easier to read and understand.

> With the proposed list-then-update approach, there's a gap between step 1 (list expired) and step 2 (mark EXPIRED) where a running job could still pick up a

what is the issue with that? What can happen to expired checkpoints? New job in a run group? But your implementation most likely also suffers from it? We share input tables and delete them after we list expired groups - so same gap, but more even more nuanced, even harder to isolate and fix if we want to

we can also do a check on the backend if needed using the same query before expiring them.

> Current approach, which is better IMO: each method does one clear thing, expire_checkpoints atomically marks + hides in a single query, no race window.

it is complicated in terms of custom logic and APIs 

it becomes more like RPC programmings - creating an API for each helper. While with APIs we still should be trying to keep it reasonably small (arguably), reuse things, move processing to the backend where is makes sense (to void too any round trips). etc, etc

src/datachain/checkpoint.py
```diff
@@ -24,6 +31,46 @@ class Checkpoint:
...
+
+    @staticmethod
+    def input_table_pattern(group_id: str) -> str:
+        """LIKE pattern for finding all input tables in a run group."""
+        return f"udf_{group_id}_%_input"
+
+    @staticmethod
+    def all_input_tables_pattern() -> str:
+        """LIKE pattern for finding all UDF input tables."""
+        return "udf_%_input"
```
dreadatour: Just a tiny comment about SQL `LIKE` syntax: underscore (`_`) is also a part of a pattern, it matches any single character. If we want to be strict, we need to escape it.

See here, for example: https://www.sqlite.org/lang_expr.html#the_like_glob_regexp_match_and_extract_operators

src/datachain/catalog/catalog.py
dreadatour: I would say yes, but to be honest I don't really like we are now using SQL-specific `LIKE` pattern outside of the warehouse, it might be a problem in future, when we will implement non standard SQL warehouses.

I am talking about `Checkpoint` class [here](https://github.com/datachain-ai/datachain/pull/1590/changes#diff-5624d8894d07edc198217327d6044635d60528b5711ad06a912c560c6fbc4828R51-R59).

src/datachain/catalog/catalog.py
dreadatour: I agree, we can simplify this a lot and in the end it might be a two atomic API calls in `Catalog.cleanup_checkpoints` method:
1. `metastore.cleanup_checkpoints` — do all the job and returning list of tables (list of patterns) needs to be deleted.
2. `warehouse.remove_tables` — remove tables from warehouse by names or patterns.

This also related to the discussion [above](https://github.com/datachain-ai/datachain/pull/1590#discussion_r2872507545) about too many new specific metastore methods.

In addition I would like to have separate cronjob to cleanup orphan tables in warehouse — in case anything goes wrong.

src/datachain/data_storage/metastore.py
```diff
@@ -574,8 +589,30 @@ def get_or_create_checkpoint(
...
+        active checkpoints are older than ttl_threshold. Expired checkpoints
+        are ignored by find_checkpoint, preventing running jobs from using
+        them while GC proceeds to clean up their tables."""
+
+    @abstractmethod
+    def get_expired_jobs(self, conn: Any | None = None) -> Iterator["Job"]:
+        """Return jobs that have checkpoints in EXPIRED status."""
+
+    @abstractmethod
+    def update_checkpoint(
```
dreadatour: Should we rename this to `update_checkpoints`, since we are passing list of checkpoint ids as an input argument here?

src/datachain/catalog/catalog.py
ilongin: Refactored based on your feedback. The cleanup flow is now down to **2 metastore API calls**:                                                   
                                                          
1. **`expire_checkpoints(ttl_threshold)`** → returns `(checkpoints, inactive_group_ids)` — atomically marks ACTIVE→EXPIRED, then collects all expired checkpoints and determines which run groups have no remaining active checkpoints.
2. **`update_checkpoint(ids, status=DELETED)`** — marks checkpoints as DELETED after warehouse tables are removed.

Removed from the public API:
- `get_expired_jobs` — no longer needed, job info is resolved inside `expire_checkpoints`
- `has_active_checkpoints_in_run_group` — now private (`_has_active_checkpoints_in_run_group`), only used internally by `expire_checkpoints`
- Extra `list_checkpoints` call from `catalog.cleanup_checkpoints`

The catalog method is now simply: expire → remove output tables → remove input tables for inactive groups → mark deleted.

Crash safety is preserved: `expire_checkpoints` returns **all** expired checkpoints (not just newly expired), so leftovers from a prior crashed
run are picked up on the next attempt.

Honestly I liked more previous version where each metastore method was doing only one thing but I guess this is ok as well.

src/datachain/checkpoint.py
ilongin: Good catch, technically `_` is a wildcard in LIKE but in practice it's harmless here since our table names always use `_` as separators in those positions (`udf_{id}_{hash}_input`), so the wildcard always matches a literal underscore. If you have strong opinion about this, I can change it, but I personally don't think it's needed...also change is not simple as CH has different way of dealing with escape characters which messes up the codebase

src/datachain/catalog/catalog.py
```diff
@@ -2094,3 +2096,90 @@ def index(
...
+
+        These can be left behind when jobs were manually deleted.
+        """
+        all_input_tables = self.warehouse.db.list_tables(
+            pattern=Checkpoint.all_input_tables_pattern()
+        )
+        if not all_input_tables:
+            return 0
+
+        active_group_ids = self.metastore.get_active_run_group_ids()
```
shcheklein: so, we still need this call ?

how can these table be left behind?

ilongin: You are right, not needed any more, removed

src/datachain/checkpoint.py
shcheklein: still smells a bit ...

PR #1593, opened 2026-02-17
UDF checkpoints for aggregator

ilongin:
Implementing UDF checkpoints for aggregator.

### DIFF
--- a/docs/guide/checkpoints.md
+++ b/docs/guide/checkpoints.md
@@ -201,5 +201,4 @@ dc.read_storage("gs://datachain-demo/dogs-and-cats/", anon=True).map(
 
 - **Script path matters:** DataChain links runs by the script's absolute path. Moving the script breaks checkpoint linking.
 - **Threading/multiprocessing:** Checkpoints are automatically disabled when Python threading or multiprocessing is detected. DataChain's built-in `parallel` setting for UDFs is not affected.
-- **Aggregations:** Unlike `.map()` and `.gen()`, aggregations (`.agg()`) do not save progress incrementally. If an aggregation fails partway through, the entire aggregation restarts from the beginning on the next run.
 - **Unhashable callables:** Built-in functions (`len`, `str`), C extensions, and `Mock` objects produce a different hash on each run, so checkpoints using these as UDFs will always recompute. Use regular `def` functions or lambdas instead.


src/datachain/data_storage/sqlite.py
```diff
@@ -417,8 +417,7 @@ def create_table(
     def drop_table(self, table: "Table", if_exists: bool = False) -> None:
         self.execute(DropTable(table, if_exists=if_exists))
         # Remove the table from metadata to avoid stale references
-        if table.name in self.metadata.tables:
-            self.metadata.remove(table)
+        self.metadata.remove(table)
```
shcheklein: Is it related change?

src/datachain/data_storage/sqlite.py
```diff
@@ -432,8 +431,9 @@ def rename_table(self, old_name: str, new_name: str):
                 f"Failed to rename table from '{old_name}' to '{new_name}': {e}"
             ) from e
         # Remove old table from metadata to avoid stale references
-        if old_name in self.metadata.tables:
-            self.metadata.remove(self.metadata.tables[old_name])
+        old_table = self.metadata.tables.get(old_name)
```
shcheklein: Same, is this related change?

src/datachain/lib/udf.py
```diff
@@ -613,9 +613,22 @@ def run(
...
+        # Check if partition_id is available (when partition_by is used)
+        partition_id_idx = None
+        if PARTITION_COLUMN_ID in udf_fields:
+            partition_id_idx = list(udf_fields).index(PARTITION_COLUMN_ID)
+
         for batch in udf_inputs:
+            # Get partition_id from first row if available (all rows in batch share it)
+            # This is used to track which partition produced each output for checkpoints
+            input_id = None
+            if partition_id_idx is not None and batch:
```
shcheklein: can batch be empty or None? what is the case for this?

shcheklein:
More tests are needed (check codecov - line by line), check CI, check all comments resolve GH automated if they do not make sense (put a comment).

src/datachain/data_storage/sqlite.py
ilongin: No, cannot remember why I added this, probably when resolving some conflict or something. Similar as above. Returning to original

src/datachain/lib/udf.py
ilongin: I think it cannot be but wanted to add this as a safeguard. Maybe I can remove it to fail fast if ever happens...removing it

src/datachain/query/dataset.py
```diff
@@ -1272,16 +1344,21 @@ def _continue_udf(
...
+            incomplete_input_ids = self.find_incomplete_inputs(parent_partial_table)
+            if incomplete_input_ids:
+                logger.debug(
+                    "UDF(%s) [job=%s run_group=%s]: Found %d incomplete inputs "
+                    "to re-process",
+                    self._udf_name,
+                    self._job_id_short(job),
+                    self._run_group_id_short(job),
+                    len(incomplete_input_ids),
+                )
```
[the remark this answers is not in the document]
ilongin: Aggregators are used with reducing functions (sum, count, mean, etc.) that produce a single output row from N input rows. Since there's  only one output row per partition, it's either fully written or not — there's no partial state to detect. The incomplete input detection is specifically designed for generators (1:N) where a single input can yield many output rows and a crash mid-way leaves a partially written  set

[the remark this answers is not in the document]
shcheklein: `Input-output relationship: N:M` per docs.

src/datachain/checkpoint.py
```diff
@@ -48,6 +48,11 @@ def input_table_name(group_id: str, _hash: str) -> str:
         """Shared UDF input table. Scoped to run group, reused across jobs."""
         return f"udf_{group_id}_{_hash}_input"
 
+    @staticmethod
+    def partition_table_name(job_id: str, _hash: str) -> str:
```
shcheklein, line 52: is it used only for checkpoints?

src/datachain/query/dataset.py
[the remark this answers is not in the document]
ilongin: Hm, wondering what are use cases for that .. anyway, I added support for incomplete inputs for aggregator as well now.

src/datachain/checkpoint.py
ilongin: Yes, it's only used when checkpoints are enabled. Without checkpoints, partition tables get auto-generated temp names. It follows the same pattern as the other table naming methods on `Checkpoint` (`output_table_name`, `input_table_name`, `partial_output_table_name`).

shcheklein:
@ilongin why is this one abandoned? what is the status of it?

ilongin:
> @ilongin why is this one abandoned? what is the status of it?

It's not abandoned, I just moved to other tasks in parallel. I still need approval or additional comments / review on it to be able to merge it. If you think it's all ok now please approve.

shcheklein:
@ilongin please review Copilot comments

tests/func/checkpoints/test_checkpoint_workflows.py
```diff
@@ -531,3 +531,23 @@ def test_ephemeral_mode_no_jobs_on_collect(test_session, nums_dataset):
 
     assert _count_rows(metastore, metastore._jobs) == jobs_before
     assert _count_rows(metastore, metastore._checkpoints) == checkpoints_before
+
+
+def test_ephemeral_mode_aggregator_with_partition_by(test_session):
+    dc.read_values(
```
shcheklein, line 539: should we have count rows like in other test?

also try to call save and get an exception?

ilongin:
> @ilongin please review Copilot comments

resolved all of them

tests/func/checkpoints/test_checkpoint_workflows.py
ilongin: added both

src/datachain/checkpoint.py
```diff
@@ -48,6 +48,11 @@ def input_table_name(group_id: str, _hash: str) -> str:
...
+        """Partition mapping table. Job-specific, maps sys__id to partition_id."""
+        return f"udf_{job_id}_{_hash}_partition"
+
```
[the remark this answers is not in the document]
shcheklein: @ilongin was this one resolved?

shcheklein:
@ilongin tests are fialing

src/datachain/query/dataset.py
```diff
@@ -1028,19 +1035,65 @@ def get_or_create_input_table(
...
+        assert checkpoint.job_id == job.rerun_from_job_id
+
+        parent_partition_table_name = Checkpoint.partition_table_name(
+            checkpoint.job_id, _hash
+        )
+
+        try:
+            parent_table = self.warehouse.get_table(parent_partition_table_name)
+        except TableMissingError:
+            raise DataChainError(
```
shcheklein: does it mean we fail the whole processing? I think we are starting from scratch in other cases, no?

shcheklein: also, do we need a test for this?

src/datachain/lib/udf.py
```diff
@@ -622,9 +622,22 @@ def run(
         download_cb: Callback = DEFAULT_CALLBACK,
         processed_cb: Callback = DEFAULT_CALLBACK,
     ) -> Iterator[Iterable[UDFResult]]:
+        from datachain.data_storage.schema import PARTITION_COLUMN_ID
+
         self.setup()
 
+        # Check if partition_id is available (when partition_by is used)
+        partition_id_idx = None
+        if PARTITION_COLUMN_ID in udf_fields:
```
shcheklein, line 631: is empty when it is effectively a single partition?

how do we handle the case when it is empty - do we have tests for this? do we set None? will the recovery logic work fine later?

shcheklein: also, if someone decide to add parition by for example - is there a guarantee (I hope by using hash?) that we won't pick up the older table? Can we also add a test for this just in case?

shcheklein (commented):
I think it looks good overall now. Let's check some final comments (resolve with a comment them).

src/datachain/query/dataset.py
ilongin: You are right, this was missing from this PR. Added fallback to run from scratch and test for it.

src/datachain/lib/udf.py
ilongin: It was working as expected but it was not explicit and was almost working by "accident". I made it explicit now - if no partition_by (single partition aggregator) that means there is no continuation and it will always run from scratch. Added a test for that.

Regarding second question: Yes, guaranteed by hash. `partial_hash` includes `hash_column_elements(partition_by)` — changing `partition_by` produces a different hash, so the old partial checkpoint won't be found. Test: `test_partition_by_change_triggers_rerun` in `test_checkpoint_invalidation.py`.

shcheklein:
@ilongin review, resolve with comments and copilot comments. (and then let's do another round after the changes you made now)

src/datachain/catalog/catalog.py
```diff
@@ -2181,6 +2181,24 @@ def cleanup_checkpoints(self, ttl_seconds: int | None = None) -> int:
...
+        # Partition tables — job-scoped, filter by expired job IDs
+        expired_job_ids = {ch.job_id for ch in checkpoints}
+        all_partition_tables = self.warehouse.db.list_tables(
+            pattern=Checkpoint.partition_table_pattern()
+        )
+        partition_tables = [
+            t
+            for t in all_partition_tables
+            if any(job_id in t for job_id in expired_job_ids)
+        ]
```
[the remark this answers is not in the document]
ilongin: jobs are uuid and this is impossible to happen

src/datachain/checkpoint.py
```diff
@@ -48,11 +48,22 @@ def input_table_name(group_id: str, _hash: str) -> str:
...
     @staticmethod
     def input_table_pattern(group_id: str) -> str:
         """LIKE pattern for finding all input tables in a run group."""
         return f"udf_{group_id}_%_input"
 
+    @staticmethod
+    def partition_table_pattern(job_id: str = "") -> str:
+        """LIKE pattern for finding partition tables. If job_id given, scoped to job."""
+        prefix = f"udf_{job_id}" if job_id else "udf_"
```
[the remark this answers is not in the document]
ilongin: Same as above — job IDs are UUIDv4, prefix collisions can't happen.

PR #1601, opened 2026-02-18
Fix to not create checkpoints and job if checkpoints are disabled

ilongin:
The issue was that there are some cases where we use `DataChain` but don't want to create job / create checkpoints. The case for this is file browser in Studio - we use `.collect()` methods from `DataChain` to calculate something ad-hoc.

This PR does 2 things:
1. Figures out if ad-hoc DataChain calls is used from Studio and disables checkpoints in this case by making sure `checkpoints_enabled()` helper method returns `False`.
2. If `checkpoints_enabled()` returns False then it avoids creating checkpoint and related job (they always go together).

### DIFF
--- a/src/datachain/query/dataset.py
+++ b/src/datachain/query/dataset.py
@@ -977,6 +977,24 @@ class UDFStep(Step, ABC):
         # Create input table from original query
         return self.warehouse.create_pre_udf_table(query, input_table_name)
 
+    def _prepare_partition_query(
+        self,
+        query: Select,
+        hash_input: str,
+        temp_tables: list[str],
+        job: "Job | None" = None,
+    ) -> Select:
+        """Create input and partition tables for partition_by, return updated query."""
+        input_table = self.get_or_create_input_table(query, hash_input, job)
+        query = self.get_input_query(input_table.name, query)
+
+        partition_tbl = self.create_partitions_table(query)
+        temp_tables.append(partition_tbl.name)
+        return query.outerjoin(
+            partition_tbl,
+            partition_tbl.c.sys__id == query.selected_columns.sys__id,
+        ).add_columns(*partition_columns())
+
     def apply(
         self,
         query_generator: QueryGenerator,
@@ -985,8 +1003,6 @@ class UDFStep(Step, ABC):
         hash_output: str,
         checkpoints_enabled: bool = True,
     ) -> "StepResult":
-        job = self.session.get_or_create_job() if checkpoints_enabled else None
-
         query = query_generator.select()
 
         # Calculate partial hash that includes output schema
@@ -996,38 +1012,33 @@ class UDFStep(Step, ABC):
             (hash_input + self.udf.output_schema_hash()).encode()
         ).hexdigest()
 
-        # If partition_by is set, we need to create input table first to ensure
-        # consistent sys__id
-        if self.partition_by is not None:
-            # Create input table first so partition table can reference the
-            # same sys__id values
-            input_table = self.get_or_create_input_table(query, hash_input, job)
-
-            # Now query from the input table for partition creation
-            # Use get_input_query to preserve SQLTypes from original query
-            query = self.get_input_query(input_table.name, query)
-
-            partition_tbl = self.create_partitions_table(query)
-            temp_tables.append(partition_tbl.name)
-            query = query.outerjoin(
-                partition_tbl,
-                partition_tbl.c.sys__id == query.selected_columns.sys__id,
-            ).add_columns(*partition_columns())
-
         if not checkpoints_enabled:
+            if self.partition_by is not None:
+                query = self._prepare_partition_query(
+                    query,
+                    hash_input,
+                    temp_tables,
+                )
             output_table, input_table = self._run_from_scratch(
                 partial_hash,
                 hash_output,
                 hash_input,
                 query,
-                job,
+                None,
                 checkpoints_enabled=False,
             )
             # No checkpoints — tables won't be reused, register for cleanup
             temp_tables.append(output_table.name)
             temp_tables.append(input_table.name)
         else:
-            assert job is not None
+            job = self.session.get_or_create_job()
+            if self.partition_by is not None:
+                query = self._prepare_partition_query(
+                    query,
+                    hash_input,
+                    temp_tables,
+                    job=job,
+                )
             if ch := self._find_udf_checkpoint(job, hash_output):
                 try:
                     output_table, input_table = self._skip_udf(
@@ -1148,8 +1159,7 @@ class UDFStep(Step, ABC):
         )
 
         partial_checkpoint = None
-        if checkpoints_enabled:
-            assert job is not None
+        if checkpoints_enabled and job:
             partial_checkpoint = self.metastore.get_or_create_checkpoint(
                 job.id, partial_hash, partial=True
             )
@@ -1178,8 +1188,7 @@ class UDFStep(Step, ABC):
             partial_output_table, UDFStep.output_table_name(run_id, hash_output)
         )
 
-        if checkpoints_enabled:
-            assert job is not None
+        if checkpoints_enabled and job:
             # Promote partial checkpoint to final and log event
             if partial_checkpoint:
                 self.metastore.remove_checkpoint(partial_checkpoint.id)


src/datachain/query/dataset.py
```diff
@@ -505,6 +505,32 @@ class UDFStep(Step, ABC):
     workers: bool | int = False
     min_task_size: int | None = None
     batch_size: int | None = None
+    _job_cached: Job | None = attrs.field(
+        init=False, default=None, repr=False, eq=False, hash=False
+    )
+
+    @property
+    def job(self) -> Job:
```
ilongin: Idea is:
1. If checkpoints are enabled -> create job as usual (checkpoints need Job object as their owner)
2. If checkpoints are disabled -> create ephermal Job instance, to avoid creating real job, with random id, to make code cleaner (UDF code here works with Job objects although it just uses 3 fields - `id`, `rerun_from_job_id` and `run_group_id`.

src/datachain/query/dataset.py
```diff
@@ -505,6 +505,32 @@ class UDFStep(Step, ABC):
...
+        if self._job_cached is not None:
+            return self._job_cached
+        if not checkpoints_enabled():
+            job = Job.create_ephemeral()
+        else:
+            job = self.session.get_or_create_job()
+        object.__setattr__(self, "_job_cached", job)
+        return job
+
+    def _reset_job_cache(self) -> None:
```
ilongin: This is needed to avoid UDF input/output table name collisions if we have same chain that is applied multiple time e.g
```
ch = dc.read_storage(...).map(...)
ch.count()
ch.count()
ch.count()
```
If checkpoints are disabled and we create ephemeral job only once on chain definition time, every time we call `.count()` map UDF is resolved(applied) and it tries to create input/output table with the same **hash** and **job_id** (ephemeral job id) which will collide with the previous one. Therefor, we need to recreate ephemeral job on every UDF apply to get new unique job ID.

src/datachain/query/dataset.py
```diff
@@ -2158,16 +2181,12 @@ def _starting_step_hash(self) -> str:
         assert self.list_ds_name
         return self.list_ds_name
 
-    @property
-    def job(self) -> Job:
```
ilongin, line 2177: No need for this property as it's used only in one place in `_last_checkpoint_hash()` method

src/datachain/utils.py
```diff
@@ -633,6 +633,13 @@ def checkpoints_enabled() -> bool:
     Returns:
         bool: True if checkpoints are enabled, False if disabled.
     """
+    # Studio internal operations (file browser, ad-hoc queries) don't use
+    # checkpoints. Actual user job execution sets DATACHAIN_JOB_ID.
+    from datachain.lib.dc.utils import is_studio
+
+    if is_studio() and not os.environ.get("DATACHAIN_JOB_ID"):
```
ilongin: This is the controversal part. Current issue is with ad-hoc `DataChain` calls from Studio backend itself that don't really need checkpoints / jobs, since the idea is not to re-run those, but since checkpoints are automatically **always** enabled they create both job (with strange name like `/usr/bin/gunicorn`) and checkpoints.  Those jobs are visible in jobs list which is strange / confusing.

This line fixes those as ad-hoc calls from Studio backend don't send `DATACHAIN_JOB_ID` (as they don't create job) so we can identify them. This still mean that locally, if user uses DataChain for example as a lib in his project, it will create both job and checkpoint although maybe it's also not needed .... the place where it is needed to create both, by contract, is when user runs specific script that has DataChain code inside with `python my_script.py` but it's impossible to identify correctly how exactly user uses it so the only way is to just enable checkpoints and create jobs on every DataChain call locally.

Alternative is to have something like `DATACHAIN_CHECKPOINTS_DISABLED=1` which user can set on his project that uses ad-hoc DataChain, if he doesn't want additional objects being created with no purpose, or we can set on Studio backend to avoid creating jobs on file browser which also uses ad-hoc DataChain.

src/datachain/query/dataset.py
shcheklein: > make code cleaner

what does it mean exactly? can you give examples? it feels a bit artificial ...

src/datachain/query/dataset.py
shcheklein: this also looks ugly :(

let's zoom out - can we redesign this stuff somehow ... really don't like all these ad-hoc manipulations

ilongin: I refactor to create job in apply() method every time. Now we don't need all that manipulation around frozen class but need to pass job to every helper method .. I guess now it's more verbose but also less complex.

src/datachain/query/dataset.py
ilongin: Means that since UDF logic now heavily relies to Job object and it's fields to create input, output tables and other, it's the easiest to continue to use Job object but since in the case of disabled checkpoints we should not actually create Job in DB I decided to create this "ephemeral" object that is just there to work with existing codebase. Alternative is to create whole different code paths for 1. checkpoint is enabled and 2. checkpoint is disabled which is not really an option IMO.

So in short words, code stayed the same, and the only change between "checkpoint is enabled" and "checkpoint is disabled" is on which job to create - actual or ephemeral and now that on beginning of `UDFStep.apply()` method

shcheklein: > but since in the case of disabled checkpoints we should not actually create Job in DB

this is not exactly clear

there can be a job with disabled checkpoints

checkpoints are checkpoints, jobs are jobs - two different things. It is not obvious, clear why disabled checkpoints automatically lead to no Job workflow ...

disabling checkpoints should be literally about that - not creating some extra tables for example when we don't need them

> So in short words, code stayed the same, and the only change between "checkpoint is enabled" and "checkpoint is disabled" is on which job to create - actual or ephemeral and now that on beginning of UDFStep.apply() method

this still feels artificial (just to patch this problem now). I am still not quite convinced that those should be affecting each other this way. 

Alternatives:

- Always create jobs (disable checkpoints if needed should be just a way to runs things faster for example w/o creating extra tables). Mark some jobs as internal. Fix them to also have meaningful details - proper query, execution time, etc. This can be even helpful to debug and analyze things.
- Create some interface where you can pass in-memory session for a Job?

ilongin: You're right that jobs and checkpoints are separate things and should not be coupled. But in Studio, Job has complex lifecycle management (celery tasks, stale detection, keepalive, log processing, webhooks). Adding 'internal' jobs means doing exclusion logic through 6+ periodic tasks and monitoring paths. 
The ephemeral job / lightweight execution context approach avoids all of that — it gives the UDF pipeline the 3 fields it needs (`id`, `rerun_from_job_id`, `run_group_id`) without creating anything in the DB or triggering any lifecycle side effects. We could formalize it as a `ExecutionContext` dataclass to make the intent clearer, but the core idea of not creating a real Job when checkpoints are disabled is the right call for Studio. 
Note that this doesn't mean that a Job won't be created — it just means that checkpoint logic won't explicitly try to create one, as it's not needed in that scenario anyway (it just needs those 3 fields for table naming).

I will try to simplify it and make more explicit about table naming

ilongin: Removed ephemeral job and adjusted logic to use just uuid4() for table naming when checkpoints are disabled

src/datachain/query/dataset.py
```diff
@@ -888,31 +884,34 @@ def _log_event(
...
     def _find_udf_checkpoint(
-        self, _hash: str, partial: bool = False
+        self, job: "Job | None", _hash: str, partial: bool = False
     ) -> Checkpoint | None:
         """
         Find a reusable UDF checkpoint for the given hash.
         Returns the Checkpoint object if found and checkpoints are enabled,
         None otherwise.
         """
+        if job is None:
```
shcheklein: so, it can be None only if checkpoints are disabled, right? can we even avoid calling _find_udf_checkpoints then?

src/datachain/query/dataset.py
shcheklein: can we look at this other way around? 

in some cases we need Job-less workflow (for whatever reason, e.g. lack of DB or something). In such workflow checkpoints are disabled (bc they require Job to be present).

or put it in a different way.

We have a scenario when DC is used as kind-a API, not a "query" to get results, save datasets, etc.

Can we name this mode somehow? Like you know - interactive / non-interactive modes in shell, etc.

Just trying to come up with a cleaner view on all of this.

So, if we have this "API" mode we then can just disable all the things we need to disable in that mode.

Ideally, we should not have `if Job` as well in checkpoints code. It should not be even triggered based on some early checkpoints_disabled - that checks that env var / mode.

Does it make sense?

ilongin: I like the "API mode" idea conceptually, but I investigated how it would affect the code and I don't see a practical benefit right now.                                                                                                       

In `apply()`, the checkpoints-disabled path still needs to create input/output tables, run the UDF, and build result  queries — the UDF pipeline fundamentally requires intermediate tables regardless of mode. So "API mode" would just be a rename of `checkpoints_enabled()`, not a structural simplification. The `job: None` handling in `_run_from_scratch()` / `get_or_create_input_table()` would remain too — we need some ID for table naming either way. The alternative would be duplicating these methods into "API mode" variants to avoid if job checks, but that's worse — same logic in two places.

There's also the detection problem: in Studio we control the environment, but locally we can't reliably distinguish "user exploring data in a notebook" from "user running a production pipeline". So it would have to be an explicit opt-in, which is essentially what `DATACHAIN_CHECKPOINTS_DISABLED` already is.

We can maybe continue to brainstorm about this but for now honestly I would avoid going into this direction until we see clear benefit

shcheklein: I don't suggest removing checkpoints_enabled. It stays - it just what does it analyze inside to decide. Also, yes, it must be also explicit.

What I'm trying to achieve is clear decouple of jobs and checkpoints and this API mode and less usage of env vars everywhere, especially when we have to setup them, then pop them, etc - just a lot of manipulations, affecting different parts of code people will have to remember about etc.

Kinda something like:

```
dc.settings(headless=True)...
```

or 

```
dc.settings(session=SomeSpecialSession)
```

something in this direction ... 

to avoid env vars, provide

ilongin: I'm not sure I follow on decoupling jobs and checkpoints — a checkpoint is fundamentally `(job_id, hash, ...)`, it records what a specific job completed.    
Without a job there's nothing to checkpoint against. That dependency is  in the model itself, not something we can decouple.                            
                                                                                                                                                             
If the goal is a true "headless" / "API mode" where nothing persists — no jobs, no checkpoints, no datasets saved, nothing, then I like that idea and it's doable (we already support in-memory SQLite in tests for example).

But in practice (e.g in Studio usage), ad-hoc DataChain usage still needs to read from existing databases and write intermediate tables to the same warehouse (e.g SQL joins and UDF requires it). So it can't be truly in-memory — we need the real DBs, just without the metadata/warehouse side effects.

So there are really two different scenarios:

1. **Real ad-hoc** — could work fully in-memory, real headless, no existing data needed - easier to implement
2. **Ad-hoc on top of existing data** (Studio filebrowser, user projects with existing datasets, etc.) — needs real DBs for reads/computation, just suppress side
effects (jobs, checkpoints)

These share the "no side effects" intent but the underlying mechanism is completely different. Trying to unify them under one `headless=True` flag would be
wrong I think.  Also, second one is significantly harder to implement and is def out of the scope of this PR.

So to conclude: I like the idea of `headless=True` but I suggest to merge this to fix immediate problem and continue on that in separate task where we can decide how would that actually work.

shcheklein: checkpoint indeed needs a job - that is fine, but job doesn't need a checkpoint (so it is one way direction).

here what I see we use DISABLE_CHECKPOINTS to actually disable Jobs (and by proxy checkpoints).

Should it be then DISABLE_JOBS at least? That's the goal, right? Not to create jobs in the first place. 

Or let's do this mental exercise.  Imagine even we had a way to create checkpoints w/o jobs. We would have been fine keeping them enabled for these internal jobs (a bit of waste and they won't reused - but who cares). But Jobs - we can't - people will see them.

So, we are solving _JOBS_ appearing and visible in the fist place. And for some reason we are solving it with DISABLE_CHECKPOINTS.

Thus at the very least you making it bidirectional and strongly coupled. 

---------

Let's zoom out a bit. 

It's not only about us. I can imagine people will use this in their environments, right? E.g. on some server they might run datachain as helper to process files. It won't be in our SaaS, so some local ad-hoc metastore will be utilized. Are we fine in that case that Job will be created with some random name? will it be created at all? are we fine that checkpoints will be created? will they be reused by mistake as well?

> Also, second one is significantly harder to implement and is def out of the scope of this PR.

any specific idea how to implement it?

> So to conclude: I like the idea of headless=True but I suggest to merge this to fix immediate problem and continue on that in separate task where we can decide how would that actually work.

Depends, I would say. Let's outline the ideas then decide if the scope it large do it later, if not do it here. I'm concerned that we keep patching things a bit ad-hoc and keep growing thus tech debt and amount of followups.

ilongin: My intent was to disable checkpoints (and since checkpoints need job and create one if doesn't exist, jobs as well as transient byproduct), but I didn't want to disable jobs in general, meaning jobs might still be created in some other way maybe  - I didn't go through all the scenarios where jobs might still be needed even in ad-hoc. If we are sure that jobs should never be created in ad-hoc scenario (and it seems that's true as well when I look at the codebase), then I agree that `DISABLE_JOB` is better.

> any specific idea how to implement it?

Actually it seems like it's not that hard .. I've added new commit in this PR that implements it. I just named setting ephemeral as I didn't really liked headless word

tests/func/checkpoints/test_checkpoint_workflows.py
```diff
@@ -470,3 +476,51 @@ def double_num(num) -> int:
         dc.read_dataset("doubled_nums", session=test_session).to_list("doubled")
     )
     assert result == [(2,), (4,), (6,), (8,), (10,), (12,)]
+
+
+def test_udf_runs_in_ephemeral_mode(test_session, nums_dataset):
+    """UDF chain works correctly in ephemeral mode."""
```
shcheklein: remove obvious comments

src/datachain/query/dataset.py
shcheklein: question stays? is there a way to not call anything related to checkpoints in some centralized way?

src/datachain/query/dataset.py
```diff
@@ -993,6 +991,10 @@ def apply(
         hash_input: str,
         hash_output: str,
     ) -> "StepResult":
+        skip_checkpoints = self.ephemeral or not checkpoints_enabled()
```
shcheklein: should checkpoints_enabled include self.ephemeral check?

src/datachain/query/dataset.py
ilongin: That condition was not needed there, removed.

In general I refactor little bit whole checkpoints enabled checks. Now the ground truth for it is in property method `DataChain._checkpoints_enabled()`. This uses that util function `checkpoints_enabled()` but also passes ephemeral setting value since it has access to it. This information about checkpoints enabled/disabled is then passed down to **DatassetQuery** where in `UDFStep.apply()` we have one if clause dealing with checkpoints not enabled and everything else is about checkpoints enabled so I think it's pretty much "localized" in right spots and not scattered around like before.

src/datachain/query/dataset.py
ilongin: Yes, added it there. More details in comment above

PR #1611, opened 2026-03-03
feat: export __version__ from datachain package

haosenwang1018:
## Summary
- add `__version__` to `datachain.__init__` using `importlib.metadata.version()`
- provide a safe fallback (`0.0.0`) when package metadata is unavailable
- include `__version__` in `__all__` for explicit public export

## Why
This makes it easier for downstream users and tooling to inspect the installed package version directly from the top-level API.

## Testing
- [x] local import path remains valid
- [x] change is additive and backward-compatible

### DIFF
--- a/src/datachain/__init__.py
+++ b/src/datachain/__init__.py
@@ -1,3 +1,5 @@
+from importlib.metadata import PackageNotFoundError, version
+
 from datachain.lib.data_model import DataModel, DataType, is_chain_type
 from datachain.lib.dc import (
     C,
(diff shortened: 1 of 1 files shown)


src/datachain/asyn.py
```diff
@@ -126,7 +126,7 @@ async def run(self) -> None:
             )
             self.gather_exceptions(done)
             assert join.done()
-        except:
+        except Exception:
```
shcheklein, line 129: @haosenwang1018 unrelated changes

shcheklein:
Branch is outdated for some reason. Closing this for now, please @haosenwang1018 clean it up, provide more details, then we can consider merging this.

PR #1613, opened 2026-03-03
docs: add version check to quick start

haosenwang1018:
## Summary
- add a short "Version check" snippet to `docs/quick-start.md`
- show how to verify installed version via `datachain.__version__`

## Why
Users often validate installation before running examples; this provides a quick, explicit check in the quick-start flow.

## Testing
- [x] docs-only change

### DIFF
--- a/docs/quick-start.md
+++ b/docs/quick-start.md
@@ -18,6 +18,13 @@ title: Quick Start
     uv add datachain
     ```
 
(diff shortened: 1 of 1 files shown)


shcheklein:
I don't think we need this in this in quick start. At best somewhere in installation guide / troubleshooting.

PR #1614, opened 2026-03-03
docs: fix Mistral client usage in quick-start example

haosenwang1018:
## Summary
- fix the quick-start "Serializing Python-objects" snippet to use `Mistral(...)` consistently
- replace `client.chat(...)` with `client.chat.complete(...)` to match current SDK usage in this doc
- add missing `import os` for `MISTRAL_API_KEY`

## Why
The existing snippet used `MistralClient()` without import and mixed API styles, which can confuse users copying the example.

## Testing
- [x] docs-only change


shcheklein:
Closing for now, no response.

PR #1621, opened 2026-03-04
feat(gc): include STALE and REMOVING dataset versions in GC cleanup

amritghimire:
## Summary

- Add `DatasetStatus.STALE` and `REMOVING` to the status filter in  `get_incomplete_dataset_versions()` so lifecycle-marked versions are picked up by the GC
- Update abstract method docstring to mention REMOVING
- Add two new tests covering REMOVING GC pickup and cleanup

## Context

Studio's lifecycle system (datachain-ai/studio#12515) marks excess dataset versions as STALE instead of deleting them directly from theStudio worker. The `datachain_worker` GC then performs the actual warehouse and metastore deletion. This change is the library-side prerequisite for that workflow.

## Test plan
- [x] `test_get_incomplete_dataset_versions_includes_stale` passes
- [x] `test_cleanup_failed_dataset_versions_removes_stale` passes
- [x] All existing `test_dataset_status_management` tests unchanged and passing

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,13 +24,13 @@ repos:
       - id: trailing-whitespace
         exclude: '^LICENSES/'
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: 'v0.15.4'
+    rev: 'v0.15.5'
     hooks:
       - id: ruff
         args: [--fix, --exit-non-zero-on-fix]
       - id: ruff-format
   - repo: https://github.com/codespell-project/codespell
-    rev: v2.4.1
+    rev: v2.4.2
     hooks:
       - id: codespell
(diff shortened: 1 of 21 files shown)


src/datachain/data_storage/metastore.py
```diff
@@ -341,14 +341,16 @@ def get_incomplete_dataset_versions(
         self, job_id: str | None = None
     ) -> list[tuple[DatasetRecord, str]]:
         """
-        Get incomplete dataset versions to clean up.
+        Get incomplete or stale dataset versions to clean up.
```
shcheklein: can we use a more straightforward status - DELETED for these? they are not stale, they are removed essentially AFAIU

amritghimire: Updated to use removing status.

PR #1623, opened 2026-03-05
Remove obsolete 'conn' and 'cursor' params from metastore methods

dreadatour:
Metastore methods `conn` and `cursor` params are obsolete:
- it is not used in most of methods — only 2 methods really uses these params (`get_or_create_checkpoint ` and `link_dataset_version_to_job `) and in both methods it was useless anyway
- is it not consistent — many methods does not have this param
- it is not working in CLI
- it is not working in SaaS

In this PR I am cleaning up `AbstractMetastore`, `AbstractDBMetastore` and `SQLiteMetastore` and remove `conn` and `cursor` params from all the methods.

Also couple tiny fixes along the way with typings and blank lines for consistency.

### DIFF
--- a/src/datachain/data_storage/sqlite.py
+++ b/src/datachain/data_storage/sqlite.py
@@ -259,8 +259,8 @@ class SQLiteDatabaseEngine(DatabaseEngine):
         return self.db.execute(*self.compile_to_args(query))
 
     @retry_sqlite_locks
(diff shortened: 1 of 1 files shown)


dreadatour:
> Seems there are some unrelated changes

There are few tiny changes related to typings and whitespaces/empty lines, but everything else related to the cleanup and make consistent of metastores/database engines methods across both repos.

PR #1628, opened 2026-03-09
Fixing output when using progress bar in REPL

ilongin:
When tqdm closes a bar with `leave=False`, it "erases" it by writing `\r` + spaces + `\r` to stderr. The cursor returns to column 0, but the space characters remain on the terminal line. Any subsequent output (DataFrame text from `show()`, or the `>>>` REPL
prompt) only overwrites as many characters as it writes — the rest of the spaces stay, producing visible corruption like  `Prvalsdoubledrows [00:00, ? rows/s]` or `>>>                               `.                                                 

Added a `Tqdm` subclass in `src/datachain/progress.py` that overrides `close()` to write the ANSI "Erase in Line" sequence (`\r\033[K`) after the bar is cleared, which properly removes the residual characters. 
Replaced all `from tqdm.auto import tqdm` imports across the codebase (11 files) with `from datachain.progress import tqdm`, so any future tqdm usage automatically gets
the fix.

### DIFF
--- a/src/datachain/cache.py
+++ b/src/datachain/cache.py
@@ -84,7 +84,7 @@ class Cache:  # noqa: PLW1641
         size = file.size
         if size < 0:
             size = await client.get_size(from_path, version_id=file.version)
-        from tqdm.auto import tqdm
+        from datachain.progress import tqdm
 
         cb = callback or TqdmCallback(
             tqdm_kwargs={"desc": odb_fs.name(from_path), "bytes": True, "leave": False},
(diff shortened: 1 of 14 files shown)


shcheklein:
@ilongin was it prioritized in Slack? why this particular issue?

we still have checkpoint followup + a bunch the list of things to do from Slack, right?

ilongin:
> pywidgets

Tested locally in Jupyter notebook both with and without ipywidgets — works fine in both cases. With ipywidgets, tqdm uses HTML widgets (not terminal writes) so the cleanup is a no-op. Without ipywidgets, tqdm falls back to text mode and our \r\033[K cleanup applies the same way as in a terminal. No artifacts or garbled output in either case.

PR #1631, opened 2026-03-10
Cleanup temp datasets

ilongin:
Add cleanup for orphaned temporary (`session_*`) datasets left behind when a worker is killed before session cleanup runs.                 
                           
- `metastore.get_temp_datasets_to_clean()` finds `session_*` dataset versions whose job is in a final state (COMPLETE/FAILED/CANCELED).    
Versions without a `job_id` are skipped.                                                                                                   
- `catalog.cleanup_temp_datasets()` removes those versions (and the parent dataset if it was the only version).
- `datachain gc` now includes this cleanup step.

### DIFF
--- a/tests/unit/test_dataset_status_management.py
+++ b/tests/unit/test_dataset_status_management.py
@@ -360,8 +360,7 @@ def test_cleanup_dataset_versions_removes_marked_for_removal(
 ):
     test_session.catalog.metastore.set_job_status(job.id, JobStatus.COMPLETE)
 
-    num_removed = test_session.catalog.cleanup_dataset_versions()
-    assert num_removed == 1
+    test_session.catalog.cleanup_dataset_versions()
 
     with pytest.raises(DatasetNotFoundError):
         test_session.catalog.get_dataset(dataset_marked_for_removal.name)


shcheklein:
CI fails (include codecov)

src/datachain/catalog/catalog.py
```diff
@@ -1166,6 +1166,34 @@ def cleanup_failed_dataset_versions(self, job_id: str | None = None) -> int:
...
+        Clean up orphaned temporary (session_*) datasets.
+
+        Removes dataset versions whose associated job has finished
+        (COMPLETE, FAILED, or CANCELED). These are left behind when
+        a worker is killed before the session cleanup runs.
+
+        Returns:
+            Number of removed dataset versions.
+        """
+        versions_to_clean = self.metastore.get_temp_datasets_to_clean()
```
shcheklein: yet again we come up with some custom API ... we have already 2-3 GC tasks - is there a real need for another one? also temporary is very confusing since we have real `tmp` datasets.

Let's zoom out, think, and make it cleaner

src/datachain/data_storage/metastore.py
```diff
@@ -1718,6 +1734,51 @@ def get_incomplete_dataset_versions(
...
+        p = self._projects
+        d = self._datasets
+        dv = self._datasets_versions
+        j = self._jobs
+
+        select_cols = (
+            *(getattr(n.c, f) for f in self._namespaces_fields),
+            *(getattr(p.c, f) for f in self._projects_fields),
+            *(getattr(d.c, f) for f in self._dataset_fields),
+            *(getattr(dv.c, f) for f in self._dataset_version_fields),
```
shcheklein: do we need versions? do we need previews and all fields for these datasets if all we want is to drop them essentially

src/datachain/catalog/catalog.py
ilongin: Consolidated the GC methods:

- `get_incomplete_dataset_versions()` → renamed to `get_dataset_versions_to_clean()` and extended to also find `session_*` COMPLETE datasets from finished jobs (same query, extra OR condition)
- `cleanup_failed_dataset_versions()` → renamed to `cleanup_dataset_versions()` — now handles both incomplete versions and orphaned session datasets in one pass
- Removed `get_temp_datasets_to_clean()` and `cleanup_session_datasets()` from metastore/catalog — no longer needed
- `datachain gc` now has one step for dataset version cleanup instead of two

src/datachain/data_storage/metastore.py
ilongin: This method is now removed as I merged two cleaning dataset operations into one as explained in above comment

Catcat855899:
謝謝辛苦幫我忙，但我完全看比不懂，我今天才知道我有設立這個帳號，因為我的設備
因我要換設備加上我真的不懂這是做什麽用的，所以因為沒辦法好好照顧，所以要把帳號刪除， 謝謝你們的💮💃

PR #1644, opened 2026-03-12
fix(delta): transform query properly to make unsafe work

shcheklein:
This change makes `delta_unsafe=True` updates work for queries with `merge` / `union`. When multiple delta-enabled sources are involved, each changed source is handled independently and its updated output is combined into the final result.

Additional fix: source dependency matching now uses full source identity (`name + project + namespace`) so same-named datasets are resolved correctly.

Closes https://github.com/datachain-ai/datachain/issues/1651

## Examples

Single-source delta still works the same:

```python
dc.read_dataset("items", delta=True, delta_on="id").save("items_delta")
```

`delta_unsafe` queries can now pick up changes from either side of a composed query:

```python
(
    dc.read_dataset("left", delta=True, delta_on="id", delta_unsafe=True)
    .merge(
        dc.read_dataset("right", delta=True, delta_on="id", delta_unsafe=True),
        on="id",
        inner=True,
    )
    .save("joined_delta")
)
```

### DIFF
--- a/src/datachain/query/dataset.py
+++ b/src/datachain/query/dataset.py
@@ -255,9 +255,8 @@ class QueryStep:
         )
 
     def hash(self) -> str:
-        return hashlib.sha256(
-            self.dataset.uri(self.dataset_version).encode()
-        ).hexdigest()
+        version = self.dataset.get_version(self.dataset_version)
+        return hashlib.sha256(version.uuid.encode()).hexdigest()
 
 
 def generator_then_call(generator, func: Callable):
--- a/tests/unit/test_datachain_hash.py
+++ b/tests/unit/test_datachain_hash.py
@@ -60,6 +60,19 @@ def mock_get_listing():
         yield mock
 
 
+def _set_stable_uuid(test_session, name, uuid):
+    """Set a stable UUID on a dataset version for deterministic hash tests."""
+    parts = name.split(".")
+    ds_name = parts[-1]
+    ns = parts[0] if len(parts) > 1 else "default"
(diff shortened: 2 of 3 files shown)


src/datachain/delta.py
```diff
@@ -17,42 +17,78 @@
...
+        operands = [
+            value
+            for value in (self, *args, *kwargs.values())
+            if isinstance(value, DataChain)
+        ]
+        if not operands:
+            raise RuntimeError(
+                "delta_disabled expected at least one DataChain operand "
+                f"for {method.__name__}"
+            )
```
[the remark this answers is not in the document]
shcheklein: defense in depth, let's keep it

src/datachain/delta.py
```diff
@@ -17,42 +17,78 @@
...
 
 T = TypeVar("T", bound="DataChain")
 
 
+def _dataset_identity(dataset: DatasetRecord) -> tuple[str, str, str]:
+    return (
+        dataset.project.namespace.name,
+        dataset.project.name,
+        dataset.name,
+    )
```
dreadatour, line 33: I think we might have this already somewhere. Anyway, will it be better to bay be move it to `DatasetRecord` class?

dreadatour (approved):
Good update and code cleanup. Looks great to me 👍

Tiny comment about dataset identity — we should finally address this one day.

src/datachain/delta.py
shcheklein: I think in this case we don't include version (it is important here) - thus it is different and it is a bit risky to share it (we need to be sure to use precise semantics here)

PR #1645, opened 2026-03-13
Using dataset version uuid for hash calculation in QueryStep

ilongin:
Before we were using dataset name + version which is not stable between runs because someone could remove version and create new one.

### DIFF
--- a/tests/unit/test_datachain_hash.py
+++ b/tests/unit/test_datachain_hash.py
@@ -60,6 +60,19 @@ def mock_get_listing():
         yield mock
 
 
+def _set_stable_uuid(test_session, name, uuid):
+    """Set a stable UUID on a dataset version for deterministic hash tests."""
+    parts = name.split(".")
+    ds_name = parts[-1]
+    ns = parts[0] if len(parts) > 1 else "default"
+    proj = parts[1] if len(parts) > 2 else "default"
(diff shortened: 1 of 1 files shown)


tests/unit/test_query_steps_hash.py
```diff
@@ -500,41 +501,51 @@ def test_udf_aggregator_hash(
...
+                _preview_data=[],
+            ),
+        ],
+        status=1,
+        schema={},
+        feature_schema={},
+        project=None,
+    )
+
+    hash1 = QueryStep(None, ds, "1.0.0").hash()
```
shcheklein, line 539: how does it work with the QuerySteps like Join, Subtract, etc?

ilongin: `SQLUnion`, `SQLJoin`, and `Subtract` all call `self.query1.hash()` / `self.query2.hash()` / `self.dq.hash()` in their `hash_inputs()`. This calls `DatasetQuery.hash()`, which goes through the full chain: `_starting_step_hash` + all step hashes. 
At the root, `_starting_step_hash` calls `QueryStep.hash()`, which now uses the UUID. So the UUID propagates automatically - no changes needed in those steps.

shcheklein:
@ilongin tests fail

PR #1650, opened 2026-03-15
[WIP] skills

dmpetrov:
datachain skill.

See new readme and how to use it: https://github.com/datachain-ai/datachain/pull/1695

ToDo:
- test Cursor
- test Codex

Note, the skills scripts were excluded from codecoverage tests. These are chacking buckets and DB directly - we need to come up with a strategy on how to test it properly.


shcheklein (approved):

dreadatour (approved):
Looks good to me! 👍

Only one question: should we think about skills update command? At least as `uninstall` + `install` may be at first?

dmpetrov:
> Only one question: should we think about skills update command? At least as `uninstall` + `install` may be at first?

Yes, `uninstall` + `install` works for now.

PR #1652, opened 2026-03-16
fix(cli): reset reconnect counter after healthy WebSocket session

amritghimire:
## Summary

When monitoring a long-running job (10+ hours), the CLI WebSocket connection drops periodically due to proxy/load-balancer timeouts — this is expected behavior. The CLI was giving up after only 10 seconds (10 retries × 1s), exiting with code 1 and failing CI even though the server-side job was still running fine.

Two root causes:
1. `retry_count` was never reset after a successful reconnect, so accumulated across multiple drops
2. Total retry window was only 10 seconds — far too short for long-running jobs

## Changes

- **`src/datachain/studio.py`**
  - Track `received_streaming_data` per WS session (actual logs or new status transitions, not just the initial handshake the server always sends on connect); reset `retry_count = 0` only when real data was received — each fresh drop gets a full retry window
  - Replace fixed 1s sleep with exponential backoff + jitter (`1s → 2 → 4 → … up to ~61s` — cap is 60s, jitter adds up to 1s on top), matching the `retry_with_backoff` pattern in `utils.py`
  - Rename constants to `RECONNECT_MAX_ATTEMPTS` / `RECONNECT_BACKOFF_BASE_SEC` / `RECONNECT_BACKOFF_MAX_SEC`; increase to 15 consecutive attempts (~10 min window of failures before giving up)
  - Show an in-place reconnect countdown on stdout (e.g. `>>>> WebSocket closed, reconnecting in 5s...`), erased automatically on successful reconnect; detailed attempt count logged at `DEBUG` level only
  - Deduplicate logs on reconnect using `last_log_id`: server replays all historical logs on every new WS connection; track highest printed log ID and skip duplicates
  - Update terminal failure message to include recovery hint: `datachain job logs <job_id>`
  - Extract `_process_logs_message()` helper to reduce function complexity

- **`tests/test_cli_studio.py`**
  - Update 3 existing tests to use new constants and messages
  - Add 2 new tests: log deduplication on reconnect, retry counter reset after streaming data

## Test plan

- [x] `test_studio_run_websocket_disconnect_job_still_running` — exits with recovery hint after max attempts
- [x] `test_studio_run_verbose_max_retry` — counter increments, debug log at limit
- [x] `test_studio_run_log_deduplication_on_reconnect` — replayed logs not printed twice
- [x] `test_studio_run_reconnect_resets_counter_on_streaming_data` — counter resets after real data, then depletes on empty sessions

### DIFF
--- a/src/datachain/studio.py
+++ b/src/datachain/studio.py
@@ -409,6 +409,7 @@ def show_logs_from_client(  # noqa: C901
             received_streaming_data = False
             session_start_id = last_log_id
             async for message in client.tail_job_logs(job_id, no_follow=no_follow):
+                reconnect_msg = _clear_line(reconnect_msg)
                 if "log_blobs" in message and not no_follow:
                     log_blobs = message.get("log_blobs", [])
                     if log_blobs and not log_blobs_processed:
@@ -429,7 +430,6 @@ def show_logs_from_client(  # noqa: C901
                     processed_statuses.add(latest_status)
                     print(f"\n>>>> Job is now in {latest_status} status.")
 
-            reconnect_msg = _clear_line(reconnect_msg)
             if received_streaming_data:
                 retry_count = 0
 
@@ -451,13 +451,13 @@ def show_logs_from_client(  # noqa: C901
                     RECONNECT_BACKOFF_MAX_SEC,
                 ) + random.uniform(0, 1)  # noqa: S311
                 retry_count += 1
-                reconnect_msg = _print_reconnect_msg(sleep_sec)
                 logger.debug(
                     "WebSocket closed, reconnecting in %.1fs (attempt %d/%d)",
                     sleep_sec,
                     retry_count,
                     RECONNECT_MAX_ATTEMPTS,
                 )
+                reconnect_msg = _print_reconnect_msg(sleep_sec)
                 await asyncio.sleep(sleep_sec)
             except KeyError:
                 break
--- a/src/datachain/remote/studio.py
+++ b/src/datachain/remote/studio.py
@@ -317,24 +317,28 @@ class StudioClient:
         if no_follow:
             ws_url += "&no_follow=true"
 
-        async with websockets.connect(
-            ws_url,
-            additional_headers={"Authorization": f"token {self.token}"},
-        ) as websocket:
-            while True:
-                try:
-                    message = await websocket.recv()
-                    data = json.loads(message)
-                    if data.get("type") == "ping":
-                        continue
-                    # Yield the parsed message data
-                    yield data
+        try:
+            async with websockets.connect(
+                ws_url,
+                additional_headers={"Authorization": f"token {self.token}"},
+            ) as websocket:
+                while True:
+                    try:
+                        message = await websocket.recv()
+                        data = json.loads(message)
+                        if data.get("type") == "ping":
+                            continue
+                        # Yield the parsed message data
+                        yield data
 
-                except websockets.exceptions.ConnectionClosed:
-                    break
-                except Exception as e:  # noqa: BLE001
-                    logger.error("Error receiving websocket message: %s", e)
-                    break
+                    except websockets.exceptions.ConnectionClosed:
+                        break
+                    except Exception as e:  # noqa: BLE001
+                        logger.error("Error receiving websocket message: %s", e)
+                        break
+        except (websockets.exceptions.WebSocketException, OSError) as e:
+            logger.debug("WebSocket connection failed: %s", e)
+            return
 
     def ls(self, paths: Iterable[str]) -> Iterator[tuple[str, Response[LsData]]]:
         # TODO: change LsData (response.data value) to be list of lists


src/datachain/studio.py
```diff
@@ -369,48 +371,78 @@ def _get_job_status(client, job_id: str) -> str | None:
     return None
 
 
+def _process_logs_message(logs: list, last_log_id: int) -> tuple[bool, int]:
+    received = False
+    for log in logs:
+        log_id = log.get("id")
+        if log_id is not None and log_id <= last_log_id:
```
amritghimire: This is so, we don't print the logs duplicate when the websocket reconnects.

src/datachain/studio.py
```diff
@@ -369,48 +371,78 @@ def _get_job_status(client, job_id: str) -> str | None:
...
                     latest_status = message["job"]["status"]
                     if latest_status in processed_statuses:
                         continue
+                    received_streaming_data = True
                     processed_statuses.add(latest_status)
                     print(f"\n>>>> Job is now in {latest_status} status.")
 
-            # After websocket closes, check actual job status via REST
+            if received_streaming_data:
+                retry_count = 0
```
amritghimire, line 434: This is one of the main change, we reset the retry count to 0 when we successfully connect once.

src/datachain/studio.py
```diff
@@ -369,48 +371,78 @@ def _get_job_status(client, job_id: str) -> str | None:
...
-                if retry_count > RETRY_MAX_TIMES:
-                    logger.debug("Max retry count reached: %s", retry_count)
+                if retry_count >= RECONNECT_MAX_ATTEMPTS:
+                    logger.debug("Max reconnect attempts reached: %d", retry_count)
                     break
-                await asyncio.sleep(RETRY_SLEEP_SEC)
+                sleep_sec = min(
+                    RECONNECT_BACKOFF_BASE_SEC * 2**retry_count,
+                    RECONNECT_BACKOFF_MAX_SEC,
+                ) + random.uniform(0, 1)  # noqa: S311
```
amritghimire, line 452: This is the main change, instead of simple sleep for 1 sec for 10 times, we now wait in exponential backoff for about 10 minutes (15 times) so that it will be more resilient.

src/datachain/studio.py
```diff
@@ -426,17 +458,21 @@ async def _run():
 
     if not job_finished:
         logger.debug("Job is not finished: %s.", final_status or "unknown")
-        print(f"\n>>>> Lost connection. Job status: {final_status or 'unknown'}.")
+        print(
+            f"\n>>>> Failed to reconnect after {RECONNECT_MAX_ATTEMPTS} attempts."
+            f" Job status: {final_status or 'unknown'}."
+            f"\nThe job may still be running. To resume monitoring:"
+            f"\n    datachain job logs {job_id}"
```
amritghimire, line 481: For UX so, user will know what to do in case user run in the CLI.

src/datachain/studio.py
shcheklein: how can be log_id None?

src/datachain/studio.py
```diff
@@ -369,48 +371,78 @@ def _get_job_status(client, job_id: str) -> str | None:
     return None
 
 
+def _process_logs_message(logs: list, last_log_id: int) -> tuple[bool, int]:
+    received = False
+    for log in logs:
+        log_id = log.get("id")
+        if log_id is not None and log_id <= last_log_id:
+            continue
+        if log_id is not None:
```
shcheklein: same question - is it possible at all?

src/datachain/studio.py
shcheklein: do we print some message to the user ? like disconnected, retrying in ...

tests/test_cli_studio.py
```diff
@@ -1165,6 +1164,93 @@ async def mock_tail_job_logs(jid, no_follow=False):
...
+                    {"id": 2, "message": "line 2\n"},
+                    {"id": 3, "message": "line 3\n"},
+                ]
+            }
+            yield {"job": {"status": "COMPLETE"}}
+
+    mocker.patch(
+        "datachain.studio.StudioClient.tail_job_logs", side_effect=mock_tail_job_logs
+    )
+    mocker.patch(
```
shcheklein: this is not a good practice .. monkey patching protected / implementation details 

find a different way to test this

src/datachain/studio.py
amritghimire: We intentionally keep reconnect attempts at DEBUG level only (visible with -v) to avoid cluttering stdout for normal users , most users running datachain job run in CI just want clean log output without reconnect
  noise. The failure message at the end (after exhausting retries) already tells the user what happened and how to resume with datachain job logs <job_id>.

I remember getting some comment on similar line from you probably where we don't want to show reconnecting information if user doesn't ask for so, kept it in debug.

tests/test_cli_studio.py
amritghimire: Updated

src/datachain/studio.py
amritghimire: Updated

src/datachain/studio.py
amritghimire: Updated

src/datachain/studio.py
shcheklein: I think it makes sense to show (and remove it when reconnect happened) - it doesn't make sense to show full stack trace, exception, repeat it on multiple lines, etc

I would probably Ctrl+C if I didn't see anything on the screen for a while ...

amritghimire: Sure will do that

src/datachain/studio.py
```diff
@@ -369,48 +371,94 @@ def _get_job_status(client, job_id: str) -> str | None:
...
 
+def _print_reconnect_msg(sleep_sec: float) -> str:
+    msg = f"\r>>>> WebSocket closed, reconnecting in {sleep_sec:.0f}s..."
+    print(msg, end="", flush=True)
+    return msg
+
+
+def _clear_line(msg: str) -> str:
+    if msg:
+        print("\r" + " " * len(msg) + "\r", end="", flush=True)
```
shcheklein, line 382: @ilongin could you review this - you recently had to deal with clearing the lines. Is it good enough?

src/datachain/studio.py
```diff
@@ -369,48 +371,94 @@ def _get_job_status(client, job_id: str) -> str | None:
...
+    logs: list, last_log_id: int, filter_up_to: int
+) -> tuple[bool, int]:
+    received = False
+    for log in logs:
+        log_id = log["id"]
+        if log_id <= filter_up_to:
+            continue
+        last_log_id = max(last_log_id, log_id)
+        received = True
+        print(log["message"], end="")
```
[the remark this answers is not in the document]
amritghimire: Too defensive

tests/test_cli_studio.py
```diff
@@ -1165,6 +1164,100 @@ async def mock_tail_job_logs(jid, no_follow=False):
...
+                    {"id": 2, "message": "line 2\n"},
+                    {"id": 3, "message": "line 3\n"},
+                ]
+            }
+            yield {"job": {"status": "COMPLETE"}}
+
+    mocker.patch(
+        "datachain.studio.StudioClient.tail_job_logs", side_effect=mock_tail_job_logs
+    )
+    mocker.patch("asyncio.sleep")
```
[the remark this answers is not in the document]
amritghimire: In Python 3.8+, unittest.mock.patch auto-detects coroutine functions and creates AsyncMock automatically (confirmed by inspect.iscoroutinefunction(asyncio.sleep) == True). Both new tests pass with the current code.

tests/test_cli_studio.py
```diff
@@ -1165,6 +1164,100 @@ async def mock_tail_job_logs(jid, no_follow=False):
...
+            yield {"logs": [{"id": 1, "message": "output\n"}]}
+            yield {"job": {"status": "RUNNING"}}
+        else:
+            yield {"job": {"status": "RUNNING"}}
+
+    mocker.patch(
+        "datachain.studio.StudioClient.tail_job_logs", side_effect=mock_tail_job_logs
+    )
+    mocker.patch("datachain.studio.RECONNECT_MAX_ATTEMPTS", 1)
+    mocker.patch("asyncio.sleep")
```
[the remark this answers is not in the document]
amritghimire: In Python 3.8+, unittest.mock.patch auto-detects coroutine functions and creates AsyncMock automatically (confirmed by inspect.iscoroutinefunction(asyncio.sleep) == True). Both new tests pass with the current code.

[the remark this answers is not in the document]
amritghimire: In Python 3.8+, unittest.mock.patch auto-detects coroutine functions and creates AsyncMock automatically (confirmed by inspect.iscoroutinefunction(asyncio.sleep) == True). Both new tests pass with the current code.

PR #1653, opened 2026-03-16
Removing transient dependencies from checkpoint hash calculation

### LINKED ISSUE #1641: checkpoints: remove transient dependency between chains
Currently each chain's checkpoint hash includes the hash of the last checkpoint in the current job. This means if a script has two independent chains:

```python
dc.read_storage("s3://a/").filter(...).save("ds1")
dc.read_storage("s3://b/").map(...).save("ds2")
```

`ds2`'s hash depends on `ds1`'s checkpoint. Changing `ds1`'s logic, reordering the chains, or wrapping one in an `if` invalidates `ds2`'s checkpoint even though its own inputs and steps haven't changed.

This transient dependency was meant to handle cases like `read_dataset("d1")` depending on a previously saved `d1`, but that dependency is already captured naturally — `read_dataset("d1")` includes the dataset name/version in its `QueryStep.hash()`.

Proposal: remove `_last_checkpoint_hash` from the hash calculation. Each chain's hash should depend solely on its own starting data and steps.

Benefits:
- Easier to reason about — no surprising invalidations from unrelated chains
- Robust to code reordering, conditionals, and script restructuring
- Enables global checkpoints — if the same chain with the same inputs ran in another job months ago, the checkpoint can be reused

ilongin:
Removes transient checkpoint dependencies from hash calculation. Previously, each chain's checkpoint hash included the hash of the last checkpoint in the current job — meaning changing, reordering, or wrapping chains in `if` blocks would invalidate unrelated chains downstream. Now each chain's hash depends only on its own starting data and steps.

- Removed `_last_checkpoint_hash` and `job_aware` parameter from `DatasetQuery.hash()`. Each chain is now self-contained.
- Removed `get_last_checkpoint` from metastore (no longer needed).
- Changed checkpoint upsert from `on_conflict_do_nothing` to `on_conflict_do_update` to handle re-activation of checkpoints with same hash in same job.
- Added `identity_hash()` to UDF — hashes function name + params + output schema (excludes function code body). This replaces `output_schema_hash()` for partial checkpoint matching. It distinguishes different UDFs while still allowing code-only bug fixes to continue from partial results. For lambdas, uses full code hash since all lambdas share the name `<lambda>`.
- Handle same-hash table collisions within a job: drop stale partial output tables before creating new ones, and drop existing final tables before renaming. This prevents data corruption in edge cases like:

```python
try:
    dc.read_dataset("data").map(process).save("results")
except:
    pass
dc.read_dataset("data").map(process).save("results_v2")
```


src/datachain/query/dataset.py
```diff
@@ -1158,9 +1158,12 @@ def _run_from_scratch(
 
         input_table = self.get_or_create_input_table(query, hash_input, job)
 
-        partial_output_table = self.create_output_table(
-            Checkpoint.partial_output_table_name(run_id, partial_hash),
+        partial_table_name = Checkpoint.partial_output_table_name(run_id, partial_hash)
+        # Drop stale partial table from earlier chain with same hash in this job
```
ilongin, line 1327: This is needed for try/except edge case - when we have 2 identical chains where first is under try/catch which means it will fail, leave partial output table and then another chain will try to create the same one. E.g
```python
  try:     
      dc.read_dataset("data").map(process).save("results")
  except:                                                                                                       
      pass
  dc.read_dataset("data").map(process).save("results_v2")
```

src/datachain/data_storage/metastore.py
```diff
@@ -2427,18 +2431,6 @@ def find_checkpoint(
             return None
         return self.checkpoint_class.parse(*rows[0])
 
-    def get_last_checkpoint(self, job_id: str) -> Checkpoint | None:
```
shcheklein, line 2622: compatibility with the old clusters again?

src/datachain/query/dataset.py
shcheklein: why dropping it though? why not just reuse it?

src/datachain/data_storage/metastore.py
```diff
@@ -2369,10 +2365,18 @@ def get_or_create_checkpoint(
...
 
-            # Use on_conflict_do_nothing to handle race conditions
-            if not hasattr(query, "on_conflict_do_nothing"):
-                raise RuntimeError("Database must support on_conflict_do_nothing")
-            query = query.on_conflict_do_nothing(index_elements=["job_id", "hash"])
+            # Use upsert to handle re-activation of deleted checkpoints
+            # (e.g. same hash in same job after previous run)
+            if not hasattr(query, "on_conflict_do_update"):
+                raise RuntimeError("Database must support on_conflict_do_update")
+            query = query.on_conflict_do_update(
```
shcheklein: how is it aligned with GC? it doesn't seem save ... the assumption in GC if checkpoint going STALE / REMOVING we never reuse it AFAIU

src/datachain/data_storage/metastore.py
ilongin: You are right, I'm returning this until we are 100% sure we can deprecate it.

src/datachain/query/dataset.py
ilongin: The partial table contains rows from a different chain execution (chain A crashed, chain B starts). Chain B processes all rows from scratch  -  if we reuse chain A's table, chain B's rows get appended on top of chain A's stale rows, resulting in duplicate/mixed data. We have a test for this: `test_try_catch_identical_chains_no_stale_partial_data`.

Main thing to notice is that we don't reuse checkpoints from current job - this simplify things a lot in lot of places and I think that's what users would expect.

src/datachain/data_storage/metastore.py
ilongin: GC only expires checkpoints for inactive jobs (done/failed). `get_or_create_checkpoint` is only called by active (running) jobs. So a conflict with an EXPIRED checkpoint for the same job_id can't happen - it would mean GC expired a checkpoint for a currently running job.
The upsert without a WHERE clause is safe.

shcheklein: okay, why are we deleting them inside the job itself then? should we delete them at then end of the job to avoid these custom logic that we should remember about and state transitions

src/datachain/data_storage/metastore.py
shcheklein: let's return in the same place to simplify diff (I don't know if there are changes to them now or not)

src/datachain/lib/dc/datachain.py
```diff
@@ -638,7 +638,7 @@ def save(  # type: ignore[override]
         # Calculate hash including dataset name and job context to avoid conflicts
         import hashlib
 
-        base_hash = self._query.hash(job_aware=True)
```
shcheklein, line 562: just to better understand - why this change?

src/datachain/query/dataset.py
shcheklein: what happens when we run it next time though? will the first chain reuse some checkpoint? is it really expected - they don't share checkpoints here, but do across the jobs - different behavior in different situations?

src/datachain/data_storage/metastore.py
ilongin: The deletion happens when promoting partial → final: UDF completes, we mark the partial as DELETED and create the final checkpoint. We could defer to end-of-job, but if the job crashes after UDF completion and before cleanup, we'd have both a partial and final checkpoint for the same UDF. It's harmless (final takes priority in lookup), but messier. The mid-execution cleanup keeps the state clean at each step.

src/datachain/lib/dc/datachain.py
ilongin: `job_aware=True` included `_last_checkpoint_hash` (the hash of the last checkpoint in the current job) in the hash calculation - this was the transient dependency. Since we're removing transient deps in this PR, the `job_aware` parameter is no longer needed / used. Each chain's hash now depends only on its own operations and starting data.

src/datachain/query/dataset.py
ilongin: On re-run, both chains find the checkpoint from the previous job and both skip -this is correct since same hash means same computation, same result.

The within-job vs across-job difference is intentional: within the same job, each `.save()` is an explicit intent to create a dataset. Across jobs, the purpose of checkpoints is to skip what was already done in a previous run.

BTW there would be a lot of issues / complications if we were reusing checkpoints from the same job. Some of those:
  - Loop case: `for i in range(3): ...save("ds")` would create 1 version instead of 3
  - Partial output table collisions in try/except (chain B writes into chain A's table)
  - No way to distinguish "my own checkpoint" from "another chain's checkpoint with same hash" without ordering - which is why transient deps existed in the first place

Simple rule: checkpoints skip what was done in a previous run, not within the same run.

src/datachain/data_storage/metastore.py
shcheklein: okay, why reactivating partial then if there is a final one? sorry, for the stupid question - I'm a bit lost already in the whole lifecycle of this

src/datachain/lib/dc/datachain.py
shcheklein: are we cleaning it up fully? removing it from hash function itself also?

src/datachain/query/dataset.py
shcheklein: Okay, let's say we have a loop like you mentioned. It fails on the first iteration.

Next time we run it - will all three invocation pick up the same partial checkpoint from the previous job?

src/datachain/data_storage/metastore.py
ilongin: There are no stupid questions and the flow **is** complex, that's a fact (currently I don't have idea how to make it simpler and still cover all use cases and edge cases).

Let go through the try/except flow step by step (this one causes partial hash collision and then we need that fix):

```python
try:
    dc.read_dataset("data").map(process).save("ds1")  # chain A
except:
    pass
dc.read_dataset("data").map(process).save("ds2")  # chain B
 ```

Quick note on the two hashes:
  - `partial_hash` = chain flow + UDF identity (function name + params + output schema). Excludes function code body.
  - `hash_output` = chain flow + full UDF hash (function name + params + output schema + function code body).

Both chains have the same **partial_hash**. They may or may not have the same **hash_output** depending on whether the function body differs — but
it doesn't matter here since the conflict is on **partial_hash**.

1. Chain A starts → creates partial checkpoint (hash=partial_hash, status=ACTIVE)
2. Chain A crashes mid-UDF
3. Chain B starts → `get_or_create_checkpoint(job_id, partial_hash, partial=True)` → hits **on_conflict_do_update** → updates existing partial (resets `created_at`, keeps ACTIVE)
4. Chain B completes → deletes partial (DELETED), creates final checkpoint (hash=hash_output)

The `on_conflict_do_update` in step 3 is needed because the partial from chain A already exists with the same (job_id, hash). Without it, the insert would fail and chain B wouldn't get a valid checkpoint reference back.

There's never a situation where we reactivate a partial alongside an existing final - the final doesn't exist yet at that point, and they use different hashes anyway.

File: src/datachain/lib/dc/datachain.py (hunk 4 above)
Comment (ilongin) [reply src/datachain/lib/dc/datachain.py:562]: Yes, fully cleaned up. `job_aware` parameter is removed from `DatasetQuery.hash()` signature as well, along with `_last_checkpoint_hash`
property that it used.

File: src/datachain/query/dataset.py (hunk 1 above)
Comment (ilongin) [reply src/datachain/query/dataset.py:1327]:
```python
for i in range(3):
    dc.read_storage("s3://a/").map(process).save("my_dataset")
```
Let's say on first run it crashed in process function of loop iteration 1 and we fix bug and re-run script.

Previous job has one partial checkpoint (hash=P) from iteration 0's crash. All three iterations have the same partial_hash=P.

1. Iteration 0: searches previous job → finds partial P → continues from it → completes → marks previous job's partial P as DELETED → creates final F in current job
2. Iteration 1: searches previous job → partial P was just deleted by iteration 0 → finds nothing → runs from scratch
3. Iteration 2: same as iteration 1 → runs from scratch

Only the chain that crashed in the previous job benefits from partial recovery. The others have no checkpoint to reuse and run normally.

Note: we are modifying the previous job's checkpoint (marking it DELETED) from the current job. This is the existing behavior for partial → final promotion in `_continue_udf`.

shcheklein: First iteration will pick up an partial checkpoint from any of those (1, 2, 3 - depends on which one failed), right?

src/datachain/data_storage/metastore.py
shcheklein: so, if we restart the job the partial checkpoint for the first chain A is lost (overwritten?)

tbh, more I look into this across these PRs more it seems to me it is easier to allow reusability within the job - same chain? reused - no complicated logic, etc - no matter same job or not ...

src/datachain/query/dataset.py
ilongin: Yes. All iterations produce the same `partial_hash`, so whichever runs first on re-run picks up the partial checkpoint. In the example where iteration 0 succeeds and iteration 1 fails:

On re-run, iteration 0 finds the final checkpoint (from its successful completion in the previous job) and skips the UDF entirely.
Iteration 1 has no final checkpoint — finds the partial and continues from it. Iteration 2 finds nothing (partial was deleted by iteration 1) and runs from scratch.

If all iterations are identical (same UDF, same input), iteration 0's final checkpoint would also match iterations 1 and 2 (same `hash_output`), so all would skip the UDF.

src/datachain/data_storage/metastore.py
ilongin: Chain B's `on_conflict_do_update` takes over chain A's partial checkpoint record and the stale partial output table is dropped and recreated fresh. On re-run, the partial reflects chain B's state.

The try/except case is an edge case but it still works correctly — no data corruption, no crashes. Chain B gets a clean start despite sharing the same hash.

Regarding within-job reuse — there are some issues with it:
1. `_skip_udf` would try to copy an output table onto itself (same job_id + same hash = same table name)
2. `_continue_udf` would build on partial data from a different chain's crashed execution
3. Loop case: `for i in range(3): ...save("ds")` creates 1 version instead of 3

Also, I think checkpoints are naturally a cross-run concept  - "did I already do this in a previous run?" Within the same run, each `.save()` / `.map()` call is an explicit action the user wants to happen.

shcheklein: my point is that even between jobs we can't predict which chain is using which checkpoint, is it correct?

ilongin: True, yes, technically all those 3 points I mentioned above can happen in situation of re-using checkpoints from previous job...with removing transient dependencies in hash calculation all kind of weird situations can occur (that's why I liked that in the first place, regardless of possible checkpoint invalidation because of the code flow).
But still I feel like it's wrong to re-use checkpoints from the same job, idk.

ilongin:
## Checkpoint Reuse Scope

**Context**: After removing transient hash dependencies, identical chains produce the same checkpoint hash regardless of their position in the script. This raises the question of checkpoint reuse scope.

The initial discussion was whether to reuse checkpoints from the **previous job only** (current PR) or from the **current job + previous job**. But this is really a broader question about what checkpoints fundamentally are. Allowing within-job reuse means we're no longer just recovering from a failed previous run — we're caching computation results. And if we accept that, there's no principled reason to stop at "current + parent job." Any checkpoint with a matching hash represents the same completed computation, regardless of which job produced it. Checkpoints could be decoupled from jobs entirely.

This leads to two mental models:

- **Recovery**: "Resume from where my previous run left off." Checkpoints are scoped to the job lineage (parent job). Each run executes everything; checkpoints only help when re-running after failure.
- **Cache**: "Has this computation been done before? Skip it." Checkpoints are keyed by hash alone. If the hash matches, the result is reused regardless of provenance.

However, as analyzed below, the cache model is fundamentally incompatible with how UDF partial checkpoints work — making the recovery model the only viable approach.

---

### Why the cache model is not viable: UDF partial checkpoint design

The UDF recovery mechanism intentionally excludes the function code body from the partial checkpoint hash:

```python
# line 1146 of query/dataset.py
partial_hash = sha256(hash_input + udf.identity_hash() + hash_column_elements(partition_by))
```

`identity_hash()` = function name + params + output schema. **No function body.** This is by design — when a user fixes a bug and re-runs, the partial_hash stays the same so they can continue processing remaining rows from where the previous run left off.

This design **requires** job lineage scoping. Without it, a global cache produces data corruption:

```python
# Job A (user's first attempt)
def process(x): return summarize(x)    # crashes at 50%

# Job B (completely different user/script, same signature)
def process(x): return translate(x)
```

Same function name, same params, same output schema → same `partial_hash`. In a global cache, Job B finds Job A's partial and continues from it. Result: 50% summarized rows (from A) + 50% translated rows (from B). **Silent data corruption.**

This works safely in the recovery model because the parent job chain guarantees the partial came from the same script — the user is re-running their own code with a bug fix, not a completely different function.

**Final checkpoints don't have this problem** — `hash_output` includes the full UDF code hash, so different function bodies produce different hashes. A global cache for finals would be hash-safe. But a hybrid approach (global cache for finals, job-scoped for partials) adds complexity without clear benefit and mixes two incompatible mental models.

---

### Approach A: Recovery model (current PR)

Checkpoints are only reused from the **parent job** (immediate previous run of the same script). Within the current job, each chain runs fresh.

**How it works:**
- `_find_udf_checkpoint` searches only `job.rerun_from_job_id` (parent)
- Within the same job: same-hash collisions are resolved by drop-before-create
- Checkpoint insert uses `on_conflict_do_update` (upsert) for within-job hash collisions

**What works well:**
- Every `.save()` / `.map()` in a run executes as the user wrote it — no surprise skips
- Loop creates correct number of versions:
  ```python
  for i in range(3):
      dc.read_storage("s3://a/").map(process).save("ds")
  # → 3 versions created ✓
  ```
- try/except: chain B gets clean state, no risk of building on chain A's crashed partial:
  ```python
  try:
      dc.read_dataset("data").map(process).save("ds1")  # chain A - crashes mid-UDF
  except:
      pass
  dc.read_dataset("data").map(process).save("ds2")  # chain B - runs from scratch, safe
  ```
- Compatible with UDF bug-fix recovery: partial_hash excludes function body, job scoping ensures the partial came from the same script
- Simple mental model: "checkpoints replay previous runs, current run always executes"

**Issues / downsides:**
- Needs drop-before-create in 4 places (partition table, partial output, final output, `_skip_udf` output) — more complex implementation
- try/except wasted work — chain B re-does chain A's exact computation:
  ```python
  dc.read_dataset("data").map(process).save("ds1")  # chain A - succeeds
  dc.read_dataset("data").map(process).save("ds2")  # chain B - reruns same UDF from scratch
  ```
- Two different behaviors for the same hash (within-job: re-execute, cross-job: skip)

**Cross-job edge cases (still present):**
- Loop crash and re-run — all iterations match the same final checkpoint:
  ```python
  for i in range(3):
      dc.read_storage("s3://a/").map(process).save("ds")
  # Run 1: iteration 0 succeeds, iteration 1 crashes
  # Run 2: all 3 iterations find iteration 0's final → all skip UDF
  ```
- try/except re-run — both chains find chain B's final from previous job:
  ```python
  try:
      dc.read_dataset("data").map(process).save("ds1")  # chain A
  except:
      pass
  dc.read_dataset("data").map(process).save("ds2")  # chain B
  # Run 1: chain A crashes, chain B completes
  # Run 2: both find chain B's final → both skip UDF
  ```

---

### Summary

| | Recovery (Approach A) | Cache (Approach B) |
|---|---|---|
| Scope | Parent job only | Any matching hash |
| UDF bug-fix recovery | Safe (job lineage guarantees same script) | **Unsafe (different function bodies can share partial_hash)** |
| Loop (3 iterations) | 3 versions ✓ | 1 version ✗ |
| try/except (A crashes) | B runs from scratch (safe, wasteful) | B reuses A's partial (risky) |
| try/except (A succeeds) | B re-executes (wasteful) | B skips (efficient) |
| Grandparent checkpoint | Lost if parent has none | Always reachable |
| Implementation complexity | 4 drop-before-create sites, upsert | Self-copy fix, GC rethink |
| Concurrent jobs | Safe (job-scoped tables) | Race conditions on shared partials |
| Mental model | "Replay failed runs" | "Skip duplicate computation" |

shcheklein:
okay, can we outline all the alternatives really? 

- we can detect and raise if collision happens within a job
- we can reuse hashes within the job and outside the job
- anything else?

let's get a complete picture with pros and cons

dmpetrov:
> ```python
> # line 1146 of query/dataset.py
> partial_hash = sha256(hash_input + udf.identity_hash() + hash_column_elements(partition_by))
> ```
> 
> `identity_hash()` = function name + params + output schema. **No function body.**

That's not enough. It also depends on it's previous fails.

dmpetrov:
Here is a summary of my discussion with @ilongin with some of my afterthoughts.

### Terminology

- **Checkpoint** - only for failed UDF (currently "partial checkpoint")
- **Chain cache** - reuse of computed steps (currently "final checkpoint")
     - **Cached datasets** - a special case of chain cache when dataset was explicitly saved and can be reused fully.

We should stop mixing these concepts. It's hard for us to discuss, it's hard to communicate with users (#1716).

### Hashing differences

These are fundamentally different systems and should be treated as such:
- **Checkpoint**
    - Hash does not include the function
    - Used only to resume execution after failure
- **Chain cache**
    - Hash does include the function
    - Used to reuse previously computed results
    - Important: recovered UDF has "rotten" chain hash - the final chain hash must include all failed attempts see: https://github.com/datachain-ai/datachain/pull/1653#issuecomment-4225917898 Ideally we should keep track of all failed functions for reproducibility of "rotten" steps.

### Scope and lifecycle

Checkpoint hashes and chain cache hashes have different meaning and lifecycle - not just different hash composition.

- **Checkpoint hashes**
    - Very limited scope. Likely scoped to user + script. Goal: fix a function and resume work
- **Chain cache hashes**
    - Broader scope, but needs clear boundaries.
    - Options for the boundaries (**This looks like the main discussion point now**):
        - Scope by script → too narrow
        - Scope by user → maybe, but still limiting
        - Time-based → usually not ideal
        - Fully global → powerful but potentially unsafe

shcheklein:
> We should stop mixing these concepts.

yep. It might be hard to do in code though.

>  recovered UDF has "rotten" chain hash - the final chain hash must include all failed attempts see:

@dmpetrov add more context pls. it is not exactly clear.

I'm not sure if the last comment also addresses this PR and the discussion tbh. Unless I'm missing something, I would still appreciate full review of actual practical options on actuals scenarios. I think a few options are missing from the table. I'm fine to do additional naming corrections and other reasonable corrections on top of it though.

dmpetrov:
> I'm not sure if the last comment also addresses this PR and the discussion tbh.

What I'm saying - we’re comparing things that aren’t actually alternatives here. Approach A is about recovery/checkpoints, while B is about caching.

These should be use together. And the only question is the scope of each (triggers):
- checkponts / Approach A - I don't think using parent job is a right thing. user+script name is much more simpler way of solving this.
- chain cache / Approach B - fully global cache of chains looks promising. We should try to find limitations/corner cases (so far all the corner cases were related to checkpoints, not chain caches).

shcheklein:
> These should be use together. And the only question is the scope of each (triggers):

I'm still lost tbh. It seems a completely separate discussion to me.

ilongin:
> > These should be use together. And the only question is the scope of each (triggers):
> 
> I'm still lost tbh. It seems a completely separate discussion to me.

I've added changes we discussed. Added some more tests and tested it manually, seem to work

shcheklein:
@ilongin tests fail?

src/datachain/data_storage/metastore.py
```diff
@@ -2561,10 +2561,18 @@ def get_or_create_checkpoint(
                 status=CheckpointStatus.ACTIVE,
             )
 
-            # Use on_conflict_do_nothing to handle race conditions
-            if not hasattr(query, "on_conflict_do_nothing"):
-                raise RuntimeError("Database must support on_conflict_do_nothing")
-            query = query.on_conflict_do_nothing(index_elements=["job_id", "hash"])
+            # Use upsert to handle re-activation of deleted checkpoints
+            # (e.g. same hash in same job after previous run)
+            if not hasattr(query, "on_conflict_do_update"):
```
shcheklein: do we really have a case like this?

src/datachain/data_storage/metastore.py
```diff
@@ -2561,10 +2561,18 @@ def get_or_create_checkpoint(
...
+                raise RuntimeError("Database must support on_conflict_do_update")
+            query = query.on_conflict_do_update(
+                index_elements=["job_id", "hash"],
+                set_={
+                    "partial": partial,
```
shcheklein: does it mean that partial status can change?

src/datachain/data_storage/metastore.py
```diff
@@ -2561,10 +2561,18 @@ def get_or_create_checkpoint(
...
+                    "status": CheckpointStatus.ACTIVE,
```
shcheklein, line 2565: could you remind please -when does it become inactive?

can it happen that we'll update active partial to non-partial by mistake?

src/datachain/lib/dc/datachain.py
```diff
@@ -645,12 +645,29 @@ def _resolve_checkpoint(
...
 
-        if (
-            self._checkpoints_enabled
-            and self.job.rerun_from_job_id
-            and not ignore_checkpoints
-            and metastore.find_checkpoint(self.job.rerun_from_job_id, job_hash)
+        if not self._checkpoints_enabled or ignore_checkpoints:
+            return None
+
+        # delta_retry reads the output's current state (error rows, missing rows)
```
shcheklein, line 654: is it a separate issue? how / why was it working before for delta_retry?

src/datachain/lib/udf.py
```diff
@@ -227,13 +226,39 @@ def hash(self) -> str:
...
         ).hexdigest()
 
-    def output_schema_hash(self) -> str:
-        """Hash of just the output schema (not including code or inputs).
+    def identity_hash(self) -> str:
+        """Hash of UDF identity: function name, inputs, and output schema.
 
-        Used for partial checkpoint hash to detect schema changes while
-        allowing code-only bug fixes to continue from partial results.
+        Excludes function code body so users can fix bugs and continue from
```
shcheklein, line 232: didn't we have a hash like this before (function w/o body) ... why adding it here? was it moved from somewhere?

src/datachain/query/dataset.py
```diff
@@ -885,6 +886,12 @@ def create_partitions_table(
             p.get_column() if isinstance(p, Function) else p for p in list_partition_by
         ]
 
+        # Drop stale partition table from earlier chain with same hash in this job
+        if table_name:
```
shcheklein, line 890: can there be a case w/o table name here?

src/datachain/data_storage/metastore.py
ilongin: Not needed, indeed. Not sure why we started doing that but it's really old code. Removed it here and in other places in metastore

src/datachain/data_storage/metastore.py
ilongin: No, `partial_hash` and `hash_output` are different hashes, so combo `(job_id, hash)` always has a fixed `partial` value and the update was a no-op. I removed `partial` from `set_` to make that explicit and safer.

src/datachain/data_storage/metastore.py
ilongin: **When does it become inactive?**
   - `DELETED`: explicit transition at `dataset.py:1362` — when a partial checkpoint is promoted to a final one after the UDF completes. 
   - `EXPIRED`: via GC, and only for completed/failed jobs — never for an active job.                                                         
                           
 **Can the upsert flip an ACTIVE partial to non-partial by mistake?**
 With the fix above (dropping `partial` from `set_`), no - we don't touch that field on conflict anymore.

src/datachain/lib/dc/datachain.py
ilongin: Old code could already break `delta_retry` — re-running a successful script would hit the previous-job checkpoint and return the cached dataset, skipping the chain entirely and with it the output-state read that `delta_retry` needs.                                              
                             
This PR just makes it fire more often. The bypass fixes the pre-existing bug.

src/datachain/lib/udf.py
ilongin: On main it was `output_schema_hash()` - just the output schema. This PR replaces it with `identity_hash()`, which adds function name + input  schema (still excluding the body).                                     
                                                                         
So it's not a move, but an expansion: the old hash was too narrow  - two different  
UDFs with the same output schema could collide on the partial-checkpoint key.

src/datachain/lib/dc/datachain.py
```diff
@@ -568,7 +568,10 @@ def save(  # type: ignore[override]
         result = self._resolve_checkpoint(name, project, _hash, kwargs)
         if bool(result):
             # Checkpoint was found and reused
-            print(f"Checkpoint found for dataset '{name}', skipping creation")
+            print(
```
shcheklein, line 571: should we use logger DEBUG or something then? do we have verbose key for this already in CLI for example?

or what is the reason to change it to stderr here?

src/datachain/lib/udf.py
shcheklein: okay, but how about code reusability - it feels I saw it already somewhere - all these lmbda calls, etc ... is there a helper for calculate it including the body for example? can it be reused here?

src/datachain/lib/dc/datachain.py
```diff
@@ -645,12 +648,29 @@ def _resolve_checkpoint(
...
+            for s in self._query.delta_sources()
+        ) or (
+            self._query.delta_spec is not None and self._query.delta_spec.delta_retry
         ):
+            return None
+
+        # Search current job first, then previous job — "done" save checkpoints
+        # act as a cache keyed by hash.
+        found = any(
+            metastore.find_checkpoint(j, job_hash)
```
shcheklein: why it it called job hash? is it actually a chain hash?

src/datachain/query/dataset.py
```diff
@@ -1321,9 +1331,12 @@ def _run_from_scratch(
 
         input_table = self.get_or_create_input_table(query, hash_input, job)
 
-        partial_output_table = self.create_output_table(
-            Checkpoint.partial_output_table_name(run_id, partial_hash),
+        partial_table_name = Checkpoint.partial_output_table_name(run_id, partial_hash)
+        # Drop stale partial table from earlier chain with same hash in this job
+        self.warehouse.db.drop_table(
```
shcheklein, line 1328: just going back here - won't it break reusability - us dropping these tables?

I think we wanted to reuse the across loop as well

tests/func/checkpoints/test_checkpoint_events.py
```diff
@@ -184,17 +184,17 @@ def buggy_double(num) -> int:
     reset_session_job_state()
     processed.clear()
 
-    def fixed_double(num) -> int:
+    def buggy_double(num) -> int:
```
shcheklein: why these changes in this file?

tests/func/checkpoints/test_checkpoint_recovery.py
```diff
@@ -75,17 +75,18 @@ def process_buggy(num) -> int:
 
     processed_nums.clear()
 
-    def process_fixed(num) -> int:
+    # Use same function name so identity_hash matches (simulates fixing the bug)
+    def process_buggy(num) -> int:
```
shcheklein, line 79: why is it buggy?

tests/func/test_catalog.py
```diff
@@ -529,7 +529,11 @@ def test_dataset_stats(test_session):
     assert dataset_version2.size == 18
 
 
-def test_ls_datasets_ordered(test_session):
+def test_ls_datasets_ordered(test_session, monkeypatch):
+    # Disable checkpoint cache — this test verifies listing order across
+    # multiple versions, so each save must create a new version.
+    monkeypatch.setenv("DATACHAIN_IGNORE_CHECKPOINTS", "true")
```
shcheklein: aren't we creating new versions each time for read_values based chains anyway?

tests/func/test_datachain.py
```diff
@@ -395,7 +395,12 @@ def test_show_without_temp_datasets(capsys, test_session):
     assert "Empty result" in normalized_output
 
 
-def test_save(test_session):
+def test_save(test_session, monkeypatch):
+    # Disable checkpoint cache — this test verifies metadata updates across
+    # multiple saves, so each save must create a new version rather than
+    # returning the cached first version.
+    monkeypatch.setenv("DATACHAIN_IGNORE_CHECKPOINTS", "true")
```
shcheklein: same here - not sure I understand this change atm

src/datachain/query/dataset.py
```diff
@@ -2477,34 +2508,19 @@ def _starting_step_hash(self) -> str:
             )
         return ""
 
-    @property
-    def _last_checkpoint_hash(self) -> str | None:
-        if not self.checkpoints_enabled:
-            return None
-        job = self.session.get_or_create_job()
-        last_checkpoint = self.catalog.metastore.get_last_checkpoint(job.id)
```
shcheklein, line 2485: is get_last_checkpoint used anywhere now?

src/datachain/data_storage/metastore.py
```diff
@@ -2561,10 +2557,15 @@ def get_or_create_checkpoint(
...
             )
 
-            # Use on_conflict_do_nothing to handle race conditions
-            if not hasattr(query, "on_conflict_do_nothing"):
-                raise RuntimeError("Database must support on_conflict_do_nothing")
-            query = query.on_conflict_do_nothing(index_elements=["job_id", "hash"])
+            # Use upsert to handle re-activation of deleted checkpoints
+            # (e.g. same hash in same job after previous run)
+            query = query.on_conflict_do_update(  # type: ignore[attr-defined]
+                index_elements=["job_id", "hash"],
```
shcheklein, line 2559: so, what happens if partial=False is requested but partial=True is returned? or otherwise around?

src/datachain/lib/udf.py
```diff
@@ -227,13 +226,39 @@ def hash(self) -> str:
...
+            is_lambda = func_to_hash.__name__ == "<lambda>"
+        except AttributeError:
+            is_lambda = False
+
+        if is_lambda:
+            # Lambdas all share '<lambda>' name, use full code hash instead
+            name_part = hash_callable(func_to_hash)
+        else:
+            # Named functions/classes: use qualified name for identity
+            qualname = getattr(func_to_hash, "__qualname__", "") or ""
```
shcheklein: should we also add module here? `__module__ + __qualname__?

src/datachain/lib/udf.py
```diff
@@ -227,13 +226,39 @@ def hash(self) -> str:
...
+        partial checkpoints. Includes function name to distinguish different
+        UDFs with the same signature. For lambdas, includes the full code hash
+        since all lambdas share the name '<lambda>' and can't be meaningfully
+        "fixed" without becoming a different lambda.
         """
-        return self.output.hash()
+        func_to_hash = self._func or self.process
+
+        try:
+            is_lambda = func_to_hash.__name__ == "<lambda>"
```
shcheklein: is there a case when __name__ is not present? is it always available on callable?

src/datachain/lib/udf.py
ilongin: Yes, it was true that there were couple of similar functions there so I unified it. `hash_callable` now takes `include_body: bool =  True`, and `UDFBase.identity_hash` is gone - the single                    `UDFBase.hash(include_body=...)` covers both uses. 

I also dropped `annotations` from the `hash_callable` extras since they don't really make sense in terms of hashing callable and in UDF context they're already captured by `params.hash()` / `output.hash()` at the caller.

src/datachain/lib/dc/datachain.py
ilongin: The print was on stdout before, and `test_query_e2e` scrapes stdout to assert on program output - the status line was leaking in and breaking it so honestly it was primarily motivated by test fix.

BUT stderr is also just the semantically right  place for a status message ("checkpoint hit, skipping") - it's not program output:                                                                      
                           
- **Piping** (`datachain ... | something`): previously leaked into the  downstream consumer; now stays out of the pipe.
- **Redirecting stdout** (`datachain ... > out.txt`): previously ended up in the file; now stays on the terminal where status messages belong.            
 
It's still visible in a normal terminal since both streams render there

Why not `logger.info`: in scripts that don't configure logging, it'd be silently dropped (default level is WARNING).  And Studio's `_run_python` captures subprocess stdout/stderr directly, so prints show up in the job output UI while logger output would not.

src/datachain/lib/dc/datachain.py
ilongin: renamed to `chain_hash`

src/datachain/query/dataset.py
ilongin: Doesn't affect reusability — reuse happens in `_skip_udf` (finds a final checkpoint earlier in the same job, e.g. across a loop) and `_continue_udf` (finds a partial). We only reach `_run_from_scratch` after confirming neither exists, so in theory no partial table should exist at this name either. 
The drop is a defensive cleanup for the edge case where a partial checkpoint was 
removed but its table lingered — it would block `create_output_table` with "already exists."

Cross-loop reuse within a job is covered by                                  `test_loop_same_chain_creates_one_version`.

tests/func/checkpoints/test_checkpoint_events.py
ilongin: The rename (`fixed_double` → `buggy_double`) is required because           `partial_hash` now includes `__qualname__` (via `udf.hash(include_body=False)`). A different function name → different partial_hash → no partial match → we'd run from scratch and the  `UDF_CONTINUED` assertion would fail. On main, the partial-identity hash was just `output_schema_hash()` so the name didn't matter; now the name is part of the identity (body can still change — that's the whole point of `include_body=False`).

tests/func/checkpoints/test_checkpoint_recovery.py
ilongin: Same reason as above - `partial_hash` now includes func name, so the second run has to keep the same function name (`process_buggy`) to match the partial from the first run. Left the name as-is; the body is what changes (fix applied)

tests/func/test_catalog.py
ilongin: You're right that `read_values` is non-deterministic across calls (it generates a fresh uuid per call). The issue here is that `chain =  dc.read_values(...)` is called **once** and the same chain object is reused across all `.save()` calls — so `chain.save("cats")` four times produces the same `chain_hash` each time in this case, and calls 2–4 hit the checkpoint cache (returning the cached version instead of creating a new one). The monkeypatch is needed to bypass that.

tests/func/test_datachain.py
ilongin: Same reason as the `test_ls_datasets_ordered` change — the chain is built once and saved twice with the same name. Both `.save(name="new_name", ...)`  calls produce the same `chain_hash`, so the second one hits the checkpoint cache

src/datachain/query/dataset.py
ilongin: No usage, but I think we should leave it for backward compatibility of the clusters just in case, not to hit those issues again...

src/datachain/data_storage/metastore.py
ilongin: Can't happen  - there's a `UniqueConstraint("job_id", "hash")`on checkpoints table,  so only one row per `(job_id, hash)` can exist. Also, callers always pair a given hash with a fixed `partial` value (`partial_hash` → partial=True, `hash_output` → partial=False), so the existing row's partial will always  match what's being requested. If it somehow didn't, `find_checkpoint(...,partial=...)` right after the upsert would return None and raise the `RuntimeError` at line 2573.

src/datachain/lib/udf.py
ilongin: Make sense, added!

src/datachain/lib/udf.py
ilongin: Not always — `functools.partial` objects, Mocks with `__name__` deleted, and some C-extension/dynamically-generated callables don't have it.              
  
It's used in two places in `hash_callable` and both handle the missing case: 
- Lambda detection: `try/except AttributeError` → falls through to `is_lambda = False` (safe default). 
- Extras: `getattr(func, "__name__", "")` → defaults to empty string.
                                                                             
For callable objects (instances with `__call__`), we redirect to `func.__call__` at the top of the function before either check runs, and  `__call__` always has `__name__`.

src/datachain/query/dataset.py
```diff
@@ -989,30 +996,31 @@ def _find_udf_checkpoint(
...
-                "hash=%s from job_id=%s",
-                self._udf_name,
-                self._job_id_short(job),
-                self._run_group_id_short(job),
-                "partial " if partial else "",
-                _hash[:8],
-                checkpoint.job_id,
-            )
-            return checkpoint
+            ):
```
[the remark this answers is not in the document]
ilongin: Not a regression — on main `_find_udf_checkpoint` only searched `rerun_from_job_id` (immediate parent). This PR actually widened the search   to also include the current job. Single-hop UDF-step rerun search is pre-existing behavior.

src/datachain/data_storage/metastore.py
shcheklein: why do we pass partial at all then? why we need this flag then?

src/datachain/query/dataset.py
shcheklein: yep, okay, let's remove it after - create a ticket / item in the checkpoints followup to track it

tests/func/test_catalog.py
shcheklein: okay, 

should we create a fixture or decorator for this? @checkpoints(False) or something

tests/func/checkpoints/test_checkpoint_events.py
shcheklein: I see - let's do a neutral name then? otherwise it is a bit confusing (you kinda expect this function to also fail / have a bug)

shcheklein:
@ilongin please check test coverage, we need to have meaningful tests for all this .. or sometimes it is obvious that if line is not covered you don't really need it

I resolved most of the comments - just a few left

tests/func/checkpoints/test_checkpoint_events.py
ilongin: Renamed

tests/func/test_catalog.py
ilongin: Created fixture `ignore_checkpoints`

src/datachain/query/dataset.py
ilongin: https://github.com/datachain-ai/datachain/issues/1729

src/datachain/data_storage/metastore.py
ilongin: True, we don't need it in `find_checkpoint()` any more - I will remove it from that method

shcheklein (approved):

PR #1656, opened 2026-03-16
Using listing dataset `uuid` for hashing instead of listing URI

### LINKED ISSUE #1655: Fix read_storage hash computed before listing resolves
Extracted from #1639 (point 1).

`read_storage()` computes its starting hash via `_starting_step_hash` before `apply_listing_pre_step()` runs. This means the hash is based on the bare listing dataset name (e.g. `lst__s3://my-bucket`) instead of the actual dataset version UUID. As a result, `update=True` with new files arriving produces the same hash — stale checkpoints are reused and new data is silently ignored.

Fix: call `apply_listing_pre_step()` in `DatasetQuery.hash()` so the listing is resolved before hash computation. This is safe because `apply_listing_pre_step()` is idempotent — if already resolved, it's a no-op. When `apply_steps()` runs later, it skips the already-resolved listing.

ilongin:
`DatasetQuery.hash()` was computing the starting hash before resolving the listing dataset, so `read_storage` chains always hashed the bare listing name (e.g. `lst__s3://my-bucket`) instead of the actual dataset version UUID. This meant `update=True` with new files arriving would produce the same hash and stale checkpoints would be reused.

Fix: call `apply_listing_pre_step()` at the start of `hash()` so the listing is resolved first. This is safe because `apply_listing_pre_step()` is idempotent — if already resolved, it skips. When `apply_steps()` runs later, the listing is already done.

Closes #1655

### DIFF
--- a/src/datachain/query/dataset.py
+++ b/src/datachain/query/dataset.py
@@ -2406,7 +2406,9 @@ class DatasetQuery:
             listing_ds = None
             try:
                 listing_ds = self.catalog.get_dataset(
-                    self.list_ds_name, include_incomplete=False
+                    self.list_ds_name,
+                    versions=None,
+                    include_incomplete=False,
                 )
             except DatasetNotFoundError:
                 pass
@@ -2415,7 +2417,9 @@ class DatasetQuery:
                 assert self.listing_fn
                 self.listing_fn()
                 listing_ds = self.catalog.get_dataset(
-                    self.list_ds_name, include_incomplete=False
+                    self.list_ds_name,
+                    versions=None,
+                    include_incomplete=False,
                 )
 
             # at this point we know what is our starting listing dataset name
@@ -2971,6 +2975,7 @@ class DatasetQuery:
                                     dep.name,
                                     namespace_name=dep.namespace,
                                     project_name=dep.project,
+                                    versions=[dep.version],
                                     include_incomplete=False,
                                 ),
                                 dep.version,
@@ -3026,6 +3031,7 @@ class DatasetQuery:
                     name,
                     namespace_name=project.namespace.name,
                     project_name=project.name,
+                    versions=[version],
                     include_incomplete=True,
                 ).has_version(version)
             ):
@@ -3101,6 +3107,7 @@ class DatasetQuery:
                                 dep.name,
                                 namespace_name=dep.namespace,
                                 project_name=dep.project,
+                                versions=[dep.version],
                                 include_incomplete=False,
                             ),
                             dep.version,
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,7 +24,7 @@ repos:
       - id: trailing-whitespace
         exclude: '^LICENSES/'
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: 'v0.15.7'
+    rev: 'v0.15.8'
     hooks:
       - id: ruff
         args: [--fix, --exit-non-zero-on-fix]
--- a/src/datachain/catalog/catalog.py
+++ b/src/datachain/catalog/catalog.py
@@ -5,8 +5,7 @@ import os.path
 import posixpath
 import sys
 import time
-import traceback
-from collections.abc import Callable, Iterable, Iterator, Sequence
+from collections.abc import Iterable, Iterator, Sequence
 from contextlib import contextmanager, suppress
 from copy import copy
 from dataclasses import dataclass
@@ -66,6 +65,7 @@ if TYPE_CHECKING:
     from datachain.data_storage import AbstractMetastore, AbstractWarehouse
     from datachain.dataset import DatasetListVersion
     from datachain.job import Job
+    from datachain.lib.dc.datachain import DataChain
     from datachain.lib.listing_info import ListingInfo
     from datachain.listing import Listing
     from datachain.remote.studio import StudioClient
@@ -77,7 +77,6 @@ DEFAULT_DATASET_DIR = "dataset"
 CHECKPOINTS_TTL = 4 * 60 * 60
 
 INDEX_INTERNAL_ERROR_MESSAGE = "Internal error on indexing"
-DATASET_INTERNAL_ERROR_MESSAGE = "Internal error on creating dataset"
 # exit code we use if query script was canceled
 QUERY_SCRIPT_CANCELED_EXIT_CODE = 11
 # exit code we use if the job is already in a terminal state (failed/canceled elsewhere)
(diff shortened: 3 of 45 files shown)


src/datachain/query/dataset.py
```diff
@@ -2329,6 +2329,10 @@ def hash(self, job_aware: bool = False) -> str:
         Args:
             job_aware: If True, includes the last checkpoint hash from the job context.
         """
+        # Resolve listing dataset if needed so since we need listing dataset uuid
+        # to calculate hash if read_storage() was used
+        self.apply_listing_pre_step()
```
shcheklein: looks like a hack ... why can't we then resolve it right away on init?

ilongin: It can't be moved to `__init__` — `listing_fn` is set by `read_storage()` after `DatasetQuery.__init__` returns, so it's not available at init time. Refactoring to pass it into init is possible but would couple listing-specific logic into the general-purpose constructor (also used by `read_dataset`, `read_values`, etc.).                                                                                                           
   
But more importantly, we want listing to stay lazy - `read_storage("s3://bucket")` shouldn't immediately list a bucket with millions of files just because the user created a chain. `hash()` is the earliest point where we actually need the resolved listing, and `apply_listing_pre_step()` is idempotent so calling it there is safe.

shcheklein: okay, all this naming for pre_listing steps is hack in the first place tbh ...

I think delta (new implementation needs to be fixed) - I'll take care of this

also, can it be the case that we don't even need listings (just some weird branch in the query), but we are still calculating the hash?

also, can someone by mistake try to get the hash before execution? w/o realizing even that it is now an expensive operation?

what I think is that we should have some separate code that explicitly triggers this transformation in the right place ... and before that we should be raising

it doesn't solve though the issue of calculating all listings, even those that we don't really need AFAIU

ilongin: Did couple of changes:
  - Removed `DataChain.hash()` public method - no use case for it for now, and it wasn't documented anyway. It was also dangerous as it was triggering listing which is heavy operation.
  - Renamed `apply_listing_pre_step()` to `resolve_listing()`

I tried removing resolve_listing from hash but it's not that easy. hash() is called recursively on sub-queries (inside union/join/subtract), so each query needs to resolve its own listing - the caller can't know about all nested sub-queries.
  
Honestly code is still not perfect but it's really constraint by our lazy listing feature which is mandatory IMO. Also, it would probably be better if we would finaly unify `DatasetQuery` and `DataChain` classes - I might do that soon as "side" project.

src/datachain/query/dataset.py
```diff
@@ -2329,6 +2332,8 @@ def hash(self, job_aware: bool = False) -> str:
         Args:
             job_aware: If True, includes the last checkpoint hash from the job context.
         """
+        self.resolve_listing()
```
shcheklein: hmm, but we still call resolve listing inside hash?

ilongin: Yes, I think it's better this way. Alternative is to call resolve_listing() from outside which moves responsibility to the caller to know this internals. Also it complicates with union/merge as I explained in above query https://github.com/datachain-ai/datachain/pull/1656#discussion_r2952642774 . 
Point is: this is all now internal code/methods and I think we should keep the logic encapsulated. Main fix here, which would significantly improve all codebase, is to remove DatasetQuery class entirely - merge it to DataChain.

shcheklein: that's the reality now (e.g. delta) - we do have to call it and be aware about two different states of the query. It is not good, but tbh considering how expensive it could be I think explicit it better in this case.

We then need to change it also (e.g. add a wrapper to "starting step" getter to automatically resolve it, or something.

But I still think hash doing more than expected / having side effects, etc is not good.

> Main fix here, which would significantly improve all codebase, is to remove DatasetQuery class entirely - merge it to DataChain.

separate discussion. And I'm not sure it will be that obvious - most likely you will still have two different states (since you need to be lazy) and more or less you will have the same question when / how call some kind of resolve.

ilongin: The issue is sub-queries inside union/join/subtract. When `hash()` is called on a chain like:

```python
dc.read_storage("s3://a/",update=True).union(dc.read_storage("s3://b/",update=True)).save("result")
```

`save()` calls `resolve_listing()` on the top-level query, but the top-level is the union result - it has no list_ds_name, so it's a no-op.
Then `hash()` reaches `SQLUnion.hash_inputs()` which calls `query1.hash()` and `query2.hash()` -  those sub-queries have unresolved listings and crash.

This also happens with `read_storage` accepting multiple URIs internally (unions them), and with any user-created chain1.union(chain2) where either side is `read_storage` with `update=True`.

One alternative to keep `hash()` pure: call `resolve_listing()` into step types that have sub-queries:

```python
def hash_inputs(self) -> str:
    self.query1.resolve_listing()  # needs to be added
    self.query2.resolve_listing()  # needs to be added
    return hashlib.sha256(
        bytes.fromhex(self.query1.hash()) + bytes.fromhex(self.query2.hash())
    ).hexdigest()
```

Same would be needed in `SQLJoin.hash_inputs()` and `Subtract.hash_inputs()`. The downside is every step type with sub-queries needs to remember this, and adding a new one could miss it.

I think the side effect in `hash()` is acceptable here. The listing must happen before any meaningful operation - you can't hash, save, or execute without knowing what data you're working with. Putting it in `hash()` is the most self contained: each query resolves itself, works recursively for sub-queries, no caller needs to know about the internal state. And since we removed the public `DataChain.hash()` API, the only callers are internal (save(), apply_steps()), both of which would trigger listing anyway.

If, regardless of this above, you still have strong opinion about moving it out of `hash()` I can do it.

shcheklein: Check how new delta logic does this. It collects the queries inside the whole query tree that need transformation (to cover multi source + merge usafe delta).

I think we can run recursive collect of the same sort (even use the same helpers?). And call resolve - again, similar how delta does this now. It also need to call resolve before doing transform. It does it explicitly. I don't see big issue in that. And it is better to my mind then taking a short cut and making hash with heavy side effects.

Let's think about the proper design for this.

ilongin: Done. Added `sub_queries` property to Step (base returns `[]`, SQLUnion/SQLJoin return `[query1, query2]`, DatasetDiffOperation returns `[dq]`). 

Also added `resolve_all_listings()` on DatasetQuery that recursively walks the step tree and resolves listings on all sub-queries. `hash()` no longer calls `resolve_listing()` — callers (`save()`, `apply_steps()`) call `resolve_all_listings()` explicitly before hashing.

PR #1657, opened 2026-03-17
fix(delta): apply listing steps to materialize starting step

shcheklein:
Follow up after recent delta refactor to make it work with delayed read storage steps.


shcheklein:
@ilongin you might need to change the call name / refactor it a bit when your PR lands.

PR #1658, opened 2026-03-17
fix(to_storage): better Ctrl-C handling

shcheklein:
Close the `to_storage()` iterator on interrupt so Ctrl-C does not leave SQLite temp-table cleanup hitting `database is locked`.

## Before

```text
Error in sys.excepthook:
  File "src/datachain/query/session.py", line 316, in except_hook
    Session.GLOBAL_SESSION_CTX.__exit__(exc_type, exc_value, exc_traceback)
  File "src/datachain/query/session.py", line 107, in __exit__
    self._cleanup_temp_datasets()
  File "src/datachain/data_storage/sqlite.py", line 259, in execute
sqlite3.OperationalError: database is locked

Exception ignored in atexit callback: <function Session._global_cleanup ...>
sqlite3.OperationalError: database is locked
```

## After

```text
  File "<stdin>", line 36, in <module>
  File "src/datachain/lib/dc/datachain.py", line 2845, in to_storage
  File "src/datachain/nodes_thread_pool.py", line 76, in run
KeyboardInterrupt
```

### DIFF
--- a/tests/func/test_to_storage.py
+++ b/tests/func/test_to_storage.py
@@ -36,10 +36,11 @@ def test_to_storage(
     file_type,
     num_threads,
 ):
-    call_count = {"count": 0}
+    call_count = 0
 
     def mapper(file_path):
-        call_count["count"] += 1
+        nonlocal call_count
+        call_count += 1
         return len(file_path)
 
     ctc = cloud_test_catalog
@@ -93,7 +94,7 @@ def test_to_storage(
         with (output_root / destination_rel).open() as f:
             assert f.read() == expected[file_obj.name]
 
-    assert call_count["count"] == len(expected)
+    assert call_count == len(expected)
 
 
 @pytest.mark.parametrize("use_cache", [True, False])
@@ -213,12 +214,19 @@ def test_to_storage_files_filename_placement_not_unique_files(tmp_dir, test_sess
 
 
 @skip_if_not_sqlite
-def test_to_storage_keyboard_interrupt_does_not_lock_cleanup(tmp_path, catalog_tmpfile):
+@pytest.mark.parametrize(
+    "num_threads",
+    [1, 5],
+    ids=["single_thread", "multiple_threads"],
+)
+def test_to_storage_keyboard_interrupt_does_not_lock_cleanup(
+    tmp_path, catalog_tmpfile, num_threads
+):
     script = tmp_path / "test_to_storage_interrupt.py"
     input_dir = tmp_path / "input"
     output_dir = tmp_path / "output"
     input_dir.mkdir()
-    for idx in range(3):
+    for idx in range(100):
         (input_dir / f"{idx}.txt").write_text(f"file-{idx}")
 
     script.write_text(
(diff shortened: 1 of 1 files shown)


tests/func/test_to_storage.py
```diff
@@ -0,0 +1,330 @@
...
+        with open(file_path, "wb") as fd:
+            fd.write(data)
+
+    df = dc.read_storage((tmp_dir / bucket_name).as_uri(), session=test_session)
+    with pytest.raises(ValueError):
+        df.to_storage(tmp_dir / "output", placement="filename")
+
+
+@skip_if_not_sqlite
+def test_to_storage_keyboard_interrupt_does_not_lock_cleanup(tmp_path, catalog_tmpfile):
```
shcheklein: C: this is new test to review

src/datachain/lib/dc/datachain.py
```diff
@@ -5,6 +5,8 @@
 import sys
 import warnings
 from collections.abc import Callable, Iterator, Sequence
+from collections.abc import Generator as IteratorGenerator
```
dreadatour, line 8: Not sure why do we want to import this as `IteratorGenerator` instead of just `Generator`.

src/datachain/nodes_thread_pool.py
```diff
@@ -105,7 +105,7 @@ def run(
         return results
 
     def cancel_all(self):
-        self.cancel = True
+        self.canceled = True
```
dreadatour, line 108: 👍

tests/func/test_to_storage.py
```diff
@@ -0,0 +1,338 @@
...
+    tmp_dir,
+    cloud_test_catalog,
+    test_session,
+    placement,
+    use_map,
+    use_cache,
+    file_type,
+    num_threads,
+):
+    call_count = {"count": 0}
```
dreadatour: I may be missing something, but why do we need dict here? Why not just to use `call_count = 0`? 🤔

tests/func/test_to_storage.py
dreadatour: Do we also want to parametrize this with `parallel` setting?

tests/func/test_to_storage.py
```diff
@@ -0,0 +1,338 @@
...
+                    ).start()
+                    time.sleep(1)
+                return original_export(self, *args, **kwargs)
+
+            File.export = patched_export
+
+            dc.read_storage({str(input_dir)!r}).to_storage(
+                {str(output_dir)!r},
+                placement="filename",
+                num_threads=1,
```
dreadatour: Also default `num_threads` is `5`, does this means it will not works with default value or test is not stable and it is hard to reproduce `SIGINT` in this case?

src/datachain/lib/dc/datachain.py
dreadatour: `datachain.lib.udf.Generator`
I see now 🤔

tests/func/test_to_storage.py
shcheklein: you mean `settings(parallel=True)` 

we don't have a mapper / UDF here, right?

dreadatour: > you mean `settings(parallel=True)`
> 
> we don't have a mapper / UDF here, right?

Yes, `parallel=True` or `parallel=2`, I just wonder will it be any difference here. Also `num_threads` in `to_storage` (comment below).

shcheklein: yep, that one is addressed now

parallel - not sure it is needed here ... resolving this one

PR #1659, opened 2026-03-18
re-do checkpoints docs

ilongin:


shcheklein:
@ilongin let's merge, I'll review after deployed and do more comments if needed.

PR #1662, opened 2026-03-20
cleanup and fix examples

shcheklein:
HF is breaking now with `AttributeError: 'NoneType' object has no attribute 'oid'` (see below). They changed at some point what they return by default and now an explicit param is required it seems to get file version (commit). Also other examples are skipped on CI and got broken.

```
File ~/Documents/eryk/datachain_example/.venv/lib/python3.11/site-packages/datachain/lib/dc/datachain.py:2128, in DataChain.show(self, limit, flatten, transpose, truncate, include_hidden)
   2125 import pandas as pd
   2127 dc = self.limit(limit) if limit > 0 else self  # type: ignore[misc]
-> 2128 df = dc.to_pandas(
   2129     flatten,
   2130     include_hidden=include_hidden,
   2131     as_object=True,
   2132 )
   2134 if df.empty:
   2135     print("Empty result")

File ~/Documents/eryk/datachain_example/.venv/lib/python3.11/site-packages/datachain/lib/dc/datachain.py:2101, in DataChain.to_pandas(self, flatten, include_hidden, as_object)
   2098 else:
   2099     columns = pd.MultiIndex.from_tuples(map(tuple, headers))
-> 2101 results = self.results(include_hidden=include_hidden)
   2102 if as_object:
   2103     df = pd.DataFrame(results, columns=columns, dtype=object)

File ~/Documents/eryk/datachain_example/.venv/lib/python3.11/site-packages/datachain/lib/dc/datachain.py:1570, in DataChain.results(self, row_factory, include_hidden)
   1568 def results(self, *, row_factory=None, include_hidden=True):
...
     82         etag=v.get("blob_id", ""),
     83         last_modified=v["last_commit"].date,
     84     )

AttributeError: 'NoneType' object has no attribute 'oid'
```

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -216,6 +216,7 @@ jobs:
       - name: Run examples
         env:
           HF_TOKEN: ${{ matrix.group == 'llm_and_nlp' && secrets.HF_TOKEN || '' }}
+          ANTHROPIC_API_KEY: ${{ matrix.group == 'llm_and_nlp' && secrets.ANTHROPIC_API_KEY || '' }}
           OPENAI_API_KEY: ${{ matrix.group == 'multimodal' && secrets.OPENAI_API_KEY || '' }}
         run: nox -s examples -p ${{ matrix.pyv }} -- -m "${{ matrix.group }}"
 


src/datachain/client/hf.py
```diff
@@ -73,14 +73,30 @@ class HfClient(Client):
     def FS_CLASS(cls):  # noqa: N802, N805
         return get_hf_filesystem_cls()
 
+    def _file_info_kwargs(self, version_id: str | None = None) -> dict[str, Any]:
+        return {
+            **self._version_kwargs(version_id),
+            "expand_info": True,
```
shcheklein, line 79: C: this is the main change here

PR #1664, opened 2026-03-21
catch proper HF exception to make example less falkey

shcheklein:
Make HF example less flakey.


examples/llm_and_nlp/hf-dataset-llm-eval.py
```diff
@@ -45,7 +45,7 @@ def eval_dialog(
                 "json_schema": {"schema": DialogEval.model_json_schema()},
             },
         )
-    except HTTPError as e:
+    except HfHubHTTPError as e:
```
shcheklein, line 48: C: main change here - HF changed their API her also

PR #1670, opened 2026-03-22
feat: support version in dataset name for read_dataset

### LINKED ISSUE #1669: read_dataset needs to support version in dataset name
### Description

I've notice agents ofter operates `my_dataset@1.0.0` format - it also convent for people. Which leads to auto generated code like:
```python
    if "@" in name_version:
        name, version = name_version.split("@", 1)
        chain = dc.read_dataset(name, version=version)
```

read_dataset() needs to support version in the dataset name to avoid this.

dmpetrov:
Fixes #1669

## Summary
- `read_dataset("my_dataset@1.0.0")` now works equivalently to `read_dataset("my_dataset", version="1.0.0")`
- Supports all version formats including specifiers: `read_dataset("my_dataset@>=1.0.0,<2.0.0")`
- Raises `ValueError` if version is specified both in the name and as a parameter
- Works with fully-qualified names: `read_dataset("namespace.project.my_dataset@1.0.0")`

## Test plan
- [x] `test_read_dataset_version_in_name` — exact version, specifier, and conflict error cases

### DIFF
--- a/src/datachain/lib/dc/datasets.py
+++ b/src/datachain/lib/dc/datasets.py
@@ -52,8 +52,8 @@ def read_dataset(
             namespace and project. Alternatively, it can be a regular name, in which
             case the explicitly defined namespace and project will be used if they are
             set; otherwise, default values will be applied. The name can also include
-            a version using the ``name@version`` format (e.g. ``"my_dataset@1.0.0"``),
-            which is equivalent to passing ``version="1.0.0"`` separately.
+            a version using the ``name@version`` format (e.g. ``"my_dataset@1.0.0"``).
+            If ``version`` is also provided explicitly, it takes priority.
         namespace: optional name of namespace in which dataset to read is created
         project: optional name of project in which dataset to read is created
         version: dataset version. Supports:
--- a/tests/func/test_read_dataset_version_specifiers.py
+++ b/tests/func/test_read_dataset_version_specifiers.py
@@ -98,7 +98,6 @@ def test_read_dataset_version_in_name(test_session):
         == "1.0.0"
     )
 
-    # Explicit version parameter takes priority over @version in name
     assert (
         dc.read_dataset(
             f"{dataset_name}@1.0.0", version="2.0.0", session=test_session


src/datachain/lib/dc/datasets.py
```diff
@@ -24,6 +24,35 @@
...
+def _parse_name_version(name: str, version: str | None) -> tuple[str, str | None]: ...
+
+
+@overload
+def _parse_name_version(
+    name: str, version: str | int | None
+) -> tuple[str, str | int | None]: ...
+
+
+def _parse_name_version(
```
shcheklein: can we use one of many existing parsers or consolidate this?

e.g. parse_dataset_with_version or parse_dataset_name

src/datachain/lib/dc/datasets.py
```diff
@@ -24,6 +24,35 @@
     P = ParamSpec("P")
 
 
+@overload
+def _parse_name_version(name: str, version: str | None) -> tuple[str, str | None]: ...
```
shcheklein: seems too complicated to have overloads for a helper, something smells here

src/datachain/lib/dc/datasets.py
```diff
@@ -50,7 +79,9 @@ def read_dataset(
         name: The dataset name, which can be a fully qualified name including the
             namespace and project. Alternatively, it can be a regular name, in which
             case the explicitly defined namespace and project will be used if they are
-            set; otherwise, default values will be applied.
+            set; otherwise, default values will be applied. The name can also include
```
shcheklein, line 54: what is the priority if version is also passed explicitly?

tests/func/test_read_dataset_version_specifiers.py
```diff
@@ -70,19 +75,74 @@ def test_read_dataset_version_specifiers_no_match(test_session):
...
+def test_read_dataset_version_in_name(test_session):
+    """Test read_dataset with version embedded in name using the @ syntax."""
+    dataset_name = "test_version_in_name"
+
+    for version in ["1.0.0", "2.0.0"]:
+        dc.read_values(data=[1, 2], session=test_session).save(
+            dataset_name, version=version
+        )
+
+    # Test exact version embedded in name
```
shcheklein: too many comments that describe obvious

src/datachain/cli/commands/show.py
```diff
@@ -19,9 +19,12 @@ def show(
     include_hidden: bool = False,
 ) -> None:
     from datachain import Session, read_dataset
+    from datachain.lib.dc.datasets import _parse_name_version
```
shcheklein: it's not a good practice to import protected methods

also, I would review if it should be somewhere in utils

src/datachain/lib/dc/datasets.py
dmpetrov: fixed

src/datachain/lib/dc/datasets.py
dmpetrov: fixed

tests/func/test_read_dataset_version_specifiers.py
dmpetrov: fixed

src/datachain/cli/commands/show.py
dmpetrov: fixed

PR #1672, opened 2026-03-22
Expose query_script via Python API and CLI

### LINKED ISSUE #1671: Expose query_script via Python API and CLI
## Summary

The `query_script` field on `DatasetVersion` stores the Python source code used to create a dataset. When a Studio job runs, the script is captured and stored — but this information is not accessible via the Python API or CLI. Additionally, locally-run datasets never capture `query_script` at all.

## Current behavior

- Studio jobs: `query_script` is populated ✓
- Local script runs: `query_script` is always `""` ✗
- No CLI command exposes `query_script`
- `DatasetInfo` (used in `dc.datasets()`) does not include `query_script`

## Root cause

1. `Session.get_or_create_job()` in `query/session.py` creates jobs with `query=""` (never reads the local script file)
2. `DatasetQuery.save()` in `query/dataset.py` calls `catalog.create_dataset()` without passing `query_script`
3. `DatasetInfo` in `lib/dataset_info.py` does not expose `query_script`
4. No CLI command to display `query_script`

### 1. Capture script content locally (`query/session.py`)

In `get_or_create_job()`, when `is_script_run()` is True, read the script file before creating the job (same pattern as Studio's `create_job()`):

```python
if is_script_run():
    script = os.path.abspath(sys.argv[0])
    try:
        with open(script) as f:
            query_content = f.read()
    except OSError:
        query_content = ""
else:
    script = str(uuid4())
    query_content = ""

job_id = self.catalog.metastore.create_job(name=script, query=query_content, ...)
```

### 2. Propagate `Job.query` → `DatasetVersion.query_script` (`query/dataset.py`)

In `save()`:
```python
dataset = self.catalog.create_dataset(..., job_id=job_id, query_script=job.query)
```

### 3. Expose in Python API (`lib/dataset_info.py`)

Add to `DatasetInfo`:
```python
query_script: str = Field(default="")
```

Update `from_models()`:
```python
query_script=version.query_script,
```

### 4. CLI flag (`datachain show --script`)

Add `--script` flag to `datachain show` that prints `dataset_version.query_script` to stdout.

```bash
datachain show my-dataset --script
datachain show my-dataset --script --version 1.0.0
```

Python API usage after fix:
```python
import datachain as dc

for ds in dc.datasets(column="dataset").to_iter("dataset"):
    if ds.name == "my-dataset":
        print(ds.query_script)
```

dmpetrov:
## Summary

- Capture local script source code in `get_or_create_job()` and store it in `job.query`
- Pass `job.query` as `query_script` when saving datasets via `DatasetQuery.save()`
- Add `query_script: str` field to `DatasetInfo` so it's accessible via `dc.datasets()`
- Add `--script` flag to `datachain show` CLI command to print the script to stdout

## Usage

Python API:
```python
for ds in dc.datasets(column="dataset").to_iter("dataset"):
    print(ds.query_script)
```

CLI:
```bash
datachain show my-dataset --script
```

Closes #1671

## Test plan
- [ ] Run a script that creates a dataset, then verify `DatasetInfo.query_script` contains the source
- [ ] Run `datachain show <dataset> --script` and verify it prints the script
- [ ] Run `pytest tests/func/test_datasets.py -x`

### DIFF
--- a/src/datachain/cli/parser/__init__.py
+++ b/src/datachain/cli/parser/__init__.py
@@ -468,7 +468,7 @@ def get_parser() -> ArgumentParser:  # noqa: PLR0915
     show_parser.add_argument(
         "--script",
         action="store_true",
-        help="Print the query script to stdout",
+        help="Print the script of source code",
     )
     add_show_args(show_parser)
 


src/datachain/cli/parser/__init__.py
```diff
@@ -465,6 +465,11 @@ def get_parser() -> ArgumentParser:  # noqa: PLR0915
         help="Dataset version",
     )
     show_parser.add_argument("--schema", action="store_true", help="Show schema")
+    show_parser.add_argument(
+        "--script",
+        action="store_true",
+        help="Print the query script to stdout",
```
shcheklein: just `Print the query script` probably enough

PR #1675, opened 2026-03-23
Reverse order remove dataset version

ilongin:
Current code can cause orphan dataset version table to appear if system crashes between removing metadata and actual dataset table.
This fixes it by first removing warehouse dataset table, and only then dataset version metadata.


src/datachain/catalog/catalog.py
```diff
@@ -1115,13 +1115,13 @@ def remove_dataset_version(
...
         Deletes one single dataset version.
-        If it was last version, it removes dataset completely
+        If it was last version, it removes dataset completely.
         """
         if not dataset.has_version(version):
             return
-        dataset = self.metastore.remove_dataset_version(dataset, version)
         if drop_rows:
             self.warehouse.drop_dataset_rows_table(dataset, version)
+        dataset = self.metastore.remove_dataset_version(dataset, version)
```
shcheklein, line 1127: I think we should set REMOVING state before touching warehouse table. Otherwise we might end up with a broken dataset if it fails in the middle.

ilongin: Good idea, added it.

PR #1677, opened 2026-03-24
fix(group_by): support inline func expressions in partition_by without label

### LINKED ISSUE #1676: group_by: partition_by does not support inline func expressions
## Description

When passing an inline `func.*` expression directly to `partition_by` in `group_by`, DataChain silently derives an internal column name and then fails with a `SignalResolvingError` because that column doesn't exist in the schema yet.

## Steps to reproduce

```python
import datachain as dc
from datachain import C, func

(
    dc.read_storage("gs://datachain-demo/", anon=True)
    .group_by(
        count=func.count(),
        total_size=func.sum(C("file.size")),
        partition_by=func.path.parent(C("file.path")),  # inline expression
    )
    .show()
)
```

## Error

```
datachain.lib.signal_schema.SignalResolvingError: cannot resolve signal name 'file__path__parent': is not found
```

## Expected behavior

Either:
- Inline expressions in `partition_by` are evaluated automatically (like they are in `mutate`), or
- A clear error message explaining that `partition_by` requires a pre-existing column name, not an expression

## Workaround

Materialize the column with `.mutate()` first:

```python
(
    dc.read_storage("gs://datachain-demo/", anon=True)
    .mutate(parent=func.path.parent(C("file.path")))
    .group_by(
        count=func.count(),
        total_size=func.sum(C("file.size")),
        partition_by="parent",
    )
    .show()
)
```

dmpetrov:
## Summary

Fixes #1676: When a `Func` expression is used in `partition_by` without `.label()`, the output column now uses the function name as its name (e.g., `"parent"` for `func.path.parent(...)`) instead of the input column's DB name (e.g., `"file__path"`).

**Before:**
```python
dc.group_by(cnt=func.count(), partition_by=func.path.parent("file.path"))
# Result column named "file__path" (confusing)
```

**After:**
```python
dc.group_by(cnt=func.count(), partition_by=func.path.parent("file.path"))
# Result column named "parent" (intuitive)

dc.group_by(cnt=func.count(), partition_by=func.path.parent("file.path").label("dir"))
# Result column named "dir" (explicit label unchanged)
```

Also fixes pre-existing test failures:
- Update hash test values that changed due to SQLAlchemy version upgrade
- Fix `datamodel-code-generator` incompatibility: `disable_future_imports` parameter was removed in newer versions; strip the misplaced `from __future__ import annotations` from generated model code instead

## Test plan

- [x] New test `test_group_by_with_unlabeled_func_in_partition_by` covers the unlabeled case with both string and `C()` column references
- [x] Existing test `test_group_by_with_functions_in_partition_by` still passes (labeled case unchanged)
- [x] All 2281 unit tests pass

### DIFF
--- a/src/datachain/lib/dc/datachain.py
+++ b/src/datachain/lib/dc/datachain.py
@@ -1804,7 +1804,7 @@ class DataChain:
 
         def _resolve(
             ds: DataChain,
-            col: str | Function | sqlalchemy.ColumnExpr,
+            col: str | Function | ColumnExpr,
             side: str | None,
         ):
             try:


src/datachain/lib/meta_formats.py
```diff
@@ -120,14 +120,21 @@ def gen_datamodel_code(
...
             use_standard_collections=True,
-            disable_future_imports=True,
             formatters=[],
         )
         epilogue = f"""
 DataModel.register({model_name})
 spec = {model_name}
 """
-        return output.read_text() + epilogue
+        # Remove 'from __future__ import annotations' lines: newer versions of
```
shcheklein: is it related?

tests/unit/test_query_steps_hash.py
```diff
@@ -182,7 +182,7 @@ def test_filter_hash(inputs, _hash):
         (
             {"new_id": func.sum("id")},
             SignalSchema({"id": int}),
-            "60f3a9e31aa77dabb8045060659e75f050133278fd613f0a0075b309747eb80e",
+            "47f57e322e79be378d12cb8823273567f93e52b245c9c0736562826b54cd6f7e",
```
shcheklein: not exactly clear why exactly a lot of hashes changed tbh

dmpetrov: same here 🙂 
@ilongin could you please explain this?

dmpetrov (commented):
@ilongin we need your help here

tests/unit/test_query_steps_hash.py
ilongin: @dmpetrov  The hash values on main are already correct for SQLAlchemy 2.0.48 (latest). You likely had a different SQLAlchemy version locally which produces different hashes from `serialize_column_element` (it uses `_traverse_internals` which varies across versions). Please revert the hash value changes and keep only the `group_by` fix. I suggest to install latest SQLAlchemy version just in case

ilongin: But I think I will refactor these tests and remove exact hash assertion on places where it depends on things like SQLAlchemy to avoid this confusion. My motivation for having exact hash values in tests was that then we are sure it will be exact hash regardless of which OS is used, which python version etc, but it is true that we have more maintenence e.g need to change them when new version of SQLAlchemy changes how it represents internal structures. We can keep exact UUID values in tests for deterministic things like where just UUID is used for hash calc.

src/datachain/lib/meta_formats.py
```diff
@@ -120,14 +120,20 @@ def gen_datamodel_code(
...
             formatters=[],
         )
         epilogue = f"""
 DataModel.register({model_name})
 spec = {model_name}
 """
-        return output.read_text() + epilogue
+        # Strip misplaced `from __future__` lines that newer datamodel-code-generator
+        # versions may emit after other imports, causing a SyntaxError on exec().
+        model_code = "\n".join(
```
shcheklein: @dmpetrov this still seems unrelated and weird change, see the `disable_future_imports=True,` above

tests/unit/test_query_steps_hash.py
dmpetrov: > But I think I will refactor these tests and remove exact hash assertion on places where it depends on things like SQLAlchemy to avoid this confusion.

Yes, please remove the values - it looks like a code smell.

dmpetrov: THank you for the details!

src/datachain/lib/dc/datachain.py
```diff
@@ -1356,7 +1356,9 @@ def group_by(  # noqa: C901, PLR0912
                             keep_columns.append(col)
                         schema_partition_by.append(col)
             elif isinstance(col, Function):
-                column = col.get_column(self.signals_schema)
+                column = col.get_column(
+                    self.signals_schema, label=col.col_label or col.name
```
shcheklein: can it collide with an existing column name somehow?

dmpetrov:
@shcheklein please take a look at the latest changes

tests/unit/lib/test_partition_by.py
```diff
@@ -588,3 +631,15 @@ def test_aggregate_after_group_by(test_session):
...
+
+def test_group_by_partition_by_name_conflicts_with_agg_column(test_session):
+    """partition_by func name clashing with an aggregation kwarg raises an error."""
+    files = [
+        File(source="s3://bucket", path="docs/a.txt", size=100),
+    ]
+    with pytest.raises(DataChainColumnError, match="conflicts with aggregation"):
+        dc.read_values(file=files, session=test_session).group_by(
+            parent=func.count(),
+            partition_by=func.path.parent("file.path"),  # also resolves to "parent"
```
shcheklein: what if `parent` already existing in the dataset as a column?

also, should we mangle the names somehow - e.g. `__func_parent` to make it less possible. E.g. in this case - `parent` is very generic it seems

src/datachain/lib/dc/datachain.py
shcheklein: what about this case:


```
dc.group_by(
    cnt=func.count(),
    partition_by=[
        func.path.parent("item.path1"), 
        func.path.parent("item.path2"),
    ],
)
```

I think we do allow multiple columns in partition by

src/datachain/lib/dc/datachain.py
```diff
@@ -1373,6 +1375,16 @@ def group_by(  # noqa: C901, PLR0912
...
         if not kwargs:
             raise ValueError("At least one column should be provided for group_by")
+
+        partition_func_names = {c.name for c in partition_by_columns}
+        overlap = partition_func_names & set(kwargs.keys())
+        if overlap:
+            name = next(iter(overlap))
+            raise DataChainColumnError(
+                name,
+                f"partition_by name '{name}' conflicts with aggregation column name",
```
shcheklein, line 1400: can we also not throw an exception here? rather be aware about column names and generate some temporary labels that don't collide

dmpetrov: Good point with the two similar function and existing column! I'll try to incorporate input signal name to the label.

Temp names approach has a downside effect - they change hash of the operation and break checkpoints/deltas. Let's not go this route and keep throwing exception - it's ok to ask user for a custom name in such cases.

shcheklein: My point is that we can see which names are taken in a particular query and just put a suffix / prefix to the names we generate. This can and should be a stable algorithm - so, we won't be breaking anything. We do this in a few other places, there is no real reason to put this burden on the users shoulders. 

This is unexpected that system can't handle a query like this. There is nothing specific to it.

Also, let's not use very generic names like `parent` - way higher risk of collisions in the first place.

dmpetrov: Temp stable hashes approach does works for expressions C("col1") since we can get `md5(str(exp))` but `str(func)` returns only func name, not params. So, adding params seems the only stable way for functions. See in the code.

dmpetrov:
@shcheklein implemented the suggestions - thanks for the comment.

A few more things were implemented:
1. auto-generated columns shouldn't stay in the result, like `SELECT *, COUNT(*) as cnt GROUP BY RIGHT(my_text, 4)` in SQL doen't keep the new sub string and keeps only newly created `cnt`.
2. expressions were also implemented like `partition_by=dc.C("a") + dc.C("b") // 100`
3. window function - partition_by were accepting only str before. fixed.

dmpetrov:
An issue was found:
4. floor division `C("a") // N` compiled to `/` in Studio/CH which returns float. It works correctly in CLI/SQLite.

dmpetrov:
Another bug was found:
5. #1687

dmpetrov:
@shcheklein now it's fully implemented with all the additional fixes I mentioed in the comments. Please take a look.

tests/func/test_listing.py
```diff
@@ -106,6 +106,6 @@ def test_parse_listing_uri(cloud_test_catalog, cloud_type):
 def test_parse_listing_uri_with_glob(cloud_test_catalog):
     ctc = cloud_test_catalog
     dataset_name, listing_uri, listing_path = parse_listing_uri(f"{ctc.src_uri}/dogs/*")
-    assert dataset_name == f"lst__{ctc.src_uri}/dogs/"
+    assert dataset_name == _sanitize_ds_name(f"lst__{ctc.src_uri}/dogs/")
```
shcheklein, line 109: this should be fixed in main now, please merge main

tests/func/test_ls.py
```diff
@@ -119,7 +119,9 @@ def test_ls_glob_sub(cloud_test_catalog, cloud_type, capsys):
 
 
 def list_dataset_name(uri, path):
-    return f"{LISTING_PREFIX}{uri}/{posixpath.join(path, '').lstrip('/')}"
+    return _sanitize_ds_name(
```
shcheklein, line 122: this - I'm not sure, if was not fixed - let's keep it here ... thanks for taking care of this

src/datachain/lib/dc/datachain.py
```diff
@@ -1356,23 +1359,53 @@ def group_by(  # noqa: C901, PLR0912
                             keep_columns.append(col)
                         schema_partition_by.append(col)
             elif isinstance(col, Function):
-                column = col.get_column(self.signals_schema)
-                col_db_name = column.name
-                col_type = column.type.python_type
-                schema_fields[col_db_name] = col_type
-                partition_by_columns.append(column)
-                signal_columns.append(column)
+                if col.col_label:
```
shcheklein: can it be passed by a user?

tests/unit/lib/test_partition_by.py
```diff
@@ -435,6 +436,46 @@ class CustomFile(DataModel):
...
+
+    ds2 = dc.read_values(
+        custom_file=custom_data,
+        session=test_session,
+    ).group_by(
+        cnt=func.count(),
+        partition_by=func.path.parent(dc.C("custom_file.path")),
+    )
+
+    assert set(ds2.signals_schema.values.keys()) == {"cnt"}
```
shcheklein: Q: do we always drop partition by columns or only in case of these dynamic "columns"?

tests/func/test_ls.py
dmpetrov: ok, keeping.

src/datachain/lib/dc/datachain.py
dmpetrov: sure. it's best practice but we cannot force it

tests/unit/lib/test_partition_by.py
dmpetrov: good question. On second thought, it's better to keep them and use counter based naming like `gr_1`, `gr_2`, ....

It would be closer to the existing DBs. Column names after `GROUP BY expenses + fees, RIGHT(filename, 3)`:
- pgsql `?column?`, `?column?`
- bigquery `f0_`, `f1_`   <-- only this is acceptable for us
- mysql `expenses + fees`, `RIGHT(filename, 3)`

dmpetrov:
@shcheklein please take another look.
The dynamic names were exposed to a simple gr_1, gr_2,... A lot of simplifications.

Also, I found issue with inference types #1692. That was actually a well know issue of mutate() but it affects partition_by as well. See `_infer_expr_type()` as a workaround but the proper fix is coming in #1693

src/datachain/func/window.py
```diff
@@ -21,12 +27,12 @@ def window(partition_by: str, order_by: str, desc: bool = False) -> Window:
     that the window function will operate on.
 
     Args:
-        partition_by (str): The column name by which to partition the result set.
-            Rows with the same value in the partition column will be grouped together
-            for the window function.
-        order_by (str): The column name by which to order the rows within
-            each partition. This determines the sequence in which the window function
-            is applied.
+        partition_by (str | ColumnElement): The column name or expression by which
```
shcheklein: can we add an example below for this (it comes really handy for people and AI also :))

tests/unit/test_func.py
```diff
@@ -254,6 +254,46 @@ def test_floordiv_mutate(chain):
...
+    sql = str((col // 100).compile(compile_kwargs={"literal_binds": True}))
+    assert "divide" in sql
+    assert "CAST" in sql
+
+    sql = str((100 // col).compile(compile_kwargs={"literal_binds": True}))
+    assert "divide" in sql
+    assert "CAST" in sql
+
+
+def test_column_floordiv_mutate():
```
shcheklein: we should be using test_session probably (to make in-memory) - like in other test

tests/unit/test_func.py
```diff
@@ -254,6 +254,46 @@ def test_floordiv_mutate(chain):
     assert res == [1, 1, 1, 1, 1]
 
 
+def test_column_truediv_uses_divide():
```
shcheklein, line 257: do we really need to test what is inside? do we care what SQL is used? 

can we use something similar to above:

```
    f = rnd1 // rnd2
    assert str(f) == "floordiv()"
```

why do these tests look different?

src/datachain/lib/dc/datachain.py
```diff
@@ -1269,11 +1270,41 @@ def select_except(self, *args: str) -> "Self":
...
+        return str
+
     @delta_disabled  # type: ignore[arg-type]
-    def group_by(  # noqa: C901, PLR0912
+    def group_by(  # noqa: C901, PLR0912, PLR0915
         self,
         *,
-        partition_by: str | Func | Sequence[str | Func] | None = None,
+        partition_by: (
+            str | Func | ColumnElement | Sequence[str | Func | ColumnElement] | None
```
shcheklein: also, add example below would be beneficial

src/datachain/lib/dc/datachain.py
```diff
@@ -1357,23 +1389,53 @@ def group_by(  # noqa: C901, PLR0912
...
+                    column = col.get_column(
+                        self.signals_schema,
+                        label=col_label,
+                    )
+                    partition_by_columns.append(column)
+                    signal_columns.append(column)
+                    schema_fields[column.name] = column.type.python_type
+            elif isinstance(col, ColumnElement):
+                col_label = f"gr_{partition_counter}"
+                partition_counter += 1
```
shcheklein, line 1375: what if we have two group bys in one after another? (just to check that colum names won't collide in some unexpected way)

shcheklein: probably good to have a test like this

src/datachain/lib/dc/datachain.py
```diff
@@ -1357,23 +1389,53 @@ def group_by(  # noqa: C901, PLR0912
...
         if not kwargs:
             raise ValueError("At least one column should be provided for group_by")
+
+        partition_output_names = {c.name for c in signal_columns}
+        overlap = partition_output_names & set(kwargs.keys())
+        if overlap:
+            name = next(iter(overlap))
+            raise DataChainColumnError(
+                name,
+                f"partition_by name '{name}' conflicts with aggregation column name",
```
shcheklein, line 1400: so, can it actually check existing column names and just avoid them?

src/datachain/lib/dc/datachain.py
```diff
@@ -1357,23 +1389,53 @@ def group_by(  # noqa: C901, PLR0912
...
+                else:
+                    col_label = f"gr_{partition_counter}"
+                    partition_counter += 1
+                    column = col.get_column(
+                        self.signals_schema,
+                        label=col_label,
+                    )
+                    partition_by_columns.append(column)
+                    signal_columns.append(column)
+                    schema_fields[column.name] = column.type.python_type
```
shcheklein: these two blocks look almost identical - we can ask to dry them and simplify most likely

shcheklein: I mean `if col.col_label:` and `esle ...`

src/datachain/lib/dc/datachain.py
```diff
@@ -1357,23 +1389,53 @@ def group_by(  # noqa: C901, PLR0912
...
+            elif isinstance(col, ColumnElement):
+                col_label = f"gr_{partition_counter}"
+                partition_counter += 1
+                labeled = cast("Column", col.label(col_label))
+                inferred = self._infer_expr_type(col, self.signals_schema)
+                sql_type = python_to_sql(inferred)
+                labeled.type = sql_type() if isinstance(sql_type, type) else sql_type
```
shcheklein: Q: what is happening here? is it a regular thing to do? why?

src/datachain/lib/dc/datachain.py
```diff
@@ -1269,11 +1270,41 @@ def select_except(self, *args: str) -> "Self":
             signal_schema=new_schema,
         )
 
+    @staticmethod
+    def _infer_expr_type(expr, signals_schema):
+        """Infer the Python type of a ColumnElement expression from the schema."""
+        from sqlalchemy.sql import visitors
```
shcheklein: review please, if some imports are redundant and / or can be moved to the top

src/datachain/lib/dc/datachain.py
```diff
@@ -1269,11 +1270,41 @@ def select_except(self, *args: str) -> "Self":
...
+
+        from datachain.lib.signal_schema import SignalResolvingError
+        from datachain.query.schema import Column
+
+        result = sql_to_python(expr)
```
shcheklein: should we just expand the helper itself - sql_to_python ? or what is the difference between them?

src/datachain/lib/dc/datachain.py
```diff
@@ -1269,11 +1270,41 @@ def select_except(self, *args: str) -> "Self":
...
+        if result is not str:
+            return result
+
+        col_types = set()
+        for elem in visitors.iterate(expr, {}):
```
shcheklein: is it recursive?

src/datachain/lib/dc/datachain.py
```diff
@@ -1269,11 +1270,41 @@ def select_except(self, *args: str) -> "Self":
...
+    def _infer_expr_type(expr, signals_schema):
+        """Infer the Python type of a ColumnElement expression from the schema."""
+        from sqlalchemy.sql import visitors
+
+        from datachain.lib.signal_schema import SignalResolvingError
+        from datachain.query.schema import Column
+
+        result = sql_to_python(expr)
+        if result is not str:
+            return result
```
shcheklein: not sure I understand what this check does tbh

src/datachain/func/window.py
dmpetrov: well.. there are 5+ functions that uses ColumnElement - join, case, ifelse, _or ... The same issue there. So, it's going to be a lot of repetitions.

I'll try to add this to docs instead. Also, it's better not to use ColumnElement since it's sqlalchemy type. I'll create a type alias with a better name for the docs.

dmpetrov: Renamed to ColumnExpr and added docs.

Also, added docs for `Function` - similar issue. And found a bigger problem #1699

tests/unit/test_func.py
dmpetrov: good catch

tests/unit/test_func.py
dmpetrov: the 1st tests columns, the 2nd - functions. these two generate different code.
No problem removing these since we already have e2e test with whole chain but I'd keep these.

src/datachain/lib/dc/datachain.py
dmpetrov: explained above

src/datachain/lib/dc/datachain.py
dmpetrov: different sql-dialect solve this differently. if user didn't assign label - it's a gray area. In our case, we just mask/ignore all previous results.

I can try using gr_N. It should do our solution 1% better I guess.

src/datachain/lib/dc/datachain.py
dmpetrov: _infer_expr_type() won't be needed after #1693

src/datachain/lib/dc/datachain.py
dmpetrov: right. in the next PR #1693

src/datachain/lib/dc/datachain.py
dmpetrov: let's ignore the whole _infer_expr_type() function

src/datachain/lib/dc/datachain.py
dmpetrov: let's ignore the whole _infer_expr_type() function

src/datachain/lib/dc/datachain.py
dmpetrov: good point! simplified.

src/datachain/lib/dc/datachain.py
dmpetrov: that's common pattern in DC, unfortunately. Not sure how to simplify.
The best you can do is: `labeled.type = python_to_sql(inferred)()` which is also pretty ugly.

src/datachain/lib/dc/datachain.py
dmpetrov: now it does - see the change related to nested-group_by / gr_N

src/datachain/lib/dc/datachain.py
shcheklein: masking is fine probably ... my only question was to confirm that it doesn't crash and has some predictable behavior I guess (thus may be also worth having a test for this)

dmpetrov: It didn't crush. And I changed it to avoid masking.

dmpetrov:
@shcheklein it was updated from main with it's dependency #1699 and ready to be merged. Please approve.

src/datachain/lib/dc/datachain.py
```diff
@@ -1296,7 +1299,7 @@ def group_by(  # noqa: C901, PLR0912
         """
         if partition_by is None:
             partition_by = []
-        elif isinstance(partition_by, (str, Func)):
+        elif isinstance(partition_by, (str, Func, ColumnElement)):
```
shcheklein: expected or should be ColumnExpr also?

src/datachain/lib/dc/datachain.py
```diff
@@ -1352,23 +1363,44 @@ def group_by(  # noqa: C901, PLR0912
...
-                schema_fields[col_db_name] = col_type
+                label = col.col_label
+                if not label:
+                    label = f"gr_{partition_counter}"
+                    partition_counter += 1
+                column = col.get_column(self.signals_schema, label=label)
                 partition_by_columns.append(column)
                 signal_columns.append(column)
+                schema_fields[column.name] = column.type.python_type
+            elif isinstance(col, ColumnElement):
```
shcheklein: ColumnElement or ColumnExpr now?

src/datachain/lib/dc/datachain.py
```diff
@@ -1352,23 +1363,44 @@ def group_by(  # noqa: C901, PLR0912
...
                         f"partition_by column {col} has type {type(col)}"
-                        " but expected str or Function"
+                        " but expected str, Function, or ColumnExpr"
                     ),
                 )
 
         if not kwargs:
             raise ValueError("At least one column should be provided for group_by")
+
+        partition_output_names = {c.name for c in signal_columns}
```
shcheklein, line 1394: do we still need this check though? is it possible to hit it now?

src/datachain/lib/dc/datachain.py
dmpetrov: it's not a big deal since it's alias and not exposed to users. but you are right, it's better to avoid using internal ColumnElement in datachain.lib. I'll try to remove if no big issues.

src/datachain/lib/dc/datachain.py
dmpetrov: it's possible if user uses assign column label/name for example

shcheklein: good point yep, would be nice to have a test if don't yet (just to make sure we won't forget why it is here)

PR #1682, opened 2026-03-24
Preserve sys__id on copy partial table

ilongin:
Added preserve_id boolean var on copy table metastore functions and using it when copying partials checkpoint tables in UDF flow.

This fixes flaky tests and hidden bug with UDF processing https://github.com/datachain-ai/datachain/actions/runs/23368936195/job/67988605801?pr=1664#step:13:250

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,7 +24,7 @@ repos:
       - id: trailing-whitespace
         exclude: '^LICENSES/'
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: 'v0.15.8'
+    rev: 'v0.15.9'
     hooks:
       - id: ruff
         args: [--fix, --exit-non-zero-on-fix]


src/datachain/data_storage/sqlite.py
```diff
@@ -852,33 +852,32 @@ def insert_into(
         table: Table,
         query: Select,
         progress_cb: Callable[[int], None] | None = None,
+        preserve_ids: bool = False,
```
shcheklein: call it preserve_sys_ids - to make it more explicit

src/datachain/data_storage/sqlite.py
```diff
@@ -852,33 +852,32 @@ def insert_into(
...
 
+        columns = (
+            list(query.selected_columns)
+            if preserve_ids
+            else [c for c in query.selected_columns if c.name != "sys__id"]
+        )
+
         # If there is no sys__id column, we cannot copy the table in batches,
         # and we need to copy all rows at once. Same if there is a group by clause.
         if col_id is None or len(query._group_by_clause) > 0:
```
shcheklein, line 862: let's put assert that this option can't be used if `len(query._group_by_clause) > 0`

src/datachain/data_storage/warehouse.py
```diff
@@ -1060,14 +1066,17 @@ def create_table_from_query(
             query: Query to populate the table from
             create_fn: Function that creates an empty table given a name
             progress_cb: Optional callback for progress updates
+            preserve_ids: If True, preserve sys__id from source
```
shcheklein: what is the reason to not preserve ids from the source if they exist?

src/datachain/data_storage/sqlite.py
ilongin: added

src/datachain/data_storage/sqlite.py
ilongin: renamed

src/datachain/data_storage/warehouse.py
ilongin: `insert_into` is used in many contexts - `.save()`, join temp tables, dataset copies - where we want fresh IDs (the destination is a new dataset with its own ID space). Preserving source IDs is only correct when copying UDF output tables where `sys__id`  must match input table IDs for the join in `create_result_query`.

shcheklein: This is not true. Generating sys_ids is a hard task in general - it requires additional code in SQL which expands it, it takes more time to run, etc. We put quite a lot of effort to keep them everywhere as much as possible and drop only when it is required - e.g. union, merge are not safe and they are dropping them.

Especially if we use `insert_into` in many different contexts it should be probably preserving them by default.

Overall, if query has them - then it must be save to use them to my mind. Again, general policy is that operation that provides query defines the schema, etc.

It might be a bit more complicated for UDFs for sure, but since it is a general helper - let's review it again, we can't be regenerating IDs where it is not needed.

ilongin: ok, reversed the default value of it

src/datachain/query/dataset.py
```diff
@@ -1987,7 +1992,7 @@ def get_query(self, dq: "DatasetQuery", temp_tables: list[str]) -> sa.Subquery:
         )
         temp_tables.append(temp_table.name)
 
-        warehouse.insert_into(temp_table, query)
+        warehouse.insert_into(temp_table, query, preserve_sys_ids=False)
```
shcheklein: do we even need it here (I mean flag)? I would expect JOIN and Union already taking care of this (they should be dropping sys_ids from the query

src/datachain/query/dataset.py
```diff
@@ -3033,7 +3038,9 @@ def save(
             )
             self.temp_table_names.append(temp_table_name)
             temp_table = self.catalog.warehouse.get_table(temp_table_name)
-            self.catalog.warehouse.insert_into(temp_table, query.select())
+            self.catalog.warehouse.insert_into(
+                temp_table, query.select(), preserve_sys_ids=False
```
shcheklein: why? is is save operation? so - does it mean that we don't preserve sys_ids in the saved dataset now?

src/datachain/query/dataset.py
ilongin: You're right, union/join already drop sys__id via `_drop_system_columns`. The query won't have `sys__id` in its columns, so the flag is irrelevant here. Removed.

src/datachain/query/dataset.py
ilongin: We need fresh IDs in save because of `order_by`. When the chain has `order_by`, rows are inserted in sorted order but carry their original `sys__id` values. Both SQLite and ClickHouse read back by `sys__id` order by default (sys__id is primary / order key in CH), so the saved dataset loses the ordering. 
This was the existing behavior before this PR — save always regenerated IDs.

But, I guess we could detect if the chain has `order_by` and only regenerate then(`any(isinstance(step, SQLOrderBy) for step in self.steps)`), but it might be overcomplicating things... I will try though.

src/datachain/data_storage/sqlite.py
```diff
@@ -869,13 +870,16 @@ def insert_into(
...
+        if preserve_sys_ids:
+            select_q = query.offset(None).limit(None)
+        else:
+            select_q = (
+                query.with_only_columns(
+                    *[c for c in query.selected_columns if c.name != "sys__id"]
+                )
+                .offset(None)
+                .limit(None)
             )
```
dreadatour, line 882: Something like this to do not duplicate the code, or will it not works this way?

```suggestion
        select_q = query.offset(None).limit(None)
        if not preserve_sys_ids:
            select_q = select_q.with_only_columns(
                *[c for c in query.selected_columns if c.name != "sys__id"]
            )
```

src/datachain/query/dataset.py
```diff
@@ -610,8 +610,13 @@ def _checkpoint_tracking_columns(self) -> list["sqlalchemy.Column"]:
           input was not fully processed and needs to be re-run. Nullable because
           mappers (1:1) don't use this field.
         """
+        sys_id_type = next(
+            c.type
+            for c in self.warehouse.dataset_row_cls.sys_columns()
+            if c.name == "sys__id"
+        )
```
dreadatour, line 617: ❤️

PR #1684, opened 2026-03-24
Refactor hash tests to not use exact hash values

ilongin:


tests/unit/test_hash_utils.py
```diff
@@ -36,317 +36,113 @@ def double_arg_and_return_annot(x: int) -> int:
...
+        )
     ],
-)
-def test_hash_column_elements(expr, result):
-    assert hash_column_elements(expr) == result
-
-
-@pytest.mark.parametrize(
-    "func,expected_hash",
+    [C("age").label("user_age")],
```
shcheklein, line 53: add minor modifications of each (value, or column name)

ilongin: added

PR #1693, opened 2026-03-30
fix: Type inference in column expression

### LINKED ISSUE #1692: Type inference in column expression
### Description

This problem exist everywhere when user define expressions like `C("a") + C("b") / C("c")` and type is required: `mutate`, partition_by in `group_by`.

The issues doesn't exist when type is not required in expression: `filter`.

```python
(
    dc.read_values(num=[5, 3, 15, 3])
    .mutate(x=dc.C("num") / 10)
    .show()
)
```
Error:
```
datachain.lib.utils.DataChainColumnError: Error for column x: Cannot infer type with expression num / CAST(:num_1 AS NUMERIC)
```

It happens with many types. In some cases it throws exception in other places it can just quietly accept a wrong type which might cause issues downstream.

PS: Workaround is using columns from table - but that's too much to ask from users. The trick is `my_ds.column()` which returns a right type:
```python
my_ds = dc.read_values(num=[5, 3, 15, 3])
my_ds.mutate(x=my_ds.column("num") / 10).show()
```

dmpetrov:
Closes #1692

### DIFF
--- a/src/datachain/lib/dc/datachain.py
+++ b/src/datachain/lib/dc/datachain.py
@@ -1446,12 +1446,12 @@ class DataChain:
         for col_name, expr in kwargs.items():
             if not isinstance(expr, (*primitives, Column, Func)):
                 if isinstance(expr, ColumnElement):
-                    kwargs[col_name] = self.signals_schema.enrich_expr_types(expr)
-                    expr = kwargs[col_name]
+                    expr = self.signals_schema.enrich_expr_types(expr)
                 if isinstance(expr.type, NullType):
                     raise DataChainColumnError(
-                        col_name, f"Cannot infer type with expression {expr}"
+                        col_name, f"Cannot infer type for expression {expr}"
                     )
+                kwargs[col_name] = expr
 
         mutated = {}
         schema = self.signals_schema
--- a/src/datachain/lib/signal_schema.py
+++ b/src/datachain/lib/signal_schema.py
@@ -20,7 +20,8 @@ from typing import (
 )
 
 from pydantic import BaseModel, Field, ValidationError, create_model
-from sqlalchemy import ColumnElement
+from sqlalchemy import Cast, ColumnElement, cast
+from sqlalchemy.sql.elements import BinaryExpression, Grouping
 
 from datachain import json
 from datachain.func import literal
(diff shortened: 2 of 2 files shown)


dmpetrov:
It's ready for review.

The only issue I see - hashes. CC @ilongin

ilongin:
> It's ready for review.
> 
> The only issue I see - hashes. CC @ilongin

I did PR of refactoring hash based tests recently but I guess I missed couple of tests to refactor. Here is new PR to refactor those as well https://github.com/datachain-ai/datachain/pull/1697

src/datachain/lib/dc/datachain.py
```diff
@@ -1449,12 +1449,14 @@ def mutate(self, **kwargs) -> "Self":
...
-            if not isinstance(expr, (*primitives, Column, Func)) and isinstance(
-                expr.type, NullType
-            ):
-                raise DataChainColumnError(
-                    col_name, f"Cannot infer type with expression {expr}"
-                )
+            if not isinstance(expr, (*primitives, Column, Func)):
+                if isinstance(expr, ColumnElement):
+                    kwargs[col_name] = self.signals_schema.enrich_expr_types(expr)
+                    expr = kwargs[col_name]
```
shcheklein: minor: a bit weird order

should it be:

```
expr = self.signals_schema.enrich_expr_types(expr)
if isinstance(expr.type, NullType):
    ...
kwargs[col_name] = expr
```

also slightly modify message:

`Cannot infer type for expression` (not with probably)

src/datachain/lib/signal_schema.py
```diff
@@ -903,6 +903,33 @@ def _find_in_tree(self, path: list[str]) -> DataType:
...
         return curr_type
 
+    def enrich_expr_types(self, expr: "ColumnElement") -> "ColumnElement":
+        """Rebuild a ColumnElement expression with typed columns from the schema.
+
+        dc.C("col") creates untyped columns (NullType). This method rebuilds the
+        expression tree replacing them with typed columns so SQLAlchemy propagates
+        types correctly through operators.
+        """
+        from sqlalchemy import Cast, cast
```
shcheklein: just to review: if we have a lot of calls to this in a loop somewhere we should probably change it to global imports

PR #1697, opened 2026-03-31
Removing exact hash assertion to hash tests

ilongin:
Continuation of refactoring hash based tests from https://github.com/datachain-ai/datachain/pull/1684 (I missed couple of tests that need refactor as well in original PR).

### DIFF
--- a/tests/unit/test_datachain_hash.py
+++ b/tests/unit/test_datachain_hash.py
@@ -7,6 +7,7 @@ from pydantic import BaseModel
 import datachain as dc
 from datachain import func
 from datachain.lib.dc import C
+from tests.utils import is_sha256_hex
 
 DF_DATA = {
     "first_name": ["Alice", "Bob", "Charlie", "David", "Eva"],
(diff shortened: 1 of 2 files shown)


tests/unit/test_datachain_hash.py
```diff
@@ -131,9 +132,9 @@ def test_read_parquet(test_session, tmp_dir):
 
 
 def test_read_storage(mock_get_listing, test_session):
-    assert dc.read_storage("s3://bucket", session=test_session).hash() == (
-        "811e7089ead93a572d75d242220f6b94fd30f21def1bbcf37f095f083883bc41"
-    )
+    h1 = dc.read_storage("s3://bucket", session=test_session).hash()
+    h2 = dc.read_storage("s3://bucket", session=test_session).hash()
+    assert h1 == h2
```
dmpetrov, line 137: what exactly it tests? these seem the same 🙂

tests/unit/test_datachain_hash.py
```diff
@@ -229,40 +229,45 @@ def agg_persons(persons):
...
-            players_chain,
-            on=["persons.name"],
-            right_on=["player.name"],
-        )
-        .hash()
-    ) == "8e5cf0a718406a94c99ab7ffafc67aa5430f79a276deb1f1d87e5a5991bc56bf"
+
+    h1 = _build_chain().hash()
+    h2 = _build_chain().hash()
+    assert h1 == h2
```
dmpetrov, line 269: same

tests/unit/test_datachain_hash.py
ilongin: We are testing hash consistency - 2 calls on the same chain should return same hash.

PR #1702, opened 2026-04-02
add cast, fix order by with expressions

shcheklein:
# Fix cast expressions and datetime behavior

## Summary

- `func.cast(...)` now works as a normal public expression in `mutate()`, `filter()`, `order_by()`, and `merge()`.
- `func.cast(...)` now accepts column expressions such as `C("num") + 1`, not just plain column references.
- datetime casts now compare, merge, and materialize consistently across SQLite and ClickHouse.
- `filter()` now resolves embedded DataChain functions inside expressions.

### Cast a column expression

```python
dc.read_values(num=[1, 2, 3]).mutate(
    num_str=func.cast(C("num") + 1, str)
).order_by("num").to_values("num_str")

# ["2", "3", "4"]
```

Before: `func.cast(...)` rejected column expressions.

### Compare a casted datetime to a Python datetime literal

```python
dc.read_values(
    row_id=[1, 2, 3],
    ts_str=[
        "2024-01-01 00:00:00",
        "2024-01-02 03:04:05.123456",
        "2024-01-03 00:00:00",
    ],
).filter(
    func.cast("ts_str", datetime) >= datetime(2024, 1, 2, 0, 0, 0)
).order_by("row_id").to_values("row_id")

# [2, 3]
```

Before: this failed during expression normalization.

### Compare a datetime column to a casted datetime expression

```python
dc.read_values(
    ts=[datetime(2024, 1, 2, 3, 4, 5, 123456)],
    ts_str=["2024-01-02 03:04:05.123456"],
).filter(C("ts") == func.cast("ts_str", datetime))
```

Before: the right-hand `Func` was treated like a bound value instead of a SQL expression.

### Merge on datetime vs string-datetime keys

```python
left.merge(
    right,
    on=left.c("ts"),
    right_on=func.cast(right.c("ts_str"), datetime),
    inner=True,
)
```

Before: the SQLite path did not line up reliably with the casted datetime side.

### Datetime string round-trip

```python
dc.read_values(ts=[datetime(2024, 1, 1, 0, 0, 0)]).mutate(
    ts_str=func.cast("ts", str),
    ts_round_trip=func.cast(func.cast("ts", str), datetime),
)

# ts_str == "2024-01-01 00:00:00"
```

Before: whole-second formatting and precision were backend-dependent.

### Filter with embedded functions

```python
chain.filter(string.length("names") > 3)
chain.filter(C("parent") == func.path.parent("path"))
```

Before: `filter()` did not resolve embedded function expressions correctly.

## Validation

- functional coverage now exercises cast in `mutate()`, `filter()`, `order_by()`, and `merge()`
- datetime coverage now includes precision, round-trip, comparison, and join cases
- added unit coverage for expression operands and `filter()` function resolution
- documented `func.cast` in the function reference


src/datachain/lib/dc/utils.py
```diff
@@ -37,25 +38,44 @@ def is_local() -> bool:
...
-        ).db_signals()
-
-        for idx, arg in enumerate(args):
-            if isinstance(arg, GenericFunction):
-                resolved_args.insert(idx, arg)  # type: ignore[arg-type]
-
-        return method(self, *resolved_args, **kwargs)
+        resolved_args: list[object] = []
+
+        def resolve_expr(expr: ColumnExpr) -> ColumnExpr:
```
shcheklein, line 60: C: Needs review

src/datachain/query/dataset.py
```diff
@@ -2130,28 +2130,52 @@ def validate_expression(self, exp: "ClauseElement", q1, q2):
...
+        right_name = self.query2.table.name
+
+        self.validate_expression(exp, q1, q2)
+
+        def replace(element, **_kwargs):
+            if not isinstance(element, ColumnClause):
+                return None
+
+            table = element.table
+            assert isinstance(table, TableClause)
```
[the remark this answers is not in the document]
shcheklein: leaving it strict for now, it can't me most likely a Subquery here

src/datachain/func/func.py
```diff
@@ -471,27 +501,114 @@ def get_col(col: ColT, string_as_literal=False) -> ColT:
...
+            str(col),
+            "A dataset context is required to infer column type",
+        )
+    elif isinstance(col, str):
+        result_type = signals_schema.get_column_type(col)
+    elif isinstance(col, Column):
+        result_type = signals_schema.get_column_type(col.name)
+    elif isinstance(col, ColumnElement):
+        result_type = sql_to_python(signals_schema.enrich_expr_types(col))
+    else:
```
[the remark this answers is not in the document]
shcheklein: fixing this even though it is not exactly relevant

src/datachain/func/aggregate.py
```diff
@@ -27,6 +29,7 @@ def count(col: str | Column | None = None) -> Func:
             count1=func.count(),
             count2=func.count("signal.id"),
             count3=func.count(dc.C("signal.category")),
+            count4=func.count(dc.C("signal.id") + 1),
```
dreadatour:
```
            count4=func.count(dc.C("signal.id") + 1),
```

I may be wrong, but does it makes any sense 🤔

src/datachain/sql/types.py
```diff
@@ -37,6 +39,34 @@
 
 NullType = types.NullType
 
+_DATETIME_EXTRA_FRACTION_RE = re.compile(r"(\.\d{6})\d+")
+_DATETIME_CAST_INPUT_TYPES = frozenset({str, bytes, date, datetime})
+
+
+def parse_datetime_text(value: str) -> datetime:
+    normalized = value.replace("Z", "+00:00") if value.endswith("Z") else value
+    return datetime.fromisoformat(_DATETIME_EXTRA_FRACTION_RE.sub(r"\1", normalized))
```
dreadatour: This is a bit sketchy IMO :(

dreadatour (approved):
looks good to me overall 👍
Couple tiny comments.

src/datachain/sql/types.py
shcheklein: probably fine, but trying to simplify it

PR #1704, opened 2026-04-02
refactor(catalog): add remove_dataset_versions 

amritghimire:
## Summary

- Add \`remove_dataset_versions(job_id, version_ids)\` on \`Catalog\` as a unified method for removing dataset versions by explicit metastore row IDs and/or job ID.
- Add \`get_dataset_versions(job_id, version_ids)\` as a new abstract method on \`AbstractMetastore\`, implemented in \`AbstractDBMetastore\`, replacing the previous \`list_dataset_versions_by_ids\`.
- Extract \`_remove_versions\` helper on \`Catalog\` — shared deletion loop used by both \`cleanup_dataset_versions\` and \`remove_dataset_versions\`.
- Extract \`_dataset_version_query_base\` and \`_fetch_version_pairs\` helpers on \`AbstractDBMetastore\` — eliminate duplicated SELECT/JOIN setup and row-parsing between the two query methods.

### \`catalog.py\`
- \`remove_dataset_versions(job_id, version_ids)\`: new public method replacing \`remove_by_version_ids\`. Accepts optional \`job_id\` and/or \`version_ids\` filters; at least one must be provided.
- \`cleanup_dataset_versions\`: unchanged semantics; now delegates to shared \`_remove_versions\` helper.
- \`_remove_versions\`: private helper encapsulating the error-tolerant deletion loop, shared by both removal methods.
- Removed \`_remove_dataset_version_pairs\` (superseded by \`_remove_versions\`).

### \`metastore.py\`
- \`get_dataset_versions(job_id, version_ids)\`: new abstract method + DB implementation replacing \`list_dataset_versions_by_ids\`. Filters by \`job_id\` and/or \`version_ids\` (AND logic); raises \`ValueError\` if neither is provided.
- \`_dataset_version_query_base()\`: private helper returning shared SELECT columns and base JOIN across namespaces/projects/datasets/versions.
- \`_fetch_version_pairs(query)\`: private helper executing a query and parsing rows into \`(DatasetRecord, version_string)\` tuples.
- \`get_dataset_versions_to_clean\`: unchanged GC logic; now uses the two shared helpers.
- Removed \`_dataset_versions_select_cols_and_base_join\` and \`_dataset_query_rows_to_version_pairs\` (superseded by the renamed helpers above).

### \`tests/unit/test_dataset_status_management.py\`
- Updated tests to use \`remove_dataset_versions(version_ids=[...])\` in place of \`remove_by_version_ids\`.
- Removed tests for \`include_complete\` mode (feature not needed).

## Testing

- [x] All tests in \`test_dataset_status_management.py\` pass
- [x] Pre-commit hooks (ruff, mypy, codespell) pass

### DIFF
--- a/src/datachain/data_storage/metastore.py
+++ b/src/datachain/data_storage/metastore.py
@@ -36,7 +36,6 @@ from sqlalchemy import (
     select,
 )
 from sqlalchemy.sql import func as f
-from sqlalchemy.sql.elements import ColumnElement
 from sqlalchemy.sql.selectable import Join
 
 from datachain import json
@@ -80,6 +79,7 @@ STALE_CREATED_THRESHOLD_HOURS = 1
 if TYPE_CHECKING:
     from sqlalchemy import CTE, Delete, Insert, Select, Subquery, Update
     from sqlalchemy.schema import SchemaItem
+    from sqlalchemy.sql.elements import ColumnElement
 
     from datachain.data_storage import schema
     from datachain.data_storage.db_engine import DatabaseEngine
@@ -1757,7 +1757,7 @@ class AbstractDBMetastore(AbstractMetastore):
 
     def _dataset_version_query_base(
         self,
-    ) -> tuple[tuple[ColumnElement[Any], ...], Join]:
+    ) -> tuple[tuple["ColumnElement[Any]", ...], Join]:
         n = self._namespaces
         p = self._projects
         d = self._datasets


src/datachain/data_storage/metastore.py
```diff
@@ -1759,6 +1782,38 @@ def get_dataset_versions_to_clean(
             .join(d, p.c.id == d.c.project_id)
             .join(dv, d.c.id == dv.c.dataset_id)
         )
+        return select_cols, base_from
+
+    def _dataset_query_rows_to_version_pairs(
```
shcheklein: why do we need this new abstraction pairs? can we just create dataset record with version records inside?

src/datachain/data_storage/metastore.py
```diff
@@ -351,6 +357,26 @@ def get_dataset_versions_to_clean(
             contains only one version (the incomplete version to clean).
         """
 
+    @abstractmethod
+    def list_dataset_versions_by_ids(
```
shcheklein: why do we need both get_dataset_versions_to_clean and list_dataset_versions_by_ids ?

src/datachain/catalog/catalog.py
```diff
@@ -1141,9 +1130,57 @@ def cleanup_dataset_versions(self, job_id: str | None = None) -> int:
...
+    ) -> int:
+        """
+        Clean up dataset versions that are no longer needed.
+
+        By default removes dataset versions that:
+        - Have status CREATED, FAILED, STALE, or REMOVING
+        - Belong to completed/failed/canceled jobs (not running)
+        - Are session_* datasets from finished jobs (orphaned intermediates)
+
+        With job_id set and include_complete=True, removes every dataset version
```
shcheklein: can't we use remove_by_version_ids in this case?

src/datachain/catalog/catalog.py
```diff
@@ -1141,9 +1130,57 @@ def cleanup_dataset_versions(self, job_id: str | None = None) -> int:
...
+
+    def remove_by_version_ids(
+        self,
+        versions: Sequence[int],
+        job_id: str | None = None,
+    ) -> int:
+        """
+        Remove dataset versions by metastore row id (datasets_versions.id).
+
+        Optional job_id restricts to versions whose job_id column matches.
```
ilongin: to me it's strange to have exact version ids to remove and then job_id to restrict them to specific ones.

src/datachain/data_storage/metastore.py
```diff
@@ -1819,16 +1874,29 @@ def get_dataset_versions_to_clean(
...
-                version = dataset.versions[0].version
-                results.append((dataset, version))
+        return self._dataset_query_rows_to_version_pairs(query)
 
-        return results
+    def list_dataset_versions_by_ids(
+        self,
+        version_ids: Sequence[int],
+        *,
+        job_id: str | None = None,
```
ilongin: again, strange additional filter. Usually if we have exact ids to do some action upon, we don't do more filters.

src/datachain/catalog/catalog.py
```diff
@@ -1141,9 +1130,57 @@ def cleanup_dataset_versions(self, job_id: str | None = None) -> int:
...
+
+        Returns:
+            Number of removed versions
+        """
+        if not versions:
+            return 0
+
+        versions_to_remove = self.metastore.list_dataset_versions_by_ids(
```
ilongin: this step seems like it shouldn't been needed. If you have already ids to remove, why need for fetching exact objects? Can this be refactored somehow

src/datachain/data_storage/metastore.py
```diff
@@ -1736,14 +1762,11 @@ def remove_dataset_version(
         dataset.remove_version(version)
         return dataset
 
-    def get_dataset_versions_to_clean(
-        self, job_id: str | None = None
-    ) -> list[tuple[DatasetRecord, str]]:
+    def _dataset_versions_select_cols_and_base_join(self) -> tuple:
```
ilongin: I don't like this method name. It seems too low level.

src/datachain/catalog/catalog.py
```diff
@@ -1102,6 +1102,30 @@ def remove_dataset_version(
...
+                self.remove_dataset_version(dataset, version)
+                num_removed += 1
+            except Exception as e:  # noqa: BLE001
+                logger.warning(
+                    "Failed to remove dataset %s version %s: %s",
+                    dataset.name,
+                    version,
+                    e,
+                )
+        return num_removed
```
amritghimire, line 1118: Extracted to avoid duplicate code.

src/datachain/catalog/catalog.py
```diff
@@ -1102,6 +1102,30 @@ def remove_dataset_version(
...
+
+    def remove_dataset_versions(
+        self, job_id: str | None = None, version_ids: list[int] | None = None
+    ) -> int:
```
amritghimire, line 1122: Entrypoint, Just pass version ids and job id.

src/datachain/data_storage/metastore.py
```diff
@@ -1736,14 +1754,11 @@ def remove_dataset_version(
         dataset.remove_version(version)
         return dataset
 
-    def get_dataset_versions_to_clean(
-        self, job_id: str | None = None
-    ) -> list[tuple[DatasetRecord, str]]:
+    def _dataset_version_query_base(self) -> tuple:
```
amritghimire: Extracted to avoid duplication

src/datachain/data_storage/metastore.py
```diff
@@ -351,6 +357,26 @@ def get_dataset_versions_to_clean(
...
+        When job_id is set, only rows with that job_id are returned.
+
+        Returns:
+            List of (DatasetRecord, version_string) tuples, same shape as
+            get_dataset_versions_to_clean. Each record contains only one version;
+            its row id is dataset.versions[0].id. Order is undefined; duplicate
+            ids in version_ids still yield at most one row per id. Missing ids are
+            omitted.
+        """
+
```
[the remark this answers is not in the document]
amritghimire: We will be adding it in studio code base too and ci will detect if missed.

src/datachain/catalog/catalog.py
```diff
@@ -1141,9 +1130,57 @@ def cleanup_dataset_versions(self, job_id: str | None = None) -> int:
...
+        Optional job_id restricts to versions whose job_id column matches.
+
+        Returns:
+            Number of removed versions
+        """
+        if not versions:
+            return 0
+
+        versions_to_remove = self.metastore.list_dataset_versions_by_ids(
+            versions, job_id=job_id
```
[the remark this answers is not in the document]
amritghimire: Updated in new change.

src/datachain/data_storage/metastore.py
amritghimire: Updated, but basically the version_ids and job_id is to allow user with two options: 

- Delete all dataset versions for a job
- Delete selected dataset versions from a job (version id)

src/datachain/catalog/catalog.py
amritghimire: We need exact objects to pass to remove dataset version it needs dataset record and version to find warehouse table.

src/datachain/data_storage/metastore.py
amritghimire: Updated.

src/datachain/catalog/catalog.py
```diff
@@ -1102,6 +1102,30 @@ def remove_dataset_version(
...
+        return num_removed
+
+    def remove_dataset_versions(
+        self, job_id: str | None = None, version_ids: list[int] | None = None
+    ) -> int:
+        versions_to_remove = self.metastore.get_dataset_versions(
+            job_id=job_id,
+            version_ids=version_ids,
+        )
+        return self._remove_versions(versions_to_remove)
```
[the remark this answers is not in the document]
amritghimire: Updated PR description.

src/datachain/catalog/catalog.py
```diff
@@ -1127,22 +1151,7 @@ def cleanup_dataset_versions(self, job_id: str | None = None) -> int:
...
-            except Exception as e:  # noqa: BLE001
-                logger.warning(
-                    "Failed to clean dataset %s version %s: %s",
-                    dataset.name,
-                    version,
-                    e,
-                )
-
-        return num_removed
+        return self._remove_versions(versions_to_clean)
```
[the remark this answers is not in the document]
amritghimire: Updated description.

src/datachain/data_storage/metastore.py
```diff
@@ -1819,16 +1851,26 @@ def get_dataset_versions_to_clean(
...
+        dv = self._datasets_versions
+
+        query = self._datasets_select(*select_cols).select_from(base_from)
+
+        if job_id:
+            query = query.where(dv.c.job_id == job_id)
+
+        if version_ids:
+            query = query.where(dv.c.id.in_(version_ids))
+
```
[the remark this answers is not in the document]
amritghimire: Empty version ids and empty job id is not allowed since we already have get all versions to get through get_datasets.

src/datachain/data_storage/metastore.py
amritghimire: The remove dataset versions expect dataset record and version string pair as already existing garbage collect implements. This just extracts it to separate function, thats all. This is outdated in recent changes.

src/datachain/data_storage/metastore.py
amritghimire: `get_dataset_versions_to_clean` and `get_dataset_versions` (renamed) seems like completely different logic. I have updated the pull request a bit to use remove_dataset_versions only. But basically `to_clean` one seems more specific to garbage collect.

src/datachain/catalog/catalog.py
amritghimire: Updated.

src/datachain/catalog/catalog.py
amritghimire: Updated, but basically the version_ids and job_id is to allow user with two options:

- Delete all dataset versions for a job
- Delete selected dataset versions from a job (version id)

shcheklein:
@ilongin please review again (and Studio PR)

src/datachain/catalog/catalog.py
dreadatour: From first view this looks like we are adding additional specific method for quite narrow use case. Worst case scenario: we may end up with hundreds methods like this for specific tasks, and it will be hard to maintain code like this.
We also have `cleanup_dataset_versions` method below doing the same but different job (quite narrow again), but with other input params.
Should we make it more flexible may be and refactor this to be one flexible universal method? I am not 100% sure, but it looks like it would be better in terms of API.

src/datachain/data_storage/metastore.py
```diff
@@ -351,6 +351,24 @@ def get_dataset_versions_to_clean(
...
+    def get_dataset_versions(
+        self, job_id: str | None = None, version_ids: list[int] | None = None
+    ) -> list[tuple[DatasetRecord, str]]:
+        """
+        Get dataset versions by job ID and/or version IDs.
+
+        Returns dataset versions filtered by job_id and/or version_ids:
+        - If job_id is provided, returns versions for that job.
+        - If version_ids are provided, returns versions matching those IDs.
+        - If both are provided, returns versions matching both filters.
```
dreadatour, line 365: What do you think of using `overload` for better type hinting?

Something like this:

```
from typing import overload

# Case 1: only job_id provided
@overload
def get_dataset_versions(self, job_id: str, version_ids: None = None) -> list[tuple[DatasetRecord, str]]: ...

# Case 2: only version_ids provided
@overload
def get_dataset_versions(self, job_id: None = None, *, version_ids: list[int]) -> list[tuple[DatasetRecord, str]]: ...

# Case 3: both provided
@overload
def get_dataset_versions(self, job_id: str, version_ids: list[int]) -> list[tuple[DatasetRecord, str]]: ...

@abstractmethod
def get_dataset_versions(
    self, job_id: str | None = None, version_ids: list[int] | None = None
) -> list[tuple[DatasetRecord, str]]:
    """..."""
```

src/datachain/data_storage/metastore.py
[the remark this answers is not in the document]
dreadatour: > Empty version ids and empty job id is not allowed since we already have get all versions to get through get_datasets.

However, this is a public metastore method, so it should not be designed only around the way we currently use it. It may be used in other scenarios in the future, and in those cases empty version IDs or an empty job ID could happen and cause failures.

In general, public methods like this should be treated as isolated, self-contained pieces of code, without relying on assumptions about how they are currently used elsewhere in the project.

src/datachain/data_storage/metastore.py
```diff
@@ -1819,16 +1851,26 @@ def get_dataset_versions_to_clean(
...
-            # Each DatasetRecord has one version (the failed one from this row)
-            if dataset.versions:
-                version = dataset.versions[0].version
-                results.append((dataset, version))
+        return self._fetch_version_pairs(query)
 
-        return results
+    def get_dataset_versions(
+        self, job_id: str | None = None, version_ids: list[int] | None = None
+    ) -> list[tuple[DatasetRecord, str]]:
```
dreadatour, line 1861: It looks redundant to have two specific methods: `get_dataset_versions` and `get_dataset_versions_to_clean`.

Considering `get_dataset_versions` is quite universal (all params are optional), should we make it more flexible and also cover the cases `get_dataset_versions_to_clean` covers?

amritghimire: Given that we need very complex query and filtering to find the versions to clean like going through job status, and other filter, I believe `get_dataset_versions_to_clean` makes sense, it taps into that specific scenario, where as where we need a simple filter or where entry, get_dataset_versions can be clean and universal. 

So, lets keep `get_dataset_versions_to_clean` as it is and use this `get_dataset_versions` as a universal one IMO.

src/datachain/catalog/catalog.py
```diff
@@ -1102,6 +1102,30 @@ def remove_dataset_version(
...
+            try:
+                self.remove_dataset_version(dataset, version)
+                num_removed += 1
+            except Exception as e:  # noqa: BLE001
+                logger.warning(
+                    "Failed to remove dataset %s version %s: %s",
+                    dataset.name,
+                    version,
+                    e,
+                )
```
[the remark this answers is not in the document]
amritghimire: This should be fine, remove implies clean and should be okay.

src/datachain/data_storage/metastore.py
amritghimire: Not sure, I mean, the return type is always same, and I just added a new case, where both job id can be none, too, so  the current single-signature approach is cleaner and less complex IMO.

src/datachain/data_storage/metastore.py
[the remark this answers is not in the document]
amritghimire: Updated.

src/datachain/catalog/catalog.py
amritghimire: https://github.com/datachain-ai/datachain/pull/1704#discussion_r3038910900

src/datachain/data_storage/metastore.py
```diff
@@ -36,6 +36,8 @@
     select,
 )
 from sqlalchemy.sql import func as f
+from sqlalchemy.sql.elements import ColumnElement
```
dreadatour: `ColumnElement` is only used as return type here, not sure why we we move it from `TYPE_CHECKING` block to top-level import.

src/datachain/data_storage/metastore.py
```diff
@@ -1819,16 +1854,23 @@ def get_dataset_versions_to_clean(
...
+        select_cols, base_from = self._dataset_version_query_base()
+        dv = self._datasets_versions
+
+        query = self._datasets_select(*select_cols).select_from(base_from)
+
+        if job_id:
+            query = query.where(dv.c.job_id == job_id)
+
+        if version_ids:
+            query = query.where(dv.c.id.in_(version_ids))
```
dreadatour: I still think this should be like this:

```suggestion
        if version_ids is not None:
            query = query.where(dv.c.id.in_(version_ids))
```

This is quite straightforward and may leads to unexpected bugs. Empty list is a correct and legitimate value, same as `None`.

Let's think about result of `get_dataset_versions` based on input params. Imagine we use this `get_dataset_versions` method somewhere else, where based on user input or some other input we pass empty list to `get_dataset_versions` and as a result we are awaiting empty list as an output. But current implementation will ignore `version_ids` param and return ALL versions.

Same for my comment in SaaS PR.

src/datachain/data_storage/metastore.py
```diff
@@ -1819,16 +1854,23 @@ def get_dataset_versions_to_clean(
...
+    def get_dataset_versions(
+        self, job_id: str | None = None, version_ids: list[int] | None = None
+    ) -> list[tuple[DatasetRecord, str]]:
+        select_cols, base_from = self._dataset_version_query_base()
+        dv = self._datasets_versions
+
+        query = self._datasets_select(*select_cols).select_from(base_from)
+
+        if job_id:
+            query = query.where(dv.c.job_id == job_id)
```
dreadatour: Same as for `version_ids` check. This should be:

```suggestion
        if job_id is not None:
            query = query.where(dv.c.job_id == job_id)
```

`job_id` equal to empty string should return nothing, empty string is a value, it is not the same as `None`.

PR #1717, opened 2026-04-08
feat(cli): add datachain bucket status command

amritghimire:
## Summary

Closes #1715

Adds `datachain bucket status <uri>` CLI command and `bucket_status()` Python API to check bucket existence and access level without listing objects. Supports S3, GCS, and Azure.

## Behavior

Exits 0 if the bucket exists, 1 if not found.

**S3**
```bash
# Public bucket
$ datachain bucket status s3://datachain-public
Status: exists
Access: anonymous

# Private bucket, no credentials
$ datachain bucket status s3://amrit-datachain-test
Status: exists
Access: denied
Error: Access denied to S3 bucket 'amrit-datachain-test' — check AWS credentials/permissions

# Non-existent bucket
$ datachain bucket status s3://amrit-does-not-exist
Status: not found
Error: S3 bucket 'amrit-does-not-exist' not found

# With correct credentials 
$ datachain bucket status s3://amrit-datachain-test
Status: exists
Access: authenticated
```

**GCS**
```bash
# Public bucket
$ datachain bucket status gs://datachain-demo
Status: exists
Access: anonymous

# Private bucket, no credentials
$ datachain bucket status gs://thomas-datachain-test
Status: exists
Access: denied
Error: Access denied to GCS bucket 'thomas-datachain-test' — check credentials/permissions

# Non-existent bucket
$ datachain bucket status gs://amrit-does-not-exists
Status: not found
Error: GCS bucket 'amrit-does-not-exists' not found

# Private bucket, authenticated
$ datachain bucket status gs://amrit-datachain-test
Status: exists
Access: authenticated

# Private bucket, credentials present but insufficient
$ datachain bucket status gs://thomas-datachain-test
Status: exists
Access: denied
Error: Forbidden: b/thomas-datachain-test/o ...
```

**Azure**
```bash
# No creds, no account name
$ datachain bucket status az://thomas-datachain-test
Status: not found
Error: unable to connect to account for Must provide either a connection_string or account_name with credentials!!

# Account name, no creds
$  datachain bucket status az://thomas-datachain-test --account-name datachain
Status: exists
Access: denied
Error: Access denied to Azure container 'thomas-datachain-test' — check credentials/configuration

# Authenticated
$ datachain bucket status az://amrit-test-az
Status: exists
Access: authenticated
```

## Python API

```python
from datachain.client import bucket_status

status = bucket_status("s3://my-bucket/")
print(status.exists)  # True / False
print(status.access)  # 'anonymous' | 'authenticated' | 'denied'
print(status.error)   # error message or None
```

## Implementation notes

- **S3/GCS**: probes anonymous access first, falls back to authenticated
- **Azure**: probes authenticated first (DefaultAzureCredential), falls back to anonymous — avoids false `anonymous` results when credentials are present in the environment
- GCS catches `google.api_core.exceptions.Forbidden` (not mapped to `PermissionError`) via a broad fallback handler
- `BucketStatus` is exported from the top-level `datachain` namespace

## Testing

- 17 unit tests covering all providers and access scenarios (mock-based)
- 2 functional tests against a live S3-compatible server
- 3 CLI parser tests


ilongin:
is it just me or `Status: exists` sounds little bit strange? ... should exsistance be one of the statuses? It feels like it should be communicated differently e.g with some error ... in console it should maybe just show message like "Bucket doesn't exist" ... on the other hand when bucket exists then this status is not needed to be shown as it will be clear by default.

dmpetrov (commented):
@amritghimire The path component in the URI is silently ignored right now - `s3://my_bucket/my_dir/dir1` only checks `my_bucket` and ignores the dirs. It should give an error when a path is presented.

Why? There is a good chance this functionality will be extended to file/directory existence checks in the future (like `test -e` / `test -f` / `test -d` in Unix), and silently discarding the path will make this comand backward incompatible.

src/datachain/client/__init__.py
```diff
@@ -16,7 +16,12 @@ def bucket_status(uri: str, **client_config) -> BucketStatus:
         'anonymous', 'authenticated', 'denied'
     """
     client_cls = Client.get_implementation(uri)
-    name, _ = client_cls.split_url(uri)
+    name, path = client_cls.split_url(uri)
+    if path:
+        raise ValueError(
+            f"URI must not contain a path component, got: {uri!r}. "
+            "Use just the bucket/container URI, e.g. s3://my-bucket/"
```
dmpetrov: not the best message. path or bucket? Also, it's too long.

Use this: "path in a bucket is not allowed, only bucket name: {uri!r}"

dmpetrov (approved):
LG but I didn't get deeper in the code

tests/unit/test_client_bucket_status.py
```diff
@@ -0,0 +1,341 @@
...
+        "gs://my-bucket/dir",
+        "az://my-container/blob",
+    ],
+)
+def test_bucket_status_rejects_path_component(uri):
+    with pytest.raises(ValueError, match="must not contain a path component"):
+        bucket_status(uri)
+
+
+# ---------------------------------------------------------------------------
```
shcheklein: cleanup - we don't use this style

tests/func/test_bucket_status.py
```diff
@@ -0,0 +1,23 @@
+import pytest
+
+from datachain.client import BucketStatus, bucket_status
+
+
+@pytest.mark.parametrize("cloud_type", ["s3"], indirect=True)
```
shcheklein: why only s3?

tests/unit/test_client_bucket_status.py
```diff
@@ -0,0 +1,341 @@
...
+# S3
+# ---------------------------------------------------------------------------
+
+
+@patch("datachain.client.s3.S3FileSystem")
+@patch("datachain.client.s3.sync")
+def test_s3_anonymous(mock_sync, mock_s3fs_cls):
```
shcheklein: we don't need these s3 tests if we have func tests already ... prefer those func tests on top of mock FSs

src/datachain/client/azure.py
```diff
@@ -1,18 +1,71 @@
...
+            return BucketStatus(exists=True, access="authenticated")
+        except PermissionError:
+            return BucketStatus(
+                exists=True,
+                access="denied",
+                error=f"Access denied to Azure container '{name}'"
+                " — check credentials/configuration",
+            )
+        except FileNotFoundError:
+            return BucketStatus(
```
shcheklein, line 80: have you tested it across all clouds? does it distinguish not found and access denied reliable for all situations?

src/datachain/client/azure.py
```diff
@@ -1,18 +1,71 @@
...
+        # is not available (az:// URIs don't carry the account name).
+        try:
+            auth_fs = cls.create_fs(**kwargs)
+            sync(get_loop(), auth_fs._info, name)
+            return BucketStatus(exists=True, access="authenticated")
+        except PermissionError:
+            return BucketStatus(
+                exists=True,
+                access="denied",
+                error=f"Access denied to Azure container '{name}'"
```
shcheklein, line 76: can this error happen if bucket allows `anon` access but we picked up some credentials that are not compatible? test this situation ... what should we do in such cases?

src/datachain/client/azure.py
```diff
@@ -1,18 +1,71 @@
...
+        except FileNotFoundError:
+            return BucketStatus(
+                exists=False,
+                access="denied",
+                error=f"Azure container '{name}' not found",
+            )
+        except ValueError:
+            pass  # No credentials configured; fall through to anonymous probe.
+
+        # Step 2: Anonymous probe — only reached when no credentials are configured.
```
shcheklein: > only reached when no credentials are configured.

is it really okay?

src/datachain/client/fsspec.py
```diff
@@ -57,6 +57,12 @@ class Bucket(NamedTuple):
     created: datetime | None
 
 
+class BucketStatus(NamedTuple):
+    exists: bool
+    access: Literal["anonymous", "authenticated", "denied"]
```
shcheklein, line 62: so, we don't really distinguish denied and anonymous, right? is it okay?

amritghimire: Why not? anonymous means it is accessed successfully without creds. Hence anon flag, otherwise it will reach next step.

src/datachain/client/azure.py
amritghimire: Yes. I have added all the scripts tested and results in pull request body itself. 

```



# S3

# Anonymous access
(.venv) ➜  datachain  (git):(main) ✗ datachain bucket status s3://datachain-public    
Status: exists
Access: anonymous

# No creds, private bucket
(.venv) ➜  datachain  (git):(main) ✗ datachain bucket status s3://amrit-datachain-test
Status: exists
Access: denied
Error: Access denied to S3 bucket 'amrit-datachain-test' — check AWS credentials/permissions


# No creds, incorrect bucket
(.venv) ➜  datachain  (git):(main) ✗ datachain bucket status s3://amrit-does-not-exist
Status: not found
Error: S3 bucket 'amrit-does-not-exist' not found



# gs

# Anonymous gcp
(.venv) ➜  datachain  (git):(main) ✗ datachain bucket status gs://datachain-demo        
Status: exists
Access: anonymous

# No creds, private bucket
(.venv) ➜  datachain  (git):(main) ✗ datachain bucket status gs://thomas-datachain-test 
Status: exists
Access: denied
Error: Access denied to GCS bucket 'thomas-datachain-test' — check credentials/permissions

# No creds, incorrect bucket
(.venv) ➜  datachain  (git):(main) ✗ datachain bucket status gs://amrit-does-not-exists 
Status: not found
Error: GCS bucket 'amrit-does-not-exists' not found

# Has creds, private bucket, has permission
(.venv) ➜  datachain  (git):(main) ✗ datachain bucket status gs://amrit-datachain-test 
Status: exists
Access: authenticated

# Has creds, private bucket, but no permission
(.venv) ➜  datachain  (git):(main) ✗ datachain bucket status gs://thomas-datachain-test 
Status: exists
Access: denied
Error: Forbidden: b/thomas-datachain-test/o
amrit-test-dc@iterative-sandbox.iam.gserviceaccount.com does not have storage.objects.list access to the Google Cloud Storage bucket. Permission 'storage.objects.list' denied on resource (or it may not exist).



# az

# No creds, no permission
(.venv) ➜  datachain  (git):(main) ✗ datachain bucket status az://thomas-datachain-test
Status: exists
Access: denied
Error: unable to connect to account for Must provide either a connection_string or account_name with credentials!!
(.venv) ➜  datachain  (git):(main) ✗ 

# Creds, with permission
(.venv) ➜  datachain  (git):(main) ✗ datachain bucket status az://amrit-test-az
Status: exists
Access: authenticated

```

src/datachain/client/fsspec.py
shcheklein: bc bucket might allow anon access but fail with credentials ... or do we have to use anon flag in that case? I'm not sure this is convenient / expected for this command

amritghimire: Thats a weird condition if it is possible with anon but fail with creds. Still, if anonymous passed, it will return anonymous and agents will use anon flag in that case. Thats the usecase anyway.

shcheklein: > Thats a weird condition if it is possible with anon but fail with creds

that's how clouds work AFAIR

if you have some default creds and they are picked up but also not compatible with anon bucket - access will be denied. Test it across different clouds.

> Still, if anonymous passed

it won't even come to that in this case, right?

amritghimire: We try anonymous access first and then go to authenticated access, except azure due to account name requirements. I will try to make azure one too consistent in approach.

src/datachain/client/azure.py
amritghimire: Verified and it works as illustrated in pull request body

src/datachain/client/azure.py
amritghimire: Updated

src/datachain/client/fsspec.py
amritghimire: Updated.

src/datachain/client/__init__.py
```diff
@@ -1,3 +1,25 @@
...
 
-__all__ = ["Client"]
+
+def bucket_status(uri: str, **client_config) -> BucketStatus:
+    """Check bucket existence and access level without listing objects.
+
+    Args:
+        uri: Bucket URI, e.g. "s3://my-bucket/", "gs://my-bucket/", "az://my-container/"
+        **client_config: Storage client configuration (anon, aws_key, etc.)
+            For Azure, pass ``account_name`` to enable anonymous access detection.
```
shcheklein: should we enforce it for Azure? always expect account name?

src/datachain/client/azure.py
```diff
@@ -1,18 +1,90 @@
...
+        anon_only = bool(kwargs.get("anon"))
+
+        # Step 1: Anonymous probe — uses BlobServiceClient directly (not adlfs)
+        # to avoid picking up credentials from environment variables like
+        # AZURE_STORAGE_CONNECTION_STRING.
+        account_name = kwargs.get("account_name")
+        if account_name:
+            try:
+                url = f"https://{account_name}.blob.core.windows.net"
+                anon_client = BlobServiceClient(account_url=url)
```
shcheklein, line 32: have you tested it? are you sure it is not picking up some credentials automatically inside the library? can we double check this

src/datachain/client/azure.py
```diff
@@ -1,18 +1,90 @@
...
 
 
 class AzureClient(Client):
     FS_CLASS = AzureBlobFileSystem
     PREFIX = "az://"
     protocol = "az"
 
+    @classmethod
+    def bucket_status(cls, name: str, **kwargs) -> BucketStatus:  # noqa: PLR0911
+        anon_only = bool(kwargs.get("anon"))
```
shcheklein: so, why do we need anon_only? what is the use case? I think with the current logic it should report back the actual state for both anon and in general?

src/datachain/client/gcs.py
```diff
@@ -36,6 +37,69 @@ def create_fs(cls, **kwargs) -> GCSFileSystem:
...
+                )
+            if anon_only:
+                return BucketStatus(
+                    exists=True,
+                    access="denied",
+                    error=f"Access denied to GCS bucket '{name}'"
+                    " — check credentials/permissions",
+                )
+
+        # Step 2: Try with configured credentials
```
shcheklein: this is a bit misleading I think ... "use fsspec that will try the crendetials in some order"

tests/unit/test_client_bucket_status.py
```diff
@@ -0,0 +1,305 @@
...
+)
+def test_bucket_status_rejects_path_component(uri):
+    with pytest.raises(ValueError, match="path in a bucket is not allowed"):
+        bucket_status(uri)
+
+
+@patch("datachain.client.s3.S3FileSystem")
+@patch("datachain.client.s3.sync")
+def test_s3_anon_only_kwarg_anonymous(mock_sync, mock_s3fs_cls):
+    anon_fs = MagicMock()
```
shcheklein: I hope we don't have to use mock - we still can use proper mock file systems - I think they support auth

src/datachain/client/azure.py
amritghimire: No because we first check anon access, and verified that it doesn't pickup anon creds before this.

src/datachain/client/__init__.py
amritghimire: Updated

src/datachain/client/azure.py
amritghimire: Yes, hence we had to select blob service client here, so it doesn't pass those value. It doesn't pick up credentials this way. 

Verified properly locally.

src/datachain/client/azure.py
amritghimire: Removed.

src/datachain/client/gcs.py
amritghimire: updated

tests/unit/test_client_bucket_status.py
amritghimire: updated

src/datachain/client/azure.py
```diff
@@ -1,18 +1,70 @@
...
+    @classmethod
+    def bucket_status(cls, name: str, **kwargs) -> BucketStatus:
+        # Step 1: Anonymous probe — uses BlobServiceClient directly (not adlfs)
+        # to avoid picking up credentials from environment variables like
+        # AZURE_STORAGE_CONNECTION_STRING.
+        account_name = kwargs.get("account_name")
+        if account_name:
+            try:
+                url = f"https://{account_name}.blob.core.windows.net"
+                anon_client = BlobServiceClient(account_url=url)
```
shcheklein, line 32: Just to double check - how we build URL - is it only bucket name?

src/datachain/cli/parser/__init__.py
```diff
@@ -506,6 +506,41 @@ def get_parser() -> ArgumentParser:  # noqa: PLR0915
...
+            "Bucket URI to check, e.g. s3://my-bucket/, gs://my-bucket/, "
+            "az://my-container/."
+        ),
+    )
+    parse_bucket_status.add_argument(
+        "--account-name",
+        dest="account_name",
+        type=str,
+        default=None,
+        help="Azure storage account name (required for anonymous access detection).",
```
shcheklein, line 541: is it required for non anon access check also? in some cases I think key doesn't include account name

tests/unit/test_client_bucket_status.py
shcheklein, line 1: We still have a lot of mocks here

I hope we can do most of the tests using mock FS implementations instead

src/datachain/client/azure.py
amritghimire: Yes, if not we raise error as requested by dmitry in one of the comment.

src/datachain/cli/parser/__init__.py
amritghimire: I would leave as it is. In case key doesn't include account name, it raises error saying the same.

tests/unit/test_client_bucket_status.py
amritghimire: It was not possible because I wanted to cover all edge cases like all types of exceptions which was only possible with mock. They are parametrized as much as possible. 
One most repeating mock is `MagicMock(bucket_cmd="status", uri="s3://b", account_name=None)` which is just to simulate the call to args for easy tests with cli arguments.

amritghimire:
cc. @shcheklein

shcheklein (approved):

PR #1718, opened 2026-04-08
Making `enlist_source` using ephemeral chain and add check for job to pre-exist if in studio

### LINKED ISSUE #1708: fix: enlist_source() creates unintended jobs in server contexts
## Problem

`catalog.enlist_source()` calls `read_storage().exec()` without ephemeral mode (`catalog.py:591`). This triggers `.save()` → `session.get_or_create_job()`, which creates a job named after `sys.argv[0]` of the host process.

In server environments (e.g. behind gunicorn/uvicorn), this produces jobs with names like `/usr/local/bin/gunicorn` and the process entry point script as the query.

## Fix

1. **Direct fix**: Add `.settings(ephemeral=True)` to the `read_storage()` call in `enlist_source()` — it's an internal catalog operation and should not create jobs.

2. **Safety net**: Guard `get_or_create_job()` in `session.py` to prevent local job creation in SaaS environments (when `is_studio()` is True and `DATACHAIN_JOB_ID` is not set), raising an error instead of silently creating jobs.

ilongin:
`enlist_source()` calls `read_storage().exec()` without ephemeral mode, which triggers `get_or_create_job()` and creates jobs named after `sys.argv[0]` (e.g. `/usr/local/bin/gunicorn` in server contexts).                         
                                                                               
- Add `.settings(ephemeral=True)` to the `read_storage()` call in `enlist_source()` to prevent job creation for this internal operation
- Add guard in `get_or_create_job()` that raises an error when `is_studio()`  is True but `DATACHAIN_JOB_ID` is not set, catching any future code paths that accidentally trigger job creation in Studio
- Change `is_studio` test fixture default from `True` to `False` — the previous default caused all tests to run as if in Studio, which was masking issues. Tests that specifically need Studio behavior now explicitly opt in via `@pytest.mark.parametrize("is_studio", [True])` and use a `studio_job` fixture to pre-create a job
- Replace `dc.create_project()` calls in test fixtures/setup with direct `metastore.create_project()` to avoid the `is_studio` dependency


tests/func/test_datasets.py
```diff
@@ -476,7 +487,7 @@ def test_move_dataset(
         assert table_row_count(catalog.warehouse.db, new_table_name) == 3
 
 
-def test_move_dataset_then_save_into(test_session):
+def test_move_dataset_then_save_into(test_session, old_new_projects):
```
shcheklein, line 490: how did it work w/o this fixture before?

tests/func/test_delta.py
```diff
@@ -477,12 +479,8 @@ def test_delta_update_unsafe_matches_same_named_aligned_sources_by_full_identity
     result_name = f"same_name_delta_result_{suffix}"
     processed: list[int] = []
 
-    dc.create_project(
```
shcheklein, line 480: why this change?

tests/unit/lib/test_datachain.py
```diff
@@ -4266,7 +4266,9 @@ def test_save_to_default_project(test_session, is_studio):
 
 
 @pytest.mark.parametrize("is_studio", [True, False])
-def test_save_to_default_project_with_read_storage(tmp_dir, test_session, is_studio):
+def test_save_to_default_project_with_read_storage(
+    tmp_dir, test_session, is_studio, studio_job
```
shcheklein, line 4167: can is_studio be providing studio_job automatically when / if it needed?

tests/unit/test_dataset_status_management.py
```diff
@@ -301,17 +301,26 @@ def test_public_api_datasets_filters_non_complete(
     assert dataset_failed.name not in dataset_names, "FAILED dataset should be hidden"
 
 
-def test_public_api_read_dataset_rejects_non_complete(
-    test_session, dataset_created, dataset_failed
-):
+@pytest.mark.parametrize("is_studio", [True])
+def test_public_api_read_dataset_rejects_non_complete(test_session, studio_job):
     """Test that dc.read_dataset() rejects non-COMPLETE datasets."""
+    ds_created = test_session.catalog.create_dataset(
```
shcheklein, line 307: why this change? why can't we reuse dataset fixtures we have anymore?

tests/func/test_datasets.py
ilongin: Before this PR, the autouse `is_studio` fixture in `conftest.py` defaulted to `True`, so all tests ran in Studio mode. That meant `.save("old.old.numbers")` could auto-create the "old" project (non default projects can be created only in Studio).

This PR flips the default to `False` because of the new guard in `session.py` — when `is_studio()` is `True` but `DATACHAIN_JOB_ID` isn't set, we now raise exception.

If we kept `is_studio=True` as the default, every test that triggers job creation would need the `studio_job` fixture to pre-create a job and set `DATACHAIN_JOB_ID`. So flipping to `False` is better IMO — only tests that actually need Studio behavior set `@pytest.mark.parametrize("is_studio", [True])` + `studio_job`.

For tests that just need non-default projects to exist (like this one), we pre-create them via `metastore.create_project()` directly to avoid that check where local mode cannot create projects.

tests/func/test_delta.py
ilongin: As mentioned in comment above, we now run all tests with is_studio to be False and then we cannot just call create_project on dc since it would hit check where local mode cannot create projects/namespaces so the workaround is to just use metastore. Project creation here is just setup anyway.

tests/unit/lib/test_datachain.py
ilongin: I did this and had a lot of issues with ordering of fixtures and errors for some reason.... but more importantly, this would mean that we would create job and use test_session all the time we need is_studio=True and there are a lot of tests where we don't need job at all but we do  need is_studio=True

tests/unit/test_dataset_status_management.py
ilongin: This test calls `read_dataset()` expecting `DatasetNotFoundError`. On CH/PG the default namespace is non-local (`@john`), so `get_dataset_with_remote_fallback` hits Studio fallback instead of raising `DatasetNotFoundError` — that's what was causing the test failure.

To prevent the fallback we need `is_studio=True` (because `no_fallback = is_studio() or is_namespace_local(namespace_name)`).

But with `is_studio=True`, the new guard in `session.py` requires `DATACHAIN_JOB_ID` to be set before any job creation. The existing `dataset_created`/`dataset_failed` fixtures use the `job` fixture which calls `get_or_create_job()` — and there's no way to guarantee `studio_job` (which sets `DATACHAIN_JOB_ID`) runs before the `job` fixture in the dependency chain. So the datasets are created inline using the `studio_job` id directly to avoid fixture ordering issues.

shcheklein (approved):

shcheklein:
@ilongin fix tests, fix codecov

shcheklein:
@ilongin also review Copilot comments, resolve them with some message

ilongin:
> @ilongin fix tests, fix codecov

Tests are still TODO and I still need to use this hash func on listing so its not complete yet

shcheklein:
> I still need to use this hash func on listing so its not complete yet

is it related to this PR?

ilongin:
> > I still need to use this hash func on listing so its not complete yet
> 
> is it related to this PR?

Sorry I thoight this was my other PR about hashing listing, please ignore

src/datachain/catalog/catalog.py
```diff
@@ -588,7 +588,9 @@ def enlist_source(
         from datachain import read_storage
         from datachain.listing import Listing
 
-        read_storage(source, session=self.session, update=update, column=column).exec()
+        read_storage(
+            source, session=self.session, update=update, column=column
+        ).settings(ephemeral=True).exec()
```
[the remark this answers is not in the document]
ilongin: valid comment, addressed

PR #1721, opened 2026-04-10
perf(metastore): eliminate O(N) merge_versions calls in list_datasets

amritghimire:
## Summary

- `list_datasets()` spent ~88% of its time (~56s on 742K rows) calling `merge_versions()` once per SQL row via `reduce()`
- Each call re-sorted a growing list of versions, causing **O(M²) sort comparisons** per dataset
- Fix: parse only the first row per dataset group fully, then extract only version-column slices from subsequent rows and `append` them directly; **sort once at the end**
- Same fix applied to `_parse_dataset` (the `DatasetRecord` path) for consistency

## Benchmark results (5,000 datasets × 50 versions = 250,000 rows)

| Implementation | Time (ms) | Per-row cost | Speedup |
|---|---|---|---|
| old `reduce/merge_versions` | 4,169 ms | 16.68 µs | 1.00× |
| new `append+sort` | 193 ms | 0.77 µs | **21.65×** |

Memory allocation is identical between both implementations.

## Line profiler breakdown

**Old** — 90% of time in `merge_versions().sort()`, called 245,000 times (49× per dataset):
```
Function: old_parse_list_dataset
  versions = [dataset_list_class.parse(*r) ...]   →   9.9%
  reduce(lambda ds, v: ds.merge_versions(v), ...) →  90.1%

Function: DatasetListRecord.merge_versions
  self.versions = list(set(...))                  →   9.3%
  self.versions.sort(key=lambda v: v.version_value) → 90.2%  ← called 245,000×
```

**New** — work split between two genuinely useful operations:
```
Function: new_parse_list_dataset
  dataset_list_class.parse(*rows[0])              →   3.3%   (first row only)
  dataset_list_version_class.parse(*row[offset:]) →  47.4%   (version slice per row)
  dataset.versions.append(version)                →   1.5%
  dataset.versions.sort(...)                      →  46.1%   (once per dataset)
```

## How it works

The SQL result rows have a fixed column order:
```
[*namespace_fields, *project_fields, *dataset_list_fields, *version_fields]
```
`version_offset` is computed dynamically from the same cached field lists used by the query builder, so it automatically adapts to studio's extended column sets without any studio-side changes.

## Test plan

- [x] Correctness verified: both implementations produce identical results across all dataset/version combinations
- [x] 3 new unit tests added to `tests/unit/test_metastore.py`:
  - `test_parse_list_dataset_single_row`
  - `test_parse_list_dataset_multiple_rows_accumulates_and_sorts`
  - `test_parse_list_dataset_empty_rows_returns_none`
- [x] All pre-commit hooks pass (ruff, mypy, codespell)
- [x] Full `tests/unit/test_metastore.py` suite passes

### DIFF
--- a/src/datachain/data_storage/metastore.py
+++ b/src/datachain/data_storage/metastore.py
@@ -1497,12 +1497,14 @@ class AbstractDBMetastore(AbstractMetastore):
         dataset = self.dataset_class.parse(
             *all_rows[0], versions_loaded=versions_loaded, preview_loaded=preview_loaded
         )
+        if not versions_loaded:
+            return dataset
         for row in all_rows[1:]:
             tmp = self.dataset_class.parse(
                 *row, versions_loaded=True, preview_loaded=preview_loaded
             )
-            dataset._versions.extend(tmp._versions)
-        dataset._versions.sort(key=lambda v: v.version_value)
+            dataset.versions.extend(tmp.versions)
+        dataset.versions = sorted(dataset.versions, key=lambda v: v.version_value)
         return dataset
 
     def _parse_list_dataset(self, rows) -> DatasetListRecord | None:
@@ -1513,7 +1515,7 @@ class AbstractDBMetastore(AbstractMetastore):
         for row in rows[1:]:
             tmp = self.dataset_list_class.parse(*row)
             dataset.versions.extend(tmp.versions)
-        dataset.versions.sort(key=lambda v: v.version_value)
+        dataset.versions = sorted(dataset.versions, key=lambda v: v.version_value)
         return dataset
 
     def _parse_dataset_list(self, rows) -> Iterator["DatasetListRecord"]:
--- a/src/datachain/data_storage/metastore.py
+++ b/src/datachain/data_storage/metastore.py
@@ -6,7 +6,7 @@ from abc import ABC, abstractmethod
 from collections.abc import Iterator, Sequence
 from contextlib import contextmanager, suppress
 from datetime import datetime, timedelta, timezone
-from functools import cached_property, reduce
+from functools import cached_property
 from itertools import groupby
 from typing import TYPE_CHECKING, Any
 from uuid import uuid4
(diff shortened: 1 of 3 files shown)


tests/unit/test_metastore.py
```diff
@@ -192,3 +221,42 @@ def test_dataset_record_versions_setter_marks_loaded(test_session):
     record.versions = loaded.versions
 
     assert record.versions == loaded.versions
+
+
+def test_parse_list_dataset_single_row():
+    metastore = SQLiteMetastore(db_file=":memory:")
```
shcheklein: all these tests smell - they use some new SQLiteMetastore (while we have test_session, etc)

they have some finally (means we need fixture)

etc, etc

review, cleanup carefully this

shcheklein (commented):
what happens to merge_versions though? (also it exists probably in both repos?)

amritghimire:
> what happens to merge_versions though? (also it exists probably in both repos?)
> 
> 

There is no other place where merge versions are called neither in datachain nor in Studio. I just kept it without removing just in case, but I believe it is to drop merge_versions altogether.

shcheklein:
> There is no other place where merge versions are called neither in datachain nor in Studio. I just kept it without removing just in case, but I believe it is to drop merge_versions altogether.

let's drop it in both repos, otherwise this PR is confusing

src/datachain/data_storage/metastore.py
```diff
@@ -1490,24 +1490,31 @@ def _parse_dataset(
...
-        if not versions:
+        all_rows = list(rows)
+        if not all_rows:
             return None
-        return reduce(lambda ds, version: ds.merge_versions(version), versions)
+
+        dataset = self.dataset_class.parse(
+            *all_rows[0], versions_loaded=versions_loaded, preview_loaded=preview_loaded
+        )
+        for row in all_rows[1:]:
```
shcheklein, line 1504: it seems we were parsing all rows as versions before, is it expected that we skip the first here?

src/datachain/data_storage/metastore.py
```diff
@@ -1490,24 +1490,31 @@ def _parse_dataset(
...
+            tmp = self.dataset_class.parse(
+                *row, versions_loaded=True, preview_loaded=preview_loaded
+            )
+            dataset._versions.extend(tmp._versions)
+        dataset._versions.sort(key=lambda v: v.version_value)
```
shcheklein: just to make sure - does this sort by version value always work correctly? proper latest version always wins? is version_value a string?

src/datachain/data_storage/metastore.py
amritghimire: Just a line above, we parse the first row and then append other rows to it. First row is already parsed to dataset at this point.

src/datachain/data_storage/metastore.py
amritghimire: version value is : 
    @property
    def version_value(self) -> int:
        return semver.value(self.version)

amritghimire:
> > There is no other place where merge versions are called neither in datachain nor in Studio. I just kept it without removing just in case, but I believe it is to drop merge_versions altogether.
> 
> let's drop it in both repos, otherwise this PR is confusing

Dropped.

src/datachain/data_storage/metastore.py
shcheklein: does it return a single int?

amritghimire: Yes, it is what we are using in almost all places for sorting include merge_versions too

shcheklein: 👍

src/datachain/data_storage/metastore.py
```diff
@@ -1490,24 +1490,41 @@ def _parse_dataset(
...
+        )
+        dataset = self.dataset_class.parse(
+            *all_rows[0], versions_loaded=versions_loaded, preview_loaded=preview_loaded
+        )
+        for row in all_rows[1:]:
+            version = self.dataset_version_class.parse(
+                *row[version_offset:], preview_loaded=preview_loaded
+            )
+            dataset._versions.append(version)
+        dataset._versions.sort(key=lambda v: v.version_value)
```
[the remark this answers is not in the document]
shcheklein: @amritghimire review / resolve this. Check if we have a way to not use protected member directly.

shcheklein (approved):
review some comments left, check tests and covecov 

otherwise LGTM

src/datachain/data_storage/metastore.py
```diff
@@ -1497,12 +1497,14 @@ def _parse_dataset(
...
         )
+        if not versions_loaded:
+            return dataset
         for row in all_rows[1:]:
             tmp = self.dataset_class.parse(
                 *row, versions_loaded=True, preview_loaded=preview_loaded
             )
-            dataset._versions.extend(tmp._versions)
-        dataset._versions.sort(key=lambda v: v.version_value)
+            dataset.versions.extend(tmp.versions)
```
shcheklein: this is might be not reliable - you are relying here on getter returning the same list, right?

should we collect first in some list here - then create dataset with versions passed as field in ctr?

or at least assign that list via `dataset.versions = sorted`

PR #1722, opened 2026-04-10
Skip creating identical bucket listing dataset if nothing is changed

### LINKED ISSUE #1700: Skip creating new listing version when listing data hasn't changed
When `read_storage(uri, update=True)` runs, it always creates a new listing dataset version even if the storage content hasn't changed. This causes checkpoint invalidation because the new version has a different UUID, producing a different chain hash.

Proposal: after listing completes, compare the new listing with the previous version using `(file_path, etag)` pairs. If identical, discard the new version and keep the old one. The chain then uses the old UUID for hash calculation, and checkpoints are reused.

This reduces unnecessary checkpoint invalidations and saves storage costs (especially on ClickHouse).

Note: this doesn't cover the glob case — `read_storage("s3://bucket/*.jpg")` lists the full directory, so adding a `.png` file would still produce a different listing even though the glob-filtered result is unchanged. Workaround: use the directory path and filter explicitly:
```python
dc.read_storage("s3://bucket/").filter(dc.C("file.path").glob("*.jpg"))
```
This way the listing is for the full directory, and the glob filter is a chain step that doesn't affect the listing comparison.

ilongin:
New SQL functions `string_hash` and `xor_agg` for computing order-independent
content fingerprints of listing datasets. Used to detect whether a new listing
version is identical to the previous one, avoiding unnecessary checkpoint
invalidation.

### How it works

```python
fingerprint = (
    listing_chain
    .group_by(fp=func.xor_agg(func.string_hash("file.path", "file.etag")))
    .to_list("fp")[0]
)
```

- `string_hash(col1, col2, ...)` — hashes column values into a 64-bit integer
  - SQLite: `sha256` (Python `create_function`)
  - ClickHouse: `sipHash64(concat(toString(col1), '|', toString(col2), ...))`
- `xor_agg(int)` — bitwise XOR aggregate, order-independent
  - SQLite: custom aggregate via `create_aggregate`
  - ClickHouse: native `groupBitXor`

Single pass over the table, no sorting, no materialization in Python.

### Benchmark setup

Realistic listing dataset with all `File` columns (`source`, `path`, `size`,
`version`, `etag`, `is_latest`, `last_modified`, `location`) populated with
synthetic S3-like data.

### Results

| Rows | DB | Save time | Fingerprint time | Memory (fingerprint) |
|---|---|---|---|---|
| 1M | SQLite | 46s | 7.5s | ~0 MB |
| 1M | ClickHouse | 20s | 0.03s | ~0 MB |
| 10M | SQLite | 408s | 77s | ~0 MB |
| 10M | ClickHouse | 188s | 0.12s | ~0 MB |

### Status

- [x] `string_hash` SQL function (SQLite + ClickHouse)
- [x] `xor_agg` SQL aggregate (SQLite + ClickHouse)
- [x] `func.string_hash` / `func.xor_agg` user-facing wrappers
- [x] Benchmark on SQLite and ClickHouse (1M + 10M rows)
- [x] Compare old vs new listing fingerprints in `DatasetQuery.save()` (when `listing=True`, after Phase 3 but before `complete_dataset_version()` — new version is in CREATED status so invisible to other queries; if fingerprints match, discard new version and keep old)
- [x] Tests

### Notes

- Fingerprint values differ between SQLite and ClickHouse (different hash
  algorithms: sha256 vs sipHash64). Comparison is always within the same DB.
- XOR is commutative — row insertion order doesn't matter.


shcheklein (commented):
let's try to calculate some hash as we do the listing (vs doing full table scan after) - i think it should be doable, it won't be expensive and we don't need expensive functions then

ilongin:
> let's try to calculate some hash as we do the listing (vs doing full table scan after) - i think it should be doable, it won't be expensive and we don't need expensive functions then

That was m first idea but code looks bad, like a hack since we have DataChain for listing. Also we would need to save that info somewhere so we would need to add new field to dataset version table. Also, performance on Clickhouse is extremely good - only 0.12 sec for 10M records which is insignificant compared to listing and in genetal I think since we utilize CH columnar design.
Another thing, we could use this to create is_equal method on DataChain class (that was my plan) that would accept arbitrary column names on which to compare.

src/datachain/lib/listing.py
```diff
@@ -34,6 +39,17 @@
 logging.getLogger("gcsfs").setLevel(logging.CRITICAL)
 
 
+def calc_fingerprint(catalog: "Catalog", dataset: "DatasetRecord", version: str) -> int:
+    """Compute an order-independent content fingerprint for a listing version.
+
+    Returns XOR of string_hash(file__path, file__etag) over all rows.
```
shcheklein: it is better to use version if it is available (less false positives since etag changes by clouds even for the same file - e.g. on GCP and Azure - just by their internal systems)

shcheklein (commented):
Was it tested on the cluster (cloud)? with some real volume and distributed tables?

src/datachain/lib/listing.py
ilongin: Updated to use `version` instead of `etag` when available (via SQL CASE), since versioning is a bucket-level setting — either all files have it or none do. This should avoid false positives I think.

```python
identifier = sa.case(
    (table.c.file__version != "", table.c.file__version),
    else_=table.c.file__etag,
)
query = sa.select(xor_agg(string_hash(table.c.file__path, identifier)))
```

Benchmark on 10M rows shows almost no performance impact from the extra column:

| DB | path+etag | path+case(version,etag) |
|---|---|---|
| SQLite | 77s | 77s |
| ClickHouse | 0.12s | 0.18s |

shcheklein: > Benchmark on 10M rows shows almost no performance impact from the extra column:

Locally?

Cloud can differ because it it distributed? Also, I wonder if it will 100% guarantee consistent results on such full scan.

src/datachain/lib/listing.py
```diff
@@ -42,11 +42,17 @@
...
 
-    Returns XOR of string_hash(file__path, file__etag) over all rows.
+    Uses file path + version (if available) or etag as the content identifier.
+    Version is preferred over etag because etags can be unstable on some cloud
+    providers (GCS, Azure) due to internal operations like storage class changes.
     """
     table_name = catalog.warehouse.dataset_table_name(dataset, version)
     table = catalog.warehouse.get_table(table_name)
-    query = sa.select(xor_agg(string_hash(table.c.file__path, table.c.file__etag)))
+    identifier = sa.case(
```
shcheklein: is it possible to avoid using internals? e.g. internal column naming conventions etc? it is another extra place we are leaking internals - harder to maintain / search, etc

src/datachain/lib/listing.py
ilongin: Yes, this is locally but I doubt cloud is much different .. I will test it though when demo-1 team cluster get's fixed. My test script checks consistency as well so we will see when I run it on cloud, but I don't think there should be an issue since xor is order independent - that's whole point of using it  - we don't need to sort data before calculating hash since sorting would be extremely slow / expensive specially on Clickhouse.

src/datachain/query/dataset.py
```diff
@@ -3233,6 +3233,27 @@ def save(
 
             self._add_dependencies(dataset, version)  # type: ignore [arg-type]
 
+            # For listing datasets, check if content is unchanged from
+            # the previous version. If so, discard the new version to
+            # preserve the old UUID and avoid checkpoint invalidation.
+            # The new version is still CREATED (invisible to other queries).
+            if kwargs.get("listing"):
+                prev_version = dataset.latest_complete_version
```
shcheklein: does dataset (object) include all versions ? (means that create_dataset should be returning an object that is populated with all versions)?

src/datachain/lib/listing.py
ilongin: switched to `Column("file.path")` /  `Column("file.version")` / `Column("file.etag") `instead of the flattened `file__*` names. Note that this file already knows about listing internals i.e. it knows which columns are in listing dataset  ... I think that's fine.

src/datachain/lib/listing.py
```diff
@@ -34,6 +39,23 @@
...
+    Uses file path + version (if available) or etag as the content identifier.
+    Version is preferred over etag because etags can be unstable on some cloud
+    providers (GCS, Azure) due to internal operations like storage class changes.
+    """
+    table_name = catalog.warehouse.dataset_table_name(dataset, version)
+    table = catalog.warehouse.get_table(table_name)
+    file_version = Column("file.version")
+    file_etag = Column("file.etag")
+    file_path = Column("file.path")
+    identifier = sa.case((file_version != "", file_version), else_=file_etag)
```
shcheklein, line 54: okay, that's better! but why can we use datchain directry? we have case, and all these functions, etc

src/datachain/query/dataset.py
```diff
@@ -3233,6 +3233,27 @@ def save(
 
             self._add_dependencies(dataset, version)  # type: ignore [arg-type]
 
+            # For listing datasets, check if content is unchanged from
+            # the previous version. If so, discard the new version to
+            # preserve the old UUID and avoid checkpoint invalidation.
+            # The new version is still CREATED (invisible to other queries).
+            if kwargs.get("listing"):
```
shcheklein, line 3241: C (not sure what to do here, some refactoring is needed)

save again is getting out of hands - too many ifs and different situations handled in some custom way

src/datachain/query/dataset.py
```diff
@@ -3233,6 +3233,27 @@ def save(
...
+                prev_version = dataset.latest_complete_version
+                if prev_version:
+                    from datachain.lib.listing import calc_fingerprint
+
+                    old_fp = calc_fingerprint(self.catalog, dataset, prev_version)
+                    new_fp = calc_fingerprint(self.catalog, dataset, version)
+                    if old_fp == new_fp:
```
shcheklein: do we need to bump expiration time for that listing btw?

src/datachain/query/dataset.py
ilongin: yes!

src/datachain/lib/listing.py
ilongin: Tried it but hit a blocker — `dc.read_dataset` only reads COMPLETE  datasets, and at this point in `save()` the just-claimed version is still in CREATED status, so it raises `DatasetNotFoundError`. Keeping  raw SQL for that reason.

src/datachain/query/dataset.py
ilongin: Extracted into helper function ... don't think we can do more atm.

src/datachain/query/dataset.py
ilongin: yes, good catch. added bump.

ilongin:
> Was it tested on the cluster (cloud)? with some real volume and distributed tables?

Ran on the distributed CH cluster:              
                                                                       
| Rows | path+etag | path+case(ver,etag) |                             
|---|---|---|                                                          
| 100k | 0.43s | 0.37s |                                               
| 10M | 0.79s | 0.78s |    
                                                                       
Consistency verified — identical fingerprints across back-to-back runs on the same table (100k: `903316409541948947`, 10M:  `2386032644197840272`). 
XOR being commutative means row order doesn't  affect the result, and `case(ver, etag)` has no measurable overhead vs plain `etag`.
Performance is little bit worse than local but still under 1 sec for 10M records which is in the area of statistical error specially considering that listing itself for 10M will take a long time.

src/datachain/lib/listing.py
shcheklein: okay, but how about everything else - use datachain function - like case? can we just replace read_dataset with something?

ilongin: Added `DataChain` but it was needed to be created via `DatasetQuery`, similar as we do on `dc/listing.py`, to override this issue of not being able to read non complete datasets.

ilongin:
@dmpetrov I've brainstormed with Claude little bit about our discussion regarding collision probabilities and Claude is pretty convinced current solution is good enough i.e. collision probability in worst case is so small that it's irrelevant. If we want more safer solution we could use `sipHash128` vs `sipHash64` but it's not really needed if this below is correct.

Here is the summary:

## Fingerprint collision analysis                                                                                                                
          
  ### Scheme                                                                                                                                       
  Per row: `h(r) = H(path + "|" + identifier)` where `identifier = file.version if non-empty else file.etag`.
                                                                                                                                                   
  - **SQLite**: `H = sha256(s)[:8]` — truncated SHA-256, 64 bits.
  - **ClickHouse**: `H = sipHash64(concat(toString(a), '|', toString(b)))` — 64-bit SipHash.                                                       
                                                                                                                                                   
  Fingerprint = bitwise XOR of all `h(r)`.
                                                                                                                                                   
  ### False negatives (two different listings, same fingerprint)                                                                                   
   
  Worst case is the simplest case: **one record changed** (e.g., etag flip). Collision requires `h_old = h_new`:                                   
                             
  > **P(silent miss) = 1 / 2⁶⁴ ≈ 5.42 × 10⁻²⁰**                                                                                                    
   
  This is flat in listing size — same for 1M, 10M, 100M, or 1B rows, because the change is localized to one row-pair.                              
                             
  Scale check: at 10¹⁰ buckets worldwide × 1 update/day, expect ~1 miss every ~5 billion years.                                                    
                             
  ### False positives (two identical listings, different fingerprints)                                                                             
   
  > **P = 0.**                                                                                                                                     
                             
  Both hash functions are deterministic, XOR is order-independent. Identical `(path, identifier)` sets always produce identical fingerprints.      
   
  ### If we want more safety                                                                                                                       
                             
  Widen the per-row hash:
  - SQLite: `sha256(s)[:16]` (128 bits, store as two Int64)
  - ClickHouse: `sipHash128`
                                                                                                                                                   
  Worst case drops from `2⁻⁶⁴` to `≈ 2⁻¹²⁸ ≈ 2.9 × 10⁻³⁹`. Past the point where anyone should care.                                                
                                                                                                                                                   
  ### Verdict                                                                                                                                      
                             
  64-bit XOR fingerprint is safe for this use case. Worst-case miss probability is 1 in 18 quintillion per update, independent of listing size.

shcheklein (approved):

PR #1723, opened 2026-04-11
remove legacy compatibility code

shcheklein:
This PR removes legacy rollout compatibility leftovers around dataset loading by tightening the metastore request contract to require explicit `versions` and `include_preview` fields, removing wire-format inference for loaded-state flags such as `_preview_loaded` and `_versions_loaded`, and aligning the Studio-side parsing and tests with the current DataChain contract so we only support the modern request and response shapes going forward.

### DIFF
--- a/src/datachain/dataset.py
+++ b/src/datachain/dataset.py
@@ -290,7 +290,7 @@ class DatasetVersion:
     _preview_data: str | list[dict] | None = field(
         default=None, metadata={"alias": "preview"}
     )
-    _preview_loaded: bool = False
+    _preview_loaded: bool = field(kw_only=True)
     sources: str = ""
     query_script: str = ""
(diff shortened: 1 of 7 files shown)


amritghimire (approved):
Looks good to me, but may be we should merge the changes in studio with the flags passing before merging this so we won't get compatibility issue with clusters when releasing?

shcheklein:
> Looks good to me, but may be we should merge the changes in studio with the flags passing before merging this so we won't get compatibility issue with clusters when releasing?

I will test it both cluster versions. But overall this PR is pretty much about removing extra code that we had to deploy previous PR on both clusters. So, we need to make this one go without hiccups and w/o followups.

dreadatour (approved):

PR #1728, opened 2026-04-20
Fix for zero output udf

### LINKED ISSUE #1705: gen - zero size output
### Description

Returning nothing seems a valid use case like parsing csv file which happened to be empty. 

Now this fails:
```python
def single(val: int) -> Iterator[str]:
    #yield f"item_{val}"
    pass

dc.read_values(val=[1, 2, 3]).gen(out=single).show()
```

While now it fails with cryptic error:
```
...
  File "/Users/dmitry/src/datachain/src/datachain/data_storage/buffer.py", line 55, in insert_many
    for entry in entries:
                 ^^^^^^^
  File "/Users/dmitry/src/datachain/src/datachain/query/dataset.py", line 515, in _insert_rows
    for row in udf_output:
               ^^^^^^^^^^
  File "/Users/dmitry/src/datachain/src/datachain/lib/udf.py", line 578, in _process_row
    for result_obj, is_last in with_last_flag(result_objs):
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dmitry/src/datachain/src/datachain/utils.py", line 755, in with_last_flag
    it = iter(iterable)
         ^^^^^^^^^^^^^^
TypeError: 'NoneType' object is not iterable
```

### Version Info

```Text
0.49.2.dev98+g5e6e1a3de.d20260402
Python 3.12.11
```

ilongin:


dmpetrov:
is it about #1705?

ilongin:
> is it about #1705?

yes

shcheklein (approved):

PR #1732, opened 2026-04-26
emit subselect if it is neeeded

shcheklein:
Fixes a few open issue in queries where `persist` is required.

Usually it means we don't generate right SQL (e.g. limit.distinct add limit _after_ distinct in the SQL), while it should be doing additional select distinct _after_ limit is applied), etc.


src/datachain/data_storage/warehouse.py
```diff
@@ -55,6 +56,8 @@
 
 SELECT_BATCH_SIZE = 100_000  # number of rows to fetch at a time
 
+QuerySelectType = TypeVar("QuerySelectType", bound=GenerativeSelect)
```
amritghimire: This seems to be  used only in this file, only in one signature. do we need a `TypeVar` here at all? `GenerativeSelect -> `GenerativeSelect`` reads fine.

src/datachain/query/dataset.py
```diff
@@ -171,19 +171,77 @@ def step_result(
...
+        self.reset()
+
+    def reset(self) -> None:
+        self.grouped = False
+        self.limited = False
+        self.offset_applied = False
+        self.distinct = False
+
+    @property
+    def has_finalized_output(self) -> bool:
```
amritghimire: the name reads like the query is done, probably , `has_finalizing_clause` or something similar?

src/datachain/query/dataset.py
```diff
@@ -1820,15 +1881,23 @@ def q(*columns):
...
         **kwargs,
     ) -> StepResult:
-        query = query_generator.select()
+        if self.requires_boundary(state):
+            query_generator = state.emit_boundary(query_generator)
+
+        query = state.warehouse.normalize_limit_offset(query_generator.select())
         new_query = self.apply_sql_clause(query)
+        new_query = state.warehouse.normalize_limit_offset(new_query)
+        self.update_query_state(state)
```
amritghimire, line 1898: `normalize_limit_offset` is called twice — once before `apply_sql_clause`, once after — and again inside `emit_boundary`. is that intentional?

src/datachain/query/dataset.py
```diff
@@ -171,19 +171,77 @@ def step_result(
...
 
+    resets_query_state_after_apply: ClassVar[bool] = True
+
     @abstractmethod
     def apply(
         self,
         query_generator: QueryGenerator,
         temp_tables: list[str],
         *args,
+        state: "QueryState",
```
amritghimire, line 232: This looks weird, a argument in between *args and **kwargs. Lets either make `state` a real keyword-only arg so missing it is a hard error or use it from kwargs.

src/datachain/data_storage/warehouse.py
shcheklein: yes, I think mypy was complaining - I'll double check

src/datachain/query/dataset.py
shcheklein: yep, addressing

src/datachain/query/dataset.py
shcheklein: yep, simplifying

src/datachain/query/dataset.py
shcheklein: yep, can be simpler I think

amritghimire (approved):

PR #1733, opened 2026-04-27
 Update logout command to revoke token server-side

amritghimire:
Context
--------
The datachain auth logout command currently only deletes the token from local config. The Studio backend now exposes POST /api/token-logout for self-revocation. The logout command should call this endpoint before clearing the local token, so the token is invalidated both locally and server-side.

This changes calls studio endpoint on logout


src/datachain/studio.py
```diff
@@ -209,6 +211,37 @@ def logout(local: bool = False):
...
+                    file=sys.stderr,
+                )
+            elif not response.ok:
+                print(
+                    f"Warning: Unexpected response from Studio "
+                    f"(HTTP {response.status_code}).",
+                    file=sys.stderr,
+                )
+        except requests.RequestException:
+            print(
```
shcheklein: let's be strict then - don't proceed with logout at all

src/datachain/studio.py
```diff
@@ -209,6 +211,37 @@ def logout(local: bool = False):
...
+                timeout=10,
+            )
+            if response.status_code == 401:
+                print(
+                    "Token was already revoked or is invalid on Studio.",
+                    file=sys.stderr,
+                )
+            elif not response.ok:
+                print(
+                    f"Warning: Unexpected response from Studio "
```
shcheklein: same here - don't proceed then

src/datachain/studio.py
amritghimire: The problem with that is in case user deletes the token from UI, then tries to clear it from cli too, it will raise error and user can't remove the token from the config. It will be a pain point for user and blocks further process.

src/datachain/studio.py
amritghimire: Same reasoning above.

src/datachain/studio.py
shcheklein: let's handle not found properly then (assuming it was deleted already)

shcheklein:
check also tests, we need to hit way more coverage

src/datachain/studio.py
0x2b3bfa0: Consider cleaning up tokens from Studio _before_ deleting them locally. Abort the logout process with an error if the Studio logout request fails, giving users the responsibility to re-try.

Studio should be able to emit distinct responses for "this token is well-formed but no longer valid" (logout can continue and local token can be deleted) versus anything else (server error, network error, ...)

shcheklein (approved):

dreadatour (approved):

PR #1735, opened 2026-04-27
Cache bug fixes

### LINKED ISSUE #1725: Cache does not work
### Description

The 2nd script will be re-downloading:

create.py:
```python
import datachain as dc

def fn(file) -> int:
    file.read()
    return 0

(
    dc.read_storage("gs://dc-readme/oxford-pets-small/**/*.jpg", anon=True)
    .limit(100)
    .settings(cache=True, prefetch=30)
    .map(x=fn)
    .save("pets")
)
```

use.py:
```python
import datachain as dc

def fn(file)->int:
    data = file.read()
    return len(data)

dc.read_dataset("pets").map(len=fn).save("pets_len")
```

### Version Info

```Text
0.49.2.dev130+gedd843756.d20260407
Python 3.12.11
```

ilongin:
Fixes #1725 — `read_dataset(...).map(...)` re-downloads files that are already in `~/.datachain/cache/`.

Two distinct bugs, both rooted in `cache=False` being treated as "ignore cache entirely" rather than "don't *write* to cache". Per @dmpetrov: `cache=` is a write switch; reads should always consult the persistent cache.

**Bug A — prefetcher ignores persistent.** With `cache=False, prefetch>0`, the prefetcher swaps to a fresh temp cache and re-downloads everything. **Fix:** `Cache` gains an optional read-only `fallback`; the temp cache uses persistent as fallback, so `contains()`/`get_path()` see hits in persistent while writes/eviction stay scoped to temp.

**Bug B — read path ignores persistent when `prefetch=0`.** `_caching_enabled=False` makes `open_object(use_cache=False)` stream from remote even on a cache hit. **Fix:** decoupled `_caching_enabled` — it now gates writes (`ensure_cached`) only; reads (`open_object`, `get_local_path`) always consult cache. 
Applied to all 5 read sites.


src/datachain/lib/arrow.py
```diff
@@ -72,7 +72,8 @@ def __init__(
     def process(self, file: File):
         if file._caching_enabled:
             file.ensure_cached()
-            cache_path = file.get_local_path()
+        cache_path = file.get_local_path()
+        if cache_path:
```
dreadatour: It looks like `cache_path` is not used below:
```suggestion
        if cache_path := file.get_local_path():
```

src/datachain/lib/file.py
```diff
@@ -516,9 +516,9 @@ def open(
                 return
             if self._caching_enabled:
                 self.ensure_cached()
-            with client.open_object(
-                self, use_cache=self._caching_enabled, cb=self._download_cb
-            ) as f:
+            # _caching_enabled is a write-side switch only; reads always consult
+            # the cache (cache hit → local, miss → remote stream).
```
dreadatour: Let's may be move it to docstrings? Also for `TarVFile` above.

src/datachain/lib/file.py
```diff
@@ -655,12 +655,12 @@ def _symlink_to(self, destination: str) -> None:
...
-        else:
-            raise OSError(errno.EXDEV, "can't link across filesystems")
+        source = self.get_local_path()
+        if source is None:
+            if self.source.startswith("file://"):
+                source = self.get_fs_path()
+            else:
+                raise OSError(errno.EXDEV, "can't link across filesystems")
 
         return os.symlink(source, destination)
```
dreadatour, line 671: From what I see `_symlink_to` is used from `File.export` only.  Are we sure we want to symlink to cache? 🤔 I think we need to dig deeper in this place.

src/datachain/lib/file.py
```diff
@@ -1665,9 +1665,9 @@ def open(self):
 
         if self.file._caching_enabled:
             self.file.ensure_cached()
-            path = self.file.get_local_path()
+        path = self.file.get_local_path()
+        if path:
             ds = dataset(path, **self.kwargs)
```
dreadatour, line 1675: I am not sure if we are following the logic "always cache on read" here. Can we check it one more time?

dreadatour (commented):
Couple comments so far.
Let's also ask CoPilot to review these changes.
I'll review this one more time when we'll fix all the tests failing.

src/datachain/lib/arrow.py
ilongin: Done

src/datachain/lib/file.py
ilongin: I think you are right, yes., we should not use cache with symlinks. Cache can be removed at any point of time and that should not break functionality.

ilongin: Reverted

src/datachain/lib/file.py
ilongin: This is reverted because copilog had a good point https://github.com/datachain-ai/datachain/pull/1735#discussion_r3186328940

src/datachain/cache.py
```diff
@@ -22,30 +24,69 @@ def try_scandir(path):
...
+
+        return async_wrapper
+
+    @functools.wraps(method)
+    def wrapper(self, *args, **kwargs):
+        if self._readonly:
+            raise RuntimeError(f"cannot call {method.__name__}() on a read-only cache")
+        return method(self, *args, **kwargs)
+
+    return wrapper
```
dreadatour: I would prefer to have
```python
class ReadonlyCache(Cache):
```
with overridden methods (`remove`, `store`, etc) instead of bringing additional complexity to existing `Cache` class.

src/datachain/lib/udf.py
```diff
@@ -429,7 +429,12 @@ def _get_cache(
     tmp_dir = cache.tmp_dir
     assert tmp_dir
     if prefetch and not use_cache:
-        return temporary_cache(tmp_dir, prefix="prefetch-")
+        # Temp primary with persistent as read-only fallback: prefetch reads
+        # check both, but writes and eviction stay in temp so cache=False
+        # never persists files.
```
dreadatour: I think this comment is redundant or needs changes.

src/datachain/lib/udf.py
```diff
@@ -429,7 +429,12 @@ def _get_cache(
...
+        return temporary_cache(
+            tmp_dir, prefix="prefetch-", fallback=cache.as_readonly()
+        )
     return nullcontext(cache)
```
dreadatour: Do we want to use read-only cache here if both prefetch and cache are disabled?

src/datachain/cache.py
ilongin: Good idea, added it.

src/datachain/lib/udf.py
ilongin: Good catch — applied. `cache=False` now returns a read-only cache whether `prefetch` is 0 or not, so the rule "`cache=` only controls writes" is enforced by the code itself and both prefetch cases behave the same.

src/datachain/cache.py
```diff
@@ -133,3 +160,31 @@ def get_total_size(self) -> int:
...
+
+class ReadonlyCache(Cache):
+    """A read-only view over an existing cache directory.
+
+    Used as a fallback for short-lived temp caches: writes/eviction stay
+    scoped to the temp cache, reads see hits in the persistent one.
+    """
+
+    def _readonly_error(self, name: str) -> RuntimeError:
+        return RuntimeError(f"cannot call {name}() on a read-only cache")
```
dreadatour, line 173: This looks redundant.

dreadatour (approved):
Looks good to me 👍

One question: do we have enough test cases for this new behavior?

dreadatour:
Studio tests are failing, not sure why 🤔

PR #1741, opened 2026-04-29
feat(file): add content type and extension metadata

yathomasi:
## Summary
Add `content_type` and `ext` to DataChain `File` metadata so storage listings can persist MIME type when providers expose it and always expose a path-derived extension.
## Details
- Add `File.content_type` and `File.ext` fields, including warehouse column types and listing table columns.
- Derive `ext` from `File.path` automatically when it is not explicitly provided.
- Resolve provider-reported MIME values from GCS, Azure/adlfs, HTTP, and S3 info dictionaries when available.
- Preserve the provider-sourced contract: `content_type` stays blank when storage listing metadata does not expose a MIME type.
## S3 Note
S3 bulk listings (`ListObjectsV2` / version listing) do not include object `ContentType`, so `content_type` is expected to be blank for normal S3 listing rows even when `ext` is available, e.g. `jpg`.
If we need accurate S3 MIME types, we can add an opt-in enrichment path that performs HEAD requests per object and reads `ContentType`. That would be significantly more expensive and slower for large buckets, so this PR avoids doing it by default.
## Testing
- Added unit coverage for extension derivation and content-type resolution across provider metadata shapes.
- Updated schema/hash fixtures for the new `File` fields.

### DIFF
--- a/src/datachain/client/azure.py
+++ b/src/datachain/client/azure.py
@@ -5,7 +5,7 @@ from adlfs import AzureBlobFileSystem
 from datachain.lib.file import File
 from datachain.progress import tqdm
 
-from .fsspec import DELIMITER, Client, ResultQueue
+from .fsspec import DELIMITER, Client, ResultQueue, resolve_content_type
 
 
 class AzureClient(Client):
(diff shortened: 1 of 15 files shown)


shcheklein:
@yathomasi thanks for the effort, but for the task we need this is over engineering. We definitely want to avoid changing File schema just to accommodate extension / content type, etc.  Let's discuss in Slack this. For now please help to polish pipelines and move preview task going forward (cc @amritghimire )

yathomasi:
Closing this for now.

Issue #1636, opened 2026-03-11
Deterministic checkpoint hashes for read_values, read_pandas, read_hf


ilongin:
## Problem

Several `read_*` entry points produce a random checkpoint hash on every run, making checkpoint reuse impossible. The root cause is `read_records()` — it creates a temp dataset via `session.generate_temp_dataset_name()` which includes a random UUID. The starting hash is derived from this random dataset name.

Affected entry points:
| Entry point | Root cause |
|---|---|
| `read_values()` | Built on `read_records()` |
| `read_pandas()` | Built on `read_values()` |
| `read_hf()` | Built on `read_values()` |
| `read_records()` with concrete data | Same random temp name |
| `read_database()` | Built on `read_records()` with a generator |
| `read_records()` with generators | Same |

Working entry points (for reference):
| Entry point | Why it works |
|---|---|
| `read_dataset()` | Hash based on dataset name + version |
| `read_storage()` | Hash based on listing dataset name |
| `read_csv()`, `read_parquet()`, `read_json()` | Built on `read_storage()` |

## Proposed solution

For entry points where data is concrete (in memory) at construction time, compute a deterministic hash from the input data upfront — before steps are applied. This preserves the current design where hash calculation is lightweight and doesn't require step application.

**Concrete data (can fix now):**
- `read_values()` — input is always concrete lists (`**fr_map` sequences)
- `read_pandas()` — input is always a DataFrame converted to lists
- `read_hf()` — input is a list of split names
- `read_records()` when called with a list/dict

Approach: compute a streaming SHA256 hash of the serialized input rows at construction time. Python's `hashlib.sha256` supports incremental `.update()` calls with constant memory. Performance for 1M rows of ~100 bytes ≈ 0.2s, negligible compared to DB insert time.

**Challenges:**
- Deterministic serialization of arbitrary Python objects (Pydantic models, datetimes, nested dicts). The `_flatten_record` + `adjust_outputs` pipeline already serializes these for DB insertion — we could reuse or mirror that logic for hashing.
- The hash needs to be available before `read_dataset()` is called on the temp dataset. Could either set it directly as the starting step hash on the `DatasetQuery`, or use a content-addressable dataset name.

**Generators/iterators (deferred — separate discussion):**
- `read_database()` passes a generator from SQL result iteration
- Direct `read_records()` calls can pass generators
[...]

PR #1742, opened 2026-04-29
Making hash calculation deterministic for single file read storage

### LINKED ISSUE #1640: read_storage: non-deterministic checkpoint hash for single-file paths
When `read_storage()` or `read_csv()` is called with a single file path (e.g. `read_csv("file.csv")` or `read_storage("s3://bucket/data.csv")`), `get_listing()` detects it's a single file and returns `list_ds_name=None`. This skips the listing path entirely and falls through to `read_values` which creates a temp dataset with a random UUID name. The starting checkpoint hash is derived from this random name, making it different on every run — so checkpoints can never be reused.

Directory paths and globs work fine since they go through the listing path which produces a deterministic hash.

Proposed fix: use the file URI + etag as the starting hash. A single `HEAD` request (S3/GCS/Azure) or `stat` call (local) to get the etag before computing the hash. This detects content changes — if the file was edited, the etag is different, so checkpoints re-run. If the file is unchanged, same etag, checkpoints reused.

This is consistent with the approach in #1639 where directory listings also use etags as the source of truth for "did the data change". Single file = one stat call, directory = streaming query over listing table.

ilongin:
Fixes #1640.

Single-file `read_storage` / `read_csv` / `read_parquet` produced non-deterministic chain hashes (random temp-dataset UUID), so checkpoints could never be reused. Directories/globs already work via the listing fingerprint (#1722).

Fix: `DatasetQuery.override_starting_hash()` lets callers pin the starting-step hash. `read_storage` sets it for single-file inputs, hashing sorted `(source, path, version_or_etag)` tuples — same identifier choice as `calc_fingerprint` (version preferred, etag fallback).

Tests (`tests/func/test_listing.py`): local + cross-cloud (file/s3/gs/azure × version_aware), CSV, Parquet, multi-URI order independence, mixed dir + single-file.

### DIFF
--- a/src/datachain/lib/dc/storage.py
+++ b/src/datachain/lib/dc/storage.py
@@ -19,7 +19,7 @@ if TYPE_CHECKING:
     from .datachain import DataChain
 
 
-def _single_file_starting_hash(files: Sequence[File]) -> str:
+def _files_starting_hash(files: Sequence[File]) -> str:
     """Return an order-independent sha256 over (source, path, version-or-etag)
     of the given files."""
     keys = sorted((f.source, f.path, f.version or f.etag) for f in files)
@@ -268,9 +268,7 @@ def read_storage(
         file_chain.signals_schema = file_chain.signals_schema.mutate(
             {f"{column}": file_type}
         )
-        file_chain._query.override_starting_hash(
-            _single_file_starting_hash(file_values)
-        )
+        file_chain._query.override_starting_hash(_files_starting_hash(file_values))
         storage_chain = storage_chain.union(file_chain) if storage_chain else file_chain
 
     assert storage_chain is not None
--- a/src/datachain/lib/dc/storage.py
+++ b/src/datachain/lib/dc/storage.py
@@ -1,3 +1,4 @@
+import hashlib
 import os
 from collections.abc import Sequence
 from functools import reduce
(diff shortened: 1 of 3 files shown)


src/datachain/lib/dc/storage.py
```diff
@@ -10,14 +11,24 @@
...
-from datachain.lib.file import FileType, get_file_type
+from datachain.lib.file import File, FileType, get_file_type
 from datachain.lib.listing import get_file_info, get_listing, list_bucket, ls
 from datachain.query import Session
 
 if TYPE_CHECKING:
     from .datachain import DataChain
 
 
+def _single_file_starting_hash(files: Sequence[File]) -> str:
```
shcheklein: why is it called single file but accepts Sequence?

src/datachain/query/dataset.py
```diff
@@ -2437,6 +2437,7 @@ def __init__(
         self.checkpoints_enabled = checkpoints_enabled
 
         self.list_ds_name: str | None = None
+        self._starting_hash_override: str | None = None
```
shcheklein, line 2570: this all looks like a lot of hacks just to patch a single niche problem 

let's think about alternatives ... what else we can do to solve this? besides introducing some special override field and handle it, etc?

tests/func/test_listing.py
```diff
@@ -203,3 +205,86 @@ def test_calc_fingerprint_order_independent(test_session):
...
+
+def test_read_storage_single_local_file_hash_changes_on_edit(test_session, tmp_dir):
+    f = tmp_dir / "a.txt"
+    f.write_text("hello")
+    uri = f.as_uri()
+
+    h1 = dc.read_storage(uri, session=test_session)._query.hash()
+
+    time.sleep(0.01)
+    f.write_text("hello world")
```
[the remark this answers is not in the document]
ilongin: Implemented

src/datachain/lib/dc/storage.py
ilongin: Bad naming, fixed it.

src/datachain/query/dataset.py
ilongin: It's a hack because we de-prioritized real fix https://github.com/datachain-ai/datachain/issues/1636 that would've fixed this one automatically as well, since `read_storage` for single files backoffs to `read_values` / `read_records`

shcheklein: it there is a case where it becomes important we might want reconsider then proper fix

let's not do hacks please 

let's think about alternative designs, including #1636 ... may be some simplified version of it

ilongin:
Already fixed with other PR

PR #1745, opened 2026-05-04
Optimize merge on sqlite

### LINKED ISSUE #1710: Merge without persists performance issue
### Description

The code below performs 10x worse without persist().

What I found so far:
1. Order doesn't matter, you can also try `for persist in (True, False):` or `for persist in (1, 2, 0, 3, 4):` - same result (0/False performs 10x worse than the others).
2. It seem doesn't happened in small scale N=700 (even after introducing delays)

```python
"""Merging unpersisted read_storage() UDF chain is ~10x slower than persisted.
Not reproducible with read_values() — the overhead is specific to file-backed chains.

    rm -rf .datachain && python repro_merge_storage.py
"""

import os
import shutil
import time

import datachain as dc

N = 7000
DIR = "/tmp/_repro"


class Label(BaseModel):
    tag: str


def classify(file: dc.TextFile) -> Label:
    return Label(tag=file.read().strip())


shutil.rmtree(DIR, ignore_errors=True)
for sub in ("a", "b"):
    os.makedirs(f"{DIR}/{sub}")
    for i in range(N):
        open(f"{DIR}/{sub}/{i}.txt", "w").write(f"{sub}_{i}")
time.sleep(1)

stem = dc.func.path.file_stem(dc.C("file.path"))

for persist in (False, True):
    t0 = time.time()
    right = (
        dc.read_storage(f"{DIR}/b/", type="text")
        .map(label=classify)
        .mutate(stem=stem)
        .select("label", "stem")
    )
    if persist:
        right = right.persist()
    r = (
        dc.read_storage(f"{DIR}/a/", type="text")
        .merge(right, on=stem, right_on="stem")
        .save(f"repro_{persist}")
    )
    print(f"persist={persist}: {time.time() - t0:.1f}s, {r.count()} rows")

shutil.rmtree(DIR)
```

### Version Info

```Text
0.49.2.dev114+gc646881e7.d20260405
Python 3.12.1
```

shcheklein:
Fixes https://github.com/datachain-ai/datachain/issues/1710

### DIFF
--- a/src/datachain/data_storage/sqlite.py
+++ b/src/datachain/data_storage/sqlite.py
@@ -20,15 +20,22 @@ from sqlalchemy import (
 from sqlalchemy.dialects import sqlite
 from sqlalchemy.schema import CreateIndex, CreateTable, DropTable
 from sqlalchemy.sql import func
-from sqlalchemy.sql.elements import BinaryExpression, BooleanClauseList
+from sqlalchemy.sql.elements import (
+    BinaryExpression,
+    BooleanClauseList,
+    ColumnClause,
+    ColumnElement,
+)
 from sqlalchemy.sql.expression import bindparam, cast
-from sqlalchemy.sql.selectable import Select
+from sqlalchemy.sql.selectable import Select, TableClause
+from sqlalchemy.sql.visitors import iterate
 
 import datachain.sql.sqlite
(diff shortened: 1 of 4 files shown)


shcheklein:
Closing this for now. I'll get back to it a bit later. It is a lower priority atm.

PR #1746, opened 2026-05-05
Fixing `read_storage` to return deterministic hash

### LINKED ISSUE #1636: Deterministic checkpoint hashes for read_values, read_pandas, read_hf

### LINKED ISSUE #1640: read_storage: non-deterministic checkpoint hash for single-file paths

ilongin:
Partially fixes #1636 (deterministic checkpoint hashes for materialized inputs); fully fixes #1640 (single-file `read_storage`).

`read_records(list/tuple/dict)` now derives a content hash from its input, stored in a new `content_hash` column on `datasets_versions`. `QueryStep.hash()` uses `content_hash` when set, falls back to `sha256(uuid)` otherwise. `uuid` retains its original meaning (random `uuid4`, globally unique). Single-file `read_storage` is rerouted through `read_records` directly so it inherits the new behavior.

## What's fixed

| Method | Status |
|---|---|
| `read_records(list / tuple / dict)` | ✅ Deterministic via content hash |
| `read_storage(single_file_uri)` | ✅ Deterministic (via `read_records`) |
| `read_storage([f1, f2, ...])` (list of single-file URIs) | ✅ Deterministic, order-independent |
| `read_storage(dir / glob)` | ✅ Deterministic (via listing fingerprint #1722) |
| `read_csv` / `read_parquet` / `read_json` | ✅ Inherits from `read_storage` |

## Still not deterministic (deferred from #1636)

| Method | Why |
|---|---|
| `read_values` | Internal seed placeholder + `.gen()` UDF closure; `hash_callable` doesn't capture closure state |
| `read_pandas`, `read_hf` | Use `read_values` → inherit |
| `read_records(generator)` / `read_database` | Lazy by design — not materialized for hashing |

## Schema

New column on `datasets_versions`: `content_hash TEXT NULL` — content-derived hash for materialized inputs; `NULL` everywhere else.

## Side fixes exposed by stable hashing

- `_resolve_checkpoint` now skips reuse when `version=` is explicit (previously a cached version could silently substitute the requested one).


src/datachain/lib/dc/datachain.py
```diff
@@ -641,6 +639,7 @@ def _resolve_checkpoint(
         project: Project,
         chain_hash: str,
         kwargs: dict,
+        version: str | None = None,
```
shcheklein: this is a bit weird - would it better to not call at all resolve_checkpoint then?

src/datachain/lib/dc/records.py
```diff
@@ -35,6 +43,31 @@ def _flatten_record(record: dict, signal_schema: SignalSchema) -> dict:
...
+) -> str:
+    """Compute a deterministic content hash for a concrete batch of records.
+
+    Stored as the temp-dataset version's `content_hash` so chains starting
+    from materialized records (read_records, single-file read_storage)
+    produce the same chain hash on identical inputs across runs — a
+    precondition for checkpoint reuse.
+    """
+    items: Iterable[dict] = [records] if isinstance(records, dict) else records
+    per_record = sorted(
```
shcheklein: can we avoid sorting? read_records now can run incrementally with a low memory footprint ... it is important to preserve

src/datachain/lib/dc/records.py
```diff
@@ -122,18 +155,28 @@ def generate_records():
         for c in signal_schema.db_signals(as_columns=True)
     ]
 
+    if isinstance(to_insert, dict):
+        to_insert = [to_insert]
+    elif not to_insert:
+        to_insert = []
+
+    is_seed_placeholder = (
```
shcheklein: why do we need this seed? if we can't calculate hash - let's not calculate it them - pass none, no?

src/datachain/lib/dc/datachain.py
ilongin: I did that at first but then wanted to encapsulate all in this method. I will return it outside this method though, don't have strong opinion about it.

src/datachain/lib/dc/records.py
ilongin: Yes, adding XOR instead of sorting

src/datachain/lib/dc/records.py
ilongin: Yea, the magic sentinel check was ugly. Here's the reason why it was there in the first place.

The seed gives `.gen()` one row to iterate over (real data comes through `.gen()` afterwards). The skip-hash is a workaround for a `hash_callable` "blind" spot: it doesn't capture closure cell contents. So `read_values(a=[1,2,3])` and `read_values(a=[4,5,6])` produce identical UDF hashes (data lives in `_func_fr`'s closure). If the seed also got a deterministic content_hash, two `read_values()` calls saved under the same name would collide and reuse the wrong checkpoint.

**Fix.** Dropped the sentinel detection. Extracted an internal `_read_records_with_hash(content_hash=...)`; `read_values` and the listing's `lst_fn` now pass `content_hash=None` explicitly with a comment naming the closure cause. Later in the code when we calculate start hash from `read_dataset` we try to take that new field `content_hash` or UUID (which is random). In this case, for `read_values` (and other `read_*` which use `read_values`) we will end up using UUID so hash will be non-deterministic all the time - which is what we want atm.

**Alternative we considered.** Make `hash_callable` closure-aware via cloudpickle. Would remove the workaround entirely, but it'd change hashes for every UDF whose closure captures anything - not sure if we want that .. also it was out of scope (more complex solution)

src/datachain/lib/dc/datachain.py
```diff
@@ -564,10 +564,12 @@ def save(  # type: ignore[override]
             (base_hash + f"{namespace_name}/{project_name}/{name}").encode("utf-8")
         ).hexdigest()
 
-        # Checkpoint handling
-        result = self._resolve_checkpoint(name, project, _hash, kwargs)
+        result = (
+            self._resolve_checkpoint(name, project, _hash, kwargs)
+            if version is None
```
shcheklein: again, what was the issue with using checkpoints with version? should we just include it into hash?

src/datachain/lib/dc/records.py
```diff
@@ -35,6 +37,104 @@ def _flatten_record(record: dict, signal_schema: SignalSchema) -> dict:
...
+    precondition for checkpoint reuse.
+    """
+    items: Iterable[dict] = [records] if isinstance(records, dict) else records
+    # XOR-combine per-record digests so order doesn't matter and we can
+    # stream records without buffering or sorting.
+    combined = 0
+    for rec in items:
+        payload = _json.dumps(
+            _flatten_record(rec, signal_schema), sort_keys=True, default=str
+        )
```
shcheklein: do we really need to use json and flatten, etc? every operation here matters - performance / memory footprint, etc

src/datachain/lib/dc/storage.py
```diff
@@ -123,8 +123,7 @@ def read_storage(
         ```
     """
     from .datasets import read_dataset
-    from .records import read_records
-    from .values import read_values
+    from .records import _read_records_with_hash, read_records
```
shcheklein: we are using protected call outside

also, I don't see a big reason for the split - looks more complicated w/o a good reason tbh

src/datachain/lib/dc/datachain.py
ilongin: The issue is that if someone asks to save specific version this basically means that we should never use checkpoints / reuse some existing dataset. Putting version into hash also does the job as checkpoint just won't be found. We will do one more DB / API call but not a huge deal so putting it there

src/datachain/lib/dc/records.py
ilongin: I refactor the code a little bit to only do once this flatten thing and use it for both hash and inserting. Before it was done twice which was wasteful

src/datachain/lib/dc/storage.py
ilongin: Reason for keeping the split: I didn't want to leak `content_hash` onto the public `read_records` signature - it's an internal  optimization that shouldn't show up in user-facing API.
I've renamed it to be without _ prefix

src/datachain/lib/dc/records.py
```diff
@@ -35,6 +36,98 @@ def _flatten_record(record: dict, signal_schema: SignalSchema) -> dict:
...
+            h.update(k.encode())
+            h.update(b"\0")
+            h.update(repr(rec[k]).encode())
+            h.update(b"\0")
+        combined ^= int.from_bytes(h.digest(), "big")
+    h = hashlib.sha256()
+    h.update(signal_schema.hash().encode())
+    h.update(b"\0")
+    h.update(combined.to_bytes(32, "big"))
+    return h.hexdigest()
```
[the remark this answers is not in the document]
shcheklein: good point I think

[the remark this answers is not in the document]
shcheklein: cc @ilongin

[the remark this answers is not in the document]
ilongin: fixed

shcheklein (approved):

shcheklein:
@ilongin review and fix conflicts + latest copilot comments.

PR #1753, opened 2026-05-07
improve docs for filter, add tests

shcheklein:
Document case-insensitive string matching in DataChain. Adds a note + example to `dc.func.string.regexp_replace` in `docs/guide/functions.md` showing the `(?i)` inline regex flag for case-insensitive replacement. Expands the `DataChain.filter()` docstring with examples for `glob`, `like`/`ilike`, `regexp` (with `(?i)`), `in_`, and `func`-based predicates, plus a quick-reference table of column-level filter operators and their case-sensitivity. Updates the internal SKILL example from `regexp_match` to the datachain-native `regexp`. Adds tests covering `like`/`ilike` filtering, `regexp` filtering with and without `(?i)`, and `regexp_replace` case-insensitive behavior. No API changes.

### DIFF
--- a/docs/guide/functions.md
+++ b/docs/guide/functions.md
@@ -112,7 +112,21 @@ String operations via `dc.func.string.*`:
 - `length(col)`: string length
 - `split(col, sep)`: split on separator
 - `replace(col, old, new)`: substring replacement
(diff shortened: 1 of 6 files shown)


dreadatour (approved):
👍

PR #1754, opened 2026-05-07
perf(file): stream File.save() to avoid OOM on large files

shcheklein:
## Problem

`File.save()` materializes the entire source file into memory via `self.read_bytes()` before handing the resulting `bytes` to `Client.upload()`, which calls `fs.pipe_file(path, data)`. For a multi-GB artifact this peaks at `file_size` of RAM and OOM-kills constrained workers — we hit this on a 7 GB k8s worker trying to save a ~6 GB `X_train.npy` artifact through DataChain Studio.

`fs.pipe_file` is bytes-only (single `f.write(value)` under the hood), and `fs.open(..., 'wb').write(huge_bytes)` would also buffer the whole payload locally before flushing — `AbstractBufferedFile` only flushes a multipart part once `self.blocksize` is reached, so a single big `.write()` accumulates everything in `self.buffer` first.

## Fix

- `Client.upload()` now accepts a bytes-like object **or** a binary readable stream:
  - bytes-like → existing fast `pipe_file` path, unchanged.
  - stream → copied into `fs.open(path, 'wb')` via `shutil.copyfileobj`. The destination `AbstractBufferedFile` flushes a multipart part every `blocksize` bytes (e.g. 5 MiB on s3fs), so **peak RAM is bounded by the backend's part size rather than the file size**.
- `File.save()` opens the source file and streams it directly into `Client.upload()`, eliminating the intermediate full-file bytes blob.
- `HTTPClient.upload` signature widened to match (still `NotImplementedError`).

## Verification

`tracemalloc` smoke test, saving a 210 MB local file via `File.save()`:

| version | peak alloc |
|---|---|
| before (`read_bytes()` + `pipe_file`) | ~210 MB (= file size) |
| after (streamed via `shutil.copyfileobj`) | **~0.2 MB** |

```python
import os, tempfile, tracemalloc
from datachain import Session

SIZE = 200 * 1024 * 1024
with tempfile.TemporaryDirectory() as a, tempfile.TemporaryDirectory() as b:
    src = os.path.join(a, 'big.bin'); open(src, 'wb').write(b'\x00' * SIZE)
    sess = Session.get()
    f = File.at(f'file://{src}'); f._set_stream(sess.catalog)
    tracemalloc.start(); f.save(os.path.join(b, 'out.bin'))
    _, peak = tracemalloc.get_traced_memory(); tracemalloc.stop()
    print(peak / 1e6, 'MB')   # ~0.2
```

## Tests

- New unit test `test_upload_accepts_binary_stream` in `tests/unit/test_client_local.py` covers the stream path.
- Existing `tests/unit/test_client_local.py` (69 passed, 6 Windows-only skipped) continues to pass — the bytes path is unchanged.

## Compatibility

- Public signature widened (`bytes` → `bytes | bytearray | memoryview | BinaryIO`); existing callers passing `bytes` are unaffected.
- All in-tree call sites continue to pass `bytes` and exercise the `pipe_file` fast path.

### DIFF
--- a/src/datachain/client/fsspec.py
+++ b/src/datachain/client/fsspec.py
@@ -469,9 +469,20 @@ class Client(ABC):
 
         if isinstance(data, (bytes, bytearray, memoryview)):
             self.fs.pipe_file(full_path, data)
-        else:
+        elif hasattr(data, "read"):
             with self.fs.open(full_path, "wb") as dst:
-                shutil.copyfileobj(data, dst)
+                # Use a larger copy buffer than shutil's 16 KiB default to
+                # reduce per-chunk overhead on multi-GB uploads. The
+                # destination's blocksize (when exposed by fsspec backends)
+                # matches the multipart part size it will flush at, making it
+                # a natural choice; fall back to 8 MiB otherwise.
+                buf_size = getattr(dst, "blocksize", None) or 8 * 1024 * 1024
+                shutil.copyfileobj(data, dst, length=buf_size)
+        else:
+            raise TypeError(
+                "Client.upload(data) expects bytes-like data or a binary "
+                f"readable stream (with a .read() method); got {type(data).__name__}"
+            )
         file_info = self.fs.info(full_path, **self._file_info_kwargs())
         return self.info_to_file(file_info, rel_path)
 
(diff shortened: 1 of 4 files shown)


dmpetrov (approved):
LG

dreadatour (approved):
🔥

PR #1756, opened 2026-05-08
Added similarity search wrapper DataChain method

### LINKED ISSUE #1752: Similarity search wrapper
### Description

We need a similarity search shortcut for docs and blog posts. Without this it requires 3 commands to implement - `.mutate(score=...).order_by().limit()` not to mention a mental overhead on new column and all the function names like `dc.func.cosine_distance`.

It should be simple like:
```python
ds.similarity_search("emb", k=5)
```

Signature:
```python
def similarity_search(
    self,
    column: str,
    k: int = 10,
    metric: str = "cosine", # or "euclidean" or "l2"
    score_column: str | None = None,  # temp name by details - do not forget to remove it.
) -> self:
```

ilongin:
New DataChain method `similarity_search` that wraps / simplifies code like:
```python
  chain = (
      chain                                                                                                                    
      .mutate(_temp_score=func.cosine_distance(C("emb"), query))
      .order_by("_temp_score")                                                         
      .limit(5) 
      .select_except("_temp_score")                                                                  
  )
```

into:
```python
chain = chain.similarity_search("emb", query, k=5)
```


src/datachain/lib/dc/datachain.py
```diff
@@ -95,6 +96,10 @@
 
 DEFAULT_PARQUET_CHUNK_SIZE = 100_000
 
+# Frozen UUID-suffixed name for the temp similarity-score column (must be
+# deterministic across runs so chain hashes stay stable for checkpoint reuse).
+SIMILARITY_SCORE_COL_NAME = "similarity_a4f74e9a1ea74dd293049359291d0c38"
```
shcheklein: long names complicate SQL (we hit different type of issue bc of it)

it is better to make it short and use suffix to avoid collisions

shcheklein (approved):
a minor comment that I would address, length of SQL is really painful,

amritghimire (approved):

dmpetrov:
@ilongin thank you for implementing this!
please close the original issue #1756 next time and reference to original issue from the description - it will close it automatically after merge.

Small feedback: `query` param name might be not the best. It might worth to rename it to sometime like `emd` or `vector`.

PR #1757, opened 2026-05-08
Fixing image info to read only header

### LINKED ISSUE #1713: get_info - parial download works only in GS
`ImageFile.get_info()` suppose to download only file header. It does work in GS but **it does not work in S3**.

The bigger issue - the code is the same for all storages but why some behave differently (S3).

**A biggest concern**: it might happening that in some storages (S3) any small `read()` leads to full file download.

```python
def get_image_info(file: dc.ImageFile) -> dc.Image:
    return file.get_info()

chain.map(info=get_image_info)
```

- [ ] Research the cloud difference for this particular method: GS, S3, AZ
- [ ] Unify storages behaviour

ilongin:
Fixes #1713.

`ImageFile.get_info()` was downloading the whole file on every backend (S3, GS, etc.), not just S3 as the issue suggested.

### Cause
`image_info` called `file.read()`, which pulls all bytes into memory and wraps them in `BytesIO` before handing them to PIL.

### Fix
Hand PIL a streaming file-like via `file.open()`. PIL reads only the header bytes it needs; fsspec fetches a single block (~5 MiB) instead of the full file.

### Verified
Ran an instrumented experiment patching `s3fs.S3File._fetch_range` / `gcsfs.GCSFile._fetch_range` to count cloud bytes and verified that code before fix was downloading whole file every time and after fix only first part of it for header read.

### Follow-up (out of scope)
`video_info` still calls `ensure_cached()` because `ffmpeg.probe()` needs a local path. Switching to `pyav` (`av.open(stream)`) would allow streaming, but MP4s without faststart layout require reading the `moov` atom from near EOF, so partial download isn't always possible. Worth a separate issue.


src/datachain/lib/image.py
```diff
@@ -10,23 +10,27 @@ def image_info(file: File | ImageFile) -> Image:
     """
     Returns image file information.
 
+    Streams the file so PIL only reads the header bytes it needs to
```
shcheklein: this is implementation detail. At least simplify it a lot if we keep it.

or can be a note at the end

src/datachain/lib/image.py
```diff
@@ -10,23 +10,27 @@ def image_info(file: File | ImageFile) -> Image:
...
     Args:
         file (ImageFile): Image file object.
 
     Returns:
         Image: Image file information.
     """
     try:
-        img = file.as_image_file().read()
+        with file.as_image_file().open(mode="rb") as stream:
+            with PILImage.open(stream) as img:
```
shcheklein, line 22: how does PIL work - is it some C code / library?

ilongin: `Image.open(stream)` it self is pure Python and does **not** touch the C decoders. 
Here we only use it to inspect the image header: `Image.open(stream)` parses just enough to give us `.width`, `.height`, and `.format`. No pixels are decoded, that only happens when you call `.load()` / `.convert()`, which we don't.

amritghimire (approved):

dreadatour (approved):

shcheklein (approved):

PR #1759, opened 2026-05-10
Fix e2e tests cloud-fixture

dreadatour:
The session-scoped `cloud_server_credentials` fixture was writing `AWS_DEFAULT_REGION=pytest-servers-region` (and AWS keys) directly to `os.environ`. In CI configurations that run cloud-fixture tests and the e2e tests in the same Python process (Studio tests), the e2e subprocess inherited those vars via `env={**os.environ, ...}` and boto3 built `https://<bucket>.s3.pytest-servers-region.amazonaws.com/`.

Fix: add `tests.utils.e2e_subprocess_env()` that strips `AWS_*` before adding DataChain overrides.
Improvement: switch `cloud_server_credentials` to `monkeypatch_session` so the AWS env is undone at session teardown instead of leaking permanently.

Example of failed tests (4 times in a row): https://github.com/datachain-ai/datachain/actions/runs/25605189540/job/75176712858?pr=1755

Note: AI was used to investigate and fix this nasty issue

### DIFF
--- a/tests/test_query_e2e.py
+++ b/tests/test_query_e2e.py
@@ -1,4 +1,3 @@
-import os
 import os.path
 import signal
 import subprocess
@@ -11,6 +10,8 @@ from threading import Thread
 from typing import IO
 
 import pytest
+
+from tests.utils import e2e_subprocess_env
 
 tests_dir = os.path.dirname(os.path.abspath(__file__))
 
@@ -185,13 +186,11 @@ def run_step(step, catalog):
         popen_args = {"start_new_session": True}
     stdin_path = step.get("stdin_file")
     with open(stdin_path) if stdin_path else nullcontext(None) as stdin_file:
-        # Build env without DATACHAIN_MAIN_PROCESS_PID so script starts fresh
-        # as its own main process (with checkpoints enabled)
-        script_env = {
-            k: v for k, v in os.environ.items() if k != "DATACHAIN_MAIN_PROCESS_PID"
-        }
-        script_env["DATACHAIN__METASTORE"] = catalog.metastore.serialize()
-        script_env["DATACHAIN__WAREHOUSE"] = catalog.warehouse.serialize()
+        # Strip DATACHAIN_MAIN_PROCESS_PID so script starts fresh as its own
+        # main process (with checkpoints enabled).
+        script_env = e2e_subprocess_env(
+            catalog, exclude_keys={"DATACHAIN_MAIN_PROCESS_PID"}
+        )
 
         process = subprocess.Popen(  # noqa: S603
             command,
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -176,6 +176,23 @@ def wait_for_condition(
     raise TimeoutError(f"Timeout expired while waiting for: {message}")
 
 
+def e2e_subprocess_env(
+    catalog: Catalog, exclude_keys: set[str] | None = None
+) -> dict[str, str]:
+    # Strip AWS_* so the subprocess talks to real public S3 instead of inheriting
+    # moto/pytest-servers credentials and region from cloud-emulator fixtures
+    # that ran earlier in the same pytest process.
+    exclude = exclude_keys or set()
+    env = {
+        k: v
+        for k, v in os.environ.items()
+        if not k.startswith("AWS_") and k not in exclude
+    }
+    env["DATACHAIN__METASTORE"] = catalog.metastore.serialize()
+    env["DATACHAIN__WAREHOUSE"] = catalog.warehouse.serialize()
+    return env
+
+
 def run_test_subprocess(command, env: dict[str, str], capture_output: bool = True):
     stdout = subprocess.PIPE if capture_output else None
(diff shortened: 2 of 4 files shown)


shcheklein:
Was it failing before? What changed?

dreadatour:
> Was it failing before? What changed?

It starts failing in this PR: https://github.com/datachain-ai/datachain/pull/1755, fix works.

Nothing was changed except for renaming durectories/modules in Studio. I wonder why it starts failing now, but reasoning about env variables overriding in tests looks solid to me.

shcheklein:
> I wonder why it starts failing now, but reasoning about env variables overriding in tests looks solid to me.

right, but is is suspicious a bit - why wasn't it failing before ... can it depend on some ordering?

dreadatour:
> > I wonder why it starts failing now, but reasoning about env variables overriding in tests looks solid to me.
> 
> right, but is is suspicious a bit - why wasn't it failing before ... can it depend on some ordering?

I was thinking about it but I can not imagine why order might changed. Another though was some tests were not running before for some reason. I can dig into it.

dreadatour:
> > I wonder why it starts failing now, but reasoning about env variables overriding in tests looks solid to me.
> 
> right, but is is suspicious a bit - why wasn't it failing before ... can it depend on some ordering?

Here is what Claude says:

`pytest-split` distributes tests across the 6 shards using `.test_durations`. The failing run was on Studio branch [rename-datachain-worker](https://github.com/datachain-ai/datachain/pull/1755), which renamed Studio modules. Renamed test IDs no longer matched entries in `.test_durations`, so `pytest-split` fell back to its default distribution for the unmatched tests. That re-shuffle landed `test_cli_e2e` and a cloud-fixture-using test (e.g., `tests/func/test_client.py`) on the same shard process for the first time. Session-scoped `cloud_server_credentials` ran first and wrote the bogus AWS env into `os.environ`; the e2e subprocess inherited it via `env={**os.environ, ...}`.

So it's an ordering/grouping change, but caused by `pytest-split` re-shuffling after the Studio rename, not by any change in datachain itself.

dreadatour:
I should also update `.test_durations` in https://github.com/datachain-ai/datachain/pull/1755 after this PR will be merged.

dreadatour:
@shcheklein check test failure: https://github.com/datachain-ai/datachain/actions/runs/25621487088/job/75208882222?pr=1759 — same as in https://github.com/datachain-ai/datachain/pull/1760. It happens too often for me recently 🤔

tests/test_query_e2e.py
shcheklein, line 251: how about this? does it affect any distribution?

tests/utils.py
```diff
@@ -176,6 +176,23 @@ def wait_for_condition(
     raise TimeoutError(f"Timeout expired while waiting for: {message}")
 
 
+def e2e_subprocess_env(
+    catalog: Catalog, exclude_keys: set[str] | None = None
+) -> dict[str, str]:
+    # Strip AWS_* so the subprocess talks to real public S3 instead of inheriting
```
shcheklein, line 182: do we really need to talk to real S3? can it then pick up some random default credential I have on my machine by default?

dreadatour: As far as I can remember we use anon=True in those tests. But I'll check.

tests/test_query_e2e.py
dreadatour: Not sure if I follow your question. With `xdist_group` we are forcing all the tests with this mark (`tmpfile`) to run on the same worker (test runner) — it affects distribution, yes.

tests/utils.py
dreadatour: > do we really need to talk to real S3?

Good question. I think we don't have to in these tests. We can now connect to real S3 (GCP, etc) from nightly benchmark runs.

tests/test_query_e2e.py
dreadatour: Oh, if you meant we can use `xdist_group` to mark all the tests to fix this — yes, we can, but this will hide the problem, not solve it. I believe manipulating `os.environ` the way it was done in these tests is not a good way of writing tests. Ideally every test should now affects anything (including the runner) — that's why we use fixtures, for example.

shcheklein: no, I'm just exploring a bit to better understand. I don't suggest anything. Curious why would we even put all tempfile tests into a single bucket tbh. Seems redundant and probably not consistent.

tests/utils.py
shcheklein: my question was - do we even talk to real S3 in these or any tests at all (not to change anything)? and what are those tests (just again better understand it)

shcheklein (approved):

tests/utils.py
dreadatour: I'll be back to this in separate PR or issue, good questions to discuss 👍

tests/test_query_e2e.py
dreadatour: I'll be back to this in separate PR or issue, good questions to discuss 👍

PR #1760, opened 2026-05-10
fix(save): roll back claimed version on post-claim failure

dreadatour:
Wrap phase 3-7 of `DatasetQuery.save()` in a context manager that removes the just-claimed dataset version if any subsequent step raises (rename, link-job, dependencies, complete). Without this, post-claim failures (e.g. SQLite contention) left orphan `CREATED` rows in the metastore, which made _next_auto_version skip ahead and caused the flaky tests.

Test failure examples: 
- https://github.com/datachain-ai/datachain/actions/runs/25605283753/job/75202371624
- https://github.com/datachain-ai/datachain/actions/runs/25619068862/job/75202286502

```
=================================== FAILURES ===================================
_________________ test_concurrent_save_fails_after_max_retries _________________
[gw1] linux -- Python 3.10.20 /home/runner/work/datachain/datachain/.nox/e2e-3-10/bin/python3

tmp_dir = PosixPath('/tmp/pytest-of-runner/pytest-1/popen-gw1/datachain-test2')
catalog_tmpfile = <datachain.catalog.catalog.Catalog object at 0x7f1c1a081e10>

    @skip_if_not_sqlite
    @pytest.mark.e2e
    @pytest.mark.xdist_group(name="tmpfile")
    def test_concurrent_save_fails_after_max_retries(tmp_dir, catalog_tmpfile):
        dataset_name = f"concurrent_save_max_{uuid.uuid4().hex}"
        barrier_dir = tmp_dir / "concurrent-save-max-barrier"
        script = os.path.join(tests_dir, "scripts", "concurrent_save.py")
        process_count = 10
    
        env = {
            **os.environ,
            "DATACHAIN__METASTORE": catalog_tmpfile.metastore.serialize(),
            "DATACHAIN__WAREHOUSE": catalog_tmpfile.warehouse.serialize(),
            "DATACHAIN_CONCURRENT_SAVE_DATASET": dataset_name,
            "DATACHAIN_CONCURRENT_SAVE_BARRIER_DIR": os.fspath(barrier_dir),
            "DATACHAIN_CONCURRENT_SAVE_PARTIES": str(process_count),
            "DATACHAIN_CONCURRENT_SAVE_SYNC_ALL": "1",
        }
    
        processes = [
            run_test_subprocess(
                (python_exc, script),
                {**env, "DATACHAIN_CONCURRENT_SAVE_WORKER": str(worker_id)},
            )
            for worker_id in range(1, process_count + 1)
        ]
    
        results = []
        for process in processes:
            _, stdout, stderr = wait_for_test_subprocess(
                process, timeout=E2E_STEP_TIMEOUT_SEC
            )
            assert stdout.strip(), stderr
            results.append(json.loads(stdout.strip().splitlines()[-1]))
    
        successes = [result for result in results if result["status"] == "success"]
        failures = [result for result in results if result["status"] == "error"]
    
        # At least one worker must succeed, and under heavy contention some may
        # fail (barrier timeouts, lock retries, etc.), so we don't assert exact
        # counts — only that every result is accounted for and consistent.
        assert len(successes) >= 1
        assert len(successes) + len(failures) == process_count
    
        # Successful versions must be unique and sequential starting from 1.0.0
        versions = sorted(result["version"] for result in successes)
        expected_versions = [f"1.0.{i}" for i in range(len(successes))]
>       assert versions == expected_versions
E       AssertionError: assert ['1.0.0', '1....0.4', '1.0.5'] == ['1.0.0', '1....0.3', '1.0.4']
E         
E         At index 1 diff: '1.0.2' != '1.0.1'
E         
E         Full diff:
E           [
E               '1.0.0',
E         -     '1.0.1',
E               '1.0.2',
E               '1.0.3',
E               '1.0.4',
E         +     '1.0.5',
E           ]

/home/runner/work/datachain/datachain/tests/test_atomicity.py:172: AssertionError
```

Note: AI was used to investigate and fix this nasty issue

### DIFF
--- a/src/datachain/query/dataset.py
+++ b/src/datachain/query/dataset.py
@@ -3389,47 +3389,50 @@ class DatasetQuery:
 
             version = version or dataset.latest_version
 
-            # Phase 3: Rename staging table to the final dataset table name.
-            final_table_name = self.catalog.warehouse.dataset_table_name(
-                dataset, version
-            )
-            self.catalog.warehouse.rename_table(temp_table, final_table_name)
-            self.temp_table_names.remove(temp_table_name)
-
-            # Link this dataset version to the job that created it
-            if job_id:
-                self.catalog.metastore.link_dataset_version_to_job(
-                    dataset.get_version(version).id, job_id, is_creator=True
+            # Phases 3-7: roll back the claim if anything below fails so we
+            # never leave an orphan CREATED row in the metastore.
+            with self._rollback_version_on_failure(dataset, version):
+                # Phase 3: Rename staging table to the final dataset table name.
+                final_table_name = self.catalog.warehouse.dataset_table_name(
+                    dataset, version
                 )
+                self.catalog.warehouse.rename_table(temp_table, final_table_name)
+                self.temp_table_names.remove(temp_table_name)
 
-            if dependencies:
-                # overriding dependencies
-                self.dependencies = set()
-                for dep in dependencies:
-                    self.dependencies.add(
-                        (
-                            self.catalog.get_dataset(
-                                dep.name,
-                                namespace_name=dep.namespace,
-                                project_name=dep.project,
-                                versions=[dep.version],
-                                include_incomplete=False,
-                            ),
-                            dep.version,
-                        )
+                # Link this dataset version to the job that created it
+                if job_id:
+                    self.catalog.metastore.link_dataset_version_to_job(
+                        dataset.get_version(version).id, job_id, is_creator=True
                     )
 
-            self._add_dependencies(dataset, version)  # type: ignore [arg-type]
+                if dependencies:
+                    # overriding dependencies
+                    self.dependencies = set()
+                    for dep in dependencies:
+                        self.dependencies.add(
+                            (
+                                self.catalog.get_dataset(
+                                    dep.name,
+                                    namespace_name=dep.namespace,
+                                    project_name=dep.project,
+                                    versions=[dep.version],
+                                    include_incomplete=False,
+                                ),
+                                dep.version,
+                            )
+                        )
 
-            if kwargs.get("listing"):
-                reused = self._reuse_listing_if_unchanged(
-                    dataset, version, name, project
-                )
-                if reused is not None:
-                    return reused
+                self._add_dependencies(dataset, version)  # type: ignore [arg-type]
 
-            # Mark as COMPLETE only after all operations succeed.
-            self.catalog.complete_dataset_version(dataset, version)
+                if kwargs.get("listing"):
+                    reused = self._reuse_listing_if_unchanged(
+                        dataset, version, name, project
+                    )
+                    if reused is not None:
+                        return reused
+
+                # Mark as COMPLETE only after all operations succeed.
+                self.catalog.complete_dataset_version(dataset, version)
         finally:
             self.cleanup()
         return self.__class__(
@@ -3439,6 +3442,26 @@ class DatasetQuery:
             version=version,
             catalog=self.catalog,
         )
+
+    @contextlib.contextmanager
+    def _rollback_version_on_failure(
+        self, dataset: "DatasetRecord", version: str
+    ) -> Iterator[None]:
+        """Remove the just-claimed dataset version on any exception so the
+        metastore never holds an orphan CREATED row when post-claim phases
+        fail (e.g. SQLite contention during rename or stats refresh)."""
+        try:
+            yield
+        except BaseException:
+            try:
+                self.catalog.remove_dataset_version(dataset, version, drop_rows=True)
+            except Exception:
+                logger.exception(
+                    "Failed to roll back claimed version %s of dataset %s",
+                    version,
+                    dataset.name,
+                )
+            raise
 
     def _reuse_listing_if_unchanged(
         self,
--- a/tests/test_atomicity.py
+++ b/tests/test_atomicity.py
@@ -6,7 +6,10 @@ import uuid
 import pytest
 import sqlalchemy as sa
 
+import datachain as dc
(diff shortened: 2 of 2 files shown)


src/datachain/query/dataset.py
```diff
@@ -3440,6 +3443,26 @@ def save(
...
+        self, dataset: "DatasetRecord", version: str
+    ) -> Iterator[None]:
+        """Remove the just-claimed dataset version on any exception so the
+        metastore never holds an orphan CREATED row when post-claim phases
+        fail (e.g. SQLite contention during rename or stats refresh)."""
+        try:
+            yield
+        except BaseException:
+            try:
+                self.catalog.remove_dataset_version(dataset, version, drop_rows=True)
```
shcheklein, line 3457: so it still can fail and we will get the same state anyways, right?

dreadatour: Yes :( Also Copilot's comment from above about the same

dreadatour: But suggestion from Copilot is one step forward, although it is not the ideal solution.

shcheklein:
I think we need better understanding of why test is actually failing, I can take a look a bit later ... since I was doing this code ... AFAIR test was quite complicated by itself (since it is testing w/o monkeypatches, real flow with some barriers, etc) - so might be just not 100% reliable by itlsef

src/datachain/query/dataset.py
shcheklein: I think it is becoming too complicated

Assumption was - it is fine to fail, it is fine to leave it all (and GC takes care of it later). I think it is a simple model, I don't think we should be complicating it tbh

shcheklein: Note also - it is a very narrow window when it can fail and leave stuff for GC - some metadata manipulation, etc. Not like the whole query.

dreadatour:
> I think we need better understanding of why test is actually failing, I can take a look a bit later ... since I was doing this code ... AFAIR test was quite complicated by itself (since it is testing w/o monkeypatches, real flow with some barriers, etc) - so might be just not 100% reliable by itlsef

Weird thing is this test was failing 2 times in the row during tomorrow, I can not find failures of this test before.

dreadatour:
> I think we need better understanding of why test is actually failing, I can take a look a bit later ... since I was doing this code ... AFAIR test was quite complicated by itself (since it is testing w/o monkeypatches, real flow with some barriers, etc) - so might be just not 100% reliable by itlsef

Closing this for now 👍 Let me know if I can help somehow

dreadatour:
JFYI: same failure happens before, 2 months ago: https://github.com/datachain-ai/datachain/actions/runs/23811665644/job/69400061045

### AI investigation

The latent bug was introduced by [PR #1603](https://github.com/datachain-ai/datachain/pull/1603) — commit [`aaa97d3e`](https://github.com/datachain-ai/datachain/commit/aaa97d3e) "fix(save): make isolated and atomic" (merged Mar 16, 2026)

That PR introduced the staging-table + multi-phase save pipeline (Phase 1 populate → Phase 2 claim → Phase 3-7 finalize) but never added rollback on post-claim failure. The test [`tests/test_atomicity.py:128`](tests/test_atomicity.py:128) was added the next day in commit `03617824` (Mar 15-16), already targeting that flow.

CI history of this exact failure (`['1.0.0', '1.0.X', ..., '1.0.5'] != ['1.0.0', ...]`):

| Date | Run | Branch / Commit | Runner |
|---|---|---|---|
| Mar 31 | [23811665644](https://github.com/datachain-ai/datachain/actions/runs/23811665644) | main / `3c5f5da5` | **Windows** |
| May 6 | [25420556398](https://github.com/datachain-ai/datachain/actions/runs/25420556398) | main / `7d7747c3` | **Linux 8-core** ← first Linux failure |
| May 6 | [25431735440](https://github.com/datachain-ai/datachain/actions/runs/25431735440) | feature branch | Linux |
| May 7 | [25498131714](https://github.com/datachain-ai/datachain/actions/runs/25498131714) | feature branch | Linux |
| May 9 | [25605283753](https://github.com/datachain-ai/datachain/actions/runs/25605283753), [25605973579](https://github.com/datachain-ai/datachain/actions/runs/25605973579), [25606643564](https://github.com/datachain-ai/datachain/actions/runs/25606643564) | various | Linux |
| May 10 | [25619068862](https://github.com/datachain-ai/datachain/actions/runs/25619068862) | your `rename-datachain-worker` | Linux |

So:
- The bug has existed since the staging-table refactor on **Mar 16** (`aaa97d3e`).
- The test caught it on Windows almost immediately (**Mar 31**) but went unnoticed.
- The same pattern started showing on **Linux 8-core** runners around **May 6** — likely because [`512bee83`](https://github.com/datachain-ai/datachain/commit/512bee83) "emit subselect if it is neeeded" (#1732, Apr 29) restructured `apply_steps()` and slightly shifted timing in `save()`'s Phase 1.

---

One more fresh failure here: https://github.com/datachain-ai/datachain/actions/runs/25621487088/job/75208882222?pr=1759

dreadatour:
@shcheklein Random thoughts: what if we will create dataset version without specifying the version (detached version) and only after successful creation (including table in warehouse) we will set version (to the latest one or specified by user)?

We will also need to run garbage collector to cleanup these detached versions after a while.

This will require refactoring, but I can not see any other robust solutions to this, other than what AI already suggests in this PR.

shcheklein:
I'm not sure how that would simplify anything tbh ... and I don't see any value in what AI suggests also in this PR so far - except it becomes even more complicated.

dreadatour:
Thank you for the fix in https://github.com/datachain-ai/datachain/pull/1762 🔥

PR #1762, opened 2026-05-10
fix: scope created dataset version returns

shcheklein:
Should fix a potential race condition in `save()`. It is leading to some tests failing from time to time - https://github.com/datachain-ai/datachain/pull/1759#issuecomment-4414656054


dreadatour (approved):
🔥❤️

PR #1763, opened 2026-05-11
Auto-detect public buckets, no `anon=True` needed

### LINKED ISSUE #1712: Public buckets require anon=True
### Description

That's not obvious that user should put anon=True for public buckets. Agents constantly make this mistakes which slowes down DC code generation.

Related to #740

I haven't check AZ, but this is correct for S3 and GC.

ilongin:
Fixes #1712.

Agents (and humans) keep forgetting to pass `anon=True` when reading from public buckets - `dc.read_storage("s3://my-public-bucket/...")` fails with auth errors instead of just working.

When `anon` is left at its default (`None`), `read_storage` now probes each unique bucket via `bucket_status` (shipped in #1717). If **every** bucket reports `access == "anonymous"`, `anon=True` is set automatically. Any other case (authenticated, denied, mixed, network error, unsupported backend) falls through to current behavior.
One extra HEAD-style request per unique bucket is needed, only when `anon` isn't passed explicitly. Negligible vs. the listing itself.


src/datachain/lib/dc/storage.py
```diff
@@ -18,6 +20,40 @@
...
+    probed: set[tuple[type[Client], str]] = set()
+    for uri in uris:
+        try:
+            client_cls = Client.get_implementation(str(uri))
+        except NotImplementedError:
+            return False
+        name, _ = client_cls.split_url(str(uri))
+        if not name:
+            return False
+        probed.add((client_cls, name))
```
amritghimire: probed here implies it is already probed, but it is not until later.

amritghimire (commented):
Won't this make the run actually slow for non anon cases with unnecessary extra api call?

tests/func/test_storage_auto_anon.py
```diff
@@ -0,0 +1,32 @@
+from unittest.mock import patch
+
+import pytest
+
+import datachain as dc
+
+
+@patch("datachain.lib.dc.storage._all_buckets_anonymous", return_value=True)
```
shcheklein: we need real tests with actual cloud mocks or even actual cloud

* bucket allows anon but some default credentials are also set and those credentials don't allow access to that bucket
* bucket allows anon read but some credentials are set for certain operation but not enough for read for example or something (need to think this through and experiment)
* bucket allows anon and no default credentials are set

src/datachain/lib/dc/storage.py
```diff
@@ -18,6 +20,40 @@
     from .datachain import DataChain
 
 
+logger = logging.getLogger("datachain")
+
+
+def _all_buckets_anonymous(uris, client_config: dict | None) -> bool:
+    """True only if every unique bucket in ``uris`` probes as anonymous.
+
+    Any failure (unsupported backend, network error, mixed access) returns
```
shcheklein: why would we ignore network errors? should we just raise them?

src/datachain/lib/dc/storage.py
```diff
@@ -18,6 +20,40 @@
...
+        if not name:
+            return False
+        probed.add((client_cls, name))
+
+    if not probed:
+        return False
+
+    for client_cls, name in probed:
+        try:
+            status = client_cls.bucket_status(name, **(client_config or {}))
```
shcheklein, line 53: double check - do we want to pass client config here or would we want to keep it empty? or ignore some keys? what does it do with it? what if crendetials are passed in it - will it still check for anon?

src/datachain/lib/dc/storage.py
ilongin: renamed

src/datachain/lib/dc/storage.py
ilongin: `client_config` is passed so the probe uses the same endpoint/region as the listing.

Each backend's `bucket_status` already strips creds for the anonymous probe, but you're right that an explicit-creds caller + public bucket would have us silently override their creds.

Fix: added `Client.has_explicit_credentials(client_config)` with per-backend `CREDENTIAL_KEYS`; `_all_buckets_anonymous` short-circuits if creds are present. Auto-anon now only fires when no creds were passed.

src/datachain/lib/dc/storage.py
ilongin: Dropped the try/catch completely. If any error happens, it will be propagated.

tests/func/test_storage_auto_anon.py
ilongin: added tests

shcheklein (approved):

PR #1764, opened 2026-05-11
fix(read_storage): always attach listing_fn so refresh paths can re-list

shcheklein:
The listing function was only attached when `update=True` or the listing dataset did not exist at chain construction time. However, `DatasetQuery.resolve_listing` also invokes `listing_fn` when the listing is detected as expired at execution time (or has been cleaned up between construction and execution, e.g. for ephemeral listings introduced in #1718). In those cases `listing_fn` was `None` and the assertion `assert self.listing_fn` failed.

Attach the listing function unconditionally - the gating logic in `resolve_listing` is unchanged, so this only affects the refresh paths that were previously broken. The placeholder `lambda: True` already used in `read_listing_dataset` confirms the project convention of always having a listing function available.

### DIFF
--- a/src/datachain/lib/dc/storage.py
+++ b/src/datachain/lib/dc/storage.py
@@ -170,7 +170,7 @@ def read_storage(
             updated_uris.add(list_uri_to_use)
             update_single_uri = True
 
-        list_ds_name, list_uri, list_path, list_ds_exists = get_listing(
+        list_ds_name, list_uri, list_path, _ = get_listing(
             list_uri_to_use, session, update=update_single_uri
         )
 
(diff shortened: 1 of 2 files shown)


dreadatour (approved):
Looks good to me 👍

src/datachain/lib/dc/storage.py
```diff
@@ -198,36 +198,40 @@ def read_storage(
...
+                .gen(
+                    list_bucket(lst_uri, cache, client_config=client_config),
+                    output={f"{column}": file_type},
+                )
+                # for internal listing datasets, we always bump major version
+                .save(ds_name, listing=True, update_version="major")
             )
 
+        # Always attach the listing function. `resolve_listing` may need to call
+        # it later not only when ``update`` is set or the dataset is missing at
```
ilongin: I would shorten this comment, seems unnecessarily verbose.

ilongin (approved):
Looks good

shcheklein:
@copilot resolve the merge conflicts in this pull request

PR #1769, opened 2026-05-13
feat(dataset): add UUID field to dataset records

amritghimire:
Add automatic UUID generation for datasets with:
- UUID column in datasets table with unique constraint
- Automatic UUID4 generation for new datasets
- Support for explicit UUID specification during creation
- Database migration support with UUID backfill
- Comprehensive test coverage for UUID functionality

This enables unique dataset identification across different DataChain instances and improves dataset tracking capabilities.


src/datachain/data_storage/metastore.py
```diff
@@ -777,6 +778,8 @@ def _datasets_columns(cls) -> list["SchemaItem"]:
         """Datasets table columns."""
         return [
             Column("id", Integer, primary_key=True),
+            Column("uuid", Text, nullable=False, default=lambda: str(uuid4())),
+            Index("uq_datasets_uuid", "uuid", unique=True),
```
dreadatour: Let's keep indexes in the end of the list?

src/datachain/data_storage/metastore.py
```diff
@@ -777,6 +778,8 @@ def _datasets_columns(cls) -> list["SchemaItem"]:
         """Datasets table columns."""
         return [
             Column("id", Integer, primary_key=True),
+            Column("uuid", Text, nullable=False, default=lambda: str(uuid4())),
```
dreadatour, line 781:
```suggestion
            Column("uuid", Text, nullable=False, default=uuid4),
```

src/datachain/data_storage/metastore.py
```diff
@@ -1215,8 +1218,24 @@ def create_dataset(
         else:
             project = self.get_project_by_id(project_id)
 
+        if uuid is not None:
+            # Validate UUID format if explicitly provided
+            if not uuid.strip():
+                raise ValueError("UUID cannot be empty or whitespace")
```
dreadatour: I think we should not allow flexible inputs for internal API — like check for empty string or string with whitespaces only. It should be strict IMO to catch errors earlier and do not propagate flaws in the code into other internal methods (callers).
```suggestion
        if uuid:
```

src/datachain/dataset.py
```diff
@@ -487,6 +488,7 @@ class DatasetRecord:
     attrs: list[str]
     schema: dict[str, SQLType | type[SQLType]]
     feature_schema: dict
+    uuid: str = field(kw_only=True, default_factory=lambda: str(uuid4()))
```
dreadatour, line 491: Just wonder, will this works here?
```suggestion
    uuid: str = field(kw_only=True, default_factory=uuid4)
```

dreadatour (approved):
Looks good to me overall, couple tiny comments.

Few important questions about this PR:

1. Do we have correlated SaaS PR already?
2. This will be a backward incompatible change in SaaS — we need to be extremely careful
3. This might be a backward incompatible change for CLI also in case Studio datasets are in local DB, because for those datasets we will generate `uuid` independently in CLI and in SaaS and they will not match. Am I right? 🤔

amritghimire:
> Looks good to me overall, couple tiny comments.
> 
> Few important questions about this PR:
> 
> 1. Do we have correlated SaaS PR already?

https://github.com/datachain-ai/studio/pull/12867

> 2. This will be a backward incompatible change in SaaS — we need to be extremely careful

Yes, it is backward incompatible changes in SaaS. 

> 3. This might be a backward incompatible change for CLI also in case Studio datasets are in local DB, because for those datasets we will generate `uuid` independently in CLI and in SaaS and they will not match. Am I right? 🤔

We don't fetch using uuid yet, and I have tested it with production saas, it works properly.

src/datachain/dataset.py
amritghimire: No it won't because it has str type and uuid4 returns UUID type

src/datachain/data_storage/metastore.py
amritghimire: The return type of uuid4 is UUID, but the column is of text type.

src/datachain/data_storage/metastore.py
```diff
@@ -1218,20 +1218,7 @@ def create_dataset(
...
-                # Validate UUID format by parsing it
-                import uuid as uuid_module
-
-                uuid_module.UUID(uuid)
-                my_uuid = uuid
-            except ValueError as e:
-                raise ValueError(f"Invalid UUID format: {uuid}") from e
-        else:
-            my_uuid = str(uuid4())
+        my_uuid = uuid.strip() if uuid and uuid.strip() else str(uuid4())
```
dreadatour: Why did we remove uuid validation? 🤔

amritghimire: I think I misunderstood https://github.com/datachain-ai/datachain/pull/1769#discussion_r3279336615. You mean to remove whitespace one?

PR #1773, opened 2026-05-18
fix(examples): guard against empty choices and message=None in multimodal LLM response

qizwiz:
## What

Guards `response.choices[0]` in `examples/multimodal/openai_image_desc_lib.py` before accessing `.message.content`:

```python
# Before
openai_description = response.choices[0].message.content or ""

# After
if not response.choices or response.choices[0].message is None:
    raise ValueError("LLM returned empty or filtered response")
openai_description = response.choices[0].message.content or ""
```

## Why

The OpenAI API (including vision/multimodal endpoints) can return `choices=[]` or `choices[0].message=None` at HTTP 200 when content is filtered or the provider short-circuits the pipeline. Without the guard, the `IndexError`/`AttributeError` is caught by `except openai.OpenAIError` — but these are not `OpenAIError` subclasses, so they propagate uncaught and crash the datachain pipeline.

## Evidence

- [Live production crash](https://github.com/HKUDS/LightRAG/issues/2551) — same pattern confirmed in production
- Gemini 2.5 Flash returns `choices[0].message = None` on `PROHIBITED_CONTENT` with HTTP 200 (verified)
- Found by [pact](https://github.com/qizwiz/pact) — 274 instances of this pattern across 759 real repos

## Scope

1 file, 2 lines added.

### DIFF
--- a/examples/multimodal/openai_image_desc_lib.py
+++ b/examples/multimodal/openai_image_desc_lib.py
@@ -37,6 +37,8 @@ def describe_image(file: File, client: openai.OpenAI) -> tuple[str, str]:
             ],
         )
 
+        if not response.choices or response.choices[0].message is None:
+            raise ValueError("LLM returned empty or filtered response")
         openai_description = response.choices[0].message.content or ""
         error = ""
 


shcheklein:
It seems to be AI generated(?) and probably is not much needed for this particular example. And if it is good to have we should be handling in a different way (catch and return default instead of raising).

qizwiz:
Thanks for the feedback, @shcheklein. Completely fair — for example code the right pattern is resilient degradation, not a raise.

The underlying issue is real (`choices=[]` at HTTP 200 does happen on content filtering, as does `choices[0].message = None`), but a raise is too harsh for an examples library. If you'd consider a re-submission, the intent would be:

```python
if not response.choices or response.choices[0].message is None:
    openai_description = ""
else:
    openai_description = response.choices[0].message.content or ""
```

Or more concisely:

```python
msg = response.choices[0].message if response.choices else None
openai_description = (msg.content or "") if msg else ""
```

Happy to file that as a separate PR if it's welcome, or leave it as-is if the example doesn't need the guard at all. Understood either way.

PR #1779, opened 2026-05-25
Skill - implement CASE methodology

### LINKED ISSUE #1771: Skill - generalize the intermediate dataset creation
### Description

**First**, The skill is optimizes creation of intermediate results. However, often it stores not useful intermediate results like embeddings that were created after some filtering - not useful outside of the context of the problem.

The skill should be optimized for general purpose intermediate results not any resutls.

**Second**, while created intermediate results it better to create them over a separate script. Now the pattern is - a single script that produces 3 intermediate results.

dmpetrov:
It implements CASE - see in the second-brain repo.
Closes #1771 as a side effect


src/datachain/skill/core/SKILL.md
```diff
@@ -45,7 +45,20 @@ Only go to raw storage (`read_storage`, `read_csv`, etc.) when no existing datas
 
 Datasets are the unit of reasoning. Chains that transform data through UDFs — or that produce a pipeline's final result — should be saved as named datasets. This creates a lineage graph where each node is reusable, inspectable, and referenceable by future pipelines and users.
 
-**Core rule: always `.save()`, never just `.show()`.** A pipeline's terminal operation should be `.save("descriptive_name")`, followed by `.show()` on the saved result for display. The only exception is one-off exploratory queries where the user explicitly asks to just "show me" or "print" without saving.
+**Core rule: always `.save()`, never just `.show()`, and NEVER write results to a local file.** A pipeline's terminal operation should be `.save("descriptive_name")`, followed by `.show()` on the saved result for display. Two exceptions: (1) one-off exploratory queries where the user explicitly asks to just "show me" or "print"; (2) Experiment-layer outputs (rankings, filters over Sense, custom analytics) per CASE per-layer reuse — persist by exception, not by default. The always-save rule remains absolute for C/A/S layers (Container, Asset, Sense) regardless of phrasing.
```
amritghimire: > and NEVER write results to a local file.**

Not sure I understand this, I mean can't user write to local file when needed using .to_storage?

amritghimire: Sorry if I am missing context here.

amritghimire: > per CASE per-layer reuse 

Do we define this prior to this?

dmpetrov: good points. fixing both in next commit.

tests/unit/test_skill_knowledge_scripts.py
```diff
@@ -482,12 +483,13 @@ def test_render_index_all_metadata_from_md(tmp_path, monkeypatch):
     (ds_dir / "my_ds.md").write_text(
         "---\nname: my_ds\nlast_version: 3.0.0\n"
         "records: 12000\nupdated: 2025-04-01T10:00:00Z\n"
-        "known_versions: [1.0.0, 2.0.0, 3.0.0]\n---\n\n"
+        "known_versions: [1.0.0, 2.0.0, 3.0.0]\n"
+        "case_layer: experiment\n"
+        "case_scope: file\n"
+        "case_source: raw_images\n"
```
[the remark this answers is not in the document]
dmpetrov: fixed

src/datachain/skill/core/SKILL.md
dmpetrov: fixed

dreadatour (approved):

PR #1783, opened 2026-05-26
Added check for anon on File read methods

### LINKED ISSUE #1778: anon does not propagate across .save() boundaries
### Description

#1712 / #1763 made read_storage() auto-detect anonymous access for public buckets, so users no longer need to type `anon=True` for the listing call. But the anon decision is purely session-scoped: it is not stored on the saved dataset, not stored on the listing dataset, not in any per-URI registry.

The moment a downstream stage runs `dc.read_dataset(...)` in a fresh Python process and a UDF calls `file.open()` / `file.read()`, the new process makes a fresh S3 HeadObject without anon and gets PermissionError: Forbidden. The original UX complaint of #1712 — "users / agents constantly make this mistake which slows down DC code generation" — reappears at every stage boundary of a multi-script pipeline.

ilongin:
Closes #1778.

When a `File` read (`open_object` / `get_file_info`) hits `PermissionError` on a cloud client that has credentials but no access to the target bucket, automatically retry with `anon=True` and cache the result per `(protocol, bucket)` so subsequent calls go straight to anonymous mode.

- Adds `_anon_fallback` decorator and `_ANON_BUCKETS` class-level cache on `Client`
- `Client.fs` honors the cache so cached buckets are created anonymously from the start
- Decorates `open_object` and `get_file_info` on the base `Client` and on `GCSClient` (which overrides both)
- Only marks the bucket as anon-needed after the retry actually succeeds; explicit `anon=True` skips the retry

### DIFF
--- a/src/datachain/client/fsspec.py
+++ b/src/datachain/client/fsspec.py
@@ -176,7 +176,7 @@ class Client(ABC):
     # True  = anon retry succeeded, use anon from the start.
     # False = anon retry also failed, don't bother retrying again.
     # Missing key = no decision yet.
-    _ANON_BUCKETS: ClassVar[dict[tuple[str, str], bool]] = {}
+    _anon_buckets: ClassVar[dict[tuple[str, str], bool]] = {}
 
     @classmethod
     def has_explicit_credentials(cls, client_config: dict | None) -> bool:
@@ -187,11 +187,11 @@ class Client(ABC):
 
     @classmethod
     def _bucket_needs_anon(cls, name: str) -> bool | None:
-        return cls._ANON_BUCKETS.get((cls.protocol, name))
+        return cls._anon_buckets.get((cls.protocol, name))
 
     @classmethod
     def _mark_bucket_anon(cls, name: str, anon: bool) -> None:
-        cls._ANON_BUCKETS[(cls.protocol, name)] = anon
+        cls._anon_buckets[(cls.protocol, name)] = anon
 
     def __init__(self, name: str, fs_kwargs: dict[str, Any], cache: Cache) -> None:
         self.name = name
--- a/src/datachain/client/fsspec.py
+++ b/src/datachain/client/fsspec.py
@@ -63,36 +63,87 @@ class BucketStatus(NamedTuple):
     error: str | None = None
 
 
-def _anon_fallback(method):
-    """Retry a Client method once with anonymous access on PermissionError.
+class AnonFallbackFS:
+    """Proxy for an fsspec ``AbstractFileSystem`` that retries any method on
+    ``PermissionError`` with ``anon=True``, and caches the decision per
+    ``(protocol, bucket)`` so future calls go straight to the right mode.
 
-    Only marks the bucket as anon-needed if the retry actually succeeds, so
-    genuinely inaccessible buckets keep raising clean errors instead of
-    being silently cached as anon.
+    Wrapping at this layer (rather than on individual ``Client`` methods)
+    means every code path that goes through ``self.fs`` is covered - both
+    sync paths (``get_file_info``, ``open_object``) and async ones
+    (``get_current_etag``, ``get_file``, ``get_size``).
     """
 
-    @functools.wraps(method)
-    def wrapper(self, *args, **kwargs):
-        try:
-            return method(self, *args, **kwargs)
-        except PermissionError:
-            if (
-                self.fs_kwargs.get("anon")
-                or self._bucket_needs_anon(self.name) is not None
-            ):
-                raise
-            saved_fs = self._fs
-            self._fs = type(self).create_fs(**{**self.fs_kwargs, "anon": True})
-            try:
-                result = method(self, *args, **kwargs)
-            except PermissionError:
-                self._fs = saved_fs
-                self._mark_bucket_anon(self.name, False)
-                raise
-            self._mark_bucket_anon(self.name, True)
-            return result
+    def __init__(self, client: "Client") -> None:
+        self._client = client
+        kwargs = dict(client.fs_kwargs)
+        if client._bucket_needs_anon(client.name) is True:
+            kwargs["anon"] = True
+        self._inner_fs = type(client).create_fs(**kwargs)
 
-    return wrapper
+    @property  # type: ignore[misc]
+    def __class__(self):
+        # Pretend to be the underlying fsspec class so isinstance checks
+        # (e.g. pyarrow's ``isinstance(fs, AbstractFileSystem)``) pass.
+        return type(self._inner_fs)
+
+    def __getattr__(self, name: str):
+        attr = getattr(self._inner_fs, name)
+        if not callable(attr):
+            return attr
+        if asyncio.iscoroutinefunction(attr):
+            return self._wrap_async(name, attr)
+        return self._wrap_sync(name, attr)
+
+    def _can_retry(self) -> bool:
+        if self._client.fs_kwargs.get("anon"):
+            return False
+        return self._client._bucket_needs_anon(self._client.name) is None
+
+    def _swap_to_anon(self) -> None:
+        self._inner_fs = type(self._client).create_fs(
+            **{**self._client.fs_kwargs, "anon": True}
+        )
+
+    def _wrap_sync(self, name, attr):
+        def call(*args, **kwargs):
+            try:
+                return attr(*args, **kwargs)
+            except PermissionError:
+                if not self._can_retry():
+                    raise
+                saved = self._inner_fs
+                self._swap_to_anon()
+                try:
+                    result = getattr(self._inner_fs, name)(*args, **kwargs)
+                except PermissionError:
+                    self._inner_fs = saved
+                    self._client._mark_bucket_anon(self._client.name, False)
+                    raise
+                self._client._mark_bucket_anon(self._client.name, True)
+                return result
+
+        return call
+
+    def _wrap_async(self, name, attr):
+        async def call(*args, **kwargs):
+            try:
+                return await attr(*args, **kwargs)
+            except PermissionError:
+                if not self._can_retry():
+                    raise
+                saved = self._inner_fs
+                self._swap_to_anon()
+                try:
+                    result = await getattr(self._inner_fs, name)(*args, **kwargs)
+                except PermissionError:
+                    self._inner_fs = saved
+                    self._client._mark_bucket_anon(self._client.name, False)
+                    raise
+                self._client._mark_bucket_anon(self._client.name, True)
+                return result
+
+        return call
 
 
 class Client(ABC):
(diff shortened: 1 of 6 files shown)


tests/unit/test_client.py
```diff
@@ -31,3 +34,119 @@ def test_parse_file_path_ends_with_slash(cloud_type):
     uri, rel_part = Client.parse_url("./animals/".replace("/", os.sep))
     assert uri == (Path().absolute() / Path("animals")).as_uri()
     assert rel_part == ""
+
+
+# Anonymous-access fallback (auto-retry on PermissionError) ---------------
```
shcheklein: we don't do subsections like this in other places (review and cleanup style everywhere)

src/datachain/client/fsspec.py
```diff
@@ -63,13 +63,45 @@ class BucketStatus(NamedTuple):
...
+        except PermissionError:
+            if self.fs_kwargs.get("anon") or self._bucket_needs_anon(self.name):
+                raise
+            saved_fs = self._fs
+            self._fs = type(self).create_fs(**{**self.fs_kwargs, "anon": True})
+            try:
+                result = method(self, *args, **kwargs)
+            except PermissionError:
+                self._fs = saved_fs
+                raise
```
shcheklein: in this case we should also save it as non-anon and don't try anon next time we hit this error

src/datachain/client/fsspec.py
```diff
@@ -63,13 +63,45 @@ class BucketStatus(NamedTuple):
...
     MAX_THREADS = multiprocessing.cpu_count()
     FS_CLASS: ClassVar[type["AbstractFileSystem"]]
     PREFIX: ClassVar[str]
     protocol: ClassVar[str]
     # client_config keys this backend treats as credentials.
     CREDENTIAL_KEYS: ClassVar[frozenset[str]] = frozenset()
+    # Process-local cache of (protocol, bucket) pairs that have been
+    # resolved as needing anonymous access. Populated only after an anon
+    # retry actually succeeds.
+    _ANON_BUCKETS: ClassVar[set[tuple[str, str]]] = set()
```
shcheklein: why is it in upper case? it is not a constant

src/datachain/client/fsspec.py
```diff
@@ -78,6 +110,14 @@ def has_explicit_credentials(cls, client_config: dict | None) -> bool:
             return False
         return any(k in client_config for k in cls.CREDENTIAL_KEYS)
 
+    @classmethod
+    def _bucket_needs_anon(cls, name: str) -> bool:
+        return (cls.protocol, name) in cls._ANON_BUCKETS
```
shcheklein: can there be a distinction by prefix - a prefix inside a bucket allows anon access, other prefix doesn't?

src/datachain/client/fsspec.py
```diff
@@ -251,6 +294,7 @@ async def get_current_etag(self, file: "File") -> str:
         info = await self.fs._info(full_path, **self._file_info_kwargs(file.version))
         return self.info_to_file(info, file.path).etag
 
+    @_anon_fallback
```
shcheklein: let's instead pretty much wrap some low level fsspec request method? so that we can catch all of them at once, w/o try to catch up and maintain these decorators everywhere

tests/unit/test_client.py
```diff
@@ -31,3 +34,119 @@ def test_parse_file_path_ends_with_slash(cloud_type):
...
+    Client._ANON_BUCKETS.clear()
+    yield
+    Client._ANON_BUCKETS.clear()
+
+
+def _gcs_client(bucket: str = "foo", **fs_kwargs) -> GCSClient:
+    return GCSClient(bucket, fs_kwargs, MagicMock())
+
+
+def test_anon_fallback_no_error_no_retry(monkeypatch, _clear_anon_cache):
```
shcheklein: use regular existing FS mocks, not monkeypatch

src/datachain/client/fsspec.py
```diff
@@ -63,13 +63,105 @@ class BucketStatus(NamedTuple):
     error: str | None = None
 
 
+class AnonFallbackFS:
```
shcheklein, line 66: ah, way to complicated for the task we are trying to solve :( really don't like the complexity and amount of code just to deal with anon that is needed only for demos / examples :(

(don't have a better solution in mind though)

src/datachain/client/fsspec.py
```diff
@@ -63,13 +63,105 @@ class BucketStatus(NamedTuple):
...
     CREDENTIAL_KEYS: ClassVar[frozenset[str]] = frozenset()
+    # Whether ``anon=True`` is a meaningful fsspec kwarg for this backend.
+    # Subclasses for S3/GCS/Azure flip this to True so ``Client.fs`` returns
+    # an ``AnonFallbackFS`` proxy.
+    _ANON_FALLBACK: ClassVar[bool] = False
+    # Process-local cache of (protocol, bucket) anon decisions.
+    # True  = anon retry succeeded, use anon from the start.
+    # False = anon retry also failed, don't bother retrying again.
+    # Missing key = no decision yet.
+    _ANON_BUCKETS: ClassVar[dict[tuple[str, str], bool]] = {}
```
shcheklein: this is not a constant

I think _ANON_FALLBACK is also not a constant

tests/unit/test_client.py
```diff
@@ -31,3 +34,126 @@ def test_parse_file_path_ends_with_slash(cloud_type):
...
+        return_value={
+            "name": "gs://foo/x.txt",
+            "size": 1,
+            "etag": "e",
+            "updated": "2024-01-01T00:00:00Z",
+        }
+    )
+
+
+def test_anon_fallback_no_error_no_retry(monkeypatch, _clear_anon_cache):
```
shcheklein, line 61: we want to use real FSs (existing mocks_, not monkey patches

src/datachain/client/fsspec.py
ilongin: True, this can be edge case. 
The problem is that implementing prefix based cache is much more complex than current one and has some open questions. Issues:
  - need to extract the path from every fsspec method call (different methods take path in different positions
  - need to pick a prefix granularity (per-segment? whole dir? configurable?) - any choice is a heuristic but maybe we can take latest directory
  - need longest-prefix matching on lookup
  - need two fs instances alive at once (one auth, one anon) or pay the cost of rebuilding on each call ~100-150 lines of code + new tests for a scenario most users don't hit (mixed-access bucket with scoped creds)

My suggestion is to just don't use cache when creds are explicitlty set and that's it.

src/datachain/client/fsspec.py
ilongin: I'm also not sure what is the less complex approach. I started adding decorator on Client methods, honestly that maybe seem less complex than this "generic" approach where more code is needed....

src/datachain/client/fsspec.py
ilongin: Renamed `_ANON_BUCKETS` to `_anon_buckets` - it's a runtime-mutated dict so lowercase makes sense.

Kept `_ANON_FALLBACK` uppercase - it's set once per subclass at class definition and never reassigned, same pattern as `CREDENTIAL_KEYS` / `FS_CLASS` / `PREFIX` / `protocol`..

tests/unit/test_client.py
ilongin: The tests use MagicMock because they need to control exactly when PermissionError is raised (first call fails, retry succeeds) and assert `create_fs` was called with `anon=True`. The cloud_server fixtures don't easily simulate per-call PermissionError

Added a separate integration test `test_anon_fallback_proxy_integration` in `tests/func/test_storage_auto_anon.py` that uses the real `cloud_server` fixture - proves the proxy correctly delegates ops to a real cloud emulator end-to-end. The actual anon-retry path stays in unit tests.

src/datachain/client/fsspec.py
```diff
@@ -63,13 +63,105 @@ class BucketStatus(NamedTuple):
...
+                saved = self._inner_fs
+                self._swap_to_anon()
+                try:
+                    result = getattr(self._inner_fs, name)(*args, **kwargs)
+                except PermissionError:
+                    self._inner_fs = saved
+                    self._client._mark_bucket_anon(self._client.name, False)
+                    raise
+                self._client._mark_bucket_anon(self._client.name, True)
+                return result
```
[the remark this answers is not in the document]
ilongin: fixed

src/datachain/client/fsspec.py
```diff
@@ -63,13 +63,105 @@ class BucketStatus(NamedTuple):
...
+                try:
+                    result = await getattr(self._inner_fs, name)(*args, **kwargs)
+                except PermissionError:
+                    self._inner_fs = saved
+                    self._client._mark_bucket_anon(self._client.name, False)
+                    raise
+                self._client._mark_bucket_anon(self._client.name, True)
+                return result
+
+        return call
```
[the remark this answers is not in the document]
ilongin: fixed

src/datachain/client/fsspec.py
```diff
@@ -63,13 +63,160 @@ class BucketStatus(NamedTuple):
...
+        ):
+            kwargs["anon"] = True
+        self._inner_fs = type(client).create_fs(**kwargs)
+
+    @property  # type: ignore[misc]
+    def __class__(self):
+        # Pretend to be the underlying fsspec class so isinstance checks
+        # (e.g. pyarrow's ``isinstance(fs, AbstractFileSystem)``) pass.
+        return type(self._inner_fs)
+
```
amritghimire, line 114: do we need this isinstance trick? feels hacky to pretend we're the underlying fs class

amritghimire: Also, can we just return class name instead of the type itself?

src/datachain/client/fsspec.py
```diff
@@ -63,13 +63,160 @@ class BucketStatus(NamedTuple):
...
+                    return result
+                except PermissionError:
+                    if self._should_use_anon_cache():
+                        self._client._mark_bucket_anon(self._client.name, False)
+                    raise
+                finally:
+                    if not keep_anon:
+                        self._inner_fs = saved
+
+        return call
```
amritghimire, line 201: Can we abstract some logic? Most part looks similar in both sync and asynchronous one?

src/datachain/client/fsspec.py
ilongin: The trick is needed because external code (pyarrow, fsspec internals) does `isinstance(fs, AbstractFileSystem)` and we can't patch those checks. Subclassing isn't clean either - the concrete class is dynamic per-client (S3FileSystem / GCSFileSystem / ...).

On returning the class name instead of the type - `__class__` has to be a class object; returning a string would break `isinstance` entirely.

src/datachain/client/fsspec.py
ilongin: The duplication is structural (Python won't let `await` be conditional, so sync and async need separate function bodies). Extracting helpers just moves the duplication into helpers without removing it, and the wrappers become indirection that hurts readability. I'd rather keep both inline and add a comment noting the deliberate parallel.

shcheklein (commented):
I think this is a bit too much for the task tbh.

What are the alternatives for this?

PR #1787, opened 2026-05-27
Add team scope and expiry argument with studio login

amritghimire:
Companion to [https://github.com/datachain-ai/studio/pull/12918 ](https://github.com/datachain-ai/studio/pull/12918)and https://github.com/treeverse/dvc-studio-client/pull/240.

Adds team scoping and expiration support to datachain auth login. 

```
# scope to specific teams
datachain auth login --team ml-team --team data-team --expires-in 90
```
Default expiration is 365 days if you don't pass anything in --expires-in.

If you log in with a single --team, it automatically sets that as your default team so you don't have to run datachain auth team afterward.

Compatibility verified manually: This DataChain version works with both old and new Studio backends. Old DataChain also continues working fine with both Studio versions , no breaking changes on either side.

Upgraded dvc studio client too


shcheklein:
@amritghimire check tests, covecov, etc

src/datachain/remote/studio.py
```diff
@@ -82,20 +83,128 @@ class StudioClient:
...
+        # Save migrated config (one-time operation)
+        with Config(ConfigLevel.GLOBAL).edit() as conf:
+            studio_conf = conf.get("studio", {})
+            studio_conf["token"] = migrated_config
+            conf["studio"] = studio_conf
+
+        return migrated_config
+
+    def _validate_token_config(self, token_config: dict) -> None:
+        """Validate token expiration."""
```
shcheklein: comment is not aligned with method name

also not informative - cleanup all these AI stuff, align and make it meaningful or remove

src/datachain/remote/studio.py
```diff
@@ -82,20 +83,128 @@ class StudioClient:
...
+    def _is_token_cache_valid(self) -> bool:
+        """Check if cached token is still valid (within 1 minute)."""
+        if not self._token_last_validated:
+            return False
+
+        cache_age = datetime.now(timezone.utc) - self._token_last_validated
+        return cache_age.total_seconds() < 60
+
+    def _get_validated_token(self) -> str:
+        """Get token from config with validation (migration at config load)."""
```
shcheklein: > migration at config load

I don't understand this

shcheklein:
> The token format in config changes from a plain string to a structured dict with value, teams, expires_at, created_at, never_expires. Existing string tokens get migrated automatically on first access — no manual steps needed.

do we really want / need to complicate this? we don't cached any other rules (and nobody does that AFAIK), I'm not sure people cache expiration time either (e.g. bc check is not reliable -depends on local clock probably which is not aligned with the server). I'm not sure what is the need to complicate this tbh

amritghimire:
> > The token format in config changes from a plain string to a structured dict with value, teams, expires_at, created_at, never_expires. Existing string tokens get migrated automatically on first access — no manual steps needed.
> 
> do we really want / need to complicate this? we don't cached any other rules (and nobody does that AFAIK), I'm not sure people cache expiration time either (e.g. bc check is not reliable -depends on local clock probably which is not aligned with the server). I'm not sure what is the need to complicate this tbh

I feel like it will be easier to show that it is expired and ask to login without connection to backend. We can warn before it expires with this too. But if you feel strongly on this we can revert this

shcheklein:
Ci is still failing?

src/datachain/cli/parser/studio.py
```diff
@@ -70,6 +70,33 @@ def add_auth_parser(subparsers, parent_parser) -> None:
...
+
+    # Expiration arguments
+    login_parser.add_argument(
+        "--expires-in",
+        type=int,
+        default=None,
+        help="Token expiration in days (default: 365)",
+    )
+    login_parser.add_argument(
+        "--never-expires",
```
shcheklein: can we use 0 for --expires-in ?

or consider this default ... it is unusual to have an explicit never-expires

src/datachain/studio.py
```diff
@@ -23,6 +23,7 @@
 
 logger = logging.getLogger("datachain")
 
+
```
shcheklein: needed?

src/datachain/studio.py
```diff
@@ -168,6 +174,25 @@ def login(args: "Namespace"):
     hostname = args.hostname or get_studio_url(config)
     scopes = args.scopes
 
+    # Parse team arguments
+    team_names = args.team
+    all_teams = args.all_teams
+
+    if all_teams and team_names:
+        raise DataChainError("Cannot specify both --team and --all-teams")
```
shcheklein: same Q: do we need both - team and all-teams? can we consider one of them as default?

amritghimire:
> Ci is still failing?

flaky test, I am resolving it

src/datachain/studio.py
dreadatour: My 2 cents: I think we need both. All teams is a bit dangerous and needs to be explicit.

shcheklein: I think we are punishing a bit most of the users (single team) (should we for example check how many team user has access to and ask? or ask to provide --team)

src/datachain/cli/parser/studio.py
amritghimire: I mean using 0 feels ambiguous to me. We have 365 days as default, if user don't specify never expires. I feel the explicit --never-expires flag is actually good UX design because, it has clear intent and is not ambiguous to me.

shcheklein: okay, let's not support never expires at all? then can specify a few years if they want? (and token type allows this)

too many options / arguments / knobs, etc

check also other CLIs, GitLab token management, GitHub

src/datachain/studio.py
amritghimire: I mean we can leave it to ui too, if all teams is not specified, so user can select the teams they want. But, making all teams default also works for me.

amritghimire: I will leave it to product decision to be honest, lets keep all teams default for minimal friction? WDYT? cc. @shcheklein

shcheklein: Can we do something like this:

- if user has a single team - we just create a token for that team
- if user has multiple we require explicit `--teams` (one or many)
- drop completely `--all-teams` - this is too wide open and dangerous (you need to go to UI and add more teams if needed)

amritghimire: I think that its a friction for user to remember team name and add it, when they add new team name they need to do reauth or modify and so on. IMO, default should be easiest option all teams while allow user to add restrictions if they want. All teams is what we have now. I think it is how gh uses too. I don't remember having to set any organization with github to access it.

shcheklein: so 99% won't have to remember it, right? we don't require it if there is a single team

if there are multiple - we can show a message with team names they have access to as part of the error message

all teams is dangerous since it silently gives also access to the new teams

> GitHub

there are different types of tokens - and it also depends on the teams settings. But it does show you which teams it will have access to, and you can even specify which repos I think.

Not sure about GH CLI though  - need to test from scratch I guess.

amritghimire: I meant about gh cli. My point is lets say user created a team, logged in with that team as it is automatically assigned as mentioned above. The problem I see is when user created the new team, user won't know that they don't have access to new team now. I think this implicit addition is confusing for user.

amritghimire: Asking them to have either team name or all teams, will make them aware the token is authorized to the specific team, not magically under the hood imo.

shcheklein: yes, but first you are punishing people with a single team this way (99% of them?), also all teams is dangerous by definition - if I join a new team it becomes automatically exposed - I don't think we need it in this form.

shcheklein: >  The problem I see is when user created the new team, user won't know that they don't have access to new team now.

it all can / should be solved on the error messages side, no?

ilongin: > Can we do something like this:
> 
> * if user has a single team - we just create a token for that team
> * if user has multiple we require explicit `--teams` (one or many)
> * drop completely `--all-teams` - this is too wide open and dangerous (you need to go to UI and add more teams if needed)

I agree with this. I have a feeling non of our users will have more than one team so by default user should not be forced to enter `--team` ... also agree that `--all-teams` is not needed at least for now

src/datachain/studio.py
```diff
@@ -181,17 +202,37 @@ def login(args: "Namespace"):
             token_name=name,
             hostname=hostname,
             scopes=scopes,
+            team_names=team_names,
+            expires_in_days=expires_in_days,
+            never_expires=False,
```
ilongin: do we need this `never_expires` flag when it's hardcoded to False anyway?

src/datachain/studio.py
```diff
@@ -168,6 +174,21 @@ def login(args: "Namespace"):
...
+
+    if all_teams and team_names:
+        raise DataChainError("Cannot specify both --team and --all-teams")
+    if not all_teams and not team_names:
+        raise DataChainError("Must specify either --team or --all-teams")
+
+    expires_in_days = args.expires_in
+
+    # Set default expiration if not specified
+    if expires_in_days is None:
```
ilongin, line 182: Should we add just a simple validation for this field, at least to be positive num?

src/datachain/studio.py
```diff
@@ -181,17 +202,37 @@ def login(args: "Namespace"):
...
             open_browser=open_browser,
             client_name="DataChain",
             post_login_message=POST_LOGIN_MESSAGE,
+            all_teams=all_teams,
         )
     except StudioAuthError as exc:
         raise DataChainError(f"Failed to authenticate with Studio: {exc}") from exc
+    except requests.HTTPError as exc:
+        response = exc.response
+        if response.status_code == 400:
```
ilongin: I think there can be cases where response is None

src/datachain/studio.py
amritghimire: One complication that we get on implementing this is Chicken-and-egg problem. To determine how many teams a user has access to, we need to authenticate first, but we're trying to decide team scope during the  authentication process. I see this option on how to achieve this, 

Leave all_teams to true first as default and ask user to select the teams if all teams from UI at this stage.
image
Because this is the first point where we know user is authenticated. From CLI, we get a token: 
```
Opening link for login at https://studio.datachain.ai/auth/device-login?code=XXXXXXX

Once you've logged in, return here and you'll be ready to start using DataChain with Studio.
```

So, we won't know how many team user have or anything at only this point. 


Your  concern about --all-teams being "dangerous" is valid. When users join new teams, tokens automatically get access without explicit consent. However, this matches GitHub CLI behavior and provides better UX for most users.

```
  Compromise suggestion: Make --all-teams require explicit confirmation:
  datachain auth login --all-teams  # Prompts: "Grant access to all current and future teams? [y/N]"
```

shcheklein: We can probably first do "no teams" token (that can only be used to get the list of teams, or we can return the list as part of the auth process). Then ask or assign that single team at the end, no?

amritghimire: We can do this, if team names is passed, pass it. Else, don't pass all teams at all. At the point in the UI above, we do know that whether all teams is passed or not, as well as passed team names.,

 So, we can make it required for user to select team there in the UI above, select the first team if one team by default and all. 
So, we can handle it from that point. WDYT? We can ask or assign there and this will be backward compatible as well since it is on our control point.

shcheklein: Yes, I like it. I think that's also how GH does in certain cases (when you install app and it is asking for permissions, etc). Probably we should skip it there if it is a single team.

amritghimire: Removed all teams from here and will ask in Studio in separate PR

src/datachain/studio.py
amritghimire: We have server side for such

shcheklein (approved):
A thing to double check (ask AI) - docs, examples, and skill - do we need any updates there?

dreadatour (approved):

PR #1788, opened 2026-05-28
Small tweak to skill prompt about JSON input

dreadatour:
Required for SaaS — rephrase a little bit — no "JSON file" there, but "JSON input". Should works for both CLI and SaaS.


dreadatour:
Closing this since it was already resolved in main 👀

PR #1789, opened 2026-05-29
Fix numpy JSON serialization at storage boundaries

shcheklein:
## Summary
- add opt-in numpy serialization to `datachain.json` without importing numpy during normal `datachain` import
- use the opt-in only where numpy values can cross JSON storage boundaries: SQLite adapters, warehouse JSON conversion, and job metrics
- add regression coverage for nested numpy arrays/scalars in JSON, SQLite adapters, warehouse conversion, and metrics

## Tests
- `tests/unit/test_json.py`, `tests/unit/sql/sqlite/test_types.py`, `tests/func/test_data_storage.py`, `tests/func/test_metastore.py`: 120 passed
- `git diff --check`

### DIFF
--- a/src/datachain/data_storage/metastore.py
+++ b/src/datachain/data_storage/metastore.py
@@ -2352,7 +2352,7 @@ class AbstractDBMetastore(AbstractMetastore):
         if finished_at is not None:
             values["finished_at"] = finished_at
         if metrics:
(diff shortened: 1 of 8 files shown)


0x2b3bfa0 (approved):
<sub>I guess there's a reason why `_build_default` exists, instead of keyword arguments with default values in function signatures. Not obvious at 01:07 UTC, though.</sub>

PR #1791, opened 2026-05-29
Optional objects

### LINKED ISSUE #1055: UDF with Optional[File] return type fails
```
from typing import Optional
from dotenv import load_dotenv
from datachain import read_storage, C, File


load_dotenv(".env.test")

def process_events(file: File) -> Optional[File]:
    return file


(
    read_storage("s3://bucket/v1")
    .settings(cache=True)
    .limit(1)
    .map(events_file=process_events)
    .show()
)
```

fails in UDF prep logic.

dmpetrov:
closes #1055

First-class `Optional[X]` with identical behavior on SQLite and ClickHouse.

It works across read/map/filter/order_by/group_by/select/mutate/aggregates/union/json/parquet - almost every function required some adjustments 🤯 Majority of the code just covers the differences between SQLite and CH. Clickhouse non-nullable is PIA 🥵 

**How:**

1. `Optional[DataModel]` uses a hidden `_is_null` sentinel + we convert leaves `T` to `Nullable(T)` on CH (tried another approach - it was much more complicated). User never sees this and cannot access, to_parquet and other - do not expose this.
2. `Optional[basic]` (int, str, etc) stores real NULL. CH: `T` to `Nullable(T)` as usual.
3. `x: T = None` is auto-promoted to `Optional[T]`.

**Fixes** of the issues we had before but didn't pay attention: 

- reading nullable value from parquet/CSV/JSON/read_values use to led to type-default (NULL in SQLite and `\0` in CH). Now it's correctly becomes NULL.

**DB difference issues** that's better to keep as is without introducing extra complexity:
- `Optional[float]` is handled different in CH and SQLite since SQLite can't tell NaN from NULL
- `Optional[list]/[dict]` is handled different because None list/dict is not unsupported on CH (use `[]`/`{}`)
- window `.over(order_by=<optional leaf>)` inconsistency: SQLite put NULL first, CH - last. Good thing - it's inconsistency happening in other DBs.

**Exceptions**:
- as expected, export commands (to_pandas/to_records/to_csv/to_json/parquet) could loose type information if optional object's fields are ALL nullable `x=Optional[DM(a=None, b=None)]`. So, what `NULL, NULL` means - `None` object or an object with `a=None` and `b=None`. However, an heuristic was implemented to properly reconstruct it if at least one is not None.
- nulls *inside* collections (`list[Optional[T]]`) error on CH (work on SQLite)
- `func.collect` inconsistency: `collect([1, None, 3])` returns `[1, None, 3]` in SQLIte and `[1, 3]` in CH. Hard to fix.

**Out of scope:**

- bare `Union[A,B]` without `None` is rejected as it was before

PS: initially i tried to avoid `T` to `Nullable(T)` conversion in CH. It led to a lot of complex logic: every read path then wrapped the leaf in CASE WHEN {parent}_is_null = 0 THEN col END to mask the type-default back to NULL, at 4 sites: filter, order_by, flat-export, and func._resolve_col. It was simplified a lot after introducing the conversion.


src/datachain/lib/udf.py
```diff
@@ -319,11 +319,22 @@ def run(
...
     def _flatten_row(self, row):
         if len(self.output.values) > 1 and not isinstance(row, BaseModel):
-            flat = []
-            for obj in row:
-                flat.extend(self._obj_to_list(obj))
+            # An Optional[DataModel] output needs flatten_value to emit its
+            # sentinel; other outputs (incl. auto-wrapper models typed as a leaf)
+            # flatten by object shape via _obj_to_list, not the declared type.
+            flat: list[Any] = []
+            for obj, anno in zip(row, self.output.values.values(), strict=False):
```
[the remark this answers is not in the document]
dmpetrov: `strict=False` is intentional here - this branch is shared with the arrow/parquet reader, where self.output.values has more entries than the per-row tuple by design. `strict=True` breaks unit test in `read_parquet` of `Optional[DataModel]`.

src/datachain/func/conditional.py
```diff
@@ -192,6 +223,9 @@ def isnone(col: str | ColumnExpr) -> Func:
     """
     Returns a function that checks if the column value is `None` (NULL in DB).
 
+    Works for an `Optional[DataModel]` column too — it checks the model's absence
```
shcheklein: minor: this comment is not very clear tbh (and probably not needed - it is kinda expected to work)

tests/unit/lib/test_optional.py
```diff
@@ -0,0 +1,374 @@
...
+def restore_model_store():
+    snapshot = copy.deepcopy(ModelStore.store)
+    ModelStore.store = {}
+    try:
+        yield
+    finally:
+        ModelStore.store = snapshot
+
+
+# -- unwrap_optional -------------------------------------------------------
```
shcheklein: minor: let's remove please all this type of breaks and splitters. We don't use them anywhere. LLMs try to innovate each time, but let's not use them - keep the same simple style

tests/unit/lib/test_optional.py
```diff
@@ -0,0 +1,374 @@
...
+    leaves = [
+        "__".join(p)
+        for p, _, has_subtree, _ in SignalSchema({"m": Optional[M]}).get_flat_tree()
+        if not has_subtree
+    ]
+    assert "m__is_null" in leaves  # the user's field
+    assert "m___is_null" in leaves  # the internal sentinel, no collision
+
+
+# ---------------------------------------------------------------------------
```
shcheklein: this might signal we need two files (but it is fine also to just drop these delimiters)

src/datachain/lib/data_model.py
```diff
@@ -105,13 +124,69 @@ def _fingerprint_tree(
...
+# could not tell NaN from None.
+NULLABLE_SCALARS: "tuple[type, ...]" = (int, str, bool, bytes, datetime)
+
+
+def promote_default_none(model: type[BaseModel]) -> None:
+    """Auto-promote non-Optional fields with `default=None` to `Optional[...]`.
+
+    `x: int = None` is treated as `x: Optional[int] = None` so the column is
+    genuinely nullable and None round-trips instead of coercing to the type
+    default on ClickHouse.
```
shcheklein: we should be mentioning CH on datachain core side - it should be some generic logic (why does in general CH do this? what is the underlying mechanics that we are fixing?)

shcheklein:
Q: which / how many models from providers we didn't handle and handle now with this? (Considering limitations like union)

src/datachain/data_storage/schema.py
```diff
@@ -94,11 +94,10 @@ def convert_rows_custom_column_types(
     for row in rows:
         row_list = list(row)
         for idx, t in custom_columns_types:
-            row_list[idx] = (
-                t.default_value(dialect)
-                if row_list[idx] is None
-                else t.on_read_convert(row_list[idx], dialect)
-            )
+            if row_list[idx] is None:
+                row_list[idx] = None if t.dc_nullable else t.default_value(dialect)
```
shcheklein, line 98: @copilot remind me with links and diagrams where this method (these changes belong to) is used and why

src/datachain/func/aggregate.py
```diff
@@ -1,20 +1,55 @@
...
+
+if TYPE_CHECKING:
+    from sqlalchemy import TableClause
+
 
 AggColT = str | Column | ColumnExpr | Func
 
 
+class _CountFunc(_SentinelAwareFunc):
+    """``count(<Optional[DataModel]>)`` counts rows where the parent is present.
```
shcheklein: @copilot so is this PR only about optional within a model? How about global non nested fields?

src/datachain/func/aggregate.py
```diff
@@ -1,20 +1,55 @@
...
+    def _sentinel_column(
+        self,
+        sentinel_path: str,
+        label: str | None,
+        table: "TableClause | None",
+    ) -> Column:
+        sentinel = Column(sentinel_path)
+        sentinel.table = table
+        # COALESCE keeps count()'s 0-for-empty contract: SUM over zero rows is NULL.
+        func_col = sa_func.coalesce(sa_func.sum(1 - sa_cast(sentinel, Integer)), 0)
```
shcheklein, line 42: @copilot what is the performance effect? Does it always now complicates counts and other sql or only for optional fields? Any other important implications considering what we do in data chain ?

src/datachain/func/aggregate.py
shcheklein: Give me links, examples @copilot

shcheklein: Why are we referring to data model here only @copilot

src/datachain/func/aggregate.py
```diff
@@ -73,6 +108,7 @@ def sum(col: AggColT) -> Func:
     Notes:
         - The `sum` function should be used on numeric columns or expressions.
         - The result column type will be inferred from the input expression type.
+        - Skips rows where the value's ``Optional[DataModel]`` parent is absent.
```
shcheklein, line 114: This quite hard to parse and understand

src/datachain/func/aggregate.py
shcheklein: @copilot how basic scalars nullable implemented in Studio considering CH doesn't support nullable fields

src/datachain/func/func.py
```diff
@@ -30,6 +31,24 @@
 ColT = Union[str, tuple, Column, ColumnExpr, "Func"]
 
 
+def _source_is_nullable(col: ColT, signals_schema: "SignalSchema | None") -> bool:
+    """True when ``col`` resolves to a column that may hold NULL — a leaf under an
+    ``Optional[DataModel]`` or an ``Optional[basic]`` column."""
+    if signals_schema is None or not isinstance(col, str):
```
shcheklein: How can it be None or not str? Cc @copilot

src/datachain/func/func.py
```diff
@@ -30,6 +31,24 @@
...
+        return False
+    db_col = ColumnMeta.to_db_name(col)
+    if signals_schema.optional_parent_sentinel(db_col) is not None:
+        return True
+    from datachain.lib.signal_schema import SignalResolvingError
```
shcheklein: Move it to the top?

src/datachain/func/func.py
```diff
@@ -30,6 +31,24 @@
...
+
+    try:
+        anno = signals_schema.get_column_type(db_col)
+    except SignalResolvingError:
```
shcheklein: When how we can get this exception? We should probably just allow to raise .. too much defensive logic

src/datachain/data_storage/schema.py
shcheklein: @copilot is it called only for data model fields? Thus customer?

src/datachain/func/aggregate.py
shcheklein: So does it mean we store also an additional column per each global non nested field to have a bool in it? @copilot

src/datachain/func/func.py
```diff
@@ -390,6 +409,22 @@ def get_col_name(self, label: str | None = None) -> str:
...
+        self,
+        signals_schema: "SignalSchema | None",
+        col_type: "DataType | None" = None,
+    ) -> bool:
+        """Whether this func's result column may hold NULL: a nullable scalar
+        result (float excluded) over a nullable source."""
+        if signals_schema is None:
+            return False
+        if col_type is None:
+            col_type = self.get_result_type(signals_schema)
```
shcheklein, line 436: @copilot do we handle case where func returns a tuple with a mix of fields some optional some not?

src/datachain/func/aggregate.py
shcheklein: @copilot handle previous comment

src/datachain/func/conditional.py
```diff
@@ -1,17 +1,48 @@
...
 CaseT = int | float | complex | bool | str | Func | ColumnExpr
 
 
+class _IsNoneFunc(_SentinelAwareFunc):
+    """``isnone`` is None-aware for ``Optional[DataModel]``: that column isn't a
+    real column on disk, so it reads the model's ``_is_null`` sentinel. Any other
+    column (including ``Optional[basic]``) keeps the plain ``col IS NULL`` check.
+    """
+
+    def is_nullable_result(self, signals_schema=None, col_type=None) -> bool:
```
shcheklein: Is there a default implementation for this?

src/datachain/func/func.py
shcheklein: @copilot are there realistic scenarios for that? Should we just raise?

src/datachain/lib/convert/flatten.py
```diff
@@ -38,8 +117,16 @@ def _flatten_fields_values(fields: dict, obj: BaseModel) -> Generator:
                 key: val.model_dump() if ModelStore.is_pydantic(type(val)) else val
                 for key, val in value.items()
             }
-        elif ModelStore.is_pydantic(anno):
-            yield from _flatten_fields_values(anno.model_fields, value)
+        elif kind.is_model:
+            if kind.is_optional:
+                if value is None:
+                    yield True
```
shcheklein: @copilot explain why yield twice

src/datachain/func/func.py
shcheklein: @copilot give me some real user level examples or some internal snippets with explanations

src/datachain/func/func.py
shcheklein: @copilot what will be the col type for a func that returns a tuple of different type? How and where will handle type by type for each of the columns there?

src/datachain/func/aggregate.py
dmpetrov: @copilot please answer

src/datachain/func/conditional.py
dmpetrov: Yes - `Func.is_nullable_result` in func/func.py (it marks a nullable-scalar result over a nullable source). 

_IsNoneFunc/_CountFunc override it to False since isnone/count never emit NULL. Aligned the override signatures with the base.

src/datachain/lib/data_model.py
dmpetrov: updated the comment.

underlying mechanics - CH x=None will return `0` (or empty string if str) which is not expected (and it's probably the only DB that does it -  not nullable by default, the other DBs are nullable by default). Instead, if there is a chance of having None - we force "promotion" of `int` to Optional[int]. Result - `x=None` is stored as `NULL` in DB and roundtrips to `None` instead of `0`.

dmpetrov:
> Q: which / how many models from providers we didn't handle and handle now with this? (Considering limitations like union)

Without Union - the same number I believe 🙂  Didn't even check since Union is a clear next step.

dmpetrov:
@shcheklein  all the comments should be addressed.

src/datachain/sql/types.py
```diff
@@ -157,15 +157,38 @@ class SQLType(TypeDecorator):
     impl: type[types.TypeEngine[Any]] = types.TypeEngine
     cache_ok = True
 
+    # Marks a column whose source annotation was an explicit Optional[scalar], so
+    # backends with non-nullable leaves emit a nullable type and None round-trips
+    # as NULL rather than the type default.
```
dreadatour: Comments is too verbose IMO.

src/datachain/sql/types.py
```diff
@@ -157,15 +157,38 @@ class SQLType(TypeDecorator):
...
+
     @property
     def python_type(self) -> StandardType:
         raise NotImplementedError
 
     def to_dict(self) -> dict[str, Any]:
-        return {"type": self.__class__.__name__}
+        d: dict[str, Any] = {"type": self.__class__.__name__}
+        # Persist the Optional[scalar] marker so a None field round-trips as NULL
+        # (not the type default) after the dataset schema is serialized and reloaded.
```
dreadatour: Also here, do we need this comment?

src/datachain/lib/udf.py
```diff
@@ -319,11 +319,22 @@ def run(
 
     def _flatten_row(self, row):
         if len(self.output.values) > 1 and not isinstance(row, BaseModel):
-            flat = []
-            for obj in row:
-                flat.extend(self._obj_to_list(obj))
+            # An Optional[DataModel] output needs flatten_value to emit its
+            # sentinel; other outputs (incl. auto-wrapper models typed as a leaf)
+            # flatten by object shape via _obj_to_list, not the declared type.
```
dreadatour: Same here, it is hard to read comments like this. Am I the only one who is struggling?

src/datachain/lib/signal_schema.py
```diff
@@ -583,14 +592,14 @@ def traverse(prefix, schema_info):
 
         return hidden_fields
 
-    def to_udf_spec(self) -> dict[str, type]:
+    def to_udf_spec(self) -> "dict[str, type[SQLType] | SQLType]":
```
dreadatour: `SQLType` is imported on a top level (not under `if TYPE_CHECKING`):

```suggestion
    def to_udf_spec(self) -> dict[str, type[SQLType] | SQLType]:
```

src/datachain/lib/signal_schema.py
```diff
@@ -908,6 +957,63 @@ def _find_in_tree(self, path: list[str]) -> DataType:
...
 
+    def _optional_sentinel_at(self, db_col: str) -> "str | None":
+        """Sentinel db-name for ``db_col`` when it is itself an
+        ``Optional[DataModel]`` node, else None."""
+        try:
+            anno = self.get_column_type(db_col, with_subtree=True)
+        except SignalResolvingError:
+            return None
+        if is_optional_model(anno):
+            return f"{db_col}{DEFAULT_DELIMITER}{self._OPTIONAL_SENTINEL_FIELD}"
```
dreadatour, line 989: I suppose it is by design, but I just wanted to mention:

`_OPTIONAL_SENTINEL_FIELD` already have `_` prefix: `_is_null`. `DEFAULT_DELIMITER` is `__` (double underscore). Result will be with three underscores: `f"{db_col}___is_null"`.

src/datachain/lib/signal_schema.py
```diff
@@ -908,6 +957,63 @@ def _find_in_tree(self, path: list[str]) -> DataType:
 
         return curr_type
 
+    def _optional_sentinel_at(self, db_col: str) -> "str | None":
```
dreadatour:
```suggestion
    def _optional_sentinel_at(self, db_col: str) -> str | None:
```

src/datachain/lib/signal_schema.py
```diff
@@ -817,19 +838,45 @@ def get_column_type(self, col_name: str, with_subtree: bool = False) -> DataType
                 return _type
         raise SignalResolvingError([col_name], "is not found")
 
+    def is_nullable_column(self, db_col: str, anno: "DataType") -> bool:
```
dreadatour:
```suggestion
    def is_nullable_column(self, db_col: str, anno: DataType) -> bool:
```

src/datachain/lib/signal_schema.py
```diff
@@ -817,19 +838,45 @@ def get_column_type(self, col_name: str, with_subtree: bool = False) -> DataType
...
+        """Whether a scalar leaf stores real NULL (``dc_nullable``) instead of the
+        type default: its base type is in ``NULLABLE_SCALARS`` and it is either an
+        ``Optional[scalar]`` or sits under an ``Optional[DataModel]`` ancestor."""
+        inner, is_optional = unwrap_optional(anno)
+        if inner not in NULLABLE_SCALARS:
+            return False
+        return is_optional or self.optional_parent_sentinel(db_col) is not None
+
+    def _db_leaf_sql_type(self, path: "list[str]", anno: "DataType") -> Any:
```
dreadatour:
```suggestion
    def _db_leaf_sql_type(self, path: list[str], anno: DataType) -> Any:
```

src/datachain/lib/signal_schema.py
```diff
@@ -817,19 +838,45 @@ def get_column_type(self, col_name: str, with_subtree: bool = False) -> DataType
                 return _type
         raise SignalResolvingError([col_name], "is not found")
 
+    def is_nullable_column(self, db_col: str, anno: "DataType") -> bool:
+        """Whether a scalar leaf stores real NULL (``dc_nullable``) instead of the
+        type default: its base type is in ``NULLABLE_SCALARS`` and it is either an
+        ``Optional[scalar]`` or sits under an ``Optional[DataModel]`` ancestor."""
+        inner, is_optional = unwrap_optional(anno)
+        if inner not in NULLABLE_SCALARS:
+            return False
```
dreadatour, line 863: I am not sure about the logic here, but is it too strict?

`NULLABLE_SCALARS` is `(int, str, bool, bytes, datetime)` tuple.

src/datachain/lib/signal_schema.py
```diff
@@ -908,6 +957,63 @@ def _find_in_tree(self, path: list[str]) -> DataType:
...
+        ``Optional[DataModel]`` node, else None."""
+        try:
+            anno = self.get_column_type(db_col, with_subtree=True)
+        except SignalResolvingError:
+            return None
+        if is_optional_model(anno):
+            return f"{db_col}{DEFAULT_DELIMITER}{self._OPTIONAL_SENTINEL_FIELD}"
+        return None
+
+    def optional_model_paths(self) -> "list[list[str]]":
```
dreadatour:
```suggestion
    def optional_model_paths(self) -> list[list[str]]:
```

src/datachain/lib/signal_schema.py
```diff
@@ -908,6 +957,63 @@ def _find_in_tree(self, path: list[str]) -> DataType:
...
+            list(path)
+            for path, type_, has_subtree, _ in self.get_flat_tree(
+                include_sentinels=False
+            )
+            if has_subtree and is_optional_model(type_)
+        ]
+        paths.sort(key=len, reverse=True)
+        return paths
+
+    def model_sentinel(self, db_col: str) -> "str | None":
```
dreadatour:
```suggestion
    def model_sentinel(self, db_col: str) -> str | None:
```

src/datachain/lib/signal_schema.py
```diff
@@ -908,6 +957,63 @@ def _find_in_tree(self, path: list[str]) -> DataType:
...
+        """DB name of the sentinel for ``db_col`` when it is itself an
+        ``Optional[DataModel]``, else None."""
+        return self._optional_sentinel_at(db_col)
+
+    def optional_parent_sentinel(self, db_col: str) -> "str | None":
```
dreadatour:
```suggestion
    def optional_parent_sentinel(self, db_col: str) -> str | None:
```

src/datachain/lib/signal_schema.py
```diff
@@ -908,6 +957,63 @@ def _find_in_tree(self, path: list[str]) -> DataType:
...
+            )
+            if has_subtree and is_optional_model(type_)
+        ]
+        paths.sort(key=len, reverse=True)
+        return paths
+
+    def model_sentinel(self, db_col: str) -> "str | None":
+        """DB name of the sentinel for ``db_col`` when it is itself an
+        ``Optional[DataModel]``, else None."""
+        return self._optional_sentinel_at(db_col)
```
dreadatour: Do we need this method? May be move `_optional_sentinel_at` here?

src/datachain/lib/signal_schema.py
```diff
@@ -908,6 +957,63 @@ def _find_in_tree(self, path: list[str]) -> DataType:
...
+        parts = db_col.split(DEFAULT_DELIMITER)
+        for i in range(len(parts), 0, -1):
+            sentinel = self._optional_sentinel_at(DEFAULT_DELIMITER.join(parts[:i]))
+            if sentinel is not None:
+                return None if sentinel == db_col else sentinel
+        return None
+
+    def order_by_column(
+        self, db_col: str, *, descending: bool = False
+    ) -> "ColumnExpr | None":
```
dreadatour:
```suggestion
    def order_by_column(
        self, db_col: str, *, descending: bool = False
    ) -> ColumnExpr | None:
```

src/datachain/lib/signal_schema.py
```diff
@@ -908,6 +957,63 @@ def _find_in_tree(self, path: list[str]) -> DataType:
...
+        """Order-by expression for a nullable column (a leaf under
+        ``Optional[DataModel]`` or a top-level ``Optional[basic]``), with explicit
+        ``NULLS LAST`` so a NULL row sorts last on every backend (SQLite would
+        otherwise sort NULLs first). None when ``db_col`` is not nullable."""
```
dreadatour: I think we can simplify this docstring.

src/datachain/lib/signal_schema.py
```diff
@@ -935,13 +1041,24 @@ def rebuild(node):
 
         return rebuild(expr)
 
+    @staticmethod
+    def _optionalize(py_type: "DataType", nullable: bool) -> "DataType":
```
dreadatour:
```suggestion
    def _optionalize(py_type: DataType, nullable: bool) -> DataType:
```

src/datachain/lib/signal_schema.py
```diff
@@ -1114,12 +1242,17 @@ def _get_flat_tree(
...
     ) -> Iterator[tuple[list[str], DataType, bool, int]]:
         for name, (type_, substree) in tree.items():
             suffix = name.split(".")
             new_prefix = prefix + suffix
             if not include_sys and new_prefix and new_prefix[0] == "sys":
                 continue
+            # The `_is_null` sentinel (leading underscore) is kept for storage and
+            # hydration but dropped from user-facing output.
+            if not include_sentinels and name.startswith("_"):
+                continue
```
dreadatour: Now I see `_` prefix is by design.

src/datachain/func/conditional.py
```diff
@@ -1,17 +1,55 @@
...
+    from datachain.lib.signal_schema import SignalSchema
 
 CaseT = int | float | complex | bool | str | Func | ColumnExpr
 
 
+class _IsNoneFunc(_SentinelAwareFunc):
+    """``isnone`` is None-aware for ``Optional[DataModel]``: that column isn't a
+    real column on disk, so it reads the model's ``_is_null`` sentinel. Any other
+    column (including ``Optional[basic]``) keeps the plain ``col IS NULL`` check.
+    """
```
dreadatour: Let's either remove this or rephrase? 🙏

src/datachain/func/conditional.py
```diff
@@ -1,17 +1,55 @@
...
+
+    def is_nullable_result(
+        self,
+        signals_schema: "SignalSchema | None",
+        col_type: "DataType | None" = None,
+    ) -> bool:
+        # isnone always returns True/False (never NULL), so its column stays bool.
```
dreadatour: `isnone`?

src/datachain/func/aggregate.py
```diff
@@ -106,6 +150,7 @@ def avg(col: AggColT) -> Func:
     Notes:
         - The `avg` function should be used on numeric columns or expressions.
         - The result column will always be of type float.
+        - Skips rows where the value's ``Optional[DataModel]`` parent is absent.
```
dreadatour, line 148: 🤔 I think we can find better description for this note? Same below.

src/datachain/func/aggregate.py
```diff
@@ -73,6 +115,8 @@ def sum(col: AggColT) -> Func:
     Notes:
         - The `sum` function should be used on numeric columns or expressions.
         - The result column type will be inferred from the input expression type.
+        - When summing a leaf of an ``Optional[DataModel]`` column, rows whose
+          parent object is absent (NULL) are skipped, like any other NULL value.
```
dreadatour: Too complicated? 👀

dreadatour (commented):
Few tiny comments. No critical findings from me.

dreadatour:
One more question: is it backward compatible? I bet it is, just wanted to be sure we have this question asked 👀

dmpetrov:
> is it backward compatible?

yes, except this bugfix:

> Fixes of the issues we had before but didn't pay attention:
> 
> reading nullable value from parquet/CSV/JSON/read_values use to led to type-default (NULL in SQLite and \0 in CH). Now it's correctly becomes NULL.

dmpetrov:
Note, I renamed `_is_null` to `_tag`  to unify Optional and coming Union - so, both use a single variable. See: https://github.com/datachain-ai/datachain/pull/1791/commits/53757f3634beca474a3603b4964b32be4980641c

tests/unit/lib/test_optional.py
```diff
@@ -0,0 +1,357 @@
...
+    schema = SignalSchema({"out": _Outer, "id": int})
+    result = schema.select_except_signals("out.addr.city")
+    leaves = result.user_signals()
+    assert "out.addr.city" not in leaves
+    assert "out.addr.zip" in leaves
+    assert not any(s.endswith("_tag") for s in leaves)
+
+
+def test_compare_signals_ignores_optional_sentinel():
+    """compare_signals() (delta/diff) must not surface the internal sentinel."""
```
shcheklein: let's review and remove "obvious" comments that just repeat test names

tests/unit/test_sqlite_add_column.py
```diff
@@ -103,3 +103,23 @@ def test_add_column_not_null_with_existing_rows(catalog):
     assert "uuid" in column_names
 
     db.execute_str(f"DROP TABLE {table_name}")
+
+
+@skip_if_not_sqlite
+def test_add_column_idempotent(catalog):
+    """Re-adding the same column is swallowed silently (mirrors CH's
```
shcheklein: obvious

src/datachain/lib/signal_schema.py
```diff
@@ -817,19 +838,45 @@ def get_column_type(self, col_name: str, with_subtree: bool = False) -> DataType
...
         Returns DB columns as strings or Column objects with proper types
         Optionally, it can filter results by specific object, returning only his signals
         """
+
         signals = [
             DEFAULT_DELIMITER.join(path)
             if not as_columns
-            else Column(DEFAULT_DELIMITER.join(path), python_to_sql(_type))
+            else Column(
+                DEFAULT_DELIMITER.join(path), self._db_leaf_sql_type(path, _type)
```
shcheklein, line 891: important: _db_leaf_sql_type calls now internally `is_nullable_column` which is by itself traversal. It becomes expensive - we keep traversing a lot. Can hurt of schemas of 500 (?)

may be some memoize needed along these helpers

src/datachain/lib/data_model.py
shcheklein, line 214: Important to consider: 

We annotate fields with None here -- all of them. It means on CH there will be a large penalty.

This and similar stuff is used in all our exporters (parquet, HF, JSON, DB) most likely. This was done to tolerate Nones in some values when we serialize as pydantic. 

It means for all our exporters in Studio we will be taking a hit across all fields.

shcheklein:
A scenario that breaks related to Optional Files. Overall it would be nice to research existing tests for files (we didn't support returning None I think file files and managed it somehow)

```python
class Wrap(DataModel):
    f: File = None                              # auto-promoted to Optional[File]

dc.read_values(w=[Wrap(f=File(path="x"))]).to_values("w")[0].f.read_text()
# >>> RuntimeError: Cannot open file: catalog is not set
```

Also

```python
dc.read_values(f=[f, None]).to_storage("/tmp/e", signal="f")
# >>> AttributeError: 'NoneType' object has no attribute 'export'   (file.py:77)
```

shcheklein:
One more scenario that fails (union on optional and non optional) + it raises sometimes since _tag is an extra field. Summarized and reproduced with AI:

---------

All outputs observed on head `53757f36` against ClickHouse 25.8 (studio test fixtures):

```python
import datachain as dc
from datachain import C

class Inner(DataModel):
    score: int = 0
    label: str = ""

mk_plain = lambda: dc.read_values(item=[Inner(score=1, label="a")])
mk_opt   = lambda: dc.read_values(item=[Inner(score=2, label="b"), None])

mk_opt().union(mk_plain()).save("u_ok")     # optional LEFT — correct
# CH DDL:  item___tag Nullable(Int64), item__score Nullable(Int64), item__label Nullable(String)

mk_plain().union(mk_opt()).save("u_bad")    # plain LEFT — left types win
# CH DDL:  item__score Int64 DEFAULT 0, item__label String DEFAULT '', item___tag Int64 DEFAULT 0

back = dc.read_dataset("u_bad")
sorted(str(r[0]) for r in back.to_list("item"))
# >>> ['None', "score=1 label='a'", "score=2 label='b'"]   ← hydration LOOKS fine (_tag=1 survived)

back.filter(C("item.label") == "").count()  # >>> 1 on ClickHouse, 0 on SQLite  ← phantom match
back.filter(C("item.score") == 0).count()   # >>> 1 on ClickHouse, 0 on SQLite  ← phantom match
```

The absent row's NULL leaves were default-filled (`''`/`0`) on insert into the NOT NULL columns, so every leaf-level operation (filters, aggregations, exports) sees phantom values on CH that don't exist on SQLite — while `to_iter` hides the problem by reading `_tag` first. And because `_tag` itself was saved `Int64 DEFAULT 0` (0 = "present"), any future outer join padding this table with NULL `_tag` will hydrate a garbage *present* object.

Found while building this repro (now P2-7): chains are **mutated in place by `union()`** (`left.signals_schema = clone_without_sys_signals()`), so using a chain as the LEFT of one union and then the RIGHT of another makes `_align_optional_for_union`'s `set(lvals) != set(rvals)` precondition fail on the missing `sys` signal → confusing `UnionSchemaMismatchError: item___tag only present in right` on both backends:

```python
opt = mk_opt()
opt.union(mk_plain()).save("u2")       # mutates `opt` in place (drops sys)
mk_plain().union(opt).save("u3")
# >>> UnionSchemaMismatchError: Cannot perform union. item___tag only present in right
```

src/datachain/lib/dc/datachain.py
```diff
@@ -126,6 +127,33 @@
 T = TypeVar("T", bound="DataChain")
 
 
+def _is_all_none(value: Any) -> bool:
+    """True for None or a (non-empty) dict whose every value is recursively None."""
+    if value is None:
```
shcheklein: important: how about floats, lists, empty dicts (that have special handling now AFAIU)?

src/datachain/lib/arrow.py
```diff
@@ -279,21 +280,42 @@ def _get_datachain_schema(schema: "pa.Schema") -> SignalSchema | None:
     return None
 
 
+def _subtree_all_none(
+    column_values: dict[str, Any], model: type["BaseModel"], prefix: str
+) -> bool:
+    """True when every scalar leaf under ``model`` (at ``prefix``) is None — i.e.
```
shcheklein: important: does it handle NaN, [], {}?

overall, we have three different "all none" checks - can it be dried?

shcheklein: issue summarized and reproduced with AI:


```python
from typing import Optional
import datachain as dc

class WithFloat(DataModel):
    a: int = 0
    x: float = 0.0

ch = dc.read_values(id=[1, 2], m=[WithFloat(a=1, x=1.5), None],
                    output={"id": int, "m": Optional[WithFloat]})
list(ch.to_iter("m"))            # [WithFloat(a=1, x=1.5), None]   ← in-memory: correct
```

**float leaf — `to_json` does not collapse (both backends), output is non-RFC JSON:**

```python
ch.to_json("out.json")
# >>> [ {"id":1,"m":{"a":1,"x":1.5}}, {"id":2,"m":{"a":null,"x":NaN}} ]
#                                            ^^^^ expected "m":null; literal NaN breaks JS/jq/RFC parsers
```

**float leaf — parquet export is unreadable by datachain itself (both backends):**

```python
ch.to_parquet("out.parquet")
dc.read_parquet("out.parquet").to_list("id", "m")
# >>> ValidationError: 1 validation error for WithFloat
#     (heuristic sees x=NaN → "present" → instantiates WithFloat(a=None, …) → a: int rejects None)
```

**list leaf — fine on SQLite, broken on ClickHouse** (`Array DEFAULT []` vs NULL):

```python
class WithList(DataModel):
    a: int = 0
    tags: list[str] = []
# SQLite:     to_json → "m":null ✓        parquet round-trip → None ✓
# ClickHouse: to_json → {"a":null,"tags":[]} ✗   parquet round-trip → ValidationError ✗
```

**Control — int/str-only models work everywhere** (`"m":null`, round-trip → `None`) — which is the only shape the PR's tests cover.

**P2-8 — single-signal chains crash regardless of leaf types (both backends):**

```python
one = dc.read_values(m=[WithFloat(a=1, x=1.5), None], output={"m": Optional[WithFloat]})
one.to_parquet("one.parquet")
dc.read_parquet("one.parquet")
# >>> UdfError: UDF returned an invalid value for output column 'm__s'. Expected str, got 1 (type: int)
#     (the m._tag column in the file shifts every leaf by one on the single-signal read path;
#      add any second column and the same round-trip works — that's why the PR's test passes)
```

shcheklein:
There might be a bunch of issues with to_parquet round trips (and other similar)

It seems we still serialize _tag? Also it seems parquet round-trip crashes whenever the optional model is the chain's only signal

tests/unit/lib/test_optional.py
```diff
@@ -0,0 +1,357 @@
...
+    model = schema.create_model("Probe")
+    annotations = {f: fi.annotation for f, fi in model.model_fields.items()}
+    assert annotations == {"id": int, "addr": _Addr, "name": str}
+    # the rebuilt schema must not have gained an Optional[DataModel] sentinel.
+    leaves = SignalSchema({"probe": model}).user_signals()
+    assert not any(s.endswith("_tag") for s in leaves), leaves
+
+
+def test_infer_optional_datamodel_from_nones():
+    # A DataModel column with some None values is inferred as Optional[DataModel]
```
shcheklein: different styles of comments, oververbose comments

src/datachain/lib/data_model.py
dmpetrov: Yes, it's intentional. We already paying a big price in code/logic to cover specific CH case - while it's the only DB that's tries to be smart about it which creates a lot of PIA for us.

Let's keep this part as is and also consider accepting CH-nullability as default to match all other warehouses and localize this issue to CH level instead of covering it everywhere over code - Created #1815

shcheklein: kk, resolving for now ... agree that CH complicates too much and overoptimizing hurts too much also - it is already too complicated w/o that ... it is a bid sad though

shcheklein:
A few more findings: 

1. `.label()` breaks sentinel-aware funcs

   ```python
   ch = dc.read_values(id=[1, 2], item=[Inner(score=5, label="abc"), None],
                       output={"id": int, "item": Optional[Inner]})
   ch.mutate(x=func.isnone("item")).to_values("x")                    # [False, True]  ← works
   ch.mutate(x=func.isnone("item").label("x")).to_values("x")
   # >>> SignalResolvingError: mutate() error - cannot resolve signal name 'item': is not found
   ch.group_by(c=func.count("item").label("c"), partition_by="id")    # same error
```

2. Composed functions break on nullable:

```python
   m = ch.mutate(ln=func.string.length("item.label"))        # direct func over nullable leaf
   m.schema["ln"], sorted(m.to_values("ln"), key=str)        # (int | None, [3, None])  ← correct
   m2 = ch.mutate(ln2=func.string.length("item.label") + 1)  # composed: func + arithmetic
   m2.schema["ln2"]                                          # int           ← schema lies
   m2.to_values("ln2")   # SQLite:     [4, None]
                         # ClickHouse: [4, 0]    ← absent row silently becomes 0
   ```

File: src/datachain/sql/types.py (hunk 16 above)
Comment (dmpetrov) [reply src/datachain/sql/types.py]: fixed. trimmed all verbose comments or removed.

File: src/datachain/lib/signal_schema.py (hunk 20 above)
Comment (dmpetrov) [reply src/datachain/lib/signal_schema.py:989]: yes, that's fine. 4 underscores are not.

File: src/datachain/lib/signal_schema.py (hunk 39 above)
Comment (dmpetrov) [reply src/datachain/lib/signal_schema.py:891]: good idea!

File: src/datachain/lib/udf.py (hunk 1 above)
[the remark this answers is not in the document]
Comment (shcheklein) [reply src/datachain/lib/udf.py]: should we just pass strict depending on a consumer. If the scenario described is right (silent drop of values) - this it not good tbh.

File: src/datachain/lib/dc/datachain.py (hunk 40 above)
Comment (dmpetrov) [reply src/datachain/lib/dc/datachain.py]: resolved. now _tag/_type_tag is used to resolve it propoerly

File: src/datachain/lib/arrow.py (hunk 41 above)
Comment (dmpetrov) [reply src/datachain/lib/arrow.py]: fixed by making absence tag-driven rather than value-guessed- _type_tag

**Comment by dmpetrov** [comment]
> breaks related to Optional Files.

Good catch! Fixed in udf level. It works now.

> One more scenario that fails (union on optional and non optional)

THat's insightful. Fixed by using Nullable.

> There might be a bunch of issues with to_parquet round trips (and other similar)

The only-signal crash is fixed  - _flatten_row double-counting the tag

And yes, we do serialize `_type_tag` in parquet, by default and it keep all the types safe. We should make it optional #1813 with some type guessing logic. By default is solid.

**Comment by shcheklein** [comment]
> And yes, we do serialize _type_tag in parquet, by default and it keep all the types safe. We should make it optional https://github.com/datachain-ai/datachain/issues/1813 with some type guessing logic. By default is solid.

it looks ugly tbh ...

how about to_database? 

how we manage not doing this for JSON (unless I'm missing something)? 

do roundtrips work properly? why do we need then some heuristics?

**Comment by dmpetrov** [comment]
> > And yes, we do serialize _type_tag in parquet, by default and it keep all the types safe. We should make it optional #1813 with some type guessing logic. By default is solid.
> 
> it looks ugly tbh ...

I see this similar to hidden File signals that also look ugly and we are hiding these in right places like show() and UI while keeping in internals.

To make it nice looking we need an ability to hide/show these hidden signals and choose a right default #1813 like ugly-but-clean-conversion or nice looking and guessing Null values vs null-objects.

> how about to_database?
> how we manage not doing this for JSON (unless I'm missing something)?

Good points. Now it's a mix now. Let me unify it - it will be visible everywhere / round-trip works. ANd after #1813 user can hide them.

> do roundtrips work properly? why do we need then some heuristics?

Yes. Now, heuristics are needed only when then _type_tag is missing.

**Comment by shcheklein** [comment]
> I see this similar to hidden File signals that also look ugly and we are hiding these in right places like show() and UI while keeping in internals.

hidden File fields do have meaningful user facing semantics, and we users don't need some of them they can select_except

_tag is really internal implementation detail (not even human readable, and not clear why is tag from the user perspective). It is more like a bunch of other pydantic internal fields that are never serialized.

I don't know why would users want to see them by default tbh.

Since we have some additional logic to infer those then we don't need them at all (?)

**Comment by dmpetrov** [comment]
> I don't know why would users want to see them by default tbh.

It was an attempt to fix an issues you pointed earlier about the to_parquet() round-trip. But yes, it does not look right and I'm reverting it.

> Since we have some additional logic to infer those then we don't need them at all (?)

Now, exports rely on the all-leaves-None **heuristic** and only rare cases are not covered (see 1st item in Exceptions in the PR description). For a safe round-trips we need to implement `include_type_tag=True` in #1813 but it should not be default.

**Comment by dmpetrov** [comment]
@shcheklein @dreadatour please take another look - I addressed all the issues.

**Comment by shcheklein** [comment]
There is still an edge case with union I think:


```python
   a = dc.read_values(x=[1, 2], output={"x": int})
   b = dc.read_values(x=[3, None], output={"x": Optional[int]})
   u = a.union(b)
   u.schema["x"], sorted(u.to_values("x"), key=str)   # (int, [1, 2, 3, None])  ← schema lies
```


also:

```
lass Inner(DataModel):
    score: int = 0
    label: str = ""

mk_plain = lambda: dc.read_values(item=[Inner(score=1, label="a")])
mk_opt   = lambda: dc.read_values(item=[Inner(score=2, label="b"), None])

mk_opt().union(mk_plain()).save("u1")    # optional LEFT
# CH: `item___type_tag` Nullable(Int64), `item__score` Nullable(Int64), `item__label` Nullable(String)

mk_plain().union(mk_opt()).save("u2")    # plain LEFT — leaves are Nullable now 👍, but:
# CH: `item___type_tag` Int64 DEFAULT 0,  `item__score` Nullable(Int64), `item__label` Nullable(String)
```

dmpetrov:
> with union I think:

good catch. It was a bigger issue with non-DataModel types. fixed.

src/datachain/lib/data_model.py
```diff
@@ -105,13 +124,69 @@ def _fingerprint_tree(
...
+
+
+def promote_default_none(model: type[BaseModel]) -> None:
+    """Auto-promote non-Optional fields with `default=None` to `Optional[...]`.
+
+    `x: int = None` is treated as `x: Optional[int] = None`, so the column is
+    nullable and `x=None` round-trips as `None`. Without this it would read back
+    as the type default (`0`/`""`) on backends with non-nullable columns.
+
+    Skipped under `skip_dc_validation()` for internally-generated models, whose
```
shcheklein: can we please improve the messaging and may be even `_skip_dc_validation` name a bit? it took me a while to understand what is going on there. May be describe a case - internally generated -> model generated by us when we read a dataset w/o model code available, when we have to reconstruct it

I wonder also is there a strict need in that case to do `default=None` or was it only because we were reusing `create_model` across dataset reader and read_parquet and other importers.

shcheklein:
Please take a look into this discrepancies.

I'm lost a bit in `float`s. `floats` is an edge case and I don't see that many tests that would clarify the expected behavior. 

We can potentially do `isnone(float_col)` on CH as `isNaN(col) OR isNull(col)` (?) to remove this discrepancy across the backends.

Or if in this case SQLite is the weird one, right? And no way to fix it. Should we do `isnone(float_col)` on CH as `isNaN(col) OR isNull(col)` (?) to remove this discrepancy across the backends? It would contaminate CH in this case, but at least make it predictable and consistent for users?

Or may be even make SQLite strict - don't allow any data corruption at all? Just raise an excception.


```python
   from datachain import func
   ch = dc.read_values(id=[1, 2, 3], score=[0.97, None, 0.55],
                       output={"id": int, "score": Optional[float]})

   ch.filter(func.isnone("score")).count()
   # SQLite: 1   ClickHouse: 0      ← same pipeline, different result
   ch.filter(func.not_(func.isnone("score"))).count()
   # SQLite: 2   ClickHouse: 3

   # contrast — Optional[int] (a real NULLABLE_SCALAR) is consistent: isnone -> 1 on both
```

Also, AFAIU `count`, `sum`, etc - they are also "broken" - return different values across bakends.

dmpetrov:
> I'm lost a bit in `float`s. `floats` is an edge case and I don't see that many tests that would clarify the expected behavior.

Confirmed and fixed. Issue: float was excluded from the nullable scalars, so Optional[float] wasn't actually nullable and CH kept the None as NaN while SQLite kept NULL.

Fixed SQLite part since CH behaves correctly - the same way as all the other data warehouses.

tests/func/test_optional.py
```diff
@@ -130,16 +130,65 @@ def txt(id: int) -> Optional[str]:
...
+    }
+    assert rows[1] == (10, "n1", False, 1.5)
+    assert rows[2] == (None, None, None, None)
+    assert rows[3] == (30, "n3", False, 4.5)
+
+
+def test_optional_float_none_consistent_across_backends(test_session):
+    """isnone/count/sum over an Optional[float] None agree on every backend (the
+    same SQL null semantics as Optional[int]), now that float is a nullable
+    scalar. Without it, ClickHouse kept the None as NaN and diverged from SQLite."""
```
shcheklein: please fix the comment, we don't want to refer to some "decision random making in the middle of some PR"

shcheklein:
Nuanced thingy found by AI, latent issue probably worth fixing if this is a one-liner:

>  **`dc_nullable` mutable state + SQLAlchemy caching is a *paired* latent trap** [V-me, corrected mechanism]. `dc_nullable` is set post-construction (`sql_type.dc_nullable = True`, 3 sites) and is not a constructor arg, so it can't appear in SQLAlchemy's auto-generated cache key. `SQLType` declares `cache_ok = True` — but verified that's currently **ineffective**: SQLAlchemy only honors `cache_ok` on the *concrete instantiated* class, and it lives on the intermediate `SQLType` base, so every datachain type (`String`, `Int64`, …) emits `NO_CACHE` and SQLAlchemy warns "will not produce a cache key" (a pre-existing per-class-`cache_ok` quirk, not from this PR). Net today: no stale-key miscompile, for **two independent reasons** (so it's deeply latent): (1) `cache_ok` is ineffective ⇒ `NO_CACHE`, nothing is cached; and more fundamentally (2) the "cache" in question is SQLAlchemy's *Engine statement-compilation cache*, used only on the `connection.execute(sa_statement)` path — but datachain never uses it: both warehouses `compile_to_args(query)` → `statement.compile(dialect=...)` (a one-shot, uncached compile) and run the **raw SQL string** on the DBAPI cursor (`db_engine.py:63`, `sqlite.py:257`, `clickhouse.py:665`). A miscompile would require *both* (a) `cache_ok` fixed per-class (verified trivial: `class StringLeaf(String): cache_ok = True` becomes cacheable) *and* (b) datachain adopting SQLAlchemy's cached execute path. Still worth defusing cheaply: thread `dc_nullable` through `__init__` (constructor args *are* in the auto cache key), which makes the type honestly cache-safe and also lets call sites do `Int64(dc_nullable=True)` instead of construct-then-mutate. (Aside: the perf half of the `NO_CACHE` warning is also moot for datachain — it recompiles every statement via `.compile()` regardless; benefiting from caching would require switching to `engine.execute`, out of scope.)

pyproject.toml
```diff
@@ -102,7 +102,9 @@ zarr = [
 ]
 tests = [
   "datachain[torch,audio,remote,vector,hf,video,postgres,zarr]",
-  "pytest>=8,<10",
+  # <9.1: 9.1.0 rejects indirect re-parametrization of params'd fixtures
```
shcheklein: can we please properly fix this?

shcheklein:
The issue with expression was fixed only for Func. We need to check all other cases.

```python
ch = dc.read_values(id=[1,2,3], a=[10,None,30], output={"id":int,"a":Optional[int]})
ch.mutate(s=C("a")+1).schema["s"]      # int   (should be int | None)
# after .save() + read_dataset:
#   ClickHouse → {1:11, 2:0,    3:31}   ← None became 0  (silent)
#   SQLite     → {1:11, 2:None, 3:31}
```

shcheklein:
Discrepancy in diff:

```python
base = dc.read_values(id=[1,2,3], x=[1,None,3], output={"id":int,"x":Optional[int]}).save("base")
new  = dc.read_values(id=[1,2,3], x=[1,5,None], output={"id":int,"x":Optional[int]}).save("new")
dc.read_dataset("new").diff(dc.read_dataset("base"), on=["id"], compare=["x"]).to_values("id")
# >>> []     ← id2 (None→5) and id3 (3→None) NOT detected as modified
```

shcheklein:
`values_to_tuples.py:120` added None-driven widening only for `Optional[DataModel]`, not scalars.

```python
dc.read_values(a=[10, None, 30]).schema["a"]     # int  → None saved as 0 on CH (verified)
# read_database with a nullable column hits the same path.
# read_json/jsonl with a key missing in some records → that field becomes REQUIRED,
#   rows lacking it are silently DROPPED (verified: 3-row jsonl, 1 missing "b" → 2 rows kept)
```

shcheklein:
We need to test `.over(order_by=Optional)`, `merge`, etc. Check also other operations.

Examples:

```python
  left  = dc.read_values(k=[1,2,3], output={"k":int})
  right = dc.read_values(k=[1], rv=[99], output={"k":int,"rv":int})
  left.merge(right, on="k", rname="r_").to_records()      # (k,rv):
  # SQLite:     [(1,99), (2,None), (3,None)]   ← unmatched rv = None
  # ClickHouse: [(1,99), (2,0),    (3,0)]      ← unmatched rv = 0  (non-optional → type default)
```

```python
  w = dc.read_values(id=[1,2,3,4], g=[1]*4, s=[3,None,1,None], output={"id":int,"g":int,"s":Optional[int]})
  win = func.window(partition_by="g", order_by="s")
  w.mutate(rn=func.row_number().over(win)).to_records()   # (id,rn) pairs:
  # SQLite:     {2:1, 4:2, 3:3, 1:4}   ← NULL rows (2,4) ranked FIRST
  # ClickHouse: {3:1, 1:2, 2:3, 4:4}   ← NULL rows ranked LAST
```

```python
  a = dc.read_values(k=[1,None,3], output={"k":Optional[int]})
  b = dc.read_values(k=[None],     output={"k":Optional[int]})
  a.subtract(b, on=["k"]).to_values("k")
  # SQLite:     [1, 3]          ← NULL-key row removed (NULL == NULL)
  # ClickHouse: [1, 3, None]    ← NULL-key row kept (NULL != NULL)
```

src/datachain/lib/data_model.py
dmpetrov: changed to `_skip_optional_promotion` to make it clearer

`Default=None` is needed. Mostly for select/to_partial/group_by - it let's the missing field to be None.

tests/func/test_optional.py
dmpetrov: That's strange - we have a corresponding rule in AGENT.md but it's a bit abstract.

Added more concrete rule:
```
Write every comment as a standalone, timeless fact about current behavior — a reader
months later on `main` has no notion of "this change". Watch for change-narrative tells:
*"now that", "without it", "previously", "used to", "we decided", "regression",
"this PR"*, and before/after framing. Git history records why it changed.
```

pyproject.toml
dmpetrov: fixed

dmpetrov:
FYI: it's mostly fixed but I'm still researching on the leftovers

dmpetrov:
> ```python
>   left  = dc.read_values(k=[1,2,3], output={"k":int})
>   right = dc.read_values(k=[1], rv=[99], output={"k":int,"rv":int})
>   left.merge(right, on="k", rname="r_").to_records()      # (k,rv):
>   # SQLite:     [(1,99), (2,None), (3,None)]   ← unmatched rv = None
>   # ClickHouse: [(1,99), (2,0),    (3,0)]      ← unmatched rv = 0  (non-optional → type default)
> ```

merge() will change the current CH behaviour. But this is kind of a bugfix, so, it worth implementing even with backward compatibility break. The same way as "reading nullable value from parquet/CSV/JSON/read_values" in the issue description.

```
left  = dc.read_values(k=[1,2,3], output={"k":int})
right = dc.read_values(k=[1], rv=[99], output={"k":int,"rv":int})
m = left.merge(right, on="k", rname="r_")   # default = left outer join

┌─────┬─────────────────────────────────────┬────────────────────┬──────────────────────────────────────┐
│  #  │             Expression              │  SQLite before →   │      ClickHouse before → after       │
│     │                                     │       after        │                                      │
├─────┼─────────────────────────────────────┼────────────────────┼──────────────────────────────────────┤
│ 1   │ m.schema["rv"]                      │ int → int | None   │ int → int | None                     │
├─────┼─────────────────────────────────────┼────────────────────┼──────────────────────────────────────┤
│ 2   │ m.to_values("rv")                   │ [99,None,None] →   │ [99,0,0] → [99,None,None]            │
│     │                                     │ same               │                                      │
├─────┼─────────────────────────────────────┼────────────────────┼──────────────────────────────────────┤
│ 3   │ m.filter(C("rv")==0).count()        │ 0 → same           │ 2 → 0 (unmatched no longer == 0)     │
├─────┼─────────────────────────────────────┼────────────────────┼──────────────────────────────────────┤
│ 4   │ m.filter(func.isnone("rv")).count() │ 2 → same           │ 0 → 2                                │
├─────┼─────────────────────────────────────┼────────────────────┼──────────────────────────────────────┤
│ 5   │ func.count("rv") (group)            │ 1 → same           │ 3 → 1 (stops counting unmatched)     │
├─────┼─────────────────────────────────────┼────────────────────┼──────────────────────────────────────┤
│ 6   │ func.sum("rv")                      │ 99 → same          │ 99 → 99 (0s and skipped-NULLs both   │
│     │                                     │                    │ sum to 99)                           │
├─────┼─────────────────────────────────────┼────────────────────┼──────────────────────────────────────┤
│ 7   │ m.save(...) → CH column type        │ INTEGER → same     │ Int64 → Nullable(Int64)              │
└─────┴─────────────────────────────────────┴────────────────────┴──────────────────────────────────────┘
```

shcheklein:
One more finding (with the help of AI):

------------------------

`Catalog.pull_dataset` builds the schema from the remote dataset's serialized column types: `schema = parse_schema(remote_ds_version.schema)` (`catalog.py:1852`).
`parse_schema deserializes each column type by calling from_dict: res[c_name] = NAME_TYPES_MAPPING[c_type["type"]].from_dict(c_type)` (`dataset.py:163`).
`SQLType.from_dict` (`sql/types.py:183-188`) is where the PR's change bites:

```python
def from_dict(cls, d):
    if d.get("dc_nullable"):
        inst = cls(); inst.dc_nullable = True
        return inst          # ← INSTANCE  (for Optional[datetime])
    return cls               # ← CLASS     (for plain datetime)
```

So an `Optional[datetime]` column — serialized as `{"type":"DateTime","dc_nullable":True}` — deserializes to a `DateTime()` instance; a plain datetime column deserializes to the `DateTime` class.

That schema `dict` is handed to `DatasetRowsFetcher(schema=schema)` (`catalog.py:1890`) and stored as `self.schema` (`catalog.py:130`).
`fix_columns` (`catalog.py:214`) — which converts the epoch timestamps in the pulled parquet back into datetimes — selects the datetime columns with a class-equality check, and fails the `t == DateTime`:

```python
for c in [c for c, t in self.schema.items() if t == DateTime]:
    df[c] = pd.to_datetime(df[c], unit="s")
```

dmpetrov:
In addition to the reported ones, I found more issues:
- `cast()` in CH - direct and indirect via read_values(..., output=...)
- null-safe in SQLite were failing in None's since these are custom functions, not built-in SQL like in CH
- `split()` crushes with Nullable in CH
- `greatest()`/`least()` over optional inconsistency
- conditional/boolean nullability - `case` + `and_`/`or_`/`not_` preserve NULL on CH
- `func.collect()` drops NULL elements on CH

I'll le you know when it's fully ready.

dmpetrov:
@shcheklein all fixed. Please take another look.

shcheklein:
@dmpetrov please review all unresolved comments  (7 of them). Same for the Studio PR.

Screenshot 2026-06-16 at 3 40 49 PM

src/datachain/func/func.py
dmpetrov: I can't - it'll create circular dependencies

src/datachain/lib/signal_schema.py
dmpetrov: It suppose to be this way. Now it also include `float` after NaN fix - less strict as it was before.

src/datachain/lib/udf.py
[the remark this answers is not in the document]
dmpetrov: good point. added the check before zip that raises an exception if returns more values than declared. the zip stays not-strict for arrow/parquet. updated tests.

src/datachain/func/func.py
dmpetrov: Agree it's over-defensive. Removed. No affect on unit tests.

shcheklein:
A few more things / scenarios to review (and fix?):

1. `array.get_element/slice` crash at definition time over `Optional[list]` (no None row needed):

```python
dc.read_values(x=[[1, 2]], output={"x": Optional[list[int]]}).mutate(y=func.array.get_element("x", 0))
# RAISES DataChainColumnError: Array column must be of type list   (at mutate(), both backends)
```

2. split on None behave differently:

```python
c = dc.read_values(t=["a-b", None], output={"t": Optional[str]})
c.mutate(s=func.string.split("t", "-")).to_values("s")
# SQLite -> [['a','b'], None]   |   CH -> [['a','b'], []]      (schema list[str], non-nullable)
c.mutate(n=func.array.length(func.string.split("t", "-"))).to_values("n")
# SQLite -> [2, None]           |   CH-save -> [2, 0]          (schema int, non-nullable)
```

3.  `array.contains` over a NULL whole-array:

```
dc.read_values(a=[None], output={"a": Optional[list[str]]}).mutate(c=func.array.contains("a", "x")).to_values("c")
# SQLite -> OperationalError (py_json_array_contains: json.loads(None))   |   CH -> [0]  (has([], 'x'))
```

4.  `concat` over an all-NULL group:

```
dc.read_values(g=[1, 1], s=[None, None], output={"g": int, "s": Optional[str]}) \
  .group_by(c=func.concat("s", ","), partition_by="g").to_values("c")
# SQLite -> [None]   |   CH -> ['']
```

5. `outer/full` merge of a plain DataModel - do we want to return optional model column there?

```python
class Inner(DataModel):
    score: int = 0
    label: str = ""
left  = dc.read_values(k=[1], output={"k": int})
right = dc.read_values(k=[2], m=[Inner(score=9, label="x")], output={"k": int, "m": Inner})
left.merge(right, on="k", full=True).order_by("k").to_values("m")
# SQLite -> [None, Inner(9,'x')]   |   CH -> [Inner(0,''), Inner(9,'x')]   (join_use_nulls=0)
```

6. `string_hash(NULL)`:

```python
dc.read_values(s=[None], output={"s": Optional[str]}).mutate(h=func.string.string_hash("s")).to_values("h")
# SQLite -> [-2039914840885289964]  (hash of "")   |   CH -> NULL/0
```

7. `Optional[list]`/`dict` empty-vs-absent missed in diff (the visible side is the correct SQLite one; CH is the bug):

```python
a = dc.read_values(id=[1], tags=[[]],   output={"id": int, "tags": Optional[list[str]]})
b = dc.read_values(id=[1], tags=[None], output={"id": int, "tags": Optional[list[str]]})
a.diff(b, on=["id"], compare=["tags"], status_col="status").to_records()
# SQLite -> [{'status': 'M', ...}]  (Modified)   |   CH -> 'S' (Same — [] and None both stored as [])
```

dmpetrov:
> please review all unresolved comments (7 of them). Same for the Studio PR.

All addressed on both PRs. Looking at the last comment...

dmpetrov:
> A few more things / scenarios to review (and fix?):

Good ones. Mostly 1 line fixes for each one but still important.

5 (fixed) - outer/full merge/join - Initially I decided not to do that because it had a large footprint and too CH-specific. But after recent changed we made it become simple change - I didn't even noticed.

2 and 7 (not fixed) - same CH-specific issue. But the change is massive. Let's solve it in docs level. There are few CH-SQLite miss-matches that will be documented. These are one of these.

Now PR is ready for another review.

shcheklein (approved):

PR #1793, opened 2026-06-01
Add Zarr support: read_zarr, ZarrStore/Array/Selection

shcheklein:
## Summary

Adds a generic Zarr core to the DataChain library.

- **`read_zarr(path, ...)`** — discovers `*.zarr` stores under a path and yields one row per store.
- **`ZarrStore` / `ZarrArray`** — data models for inspecting a store's arrays and reading array data (using the `zarr` package).
- **`ZarrSelection`** — a lazy, bounded region inside an array that travels through a chain as a column without materializing its bytes. Provides `read()` and an image `read_bytes()` encoder used for streamed previews.

Zarr is an **optional** dependency, kept out of the core install so the
project keeps supporting Python >= 3.10 (the `zarr` 3.x package itself
requires Python >= 3.11). Install it with the `zarr` extra:

```
pip install 'datachain[zarr]'
```

Importing `datachain.lib.zarr` works without the extra; the Zarr ops
raise a clear `ImportError` pointing at `datachain[zarr]` (and the
Python >= 3.11 requirement) only when actually used.

## Test plan

- `tests/unit/lib/test_zarr.py` covers store discovery (one row per store, any depth), single-store reads, array access, and error cases.

### DIFF
--- a/src/datachain/lib/zarr.py
+++ b/src/datachain/lib/zarr.py
@@ -11,6 +11,7 @@ Reading the actual data requires the optional ``zarr`` dependency:
 """
 
 import posixpath
+import sys
 from collections.abc import Iterator
 from typing import Any, ClassVar, Literal
 
@@ -31,10 +32,13 @@ def _import_zarr() -> Any:
     try:
         import zarr
     except ImportError as exc:
-        raise ImportError(
-            "Missing dependencies for Zarr support.\n"
-            "To install run:\n\n"
+        hint = (
             "  pip install 'datachain[zarr]'\n"
+            if sys.version_info >= (3, 11)
+            else "Zarr support requires Python 3.11 or newer.\n"
+        )
+        raise ImportError(
+            "Missing dependencies for Zarr support.\nTo install run:\n\n" + hint
         ) from exc
     return zarr
 
--- a/src/datachain/lib/zarr.py
+++ b/src/datachain/lib/zarr.py
@@ -11,11 +11,31 @@ from collections.abc import Iterator
 from typing import Any, ClassVar, Literal
 from urllib.parse import urlsplit, urlunsplit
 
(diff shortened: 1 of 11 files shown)


dmpetrov (approved):
Looks great

src/datachain/lib/zarr.py
```diff
@@ -0,0 +1,252 @@
...
+
+    ``file`` is a store-root metadata marker.  The store root is the directory
+    that contains it.  When that directory coincides with the listing source
+    (so the marker's relative path has no parent), the source is lifted one
+    level so the root is always addressable as ``source`` + ``path``.
+    """
+    parent = posixpath.dirname(file.path)
+    if parent:
+        return file.source, parent
+    base, _, name = file.source.rstrip("/").rpartition("/")
```
ilongin: Will this work for all the cases? E.g `source="s3://bucket/"` or `source="file:///"` ?

src/datachain/lib/zarr.py
```diff
@@ -0,0 +1,252 @@
...
+        raise ImportError(
+            "Missing dependencies for Zarr support.\nTo install run:\n\n" + hint
+        ) from exc
+    return zarr
+
+
+class ZarrInfo(DataModel):
+    """Summary metadata for a Zarr store."""
+
+    zarr_format: int = Field(default=0)
```
ilongin: From what I can see 0 is not valid zarr_format. Maybe we can set to None for default case.

shcheklein: yep, addressed ... this whole PR will require some rework (it is expected)

src/datachain/lib/zarr.py
shcheklein: addressed

PR #1796, opened 2026-06-01
Added ability to add multiple functions to single map

### LINKED ISSUE #1698: Multiple functions in single udf
### Description

Claude Code like using this pattern: `.map(a=fn1, b=fn2)` instead of `.map(a=fn1).map(b=fn2)`.

That's seem a good syntax sugar and needs to be supported.

- [ ] map
- [ ] gen
- [ ] agg

ilongin:


dmpetrov:
@ilongin this implementation adds syntax sugar without proper optimization - see https://github.com/datachain-ai/datachain/issues/1698#issuecomment-4608446299

I don't think this feature makes sense without it - people and agents will expect all the functions in map to be in a single stage without intermediate result while the current implementation transforms `map(a=f1, b=f2)` to `map(a=f1).persist().map(b=f2)`.

If that's not easy to implement - it's better to postpone this issue since the priority is not high for now.

PR #1797, opened 2026-06-04
Skill knowledge base: refactor to reuse code in SaaS

dreadatour:
Few changes to be able to reuse knowledge base skill in SaaS.


dmpetrov (approved):
Looks good

PR #1800, opened 2026-06-05
Refactor `datachain.schema` to make it more useful

shcheklein:
## Motivation

`DataChain.schema` is already the natural public entry point for schema introspection, but it only exposed top-level signals. Anything that needed leaf columns for nested models had to rely on lower-level internals, while `print_schema()` only wrote a formatted tree to stdout or a file-like object. That made schema-aware scripts, CLIs, examples, and agent skill workflows harder to compose than they need to be.

This PR makes `schema` useful both as structured data and as a printable representation, while keeping the old printed schema semantics as the default. In particular, hidden fields remain included by default so `print(chain.schema)` is a drop-in replacement for `chain.print_schema()`.

## What changed

- Adds public `DataChainSchema`, returned by `chain.schema`.
- Keeps existing dict-like top-level behavior: `chain.schema["file"]`, `chain.schema.items()`, equality with a plain dict, etc.
- Adds `chain.schema.flatten()` for leaf field names and Python types.
- Adds `chain.schema.to_string()` for the nested schema tree format.
- Makes `str(chain.schema)` render the schema tree, so callers can use `print(chain.schema)`.
- Includes hidden fields by default in `str()`, `to_string()`, and `flatten()` to match existing schema-printing semantics; callers can pass `include_hidden=False` when they want the compact public data view.
- Deprecates `DataChain.print_schema()` in favor of `print(chain.schema)`.
- Updates docs, README, examples, CLI output, and the bundled knowledge skill to use the public schema API.

## Examples

Top-level schema behavior stays the same:

```python
ds.schema == {"file": File, "label": str}
ds.schema["file"] is File
```

Print the schema as a drop-in replacement for `print_schema()`:

```python
print(ds.schema)
```

Get leaf columns as structured data, including hidden fields by default:

```python
ds.schema.flatten()
# {
#     "file.source": str,
#     "file.path": str,
#     "file.size": int,
#     "file.version": str,
# }
```

Opt into the compact visible-only view when needed:

```python
ds.schema.flatten(include_hidden=False)
print(ds.schema.to_string(include_hidden=False))
```


dmpetrov (approved):
Looks good. Good idea with print(schema)

PR #1807, opened 2026-06-08
Support multiple functions in a single `.map()` call

### LINKED ISSUE #1698: Multiple functions in single udf

ilongin:
Closes #1698.

`chain.map(stem=f1, ext=f2)` now runs both functions as a single UDF stage - one row iteration, no intermediate column materialized between f1 and f2.

Before this change multi-kwarg was rejected at the signature layer with "multiple signals not supported in processors". Users had to chain `.map(stem=f1).map(ext=f2)`, which adds a separate UDF stage per function.

## How it works

- `_MultiSignalMapper(Mapper)` holds the dict of functions, computes the union of their parameters, and dispatches per row - each function gets only its declared params.
- `DataChain.map()` detects `len(signal_map) > 1`, builds the combined Mapper, derives the output schema from each function's return annotation, and hands the whole thing to the existing single-UDF infrastructure.
- Output flattening, parallelism, prefetch, and pickling all reuse what the regular tuple-returning Mapper path already does.

## Constraints

The following kwargs raise when combined with multi-kwarg:

- `func=` - ambiguous role alongside per-signal functions.
- `output=` - each function's return annotation is the source of truth for its column type.
- `params=` - genuinely ambiguous in a multi-function context. With a single function `params=["a", "b"]` means "pass column a and b positionally as the function's two args". With two functions, the semantics aren't clear: do both functions get the same args (only works if both share the signature)? Do we split by per-functions (no obvious split rule)? Rather than pick one and surprise users later, we reject and tell them to extract nested columns to a top-level column first with a single `.map()`.


amritghimire (approved):
LGTM

dreadatour:
Two comments:

1. Am I right we can not use result of the first function as argument in second function? To me it looks like it will be used often. Do we need to support it or at least add a comment/check/verbose warning about this case?
2. Too few tests for such an important change IMO. Did you consider to add more tests?

dmpetrov:
Right, all such use cases should be supported. Please ask AI what other possible use cases might be missing.

One useful use case that might be a bit out of the scope - removing intermediate results without saving it to table. It's like select_expect() that also modifies the previous map() result. It should be helpful when you need just a couple of columns from a whole LLM response - heavy LLM response should never be saved to table.

src/datachain/lib/udf.py
```diff
@@ -515,32 +516,59 @@ class _MultiSignalMapper(Mapper):
...
+        # For each function: which of its params come from another
+        # function's output (dependencies) vs from an input row column.
+        deps: dict[str, set[str]] = {}
+        for name, fn in signal_map.items():
+            params = list(inspect.signature(fn).parameters.keys())
+            self._per_func_params[name] = params
+            deps[name] = {p for p in params if p in output_names and p != name}
+
+        try:
+            self._exec_order = list(TopologicalSorter(deps).static_order())
```
dreadatour, line 566: Why do we need this? We know the order of map functions — it is defined by user — it is kind of intuitive and obvious, why do we need to guess and use `TopologicalSorter`?

I mean it is great we do think about this, but, since (AFAIK) `TopologicalSorter` does not guarantee the order and in case of `.map(a=f1, b=f2)` it can be run `f1` first and it can be run `f2` first. Each of these UDFs might have side-effects and result might be unexpected in case `f2` will run before `f1`.

Should we simplify the code and let user to take care of correct order?

ilongin: This way we don't enforce user to write functions in the order arguments flow from one to another, which is what we want IMO since it would be strange to enforce ordering in kwargs. 
What do you mean by "TopologicalSorter does not guarantee the order"? The idea of topological sort is to sort by dependencies i.e. if one map function depends on another map function result, then that second one will be ran first. It also finds circular dependencies which is user's mistake and then error is thrown. 
If functions don't have dependencies, then it's irrelevant which one runs first....I just don't see why would that be important.

tests/func/model/test_yolo.py
```diff
@@ -1706,13 +1706,13 @@ def test_yolo_seg_from_results(running_img, running_img_masks):
                 166,
                 165,
                 165,
+                165,
                 164,
-                164,
```
shcheklein, line 1710: @ilongin why do we have these changes in this file at all?

ilongin: tests started to fail because libs used by that code have new version. I fixed it just to pass CI but I also created new PR which should fix this test for good https://github.com/datachain-ai/datachain/pull/1862

src/datachain/lib/udf.py
```diff
@@ -533,6 +535,77 @@ def _prepare_rows(udf_inputs) -> "abc.Generator[Sequence[Any], None, None]":
...
+
+        # combined_params = union of params NOT resolved from other outputs,
+        # in first-seen order. Consumed positionally by the UDF layer.
+        seen: set[str] = set()
+        self.combined_params: list[str] = []
+        for name in signal_map:
+            for p in self._per_func_params[name]:
+                if p in output_names:
+                    continue
+                if p not in seen:
```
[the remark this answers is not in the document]
ilongin: The scenario in the example isn't reachable in practice. Overwriting an existing column via .map() (single or multi-signal) is rejected at the query layer with ValueError: Column already exists or added in the previous steps: stem. And .map(stem=lambda stem: ...) when no stem column exists fails earlier with SignalResolvingError: cannot resolve signal name 'stem'. Either way the _MultiSignalMapper never runs, so the KeyError in combined_params can't be triggered.

tests/func/model/test_yolo.py
```diff
@@ -293,24 +293,22 @@ def test_yolo_seg_from_results(running_img, running_img_masks):
             15,
             15,
             15,
+            15,
+            15,
```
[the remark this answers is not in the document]
ilongin: It's about fixing flaky test that depends on lib version. Opened another PR for proper fix

src/datachain/lib/udf.py
dreadatour: > What do you mean by "TopologicalSorter does not guarantee the order"?

I meant exactly this:

> `TopologicalSorter` does not guarantee the order and in case of .map(a=f1, b=f2) it can be run `f1` first and it can be run `f2` first.

tests/func/model/test_yolo.py
dreadatour: Tests are still broken 👀

src/datachain/lib/dc/datachain.py
```diff
@@ -1003,6 +1015,48 @@ def map(
...
+
+        multi_mapper = _MultiSignalMapper(signal_map)
+        output_dict: dict[str, Any] = {}
+        for name, fn in signal_map.items():
+            anno = inspect.signature(fn).return_annotation
+            output_dict[name] = (
+                anno
+                if anno is not inspect.Signature.empty
+                else UdfSignature.DEFAULT_RETURN_TYPE
+            )
```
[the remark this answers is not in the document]
ilongin: implemented

PR #1814, opened 2026-06-10
feat: built-in per-dataset statistics

shcheklein:
## What & why

Adds **built-in per-dataset statistics** as a first-class, cached, backend-aware feature. Today this is a user-land script (one `group_by` per column → N+1 queries, categorical-only, not cached or surfaced). This makes it native across both warehouse backends (SQLite here; ClickHouse in SaaS via overridable hooks), cached on the dataset version, and exposed in the CLI and Python SDK.

## Design (3 layers + adaptive timing)

- **Compute** — `AbstractWarehouse.compute_dataset_stats()` classifies leaf columns by SQL type and runs a few aggregate passes (batched scalar aggregates; per-column numeric histograms; categorical top-K + distinct). Portable SQLAlchemy both backends inherit; small overridable hooks (`_stats_distinct_count`/`_stats_top_k`/`_stats_numeric_histogram`) let the SaaS ClickHouse warehouse use native `uniq`/`topK`/`histogram`.
- **Storage** — a nullable `stats` JSON column on `_datasets_versions` (auto-migrated). `None` = "not computed yet". `DatasetVersion.stats`; optional in `from_dict` → backward compatible with remote/Studio payloads.
- **Timing (adaptive)** — `Catalog.get_dataset_stats()` computes-and-caches lazily; `stats_compute_is_cheap()` lets cheap backends (ClickHouse) compute **eagerly on save**, while SQLite stays lazy.

## Stats per column (kind-aware)

numeric → min/max/avg/stddev + equal-width histogram · categorical → top-K + (approx) distinct + null count · boolean → true/false counts · temporal → min/max (ISO). Complex/array/JSON/embedding/File-internal columns are recorded in `skipped_columns`.

## Surface

- **CLI**: `datachain dataset stats <name> [--version V] [--force] [--json]`
- **Python**: `dc.read_dataset("x").stats()` and `DatasetVersion.stats`
- **JSON contract** is versioned (`stats_version`) — the integration point for a future Studio UI.

## Tested

`tests/func/test_dataset_stats.py` (17 tests): numeric/categorical/boolean/temporal, nulls, empty dataset, single-value histogram, `max_columns` truncation, `columns` filter, lazy-vs-eager, force recompute, CLI table+JSON, temporal JSON-serializability. ruff + mypy clean.

## Reviewed

Went through an independent multi-angle review pass. Fixes folded into the second commit: temporal ISO coercion (JSON-serializable + cache-consistent), clamp `top_k`/`bins`/`max_columns` ≥1 (bins=0 guard), dropped the unused/untested sampling path (deferred), CLI catches `DatasetVersionNotFoundError`.

## Out of scope / follow-ups

- SaaS: ClickHouse hook overrides + `stats_compute_is_cheap → True`
- Studio: UI rendering the JSON contract
- Sampling for very large tables on slow backends
- Stats as a queryable DataChain

### DIFF
--- a/src/datachain/catalog/catalog.py
+++ b/src/datachain/catalog/catalog.py
@@ -1058,10 +1058,64 @@ class Catalog:
                 preview_rows,
             )
 
+        # Compute distribution stats eagerly on backends where it is cheap
+        # (e.g. ClickHouse).  Slow backends (SQLite) leave this for lazy
+        # computation on first access via ``get_dataset_stats``.
+        if self.warehouse.stats_compute_is_cheap() and dataset_version.stats is None:
+            try:
+                values["stats"] = self.warehouse.compute_dataset_stats(dataset, version)
+            except Exception:
+                # Stats are best-effort and must never break dataset creation.
+                logger.warning(
+                    "Failed to compute stats for %s@%s",
+                    dataset.name,
+                    version,
+                    exc_info=True,
+                )
+
         if not values:
             return
 
         self.metastore.update_dataset_version(dataset, version, **values)
+
+    def get_dataset_stats(
+        self,
+        name: str,
+        version: str | None = None,
+        *,
+        namespace_name: str | None = None,
+        project_name: str | None = None,
+        force: bool = False,
+        **compute_kwargs,
+    ) -> dict:
+        """Return per-column distribution statistics for a dataset version.
+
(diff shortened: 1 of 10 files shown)


shcheklein:
Closing for now, don't like the complexity, I'll get to it a bit later

PR #1817, opened 2026-06-14
dc-knowledge: improve typings and unify prompts/data between CLI and SaaS

dreadatour:
Changes related only to Knowledge Base DataChain skill.

- New `types.py` with `TypedDicts` for the bucket/dataset snapshots the prompts consume, and the scripts are now annotated against them — so the JSON shape and the prompt spec stay in sync in both CLI (DataChain) and SaaS (Studio).
- Dependency-entry building consolidated into `utils.dep_entry`, which now also links each dependency to its own knowledge doc so the lineage graph is followable by link.
- Fetch direct dependencies only (`indirect=False`) and drop the catalog-based script-diffing, which was heavy and rarely useful.
- De-duplicate query scripts across versions so the same script isn't rendered twice, and fix `merge_versions` so a version present in both the file and a refetch no longer shows up twice.
- Prompts now document the warnings field and tell the model to surface completeness caveats without echoing raw internal error strings.
- Added unit tests for the new helpers and the merge fix.


tests/unit/test_skill_knowledge_scripts.py
```diff
@@ -425,6 +471,22 @@ def test_build_changes_script_unchanged():
     assert result["previous_script"] is None
 
 
+# ---------------------------------------------------------------------------
```
shcheklein, line 475: again, weird delimiters, comments, style - please cleanup

amritghimire (approved):
LGTM aside from some cosmetic changes asked by ivan.

tests/unit/test_skill_knowledge_scripts.py
dreadatour: I was following the style of these tests. Don't want to have unrelated "cleanup" changes in this PR.

Let me do the proper cleanup in separate PR on top of this.

dreadatour: See this follow-up PR: https://github.com/datachain-ai/datachain/pull/1821

shcheklein:
@dreadatour what is the status here?

dreadatour:
> @dreadatour what is the status here?

I'll merge this along with Studio PR.

src/datachain/skill/knowledge/collect.py
```diff
@@ -0,0 +1,67 @@
+"""Build a dataset snapshot from any `AbstractMetastore`."""
```
dmpetrov: We do not do file headers. Please remove all if no real need.

src/datachain/skill/knowledge/collect.py
```diff
@@ -0,0 +1,67 @@
...
+    from datachain.dataset import (
+        DatasetDependency,
+        DatasetRecord,
+        DatasetVersion,
+        Project,
+    )
+    from datachain.skill.knowledge.types import DatasetSnapshot, DependencyEntry
+
+
+def collect_dataset_snapshot(
```
dmpetrov, line 18: good idea! it's about the whole file

src/datachain/skill/knowledge/scripts/summary.py
```diff
@@ -451,11 +431,6 @@ def _summarize_list(values):
     }
 
 
-# ---------------------------------------------------------------------------
-# Formatting helpers
-# ---------------------------------------------------------------------------
```
dmpetrov, line 456: thanks for cleanning up all these crappy comments :)

src/datachain/skill/knowledge/scripts/summary.py
```diff
@@ -528,19 +498,13 @@ def _build_overview(total_rows, columns_info, col_results, schema):
     return SEP.join(parts)
 
 
-# ---------------------------------------------------------------------------
-# Main entry points
-# ---------------------------------------------------------------------------
-
-
-def dataset_summary_from_chain(chain) -> dict:
+def dataset_summary_from_chain(chain: "DataChain") -> dict:
```
dmpetrov: do we need a class/TypedDict for the summary?

dmpetrov (approved):
Amazing change - thank you. just a few minor comments!

dmpetrov:
> > @dreadatour what is the status here?
> 
> I'll merge this along with Studio PR.

I'd suggest merging it sooner. It blocks other changes in the skill.

If you think more changes might be required for the Studio part - it's better to create another PR.

dreadatour:
> > > @dreadatour what is the status here?
> 
> > 
> 
> > I'll merge this along with Studio PR.
> 
> 
> 
> I'd suggest merging it sooner. It blocks other changes in the skill.
> 
> 
> 
> If you think more changes might be required for the Studio part - it's better to create another PR.

Deal! Let me review it one more time and I'll merge it 👍

PR #1818, opened 2026-06-15
Fix broken tests after pytest update

dreadatour:
Pytest 9.1 rejects a pattern we use everywhere: a fixture with default `params` that individual tests narrow via an indirect `parametrize` of the same arg. It now fails at collection with `duplicate parametrization` ([pytest #13974(https://github.com/pytest-dev/pytest/issues/13974)).

The defaults move off the fixtures into a small `pytest_generate_tests` hook that applies them only when a test hasn't parametrized that arg itself. The per-test overrides stay `indirect` (a direct override doesn't work with our session-scoped `cloud_type → cloud_server` chains — the value never reaches the dependent fixtures).

Behavior is unchanged — same matrix, same overrides, same `--disable-remotes` skips; only test-ID segment ordering shifts. Collection is clean again and the unit suite is green under 9.1.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -416,7 +416,7 @@ def pytest_configure(config):
     )
 
 
-@pytest.fixture(scope="session", params=[DEFAULT_TREE])
+@pytest.fixture(scope="session")
 def tree(request):
     return request.param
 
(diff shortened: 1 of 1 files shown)


shcheklein (approved):

ilongin:
Merging this to unblock other PRs

PR #1822, opened 2026-06-18
Added unnest

### LINKED ISSUE #1820: Add unnest() to explode a list column into rows
DataChain has no list→rows primitive (like SQL `UNNEST`, Spark `explode`, pandas `.explode()`).

Today you hand-write a generator UDF:

```python
chain.gen(lambda row: ((row.id, x) for x in row.items), output={"id": int, "item": int})
```

which is verbose, drops sibling columns unless you re-`yield` them, and makes you restate the element type even though it's known from the list's declared type.

Note: the existing `explode()` is unrelated — it parses a JSON-object column into columns (a `map`, 1 row → 1 row), not list → rows.

**Proposed:**

```python
chain.unnest("items")   # list[T] -> one row per element (item: T); siblings repeated; element type inferred
```

Motivating case: exploding `content: list[Union[Block, ...]]` from LLM responses into one row per block.

ilongin:
Adds `DataChain.unnest(col)` - expands a list column into rows, one row per element. Sibling columns are repeated. Equivalent to SQL `UNNEST`, Spark `explode`, pandas `.explode()`. Empty lists and `None` drop the source row (inner semantics).

## Naming

The industry-standard name for this is `explode`, but datachain already has an `explode()` that does something different (JSON-string-to-columns). To get to the standard name without breaking users, the plan is:

1. Ship `unnest` now (this PR).
2. In a follow-up, rename the existing `explode` to something descriptive (e.g. `expand_json`) and keep `explode` as a deprecated alias that warns.
3. In a later release, once `explode` is no longer the JSON helper, alias `unnest` as `explode` and deprecate `unnest`.
4. After a deprecation window, drop the old names.

This avoids a single confusing rename and lets users migrate one step at a time.


dmpetrov:
@ilongin it might be a little bit premature implementing this. I expect a lot of unset-challenges with the coming `Union[]` #1824 that's required for proper LLM resposes decoding.

I'd wait for Union first.

Also, as we discussed before, it’s better not to pick up issues that haven’t been prioritized yet - or ask about it (I'm not sure if you had some other discussions about this issue).

amritghimire:
What is the status of this PR? cc. @ilongin

PR #1824, opened 2026-06-18
Union types

### LINKED ISSUE #1812: Union type
### Description

In order to support serialization/deserialization of full LLM responses into DB we need to support Union in types properly:

```
value: Union[str, int]
```

It should also support nullable:

```
value: Union[str, int] | None
```

The union and Optional #1791 need to use similar approach. If needed Optional approach might be changed.

### Version Info

```Text

```

dmpetrov:
Closes #1812
Adds multi-arm `Union[...]` signals — `Union[int, str, float]`, `Union[ModelA, ModelB]`, mixed, and nullable.

Storage reuses the hidden `_type_tag` discriminator: `_type_tag` holds the active arm's name (`"int"`, `"Text"`), `NULL` means `None`. `Optional[Model]` is just the single-arm case.

**Compatibility:** the old numeric `_type_tag` (arm index) is still read, with a `FutureWarning`; its removal is tracked in #1949. Mixing a legacy dataset with a new one in a single query is not supported — the discriminator column type changed `Int64` → `String`. Arms are now serialized in a fixed order, so the first run after upgrading recomputes once instead of hitting its checkpoint.

**Now rejected:** same-shaped union arms inside a `list`/`dict` — one JSON cell has no per-element tag, so `main` silently reads back the wrong arm. Workaround: a distinguishing field (`kind: Literal["..."]`). A per-element tag could lift this later — no issue filed yet.

**Out of scope (follow-up):** collection arms (`Union[str, list[str]]`) — #1952


src/datachain/lib/convert/flatten.py
```diff
@@ -32,49 +37,107 @@ class FlatColumn(NamedTuple):
...
-    """
+def union_value_match(obj, anno) -> bool:
+    layout = union_layout(anno)
+    if layout is None:
+        return False
+    if obj is None:
+        return layout.has_none
+    if any(_arm_matches(obj, arm, exact=False) for arm in layout.arms):
+        return True
+    return not isinstance(obj, BaseModel)
```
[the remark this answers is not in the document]
dmpetrov: Not a bug but fragile - valid outputs flattened fine. Fixed anyway.

src/datachain/lib/convert/flatten.py
```diff
@@ -32,49 +37,107 @@ class FlatColumn(NamedTuple):
...
+            yield from _flatten_fields_values(fr.model_fields, value)
+    else:
+        yield _flatten_scalar(value)
+
+
+def _match_union_arm(value, layout: UnionLayout) -> int | None:
+    """Arm index for ``value`` (None for the None arm). Exact-type match beats
+    ``isinstance`` so ``bool`` isn't swallowed by an ``int`` arm."""
+    if value is None:
+        return None
```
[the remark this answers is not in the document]
dmpetrov: Not a bug on its own - pydantic won't put None in a non-nullable union but related to the previos one. fixed.

src/datachain/lib/dc/datachain.py
```diff
@@ -318,6 +323,12 @@ def column(self, name: str) -> Column:
             name_path = [name]
         for path, type_, _, _ in self.signals_schema.get_flat_tree():
             if path == name_path:
+                if union_layout(type_) is not None:
+                    raise DataChainColumnError(
+                        name,
+                        "a multi-arm Union signal has no single column; reference a "
+                        "specific arm (e.g. C('col.int')) or a scalar key instead",
+                    )
```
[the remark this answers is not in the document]
dmpetrov: Message was improved.

src/datachain/lib/convert/flatten.py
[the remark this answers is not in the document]
dmpetrov: Reverting - this breaks valid Optional[Model] parquet round-trip.

Kept only the real fix: None into a non-nullable union now errors.

src/datachain/func/aggregate.py
```diff
@@ -209,7 +208,7 @@ def max(col: AggColT) -> Func:
     Notes:
         - The `max` function can be used with numeric, date, and string inputs.
         - The result column will have the same type as the input expression.
-        - Skips rows where the value's ``Optional[DataModel]`` parent is absent.
+        - Skips rows where the value is absent (its tagged-union arm is inactive).
```
shcheklein: I think it is impossible to understand for end users: `its tagged-union arm is inactive` 🤯

src/datachain/lib/data_model.py
```diff
@@ -189,7 +260,7 @@ def promote_default_none(model: type[BaseModel]) -> None:
         model.model_rebuild(force=True)
 
 
-def is_chain_type(t: type) -> bool:
+def is_chain_type(t: type) -> bool:  # noqa: PLR0911
```
shcheklein: can we try to avoid PLR0911 ? we are getting more and more of those (exceptions for the whole file second time, here is other place (?) functions are getting complicated). Split, review and refactor?

dmpetrov: fixed

src/datachain/func/aggregate.py
dmpetrov: fixed everywhere

dmpetrov:
@datachain-ai/all any update on this. it's waiting for a while.

CLAUDE.md
```diff
@@ -7,3 +7,8 @@ change here.
 
 Keep general product insights in `AGENT.md` (not here); this file is only for
 Claude-specific notes.
+
+## Rules I repeatedly miss
+
+The AGENT.md section I most often violate: **"Comments and docstrings"** — weight
+it accordingly.
```
dreadatour: Related? 🤔

dreadatour:
> Storage reuses the hidden `_type_tag` discriminator: `_type_tag >= 0` is the active arm's index, `NULL` means `None`. `Optional[Model]` is just the single-arm case.
> 
> **Compatibility:** drops the old `_type_tag > 0` Optional-present flag in favor of this scheme.

Current implementation in `main` branch: https://github.com/datachain-ai/datachain/blob/87246454cbc14584781d9bf2d04f786cea29f00c/src/datachain/lib/data_model.py#L229-L236

This means `0` — model is not `None`, `1` — model is `None`.

And it looks like there is an issue here: `_type_tag == 0` meant that an optional model was present in current implementation. The new reader only recognizes string tags. Every previously stored present `Optional[Model]` value becomes `None` after this PR. Reproduce:

```python
from typing import Optional

class M(DataModel):
    x: int


schema = SignalSchema({"m": Optional[M]})

# Physical representation written before this PR:
# 0 = model present, 7 = M.x
result = schema.row_to_objs((0, 7))

print(result)
```

Actual: `[None]`
Expected: `[M(x=7)]`

<sub>Found this one using LLM</sub>

src/datachain/lib/udf.py
```diff
@@ -335,10 +334,7 @@ def _flatten_row(self, row):
...
             # strict=False as shorter row is allowed for arrow/parquet (guarded above)
             for obj, anno in zip(row, annos.values(), strict=False):
-                # tag is added only when obj IS the Optional value, not a wrapper.
-                if is_optional_model(anno) and (
-                    obj is None or isinstance(obj, classify_field(anno).inner)
-                ):
+                if union_value_match(obj, anno):
                     flat.extend(flatten_value(obj, anno))
                 else:
                     flat.extend(self._obj_to_list(obj))
```
dreadatour: A multi-output UDF can silently reinterpret an unrelated BaseModel as a declared union arm.
This only succeeds when the unrelated model’s flattened fields happen to match the union’s physical column count/order/types. That constraint makes it less common, but the outcome is silent data corruption rather than an error.

Minimal reproducer:
```python
from typing import Union

import datachain as dc


class Foo(DataModel):
    x: int


class Bar(DataModel):
    y: int


class Wrong(DataModel):
    tag: str
    bar_y: int | None
    foo_x: int


def bad(n):
    return n, Wrong(tag="Foo", bar_y=None, foo_x=123)


session = Session.get(in_memory=True)

out = dc.read_values(n=[1], session=session).map(
    bad,
    output={"n2": int, "value": Union[Foo, Bar]},
)

print(out.to_values("value"))
```

Actual: `[Foo(x=123)]`
Expected: `DataChainParamsError: Wrong(...) does not match any arm of Union[Foo, Bar]`

The same `Wrong` value is correctly rejected when the union is the UDF’s only output. The bug is specific to the multi-output path.

Could we distinguish a per-output union value from the wrapper/Parquet case here, reject unrelated models, and add a multi-output regression test?

tests/func/test_union_types.py
```diff
@@ -0,0 +1,666 @@
...
+    left = dc.read_values(
+        id=[1, 2, 3],
+        value=["a", 1, "b"],
+        output={"id": int, "value": Union[str, int]},
+        session=test_session,
+    )
+    right = dc.read_values(
+        id=[1, 2], k=["x", "y"], output={"id": int, "k": str}, session=test_session
+    )
+    merged = left.merge(right, on="id", inner=False)
```
dreadatour, line 532: This test keeps `value` on the preserved left side, so it does not exercise a missing union value. If `value: Union[int, str]` is on the right, this left merge produces `None` for the unmatched row while the schema remains `Union[int, str]`:

```python
left = dc.read_values(id=[1, 2], session=test_session)
right = dc.read_values(
    right_id=[1],
    value=["x"],
    output={"right_id": int, "value": Union[int, str]},
    session=test_session,
)

result = left.merge(
    right, on="id", right_on="right_id", inner=False
).order_by("id")

assert result.to_list("id", "value") == [(1, "x"), (2, None)]
assert result.signals_schema.values["value"] == Union[int, str, None]
```

Currently the second assertion fails because `SignalSchema.merge()` only widens direct scalar types. Could we widen a tagged union to `Union[..., None]` when its merge side is nullable and add this reverse-direction regression test?

src/datachain/lib/signal_schema.py
```diff
@@ -1401,14 +1583,34 @@ def _model_subtree(
...
+        subtree: dict[str, tuple[DataType, dict | None]] = {
+            SignalSchema._TYPE_TAG_FIELD: sentinel
+        }
+        for arm in layout.arms:
+            arm_sub = (
+                SignalSchema._build_tree_for_model(fr)
+                if (fr := ModelStore.to_pydantic(arm)) is not None
+                else None
+            )
+            subtree[arm_selector(arm)] = (arm, arm_sub)
```
dreadatour, line 1800: If a model arm’s stable name is _type_tag, this assignment overwrites the discriminator inserted above:

```python
class _type_tag(DataModel):
    x: int

annotation = Union[_type_tag, str]
schema = SignalSchema({"value": annotation})

assert schema.db_signals() == [
    "value___type_tag",
    "value___type_tag__x",
    "value__str",
]
```

Currently the first column is missing, and ingestion fails later with a misleading type error because the discriminator value is written into `value___type_tag__x`.

Could we reject `_type_tag` as a reserved arm selector during union-layout validation and add a regression test? The validation should use the stable `ModelStore` base name, since that is what `arm_selector()` returns.

src/datachain/lib/data_model.py
```diff
@@ -195,22 +265,43 @@ def is_chain_type(t: type) -> bool:
...
-    args = get_args(t)
-    if orig is list and len(args) == 1:
-        return is_chain_type(get_args(t)[0])
 
+def _is_chain_container_type(t: type) -> bool:
+    """Whether a union / list / dict annotation is a supported DataChain type."""
+    arms, _ = union_arms(t)
+    if len(arms) >= 2:
+        # multi-arm union: supported only as a tagged union (scalar/DataModel arms)
+        return union_layout(t) is not None and all(is_chain_type(arm) for arm in arms)
```
dreadatour, line 415: Returning False here is expected for a collection arm, but on Python 3.10–3.13:

```python
dc.read_values(
    value=[["x"]],
    output={"value": str | list[str]},
    session=test_session,
)
```

raises: `AttributeError: 'types.UnionType' object has no attribute '__name__'`

The equivalent `typing.Union[str, list[str]]` correctly raises `UdfSignatureError`. The difference comes from `_validate_output()` formatting the rejected annotation with `value.__name__`.

Could we format unsupported annotations safely and add both union spellings to a regression test? Collection-arm support can remain out of scope.

src/datachain/lib/signal_schema.py
```diff
@@ -944,6 +967,8 @@ def resolve(self, *names: str) -> "SignalSchema":
         for field in names:
             if not isinstance(field, str):
                 raise SignalResolvingTypeError("select()", field)
+            # readable union-arm path (value.int) -> positional slot (value._0)
```
dreadatour: The final layout is selector-keyed (`value__int` / `value__str`), so this no longer resolves to a positional `value._0` slot, and `_type_tag` stores the selector string rather than an index.
Could we sweep the other stale positive descriptions in `datachain.py:1549`, `signal_schema.py:1191/1670`, `test_union_adversarial.py:55/63/301`, and `test_union_types.py:239/246` plus the _type_tag index comment in the unit suite? Assertions that `_0`/`_1` never appear publicly can remain as regression checks.

Current representation:

```python
from typing import Union

schema = SignalSchema({"value": Union[int, str]})

print(schema.db_signals())
```

Output:

```
[
    "value___type_tag",
    "value__int",
    "value__str",
]
```

There are no `value._0` or `value._1` slots, and `_type_tag` stores `"int"` or `"str"`, not an arm index.

dreadatour (commented):
@dmpetrov sorry for the delay, there are few conflicts with `main` now needs to be solved. Also description update is needed (see first comment).

PR is quite big, I have read though it — looks good to me!

Also ask LLM to check and post all the findings above.

dmpetrov:
> there is an issue here: `_type_tag == 0` meant that an optional model was present in current implementation. The new reader only recognizes string tags. Every previously stored present `Optional[Model]` value becomes `None` after this PR.

Yes, it was discussed in Slack as I remember. So, we need to break the backward-compatability.
The best we can do for now: old `Optional[DataModel]` datasets still read, now behind a warning; removal is tracked in #1949. I'm implementing this.

Caveat: only reading is covered - the column type changed Int64 → String, so mixing a legacy dataset with a new one in one query still fails on CH.

src/datachain/lib/signal_schema.py
dmpetrov: Good catch - it was actually fixed in one place but not the other. 

Fixed: added a validation with an exception.

src/datachain/lib/signal_schema.py
dmpetrov: ha, right! Fixed all the comments with old meaning.

CLAUDE.md
dmpetrov: not really but it's better to keep 🙂 
however, it's still messing up with comments 😅

tests/func/test_union_types.py
dmpetrov: Fixed: SignalSchema._nullable() now widens a tagged union as well, and your reverse-direction case is in as test_merge_widens_union_on_the_padded_side.

src/datachain/lib/udf.py
dmpetrov: fixed. A cover model is now recognised by its fields matching the remaining declared outputs, so a non-arm model raises DataChainParamsError exactly as in the single-output path.

src/datachain/lib/data_model.py
dmpetrov: `str | list[str]` now raises UdfSignatureError like the `typing.Union` spelling

dmpetrov:
@dreadatour thank you for the review. all fixed - please take a look.

dreadatour:
One more finding: a subclass of a declared union arm can be matched to a less specific arm, causing silent data loss.

```python
from typing import Union

import datachain as dc


class Animal(DataModel):
    legs: int = 4


class Zebra(Animal):
    stripes: int = 1


class BabyZebra(Zebra):
    age: int = 0


chain = dc.read_values(
    value=[BabyZebra(stripes=9, age=2)],
    output={"value": Union[Animal, Zebra]},
    session=Session.get(in_memory=True),
)

value = chain.to_values("value")[0]
print(type(value).__name__, value)
```

Actual:

```text
Animal legs=4
```

Expected:

```text
Zebra legs=4 stripes=9
```

`BabyZebra` matches both `Animal` and `Zebra` through `isinstance()`. The fallback currently selects the first matching arm in canonical type-name order, so `Animal` wins and the `stripes` field is silently discarded.

Could we select the most specific matching declared superclass and add a regression test using a subclass of `Zebra`?

dmpetrov:
> One more finding: a subclass of a declared union arm can be matched to a less specific arm, causing silent data loss.

Fixed: the isinstance fallback now picks the narrowest matching arm. Exact-type match still wins first, so bool vs int is unaffected.

dreadatour (approved):
Looks good to me overall as far as I can tell — the diff is too big to be able to review it properly.

One comment before merge — let's, please, cleanup `CLAUDE.md` file? 😀

pyproject.toml
```diff
@@ -340,6 +340,9 @@ max-complexity = 15
 "tests/unit/lib/test_signal_schema.py" = ["UP006", "UP007", "UP035", "UP045"]
 "tests/unit/lib/test_optional.py" = ["UP007", "UP045"]
 "tests/func/test_optional.py" = ["UP007", "UP045"]
+"tests/unit/lib/test_union_types.py" = ["UP007", "UP045"]
```
shcheklein: minor: can we actually fix this instead of keep growing exclusions?

tests/unit/lib/test_union_types.py
```diff
@@ -0,0 +1,328 @@
...
+    assert by_name["value___type_tag"].type.dc_nullable
+
+
+def test_arm_selector():
+    assert arm_selector(int) == "int"
+    assert arm_selector(str) == "str"
+    assert arm_selector(Foo) == "Foo"
+
+
+# ---- flatten / unflatten round-trips ---------------------------------------
```
shcheklein: minor: cleanup style - I don't think we use sections like this (also review comments, better names, etc)

pyproject.toml
dmpetrov: Done - all three entries gone, incl. N806.

tests/unit/lib/test_union_types.py
dmpetrov: Done - banners gone. Comments that restated the test name dropped. test_signal_schema_union_path_edges split into two named tests.

dmpetrov:
Thanks @shcheklein - all fixed. Plus one extra: a separate review pass caught a nested union restoring the wrong arm (the tag was written, then ignored).

I only ran this through Claude Code, no other models. Enough, or would you like a separate Codex pass? Otherwise I'd just merge it.

shcheklein:
@dmpetrov check the Codex output please https://claude.ai/code/artifact/345d4fe5-57b2-47fb-8579-6d5a604327be

dmpetrov:
@shcheklein thank you! Fixed 1-4, thanks. 5 and 6 are in the description now. 7 reproduces on main, leaving it out.

PR #1825, opened 2026-06-19
Video rotation

### LINKED ISSUE #1823: VideoFile.get_np()/get_info() ignore stream rotation
`VideoFile.get_np()` and `get_info()` ignore the stream's display rotation (`DISPLAYMATRIX`). For any rotated video (phones, many dashcams), DataChain returns the raw coded frame, i.e. sideways, and reports `width`/`height` swapped. ffmpeg/OpenCV autorotate; we don't. A model run over these frames silently sees sideways images.

Only affects videos that carry a rotation flag; normal videos with no rotation metadata decode correctly.

### Exact repro file (public domain, CC0)

A single BDD100K dashcam clip with a 90° `DISPLAYMATRIX`. Fetch it (no login):

```bash
curl -sL "https://www.kaggle.com/api/v1/datasets/download/robikscube/driving-video-with-object-tracking/bdd100k_videos_train_00%2Fbdd100k%2Fvideos%2Ftrain%2F0000f77c-6257be58.mov" -o clip.zip
unzip clip.zip   # -> 0000f77c-6257be58.mov   (coded 720x1280, displays 1280x720)
```

```bash
$ ffprobe -v error -select_streams v:0 -show_entries stream=width,height -show_entries side_data=rotation 0000f77c-6257be58.mov
width=720
height=1280
rotation=-90
```

### Repro (verified, run as-is)

```python
import datachain as dc

vf = dc.read_storage("file:///abs/path/to/0000f77c-6257be58.mov", type="video").to_values("file")[0]

info = vf.get_info()
print(info.width, info.height)              # 720 1280       -> swapped; clip displays as 1280x720
print(vf.get_frame(150).get_np().shape)     # (1280, 720, 3) -> sideways (h, w, c)
```

For the same clip OpenCV returns `(720, 1280, 3)` (it applies the rotation).

### Cause

`src/datachain/lib/file.py:1563,1572` call `to_ndarray(format="rgb24")` without applying the display matrix (PyAV doesn't autorotate; ffmpeg/cv2 do). `get_info()` reads coded dims for the same reason.

### Fix

Apply the display-matrix rotation on decode (ffmpeg `autorotate` equivalent).

(Minor: `get_np()` returns RGB; consumers expecting BGR, e.g. Ultralytics, get swapped channels. Worth a docstring note.)

dmpetrov:
Closes #1823

### DIFF
--- a/src/datachain/lib/video.py
+++ b/src/datachain/lib/video.py
@@ -25,6 +25,7 @@ from datachain.lib.file import (
 try:
     import av
     import ffmpeg
+    from av.sidedata.sidedata import Type as _SideDataType
 except ImportError as exc:
     raise ImportError(
         "Missing dependencies for processing video.\n"
@@ -32,14 +33,7 @@ except ImportError as exc:
         "  pip install 'datachain[video]'\n"
     ) from exc
 
-_DISPLAYMATRIX: Any = None
-try:
-    from av.sidedata.sidedata import Type as _SideDataType
-
-    _DISPLAYMATRIX = _SideDataType.DISPLAYMATRIX
-except Exception:  # noqa: BLE001, S110  # PyAV without this enum -> no rotation
-    pass
-
+_DISPLAYMATRIX = _SideDataType.DISPLAYMATRIX
 
 VIDEO_TIME_BASE = 1_000_000
 # DISPLAYMATRIX rotation entries are 16.16 fixed point (FFmpeg libavutil/display.h).
@@ -58,11 +52,12 @@ class _SeekLandedAfterStartError(Exception):
 def _display_matrix_rotation(frame) -> int:
     """
     Clockwise display rotation (0/90/180/270) from a frame's DISPLAYMATRIX,
-    matching FFmpeg ``autorotate``/OpenCV (PyAV does not autorotate). Mirrored
-    matrices reduce to their nearest rotation; returns 0 when absent.
+    matching what FFmpeg/OpenCV apply on playback (PyAV does not autorotate).
+
+    Only right-angle rotation is applied: scale is normalized out and any
+    flipped or off-axis matrix is snapped to the nearest 0/90/180/270, as
+    OpenCV does. Returns 0 when there is no rotation metadata.
     """
-    if _DISPLAYMATRIX is None:
-        return 0
     try:
         side_data = frame.side_data.get(_DISPLAYMATRIX)
         if side_data is None:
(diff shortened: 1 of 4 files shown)


src/datachain/lib/video.py
```diff
@@ -31,8 +32,20 @@
         "  pip install 'datachain[video]'\n"
     ) from exc
 
+# Resolved lazily so a PyAV without this enum degrades to "no rotation".
+_DISPLAYMATRIX: Any = None
+try:
+    from av.sidedata.sidedata import Type as _SideDataType
+
+    _DISPLAYMATRIX = _SideDataType.DISPLAYMATRIX
+except Exception:  # noqa: BLE001, S110  # older/newer PyAV without this enum
```
shcheklein: let's not support them? and simplify the whole import

src/datachain/lib/video.py
```diff
@@ -73,6 +152,10 @@ def video_info(file: File | VideoFile, video_stream_index: int = 0) -> Video:
 
                 width = int(video_stream.width or 0)
                 height = int(video_stream.height or 0)
+                # Report display dimensions: a 90/270 rotation swaps them, just
+                # as the decoded frames returned by get_np() are rotated.
+                if _video_stream_rotation(container, video_stream) in (90, 270):
```
shcheklein, line 138: do we need to handle other cases? (scaling? some non 90 rotations?) asking since I see above it the whole matrix with some weird stuff happening

shcheklein: is it rotation per frame or per stream btw?

src/datachain/lib/video.py
```diff
@@ -44,6 +57,72 @@ class _SeekLandedAfterStartError(Exception):
     pass
 
 
+def _display_matrix_rotation(frame) -> int:
+    """
+    Clockwise display rotation from a frame's DISPLAYMATRIX side data, snapped
+    to ``0``/``90``/``180``/``270`` (matching FFmpeg ``autorotate`` and OpenCV).
+
+    PyAV decodes the coded orientation without autorotating, so callers apply
+    this. Like OpenCV, only the rotation component is used (mirrored matrices
```
shcheklein: what are the mirrored matrices?

src/datachain/lib/video.py
dmpetrov: done

src/datachain/lib/video.py
dmpetrov: 90 / not-90 -No need,  scaling is normalized out. Non-right angle is converted to the nearest 0/90/180/270. That matches FFmpeg autorotate and OpenCV.

Per stream - ffmpeg copies it onto every frame.

src/datachain/lib/video.py
dmpetrov: simplified comment

shcheklein (approved):

PR #1827, opened 2026-06-19
Use logical names in cast complex object errors

### LINKED ISSUE #1826: func.cast() error shows internal db-name instead of readable column name
**Bug.** `func.cast()` over a complex-object column reports the internal `__`-joined db-name, not the user's dotted name.

```py
func.cast("item.Foo", str)
# DataChainParamsError: Function cast doesn't support complex object columns like 'item__Foo'.
# Use a leaf field (e.g., 'item__Foo.path') ...
```

Should read `item.Foo` / `item.Foo.path`. Source: `_validate_sql_func_columns`, `src/datachain/func/func.py`.

Extra step:
- [ ] check other functions when `__` is used in user facing messages or exceptions

gaoflow:
Fixes #1826.

## Summary
- Render complex-object column names with the user-facing dotted form in `func.cast()` errors.
- Keep the schema lookup and SQL path unchanged; only the error message changes.
- Add regression coverage for `item.Foo` so the message suggests `item.Foo.path` instead of the internal `item__Foo.path`.

## Tests
- `.venv/bin/python -m pytest tests/func/functions/test_cast.py -q`
- `.venv/bin/python -m pytest tests/unit/test_func.py -q`
- `.venv/bin/python -m pytest tests/func/functions/test_aggregate.py -k complex_object -q`
- `.venv/bin/python -m pytest tests/func/functions/test_conditional.py -k complex_object -q`
- `.venv/bin/python -m ruff format --check src/datachain/func/func.py tests/func/functions/test_cast.py`
- `.venv/bin/python -m ruff check src/datachain/func/func.py tests/func/functions/test_cast.py`
- `git diff --check`

AI assistance was used under my direction.

### DIFF
--- a/src/datachain/func/func.py
+++ b/src/datachain/func/func.py
@@ -14,7 +14,7 @@ from datachain.lib.convert.sql_to_python import sql_to_python
 from datachain.lib.data_model import NULLABLE_SCALARS, unwrap_optional
 from datachain.lib.model_store import ModelStore
 from datachain.lib.utils import DataChainColumnError, DataChainParamsError
-from datachain.query.schema import Column, ColumnExpr, ColumnMeta
+from datachain.query.schema import DEFAULT_DELIMITER, Column, ColumnExpr, ColumnMeta
 from datachain.sql.functions import numeric
 from datachain.sql.functions.conversion import datetime_to_string
 from datachain.sql.types import SQLType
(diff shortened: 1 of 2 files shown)


dreadatour (approved):
Looks good to me 👍

I wonder if we do have helper method to do not use this everywhere:

```python
user_col = arg.replace(DEFAULT_DELIMITER, ".")
```

PR #1830, opened 2026-06-20
Reflect only the requested table in SQLiteWarehouse.get_table

dreadatour:
Fix for something I hit while investigating #1829. Small change, but worth applying IMO.

---

`SQLiteWarehouse.get_table` re-reflected the whole schema on every call — `metadata.reflect(only=...)` introspects every table even with a name filter. So it was O(number of tables), and `get_table` runs a few times per query. Now it reflects just the one table, so cost no longer scales with how many tables are in the DB.

## Performance

Local in-memory SQLite — absolute numbers are machine-specific; what matters is **linear → flat**.

Per-query time in a `read_values().map().sum()` loop as tables pile up (datasets / temp / UDF):

| tables | before | after |
|--------|--------|-------|
| 40     | 58 ms  | 28 ms |
| 115    | 109 ms | 37 ms |
| 215    | 217 ms | 50 ms |
| 315    | 284 ms | 57 ms |
| 365    | 370 ms | 59 ms |

Before climbs with table count (~6× at 365 and rising); after stays flat-ish.

Same refresh behavior (`extend_existing`), missing tables still raise `TableMissingError`. On a small DB the difference is negligible — it only shows up once tables accumulate.

🤖 Made with the help of Claude Code

### DIFF
--- a/src/datachain/data_storage/sqlite.py
+++ b/src/datachain/data_storage/sqlite.py
@@ -767,20 +767,6 @@ class SQLiteWarehouse(AbstractWarehouse):
         (db_class, db_args, db_kwargs) = db_clone_params
         return cls(db=db_class(*db_args, **db_kwargs))
 
-    def _reflect_tables(self, filter_tables=None):
-        """
-        Since some tables are prone to schema extension, meaning we can add
-        additional columns to it, we should reflect changes in metadata
-        to have the latest columns when dealing with those tables.
-        If filter function is defined, it's used to filter out tables to reflect,
-        otherwise all tables are reflected
-        """
-        self.db.metadata.reflect(
-            bind=self.db.engine,
-            extend_existing=True,
-            only=filter_tables,
-        )
-
     def create_dataset_rows_table(
         self,
         name: str,
@@ -847,11 +833,14 @@ class SQLiteWarehouse(AbstractWarehouse):
         return cast(func.instr(source, target), sqlalchemy.Boolean)
 
     def get_table(self, name: str) -> sqlalchemy.Table:
-        # load table with latest schema to metadata
-        self._reflect_tables(filter_tables=lambda t, _: t == name)
         try:
-            return self.db.metadata.tables[name]
-        except KeyError:
+            return Table(
+                name,
+                self.db.metadata,
+                autoload_with=self.db.engine,
+                extend_existing=True,
+            )
+        except sqlalchemy.exc.NoSuchTableError:
             raise TableMissingError(f"Table '{name}' not found") from None
 
     def python_type(self, col_type: Union["TypeEngine", "SQLType"]) -> Any:


src/datachain/data_storage/sqlite.py
```diff
@@ -847,11 +833,14 @@ def instr(self, source, target) -> "ColumnElement":
         return cast(func.instr(source, target), sqlalchemy.Boolean)
 
     def get_table(self, name: str) -> sqlalchemy.Table:
-        # load table with latest schema to metadata
-        self._reflect_tables(filter_tables=lambda t, _: t == name)
         try:
-            return self.db.metadata.tables[name]
-        except KeyError:
+            return Table(
```
shcheklein, line 837: aren't we trading in some cases now one single call vs many calls happening a lot?

dreadatour: Before:

```python
    def get_table(self, name: str) -> sqlalchemy.Table:
        # load table with latest schema to metadata
        self._reflect_tables(filter_tables=lambda t, _: t == name)
        try:
            return self.db.metadata.tables[name]
        except KeyError:
            raise TableMissingError(f"Table '{name}' not found") from None
```

We do call `self._reflect_tables(filter_tables=lambda t, _: t == name)` on every `get_table` call.

In `self._reflect_tables` we do call `metadata.reflect(only=…)`, which does every table list first and then filtering out results using `only` callable (our `filter_tables=lambda t, _: t == name`).

Source code for `metadata.reflect` is here: https://github.com/sqlalchemy/sqlalchemy/blob/rel_2_0_51/lib/sqlalchemy/sql/schema.py#L5721

We needed to call `_reflect_tables` because new table may appears.

After:

```python
    def get_table(self, name: str) -> sqlalchemy.Table:
        try:
            return Table( 
                name,
                self.db.metadata,
                autoload_with=self.db.engine,
                extend_existing=True,
            )
        except sqlalchemy.exc.NoSuchTableError:
            raise TableMissingError(f"Table '{name}' not found") from None
```

No listing of all tables happens, everything else stays the same.

dreadatour: Good question btw!

shcheklein: I means exactly SQL calls

Can be one large vs many. Or can be that now we hit it w/o caching, etc, etc. I don't know exactly, but it is worth comparing actual logs.

can we confirm the sequence in the first and the second case (I feel this "populate whole cache once" was done at some point quite deliberately - it is not random. You can can even ask AI to fully check dvcx repo history for it).

dreadatour: As far as I understand it we are making one less SQL call (list tables), but let me double check to be on a safe side 👍

shcheklein: it was list tables, but then it was using cache, no? and I wonder if for some at least tables initially it was not attempting to re-reflect them ... not sure about now

what we probably want in general is to have cache and for read-only tables (e.g. older datasets? etc) - keep using cache - no reflection. May be not even in get_table itself, but in some other places (?)

dreadatour: It wasn't using cache, it listed all the tables every time:
- https://github.com/sqlalchemy/sqlalchemy/blob/rel_2_0_51/lib/sqlalchemy/sql/schema.py#L5838
- https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/dialects/sqlite/base.py#L2310-L2313

I don't think we need cache to be honest, since we don't really know when new table will be added to the db. Also it is fast operation to get table meta by table name — should we try to optimize it?

dreadatour: New table added or removed to be more specific.

shcheklein: hmm .. if we don't use cache it means each get_table is a SQL query? and we do like 1000s of them ... I'm a bit surprised tbh, i thought `return self.db.metadata.tables[name]` serves as a cache pretty much

shcheklein: ant that's why we pass `self.db.metadata` everywhere, no?

dreadatour: Please, take a look at original function one more time:

```python
    def get_table(self, name: str) -> sqlalchemy.Table:
        # load table with latest schema to metadata
        self._reflect_tables(filter_tables=lambda t, _: t == name)
        try:
            return self.db.metadata.tables[name]
        except KeyError:
            raise TableMissingError(f"Table '{name}' not found") from None
```

Before `return self.db.metadata.tables[name]` it always calls `self._reflect_tables(filter_tables=lambda t, _: t == name)` and `self._reflect_tables` always calls `metadata.reflect(only=…)` and it always selects all the tables. It depends on DB implementation, but for SQLite it is so (see [here](https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/dialects/sqlite/base.py#L2307-L2314)).

shcheklein: > Please, take a look at original function one more time:

it is not so much about original function ... I more interested in the original history / intention here. Someone might have broken it along the way.

also, how about CH for example?

ilongin: I took a closer look on this.

**SQL call count.** I counted queries on both paths. The diff is exactly 2 queries per `get_table` call: the new code skips the `SELECT name FROM sqlite_master` + `SELECT name FROM sqlite_temp_master` upfront. The remaining 12 PRAGMA queries (`table_xinfo`, `foreign_key_list`, `index_list`, `table_info`, plus a few `SELECT sql FROM sqlite_master`) are byte-identical before and after. So it's strictly fewer calls, not more. Not a "one big vs many small" trade-off.

**CH parity.** The CH warehouse already uses this exact pattern (`Table(name, metadata, autoload_with=engine, extend_existing=True)`), no listing step. This PR just brings SQLite in line with it.

**On the benchmark.** Side note - I tried to reproduce the per-call cost of the old `get_table` in isolation and got ~1ms at 365 tables, not 370ms. So the headline numbers (58 -> 370ms) are full `read_values().map().sum()` wall-clock, not `get_table` alone. This PR shaves a real but small piece of that (skipping the 2 list queries). Most of the linear growth in the benchmark must be coming from somewhere else - probably worth a follow-up to track it down, since the bulk of the win is still on the table.

LGTM either way - the change itself is clean and strictly better.

ilongin (approved):

src/datachain/data_storage/sqlite.py
dreadatour: Thank you for the review and for the measurements! 🙏

shcheklein (approved):

shcheklein:
Following up on the benchmark discussion above — I think this PR may be fixing **the whole O(number of tables) cost**, not just shaving ~2 list queries per call. Worth confirming independently, but here's what local profiling (in-memory/file SQLite, SQLAlchemy 2.0.49) shows:

The expensive case is `get_table` on a table that **doesn't exist yet**. `_reflect_tables` calls `metadata.reflect(only=lambda t, _: t == name)`; when `name` matches nothing in the DB, the `only` filter is empty and SQLAlchemy falls back to reflecting **every** table in the DB. Same root cause as the ClickHouse-side fix in iterative/studio#12803 (`has_table` guard) — SQLite never got that guard.

Single `get_table` call, queries vs. number of tables in the DB:

| DB tables | `get_table(`existing`)` | `get_table(`missing`)` |
|-----------|------------------------|-----------------------|
| 51        | 12                     | 512                   |
| 366       | 12                     | 3,662                 |
| 1001      | 12                     | 10,012                |

In a `read_values().map().sum()` run, **both** `get_table` calls are on not-yet-created UDF output tables (`exists_at_call=False`), so each fans out across the whole DB. Whole-query SQL and wall time (365 padding tables):

| | total SQL / query | wall / query |
|---|---|---|
| before (this PR's base) | 9,200 | ~221 ms |
| after (this PR) | **8** | ~34 ms |

And it's **flat** with table count after the change (8 SQL at 10/100/365 tables), vs. linear before. 100% of the before-queries are issued inside the two `_reflect_tables` calls.

This reconciles the earlier numbers: the ~1 ms / "2 fewer queries" measurement is the *existing-table* path (flat 12), while the 58→370 ms benchmark is the *missing-table* fan-out. So the linear growth wasn't elsewhere — it's exactly this call, and switching to `Table(name, ..., autoload_with=engine)` (which raises `NoSuchTableError` immediately on a missing table) removes it.

**Needs confirming** before we treat it as the full fix: (a) someone re-running the profiling independently, and (b) checking the CI SQLAlchemy version behaves the same. If it holds, this PR closes the O(#tables) concern rather than leaving "the bulk on the table."

<sub>🤖 numbers from a Claude Code profiling run</sub>

dreadatour:
> **Needs confirming** before we treat it as the full fix: (a) someone re-running the profiling independently, and (b) checking the CI SQLAlchemy version behaves the same. If it holds, this PR closes the O(#tables) concern rather than leaving "the bulk on the table."

Re-ran the profiling independently on my machine — confirms once again.

**SQLAlchemy version:** CI installs 2.0.51 (checked the CI job logs; it's also the latest on PyPI). All numbers below are on 2.0.51, so same behavior as your 2.0.49 run.

**Profiling re-run** (main @ 57f6f6ae vs this PR):

Single `get_table` call, SQL queries:

| DB tables | main, existing | main, missing | PR, existing | PR, missing |
|-----------|---------------|---------------|--------------|-------------|
| 51        | 14            | 614           | 12           | 4           |
| 366       | 14            | 4,394         | 12           | 4           |
| 1001      | 14            | 12,014        | 12           | 4           |

Same shape as yours (per-table constant differs a bit — depends on column count). The missing-table case really does reflect the whole DB, and the existing-table path is flat — which is why the earlier ~1 ms measurement looked fine.

`read_values().map().sum()`, avg of 5 runs:

| pad tables | main    | PR     | main SQL/query | PR SQL/query |
|------------|---------|--------|----------------|--------------|
| 10         | 69 ms   | 36 ms  | 920            | 8            |
| 100        | 129 ms  | 40 ms  | 3,080          | 8            |
| 365        | 344 ms  | 71 ms  | 9,440          | 8            |
| 1000       | 1,134 ms| 127 ms | 24,680         | 8            |

Flat 8 SQL/query at every table count — matches your number exactly.

One nuance: PR wall time still grows a little (~0.09 ms/table vs ~1.1 before). SQL count is constant, so it's not reflection — most likely SQLite re-parsing its schema after each query's DDL (temp/UDF table creates bump the schema cookie). C-level, ~12× cheaper, negligible at realistic table counts.

🤖 numbers from a Claude Code profiling run

PR #1832, opened 2026-06-22
Avoid waking the warehouse on metadata-only dataset edits

amritghimire:
edit_dataset (and update_dataset under it) always called warehouse.rename_dataset_tables, even when only the description or attrs changed. A dataset's table name is built solely from its namespace, project, name and version, so metadata-only edits can never trigger a rename - but the call still reached into the warehouse, which for remote/Studio backends means waking up compute we don't need.

We now only rename the tables when a field that actually feeds the table name changes (name or project_id), so a pure metadata update no longer touches the warehouse at all. The move path keeps working since it passes project_id, and rename_dataset_tables still no-ops when the name is unchanged.

### DIFF
--- a/src/datachain/catalog/catalog.py
+++ b/src/datachain/catalog/catalog.py
@@ -1095,7 +1095,11 @@ class Catalog:
     def update_dataset(self, dataset: DatasetRecord, **kwargs) -> DatasetRecord:
         """Updates dataset fields."""
         dataset_updated = self.metastore.update_dataset(dataset, **kwargs)
-        self.warehouse.rename_dataset_tables(dataset, dataset_updated)
+        # Renaming dataset tables touches (and may wake) the warehouse, so only
+        # do it when a field that feeds the table name changes. See
+        # AbstractWarehouse._construct_dataset_table_name for those fields.
+        if kwargs.keys() & {"name", "project_id"}:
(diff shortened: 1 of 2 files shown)


src/datachain/catalog/catalog.py
```diff
@@ -1095,7 +1095,11 @@ def complete_dataset_version(
     def update_dataset(self, dataset: DatasetRecord, **kwargs) -> DatasetRecord:
         """Updates dataset fields."""
         dataset_updated = self.metastore.update_dataset(dataset, **kwargs)
-        self.warehouse.rename_dataset_tables(dataset, dataset_updated)
+        # Renaming dataset tables touches (and may wake) the warehouse, so only
+        # do it when a field that feeds the table name changes. See
+        # AbstractWarehouse._construct_dataset_table_name for those fields.
+        if kwargs.keys() & {"name", "project_id"}:
+            self.warehouse.rename_dataset_tables(dataset, dataset_updated)
```
ilongin, line 1102: I think it's better to move that decision of whether to update table name or not into warehouse method itself. This way we are leaking internal logic outside to Catalog.

amritghimire: I am not so sure, I mean, here we were calling the warehouse to rename the dataset tables, even when it was not needed.

shcheklein (approved):

PR #1841, opened 2026-06-29
Fixing list duplication

### LINKED ISSUE #1836: Duplications in listing
### Description

Duplications in output.
```bash
$ datachain ls s3://dc-readme/
fiona.jpg
fleet-cameras/
models/
oxford-pets-micro/
oxford-pets-small/
s3://dc-readme/:
fiona.jpg
fleet-cameras/
models/
oxford-pets-micro/
oxford-pets-small/
```

### Version Info

```Text
0.57.4.dev6+g897a87629.d20260623
Python 3.12.11
```

ilongin:
Fixes #1836.

Before, `datachain ls` and `datachain dataset ls` would list both local and Studio when you were logged in, which caused duplicated output.

Now:
- No flag (default): local only.
- `--studio`: Studio only.
- `--local`: local only.
- `--all`: both, with `Local:` and `Studio:` headers between sections.

Tested locally:

```
$ datachain ls s3://dc-readme/
fiona.jpg
fleet-cameras/
models/
oxford-pets-micro/
oxford-pets-small/
```

```
$ datachain ls --all s3://dc-readme/
Local:
fiona.jpg
fleet-cameras/
models/
oxford-pets-micro/
oxford-pets-small/

Studio:
s3://dc-readme/:
  ...
```

```
$ datachain dataset ls
Name                               Latest Version
---------------------------------  ----------------
local.local.nums                   v1.0.8
local.local.nums_doubled           v1.0.3
...
```

```
$ datachain dataset ls --all
Name                               Studio    Local
---------------------------------  --------  -------
@ilongin.default.pagination_test   v1.0.1    ✖
local.local.nums                   ✖         v1.0.8
local.local.nums_doubled           ✖         v1.0.3
...
```

### DIFF
--- a/src/datachain/cli/commands/ls.py
+++ b/src/datachain/cli/commands/ls.py
@@ -24,7 +24,7 @@ def ls(
     all, local, studio = determine_flavors(studio, local, all, token)
 
     show_local = all or local
-    show_studio = (all or studio) and token
+    show_studio = (all or studio) and bool(token)
     label_sections = show_local and show_studio
 
     if show_local:
--- a/src/datachain/cli/commands/ls.py
+++ b/src/datachain/cli/commands/ls.py
@@ -24,7 +24,7 @@ def ls(
     all, local, studio = determine_flavors(studio, local, all, token)
 
     show_local = all or local
-    show_studio = (all or studio) and bool(token)
+    show_studio = all or studio
     label_sections = show_local and show_studio
 
     if show_local:
--- a/src/datachain/cli/commands/ls.py
+++ b/src/datachain/cli/commands/ls.py
@@ -16,17 +16,25 @@ def ls(
     long: bool = False,
     studio: bool = False,
     local: bool = False,
-    all: bool = True,
+    all: bool = False,
     team: str | None = None,
     **kwargs,
 ):
     token = Config().read().get("studio", {}).get("token")
     all, local, studio = determine_flavors(studio, local, all, token)
 
-    if all or local:
+    show_local = all or local
+    show_studio = (all or studio) and token
+    label_sections = show_local and show_studio
+
+    if show_local:
+        if label_sections:
+            print("Local:")
         ls_local(sources, long=long, **kwargs)
 
-    if (all or studio) and token:
+    if show_studio:
+        if label_sections:
+            print("\nStudio:")
         ls_remote(sources, long=long, team=team)
 
 
--- a/src/datachain/cli/parser/__init__.py
+++ b/src/datachain/cli/parser/__init__.py
@@ -249,14 +249,14 @@ def get_parser() -> ArgumentParser:  # noqa: PLR0915
         "--studio",
         action="store_true",
         default=False,
-        help="List datasets from Studio only",
+        help="List the datasets in the Studio",
     )
     datasets_ls_parser.add_argument(
         "-L",
         "--local",
         action="store_true",
         default=False,
-        help="List local datasets only (default)",
+        help="List local datasets only",
     )
     datasets_ls_parser.add_argument(
         "-a",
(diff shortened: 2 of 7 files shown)


src/datachain/cli/parser/__init__.py
```diff
@@ -249,21 +249,21 @@ def get_parser() -> ArgumentParser:  # noqa: PLR0915
...
-        help="List the files in the Studio",
+        help="List the datasets in the Studio",
     )
     datasets_ls_parser.add_argument(
         "-L",
         "--local",
         action="store_true",
         default=False,
-        help="List local files only",
+        help="List local datasets only",
```
dreadatour: Should we add this will be the default?

src/datachain/cli/commands/ls.py
```diff
@@ -16,17 +16,25 @@ def ls(
...
+    all: bool = False,
     team: str | None = None,
     **kwargs,
 ):
     token = Config().read().get("studio", {}).get("token")
     all, local, studio = determine_flavors(studio, local, all, token)
 
-    if all or local:
+    show_local = all or local
+    show_studio = (all or studio) and token
```
dreadatour: What if `--studio` flag will be set and no token found/set? Will result be empty? Any warnings or message in this case?

src/datachain/cli/parser/__init__.py
ilongin: Yes, adding it

src/datachain/cli/commands/ls.py
ilongin: No silent empty - `determine_flavors` raises `DataChainError("Not logged in to Studio. Log in with 'datachain auth login'.")` when `--studio` is passed without a token. That existed before this PR and still triggers.

src/datachain/cli/commands/ls.py
```diff
@@ -16,17 +16,25 @@ def ls(
...
+    all: bool = False,
     team: str | None = None,
     **kwargs,
 ):
     token = Config().read().get("studio", {}).get("token")
     all, local, studio = determine_flavors(studio, local, all, token)
 
-    if all or local:
+    show_local = all or local
+    show_studio = (all or studio) and bool(token)
```
dreadatour: Wanted to get back to this one more time. Just wonder, if `determine_flavors` will raise an exception in case `all` or `studio` is set and `token` is not set as it was mentioned in [this comment](https://github.com/datachain-ai/datachain/pull/1841#discussion_r3495281964), why do we need to check `and token` here? Because from the code below if both `show_local` and `show_studio` are `False`, then it will print nothing and logic in this code here looks a little bit odd.

src/datachain/cli/parser/__init__.py
```diff
@@ -249,21 +249,21 @@ def get_parser() -> ArgumentParser:  # noqa: PLR0915
         "--studio",
         action="store_true",
         default=False,
-        help="List the files in the Studio",
+        help="List the datasets in the Studio",
```
dreadatour, line 252: Add "only"? Or remove "only" from `--local` for consistency? Not sure, to be honest, but looks a little bit off.

src/datachain/cli/utils.py
```diff
@@ -80,6 +80,7 @@ def determine_flavors(studio: bool, local: bool, all: bool, token: str | None):
     if local or studio:
         all = False
```
dreadatour, line 81: Does this means if I'll use `--local --studio` params this will not be the same as `--all`? 🤔

dreadatour (approved):
Looks good to me overall, couple comments I'd like to address.

src/datachain/cli/commands/ls.py
ilongin: Pushed the token check into `determine_flavors` - on `--all` without a token it now returns `all=False, local=True` (silent downgrade). Callers just do `show_studio = all or studio` now. Same behavior as before but I guess code is cleaner

src/datachain/cli/parser/__init__.py
ilongin: Dropped "only" from `--local` on both `dataset ls` and `ls`. The "(default)" already covers it and it's consistent with `--studio` now.

src/datachain/cli/utils.py
ilongin: Output to the user is identical - both show the same "Local:" / "Studio:" sections with the same content. Internally `--all` returns `all=True`, `--local --studio` returns `local=True, studio=True`, but callers do `all or local` / `all or studio` so the branches taken are the same.

PR #1844, opened 2026-06-30
dc.llm - LLM operations

### LINKED ISSUE #1840: Add dc.llm - LLM operations over files
### Description

Add a small LLM library `dc.llm` (parallel to `dc.func`) for model calls over files. Sugar over existing verbs, fully operational (any model).

Functions (used inside map/gen/agg by cardinality):
- `llm.complete(col, prompt=, schema=) -> str | Pydantic model`
- `llm.classify(col, into=[...]) -> str`
- `llm.score(col, prompt=) -> float`
- `llm.embed(col) -> list[float]`

No fused `if`: materialize a column instead, then plain filter.

Components:
- [ ] dc.llm module + 4 functions; each carries its output type so map/gen/agg lift it (no `output=`, no
  double schema)
- [ ] Model select via `.settings(llm="provider/model")`, inherited downstream; per-call `model=` override; `DATACHAIN_AI_MODEL` env fallback
- [ ] Routing via LiteLLM (provider-prefix string; no Router, no config files, models always external)
- [ ] Provider params via `.settings(llm_params=dict|callable)`; callable resolved per-worker (no key in payload); default env/IAM
- [ ] `retries=`, `fallback=` mapped to LiteLLM num_retries/fallbacks
- [ ] schema= -> LiteLLM response_format (Pydantic) -> validate -> typed column; validation failure triggers retry
- [ ] async + batch-aware execution (ride AsyncMapper + UDF batching)
- [ ] Model identity (provider/model/version/params/prompt) into dataset fingerprint + Sense cache key
- [ ] Images: encode file to data-URI. Video: get_frames() in gen, then llm.* per frame
- [ ] Keep `.setup()` + `.map()` BYO-client path working
- [ ] Ship as optional extra `datachain[llm] = ["litellm"]`; lazy-import litellm in dc.llm with a clear install hint; core and `.setup()` BYO-client path stay dependency-free

dmpetrov:
Closes #1840

It's mostly AI generated but I put a lot of effort to make it nice and stable - see commit history. The remaining open point is decision (1) related to types: it's hard to make "file.path" acceptable in the llm function because we cannot resolve it in a function definition like `def fn(file.path p) -> int:`.

```python
from datachain import llm

class Scene(BaseModel):
    objects: list[str]
    risk: float

(
    dc.read_storage("s3://frames", type="image")
      .settings(llm="anthropic/claude-haiku-4-5")          # set once, inherited
      .map(label=llm.classify("file", into=["accident", "ok"]))  # -> str
      .map(risk=llm.score("file", "accident risk 0..1"))         # -> float
      .map(scene=llm.complete("file", schema=Scene))             # -> Scene
      .gen(chunk=llm.complete("file", schema=list[Chunk]))       # 1:N -> one row per item
      .map(vec=llm.embed("caption"))                             # -> list[float]
)

# opt-in token usage (native multi-output, no wrapper)
.map(llm.complete("file", schema=Scene, include_usage=True),
     output={"res": Scene, "tok": dc.llm.Usage})
```

Two ways to set the model (resolution order: per-call `llm=` › `.settings(llm=)`):
```python
  .settings(llm="anthropic/claude-haiku-4-5")                 # set once, inherited by all llm.* below
  .map(x=llm.complete("file", schema=Scene, llm="openai/gpt-5-mini"))  # per-call override
```

Forward params to the model call (LiteLLM) per chain:
```python
    .settings(
        llm="openai/gpt-5-mini",
        llm_params={               # forwarded to the model call
           "temperature": 0,
           "max_tokens": 512,
           "api_base": "https://...",
           "api_key": "sk-...",     # credential keys are stripped from the cache key
        },
    )
```
(`type=`, `retries=`, `fallback=` are verb kwargs, not `llm_params`.)

Verbs complete / classify / score / embed; shared kwargs `prompt`, `context=`, `type=("text"|"image"|"document")`, `llm=`, `retries=`, `fallback=`, `include_usage=`, `**params` (classify also `into=`).

Design decisions
1. Binding & typing (core touch — review here). A verb gets an `LLMSpec`, not a finished UDF, because the model/credentials live on `.settings()` and aren't known when `llm.complete(...)` is constructed. So `_udf_to_obj` resolves any `BoundSpec` (a small core ABC) by calling `spec.bind(BindContext(settings, target))` the moment it's attached to a verb, baking the resolved model into the cache key. (`BoundSpec`/`BindContext` replaced an earlier duck-typed `__datachain_bind__` hook — a typed, extensible seam for any lib that needs settings/target to build a UDF.) The bound closure carries its output type as a normal return annotation; input columns travel via `__datachain_params__` because they can be dotted (`"file.path"`), which can't be Python parameters. The target verb decides cardinality/shape: `.map()` 1:1, `.gen()` 1:N, and `list[Model]` becomes either one `list[Model]` column (`.map()`) or fanned-out rows (`.gen()`).
2. LiteLLM routing — any provider/model string; optional dep `datachain[llm]`. Model = per-call `llm=` › `.settings(llm=)`.
3. Inputs by column type — text/image auto-detected from the File type; `type=` (text/image/document) overrides the modality for raw bytes / untyped `File` (PDF → document-capable model). Named to match `read_storage`'s `type=`; distinct from an object's `content_type` (MIME).
4. Structured output — `response_format` + our own schema-validation reask (feeds the bad output back); a single `retries` budget that doesn't multiply with LiteLLM's, and non-transient errors (bad auth/request) fail fast instead of burning it.
5. Caching — invalidates on any output-affecting input (model, prompt, schema fields, column, `type`, params). Reliability knobs (`retries`, `fallback`) are excluded, so tuning them resumes from the checkpoint instead of re-running every row. A dict `llm_params` is part of the key; a callable (per-worker secrets) is not, and well-known credential keys are stripped so secrets never enter the UDF hash.
6. No fused predicate — materialize an `llm.*` column, then `.filter()` it (cheap recall, not a re-call).


dmpetrov:
> The biggest concern I have is decision (1) related to types. But I'm not sure how to implement it better

Heads up. I reworked the commit datachain-ai/datachain@3edf3f8: the llm.*→UDF binding from a duck-typed `__datachain_bind__` hook (via getattr), which read like a hack. Switched it to a proper OOP-style BoundSpec ABC + BindContext in lib/udf.py (they way udfs are designed). It fits much better to our class-based Mapper/Aggregator design.

The major benefit - it defines a clear path to pass information to build UDFs. So, if other lib require something beyond Settings to build a proper UDF - it's streight forward how to implement without introducing another hack like `__datachain_bind__`.

If no objections, I'll keep it.

docs/references/llm.md
```diff
@@ -0,0 +1,156 @@
...
+```python
+chain.agg(in_tok=dc.func.sum("tok.input_tokens"))
+chain.filter("tok.retries > 0")
+```
+
+Notes:
+
+- `retries` is the count of `datachain.llm`'s schema-validation reasks (it feeds
+  the failed output back and asks again), not LiteLLM's transient retries; only
+  structured output ever exceeds 0.
```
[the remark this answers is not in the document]
dmpetrov: yes. Updated the docstring.

src/datachain/llm/spec.py
```diff
@@ -0,0 +1,285 @@
...
+        params in the dict form of ``llm_params`` or in per-call kwargs so they are
+        captured here.
+        """
+        schema_repr: Any = None
+        if self.schema is not None:
+            elem, is_list = _element_type(self.schema)
+            if hasattr(elem, "model_json_schema"):
+                schema_repr = (str(elem.model_json_schema()), is_list)
+            else:
+                schema_repr = str(self.schema)
```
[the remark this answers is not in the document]
dmpetrov: changed to `_canonical()`

dmpetrov:
@shcheklein the issues were addressed. please take a look.

src/datachain/llm/content.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+def _as_text(value: Any) -> Text:
+    if isinstance(value, File):
+        return Text(_read_text(value))
+    if isinstance(value, BaseModel):
+        return Text(value.model_dump_json())
+    if isinstance(value, bytes):
+        return Text(_decode(value))
+    return Text(str(value))
```
shcheklein, line 101: can value be None? does it mean we will send it to LLM?

here is what AI gives:


```python
scored = users.merge(reviews, on="id", right_on="user_id")  # unmatched rows -> review=None
scored.map(risk=llm.score("review.text", "toxicity 0..1"))
# each unmatched row => model receives "None", returns a "score" for it, you pay for it
```

shcheklein: not sure what to do here - raise? what do we do with empty strings?

src/datachain/llm/spec.py
```diff
@@ -0,0 +1,285 @@
...
+        return (
+            self.kind,
+            model,
+            self.prompt,
+            schema_repr,
+            tuple(self.into) if self.into else None,
+            self.col,
+            self.context_col,
+            self.media,
+            self.retries,
```
shcheklein: another good finding:

`retries`/`fallback` params invalidate checkpoints, right? 

that's what AI gives:


```python
chain.map(scene=llm.complete("file", schema=Scene, retries=1)).save("scenes")  # 1M rows billed
# a few hundred rows failed on provider hiccups; bump the budget:
chain.map(scene=llm.complete("file", schema=Scene, retries=3)).save("scenes")
# new UDF hash => checkpoint invalid => ALL 1M rows re-run and re-billed
```

docs/references/llm.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Token usage
+
+Pass `include_usage=True` to capture token counts for a call. The function then
+returns a tuple `(value, dc.llm.Usage)`, so you name both columns with the
+multi-output form (the value column keeps its plain type, no wrapper):
+
+```python
+.map(
+    llm.complete("file", schema=Scene, include_usage=True),
```
shcheklein, line 105: is this expected:

```python
# Form A — positional + output= (the ONLY form the docs and PR description show):
.map(llm.complete("file", schema=Scene, include_usage=True),
     output={"res": Scene, "tok": Usage})
# resulting schema: res: Scene, tok: Usage           <- as documented

# Form B — keyword (signal-name) form, i.e. how every OTHER llm.* call is written:
.map(scene=llm.complete("file", schema=Scene, include_usage=True))
# resulting schema: scene_0: Scene, scene_1: Usage   <- no "scene" column, no error, no warning
```

dmpetrov: Yes. THat's how standard behaviour for tuple. It works the same for any regular function that returns tuple.

dmpetrov: it turned out that this tuple stuff is not documented. I'm adding the docs (not dc.llm specfic).

src/datachain/llm/spec.py
dmpetrov: that's good one! fixing.

src/datachain/llm/content.py
dmpetrov: good catch. we should return None without any LLM call. This is how SQL functions usually work. fixing.

docs/references/llm.md
shcheklein: right, but in this case, would it make sense to make it behave as in the first case though? AFAIU we control the output type, right? and probably names, etc? (I'm not sure if that would lead to more code). But second case returning doing the first behavior is what I think most users would expect automatically in this llm-specific case I think.

(not a blocker from my end, I just think it can be a nice improvement)

src/datachain/llm/content.py
shcheklein: resolving, considering this is done

docs/references/llm.md
dmpetrov: Do you propose introducing a second column like scene_tok that was not specified?

I don't like a not specified signal approach
Because it creates an exception /special-case in udf-s. The tuple exception seems enough and actually needed.

Another related approach - wrap output to a class with two variables like value and tokens. It has its own downsides:
- scene.value instead of scene
- auto generated class with "random" name like SceneWithIsage that used didn't asked for.


So, if we keep the tuple logic it's better to use it (it's useful but not typical). If we remove the tuple stuff from the core - use one of the approaches above.

src/datachain/llm/engine.py
```diff
@@ -0,0 +1,342 @@
...
+
+def _has_document(messages: list[dict[str, Any]]) -> bool:
+    return any(
+        isinstance(m.get("content"), list)
+        and any(p.get("type") == "file" for p in m["content"])
+        for m in messages
+    )
+
+
+def _check_document_support(model: str, messages: list[dict[str, Any]]) -> None:
```
shcheklein: AFAIU this check happens per row, runtime, right?

* Why is it only about PDF (document), we can do the same vision (and / or other modalities)?
* Code is quite complicated - geattr, try, etc, etc - can it be simplified?
* More important question: can we detect this `bind` time?
* Untyped `File` holding a PDF without an explicit `media=`  errors ... not sure if this is fine, we can try to infer in resolve?

E.g, can we avoided I think even before we run it:

```python
frames.settings(llm="openai/gpt-4o-mini-text-only")  # hypothetical text-only model
      .map(cap=llm.complete("file"))                 # ImageFile column
# builds fine; then every row: provider 400, retried num_retries times, then row error
```

docs/references/llm.md
shcheklein: No, I think we can do same columns, this is about naming. I think it just unexpected for users to go from 

```
.map(scene=llm.complete("file", schema=Scene))
```

with a single column `scene`

to 

```
.map(scene=llm.complete("file", schema=Scene, include_usage=True))
```

which gets now scene_0, scene_1

naming is bad here ^^

I would expect same stable name `scene` + some additional name (e.g. `usage`).

Since we pretty much control output in this case, I think we should be able to do this. I don't think anyone would understand anything about `tuple` semantics in this case.

src/datachain/llm/engine.py
```diff
@@ -0,0 +1,342 @@
...
+
+def parse_one(schema: type[T], content: str) -> T:
+    last_error: ValidationError | None = None
+    for candidate in (content, _strip_fences(content)):
+        try:
+            return schema.model_validate_json(candidate)
+        except ValidationError as exc:
+            last_error = exc
+    raise LLMError(
+        f"stored output could not be parsed as '{schema.__name__}'"
```
shcheklein: what is "stored" output?

shcheklein (commented):
Can we please ask AI to cleanup and update description (and probably should make it part of some skill). Not a big deal - but I think it is beneficial even for agents later when they read the commit summary

examples/llm_and_nlp/dc-llm-functions.py
```diff
@@ -0,0 +1,172 @@
...
+def _pdf(text: str) -> bytes:
+    img = PILImage.new("RGB", (480, 160), "white")
+    ImageDraw.Draw(img).text((20, 60), text, fill="black")
+    buf = BytesIO()
+    img.save(buf, format="PDF")
+    return buf.getvalue()
+
+
+def check_text() -> None:
+    rows = dc.read_values(review=[POSITIVE, NEGATIVE]).settings(llm=CHAT, cache=True)
```
shcheklein: minor: probably cache True is not needed

src/datachain/__init__.py
shcheklein, line 57: double check - do we need to include `func` and `llm` into __all__?

docs/references/llm.md
```diff
@@ -0,0 +1,158 @@
...
+For raw `bytes` or an untyped `File`, declare the modality with `media=`:
+
+```python
+.map(cap=llm.complete("frames", media="image"))     # bytes / File -> image
+.map(ext=llm.complete("file",  media="document"))    # File / bytes -> PDF (document-capable model)
+```
+
+Rules to keep in mind:
+
+- `media` is validated up front: `media="image"` on non-image bytes, or
```
shcheklein: it is not, right?

docs/references/llm.md
```diff
@@ -0,0 +1,158 @@
...
+  `media="document"` on a non-PDF, raises a clear error.
+- A `str` is sent **verbatim**, so a column holding a *path* sends the path, not the
+  file's contents; read it as a `File` (`read_storage(...)`) to send the content.
+- `media="document"` covers "summarize/extract from this PDF"; heavy document work
```
shcheklein: minor observation: we have `type` in `read_storage` and `media` here - two different names with overlapping meaning ... it would nice to have them aligned

shcheklein: also, observation is it called `content_type` in buckets / HTTP

src/datachain/llm/content.py
```diff
@@ -0,0 +1,184 @@
...
+        return None
+    return f"image/{fmt.lower()}" if fmt else None
+
+
+def _read_text(file: File) -> str:
+    try:
+        return file.read_text()
+    except UnicodeDecodeError as e:
+        raise LLMError(
+            f"cannot read '{file.path}' as text (it looks binary); if it is an "
```
shcheklein: minor: same error code is shared with `embed` which seems doesn't support media image / document (why btw??), error message should not be including `set media='image'/'document'`

src/datachain/llm/spec.py
```diff
@@ -0,0 +1,290 @@
...
+            model,
+            self.prompt,
+            schema_repr,
+            tuple(self.into) if self.into else None,
+            self.col,
+            self.context_col,
+            self.media,
+            tuple(self.fallback) if isinstance(self.fallback, list) else self.fallback,
+            self.include_usage,
+            _canonical(params),
```
shcheklein, line 203: AFAIU for the case like:

```python
.map(x=llm.complete("t", client_opts=MyOpts()))
```

it will be returning something like 

```
# repr: <MyOpts object at 0x105f3...>
```

(address)

It will be different each time - so hash will be different (?). So full recompute. Not sure what exactly cases it will impact, wort running through AI carefully

pyproject.toml
```diff
@@ -101,8 +101,11 @@ postgres = [
 zarr = [
   "zarr>=3.0.0 ; python_version >= '3.11'"
 ]
+llm = [
+  "litellm>=1.50.0,<2"
```
shcheklein: it is extremely old, they had some major security issues, also it doesn't support a lot of stuff
I think we can bump it to the latest, and remove a lot of weird stuff - like defensive checks on supports_pdf_input, etc

src/datachain/llm/engine.py
```diff
@@ -0,0 +1,342 @@
...
+    fallback: str | list[str] | None,
+    params: dict[str, Any],
+) -> tuple[str, Usage]:
+    litellm = _litellm()
+    kwargs = _completion_kwargs(model, messages, retries, fallback, params)
+    resp = litellm.completion(**kwargs)
+    return _content(resp), _usage(resp)
+
+
+def _parse_with_retries(
```
shcheklein: @dreadatour do you handle any retries in your litellm code? can we consolidate all this code between implementations?

shcheklein: also, why do we need this complicated / custom loop?

src/datachain/llm/engine.py
```diff
@@ -0,0 +1,342 @@
...
+    litellm = _litellm()
+    messages: list[dict[str, Any]] = list(kwargs["messages"])
+    kwargs = {**kwargs, "messages": messages, "num_retries": 0}
+    attempts = max(retries, 0) + 1
+    total_in = total_out = 0
+    last_error: Exception | None = None
+    for attempt in range(attempts):
+        try:
+            resp = litellm.completion(**kwargs)
+        except Exception as exc:  # noqa: BLE001 - transient/provider error; retry
```
shcheklein: this can be literally any error here? e.g. bad auth - and we will keep hitting it

does it burn tokens on each retry? do we count them?

can LLMError provide a reason for transient errors?

src/datachain/llm/__init__.py
```diff
@@ -0,0 +1,207 @@
...
+        params (Any): Extra arguments forwarded to the underlying model call.
+
+    Returns:
+        LLMSpec: A spec whose output type is ``str``.
+    """
+    return LLMSpec(
+        kind="classify",
+        col=col,
+        prompt=prompt,
+        into=list(into),
```
shcheklein, line 115: if someone pass a string - this will convert it into a list of chars:

```python
spec = llm.classify("text", into="yes")
assert spec.into == ["y", "e", "s"]
assert spec._build_prompt() == "Classify the input into exactly one of: y, e, s."
```

before it has a chance to validate in `LLMSpec.__post_init__`

src/datachain/llm/engine.py
```diff
@@ -0,0 +1,342 @@
...
+        and any(p.get("type") == "file" for p in m["content"])
+        for m in messages
+    )
+
+
+def _check_document_support(model: str, messages: list[dict[str, Any]]) -> None:
+    # Gate only the primary; a fallback handles unrelated transient failures.
+    if not _has_document(messages):
+        return
+    supports = getattr(_litellm(), "supports_pdf_input", None)
```
shcheklein: check also if `getattr(_litellm(), "supports_pdf_input", None)` actually works even in the recent versions - 1.91.1

docs/references/llm.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Token usage
+
+Pass `include_usage=True` to capture token counts for a call. The function then
+returns a tuple `(value, dc.llm.Usage)`, so you name both columns with the
+multi-output form (the value column keeps its plain type, no wrapper):
+
+```python
+.map(
+    llm.complete("file", schema=Scene, include_usage=True),
```
shcheklein, line 105: minor: should we do Scene | None and Usage | None in types, btw (considering that we return Nones on Nones now?)

src/datachain/llm/spec.py
```diff
@@ -0,0 +1,290 @@
...
+        return (
+            self.kind,
+            model,
+            self.prompt,
+            schema_repr,
+            tuple(self.into) if self.into else None,
+            self.col,
+            self.context_col,
+            self.media,
+            tuple(self.fallback) if isinstance(self.fallback, list) else self.fallback,
```
shcheklein: consider making `fallback` make transparent for hashes:

```python
chain.map(x=llm.complete("file", fallback="openai/a")).save("out")
# later, replace an unavailable fallback:
chain.map(x=llm.complete("file", fallback="openai/b")).save("out")
# new hash -> successful primary-model rows are called/billed again
```

src/datachain/llm/spec.py
```diff
@@ -0,0 +1,290 @@
...
+            self.include_usage,
+            _canonical(params),
+        )
+
+    def _resolve_model(self, settings: "Settings") -> str:
+        model = self.llm or settings.llm
+        if not model:
+            raise LLMConfigError(
+                f"no model configured for llm.{self.kind}(); set one with "
+                '.settings(llm="provider/model") or a per-call llm=.'
```
shcheklein: a broken message? (ending is weird - should there be some example?)

src/datachain/llm/spec.py
```diff
@@ -0,0 +1,290 @@
...
+                raise ValueError(
+                    "llm.classify(into=...) must be a non-empty list of strings"
+                )
+            if len(set(self.into)) != len(self.into):
+                raise ValueError("llm.classify(into=...) categories must be distinct")
+        if self.media is not None and self.media not in MEDIA_VALUES:
+            raise ValueError(
+                f"media must be 'text', 'image', or 'document', got {self.media!r}"
+            )
+        if not isinstance(self.retries, int) or isinstance(self.retries, bool):
```
shcheklein, line 114: minor:

```python
llm.complete("t", retries=-5)       # accepted; silently clamped to zero
llm.complete("t", fallback=123)     # accepted; TypeError inside _fallbacks at runtime

def bad_factory():
    return None
rows.settings(llm_params=bad_factory).map(...)  # mapping error on first row/worker
```

improve checks

docs/references/llm.md
dmpetrov: Good points but this issue needs to be addressed for tuple-result in general, outside of this LLM issue.
Created #1851.

dmpetrov: so, let's keep tuple as is in this PR

docs/references/llm.md
dmpetrov: yes

docs/references/llm.md
dmpetrov: renamed it to type and improved docs a bit

`content_type` is more tricky - we don't have this information for now when reading files. and it's expensive to get in S3 (an additional call per file while listing). So, i don't think we will go this direction in near future.
Alternative - guessing type by file extention that we already do.

src/datachain/llm/spec.py
dmpetrov: fixed

src/datachain/llm/spec.py
```diff
@@ -0,0 +1,290 @@
...
+        if self.schema is not None:
+            elem, is_list = _element_type(self.schema)
+            if hasattr(elem, "model_json_schema"):
+                schema_repr = (_canonical(elem.model_json_schema()), is_list)
+            else:
+                schema_repr = str(self.schema)
+        params = self.params
+        if isinstance(llm_params, dict):
+            params = {**llm_params, **self.params}
+        return (
```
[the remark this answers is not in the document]
dmpetrov: fixed

src/datachain/llm/spec.py
```diff
@@ -0,0 +1,290 @@
...
+        )
+
+    def _resolve_model(self, settings: "Settings") -> str:
+        model = self.llm or settings.llm
+        if not model:
+            raise LLMConfigError(
+                f"no model configured for llm.{self.kind}(); set one with "
+                '.settings(llm="provider/model") or a per-call llm=.'
+            )
+        return model
```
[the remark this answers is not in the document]
dmpetrov: fixed in PR description

pyproject.toml
dmpetrov: fixed

src/datachain/llm/engine.py
dmpetrov: fixed

src/datachain/llm/__init__.py
dmpetrov: good one. fixed

docs/references/llm.md
dmpetrov: yes! changed.

src/datachain/llm/spec.py
dmpetrov: fixed

src/datachain/llm/spec.py
dmpetrov: good catch. fixed

src/datachain/llm/engine.py
dmpetrov: fixed

examples/llm_and_nlp/dc-llm-functions.py
dmpetrov: fixed

src/datachain/llm/engine.py
dmpetrov: Fixed: 
- Per-row runtime check / retry-storm
- Detect at bind time - added for video/audio

Not-fixed:
- text-only-model + ImageFile - it would require an extra LLM call to check. but retries are fixed.
- Untyped File with pdf - kept explicit. It can be text or a PDF and we don't know which at bind.

src/datachain/__init__.py
dmpetrov: fixed

src/datachain/llm/engine.py
dmpetrov: fixed

src/datachain/llm/content.py
dmpetrov: fixed

dmpetrov:
> cleanup and update description (and probably should make it part of some skill)

Description was updated. Skill is followup since sill is generated from second-brain repo - I need to do RI/FI first.

@shcheklein all changes were addressed - please take a look.

**Important!** I included llm into a default package -this functionality and LLM-processing is must have these days. The overhead is reasonable:

```
┌────────────────────────────────────────────────┬──────────┬──────────┬──────────────┐
│  pip install datachain[…]                      │   Size   │  Δ core  │ Install time │
├────────────────────────────────────────────────┼──────────┼──────────┼──────────────┤
│  core (no extras)                              │ ~490 MB  │    —     │   ~16 s      │
│  core + llm             ◀ current default      │ ~597 MB  │  +22%    │   ~20 s      │
│  core + llm + light     ◀ suggested default    │ ~669 MB  │  +37%    │   ~24 s      │
│  [hf]     (core + hf)                          │ ~1197 MB │  +144%   │   minutes    │
│  [torch]  (core + torch)                       │ ~1157 MB │  +136%   │   minutes    │
└────────────────────────────────────────────────┴──────────┴──────────┴──────────────┘

light extras (Δ core): video +10%, postgres +2%, zarr +1%, vector +1%, audio +1%, remote ~0%
                          (all light together = +15%)
```

We should also move video and audio default since it's a core use cases for us. In a separate PR. Created #1852

src/datachain/llm/engine.py
shcheklein: another question btw @dmpetrov @dreadatour will it pick up LLM credentials in Studio automatically?

also, doest it all / will it work with a locally deployed (e.g. on the GPU worker) model. E.g. Qwen?

docs/references/llm.md
shcheklein: kk, again not a blocker

I think though, this feedback is unrelated tbh. (unless I'm missing something). Users don't see tuples here, they don't care about them and their semantics. They don't even "see" udfs in this case. That's the point of these helpers AFAIU. So, we have full control, we can easily create a schema that we want. I'm also not sure that the tuple's solution will be optimal in this case, but it might be. The benefit of potentially solving it before hand is backward compatibility.

pyproject.toml
```diff
@@ -56,6 +56,7 @@ dependencies = [
   "dvc-studio-client>=0.23.0,<1",
   "tabulate",
   "websockets",
+  "litellm>=1.77.0,<2",
```
shcheklein: It permits malicious PyPI releases `1.82.7` and `1.82.8`, which were part of a credential-stealing supply-chain compromise. This is still not safe ... 

why is it trying to be smart about the lower boundary, ughhh ??

src/datachain/llm/spec.py
```diff
@@ -0,0 +1,340 @@
...
+                base = llm_params() if callable(llm_params) else llm_params
+                if base is not None and not isinstance(base, dict):
+                    raise TypeError(
+                        f"llm_params must resolve to a dict, got {type(base).__name__}"
+                    )
+                resolved.append({**(base or {}), **spec.params})
+            return resolved[0]
+
+        # `_id` default arg bakes the cache key into the UDF hash (via __defaults__).
+        _id = self.identity(model, llm_params)
```
shcheklein, line 333: not sure if it needs a fix, but it has a point / gotchas in some cases, some users might decide to mutate it (?) ... it might break because of caching, etc

```python
spec = llm.complete("text", "OLD")
run = spec.bind(BindContext(Settings(llm="m")))
before = hash_callable(run)

spec.prompt = "NEW"
assert hash_callable(run) == before

run("value")
# model receives "NEW\n\nvalue", although the bound UDF keeps the OLD identity
```

src/datachain/llm/engine.py
```diff
@@ -0,0 +1,322 @@
...
+    if not fallback:  # None, "", or []
+        return None
+    return [fallback] if isinstance(fallback, str) else list(fallback)
+
+
+def _is_transient(exc: Exception) -> bool:
+    """Worth retrying: timeout (408), rate limit (429), and server errors; other
+    client errors (bad auth/request) are fatal."""
+    status = getattr(exc, "status_code", None)
+    if not isinstance(status, int):
```
shcheklein: this will include ValueError, TypeError, etc - probably not good retrying them

src/datachain/llm/engine.py
```diff
@@ -0,0 +1,322 @@
...
+    messages: list[dict[str, Any]] = list(kwargs["messages"])
+    kwargs = {**kwargs, "messages": messages, "num_retries": 0}
+    attempts = max(retries, 0) + 1
+    total_in = total_out = 0
+    last_error: Exception | None = None
+    for attempt in range(attempts):
+        try:
+            resp = litellm.completion(**kwargs)
+        except Exception as exc:
+            if not _is_transient(exc):
```
shcheklein: so, on 429 or some other timeout - will send 10 request in a very short span, making it even worse ...

I feel we are reinventing a wheel here ... there is @retry decorator with some proper backoff policies, can it be used here? can litellm handle this automatically?

src/datachain/llm/spec.py
```diff
@@ -0,0 +1,340 @@
...
+    if _DEFAULT_REPR.search(repr(value)):
+        warnings.warn(
+            f"llm param {type(value).__name__!r} has no stable repr; it breaks "
+            "caching (full recompute every run).",
+            stacklevel=2,
+        )
+    return value
+
+
+# Credential params kept out of the cache key (no secrets in the UDF hash).
```
shcheklein: It still probably taking secrets:

Common valid LiteLLM shapes such as `extra_headers={"Authorization": ...}`, `api_token`, `access_token`, `client_secret`, and provider-specific token names remain part of identity. Rotating one of them still changes the UDF hash and can re-bill the entire dataset.

```python
spec = llm.complete("t")

assert spec.identity(
    "m", {"extra_headers": {"Authorization": "Bearer OLD"}}
) != spec.identity(
    "m", {"extra_headers": {"Authorization": "Bearer NEW"}}
)

assert spec.identity("m", {"api_token": "OLD"}) != spec.identity(
    "m", {"api_token": "NEW"}
)
```

Need recursive sanitazation?

docs/references/llm.md
```diff
@@ -0,0 +1,166 @@
...
+    output={"res": Scene, "tok": dc.llm.Usage},
+)
+# res: Scene   (res.risk works),  tok: Usage
+```
+
+`Usage` has `input_tokens`, `output_tokens`, and `retries` (extra attempts beyond
+the first; 0 on a first-try success). Aggregate or filter on it afterwards:
+
+```python
+chain.agg(in_tok=dc.func.sum("tok.input_tokens"))
```
shcheklein, line 114: in case of `gen` it seems we replicate usage object (?) - does it mean this formula might overcount then?

pyproject.toml
dmpetrov: oh, sorry. fixed

src/datachain/llm/engine.py
dmpetrov: right! fixed

src/datachain/llm/spec.py
dmpetrov: let's do this

docs/references/llm.md
dmpetrov: well... i thought it's good to have this - both approaches have pros and cons.
after some extra research decided to generate it only for 1st row and keep Usage(0,0,0) for the rest to make aggregation work.
fixed

src/datachain/llm/engine.py
dmpetrov: the wheel was reinvented to cover LLM-usual retries together with schema-mismatches (real issues for models that does not natively support structured output like Qwen).

The decision - simplify to the ground. Not-structured output now on user shoulders (job will fail) and only traditional LLM-retries are working which is fully covered by the llm lib.

fixed

src/datachain/llm/spec.py
dmpetrov: These are kind of internals. If user touches this part - it's on them. I don't think we should be protecting them from these types of tricks.

Let's keep as is.

src/datachain/llm/engine.py
dmpetrov: > will it pick up LLM credentials in Studio automatically?

it won't. the creds has to be set up in `settings()`.

Ideally, it should be able to pick up studio creds and studio should track it's usage as a separate activity. So, user will use an aggregated usage with the Studio's AI-enrichment part as well as this LLM-processing part. `.map(x=llm.complete("file", schema=Scene, llm="datachain/Prod demo-1 team"))`

It's a followup issue - not related to this PR.

dmpetrov: > doest it all / will it work with a locally deployed (e.g. on the GPU worker) model. E.g. Qwen?

Yes. 
```python
.settings(llm="openai/Qwen2.5-7B-Instruct",
          llm_params={"api_base": "http://gpu-worker:8000/v1"})
```

dmpetrov:
@shcheklein all new comments were addressed. please check.

shcheklein (approved):
It would be great for @dreadatour  to also review the engine, credentials management, how we can reuse and integrate all of this together with Studio

dreadatour:
> It would be great for @dreadatour to also review the engine, credentials management, how we can reuse and integrate all of this together with Studio

Looks good to me! We can reuse everything, couple tiny tweaks may be needed (e.g. setting model name from Studio side), but we can do it in a follow-up 👍

dreadatour (approved):

PR #1846, opened 2026-07-05
feat: object write metadata for File save/upload/open/export

shcheklein:
## What

Adds a way to pass object metadata (content-disposition, content-type, cache-control, content-encoding, custom metadata, plus a raw escape hatch) to every `File` write path.

```py
file.save(
    "s3://bucket/report.pdf",
    content_type="application/pdf",
    content_disposition='attachment; filename="report.pdf"',
    cache_control="max-age=3600",
    metadata={"run_id": "abc"},
)

File.upload(pdf_bytes, "s3://bucket/report.pdf", content_disposition="attachment")

f = File.at("s3://bucket/out.pdf")
with f.open("wb", content_type="application/pdf") as h:
    h.write(data)
```

The new keyword args are available on `File.upload`, `File.save`, `File.open` (write modes), `File.export`, and every subclass save (`TextFile` / `ImageFile` / `VideoFile` / `AudioFile` / `AudioFragment` / `VideoFrame` / `VideoFragment`) plus the `save_audio` / `save_video_fragment` helpers. `File.at` is unchanged — it only constructs a `File`, so its metadata lives on the subsequent `open("wb", ...)`.

## How

A small `WriteConfig` (in the client layer) carries the normalized fields and a raw `extra`. Each `Client` maps it to its backend's native parameters, set **inline in the single write commit** on every backend (atomic):

`Client.upload` / `File.open` delegate the actual write to overridable hooks (`_write_object` / `open_for_write`); each backend maps `WriteConfig` to its native write:

| Backend | Mapping |
|---|---|
| **S3** | PascalCase → `s3_additional_kwargs` (`put_object` / multipart) |
| **GCS** | `content_type` / `metadata` / `fixed_key_metadata` (resumable upload) |
| **Azure** | `ContentSettings` + `metadata` via azure-storage-blob `upload_blob` |
| **local** | ignored (no HTTP-header concept) |

The `write_options` escape hatch is a genuine passthrough only on S3; the GCS/Azure `fsspec` backends have no raw write-kwargs passthrough, so it raises `NotImplementedError` there. The base `_write_kwargs` also rejects it, so a backend without an override (e.g. `HfClient`) can't silently leak it. Read mode with any write metadata raises `ValueError`.

The existing `content_disposition` support for **signed download URLs** is unrelated and untouched — this is purely the **write** side.

### Azure

adlfs doesn't forward content settings on write (unlike s3fs/gcsfs), so `AzureClient` writes objects through **azure-storage-blob's public `upload_blob` API** directly, which sets content settings + metadata inline in a single atomic upload. This keeps Azure writes atomic (no post-write `set_http_headers`, no concurrent-overwrite race, correct version returned) and — importantly — **needs no adlfs changes**: datachain is self-contained, with no vendored internals and no version-gating. Streamed writes (`File.save`, `File.open`) are buffered to a temp spool so the fsspec event loop never reads a loop-backed stream. (The corresponding adlfs improvement is upstreamed as **fsspec/adlfs#554** for the ecosystem, but datachain no longer depends on it.)

## Testing

Verified **read-back of the actually stored metadata** across the backend matrix (moto / fake-gcs / azurite) × `version_aware`:

- `tests/func/test_file.py` — upload / save / open / export / TextFile.save / ImageFile.save on S3 / GCS / Azure; `write_options` applied on S3 and rejected on GCS/Azure.
- `tests/unit/test_write_config.py` — exact per-backend kwarg mappings + `WriteConfig` semantics (incl. base/`HfClient` rejecting `write_options`).
- `tests/unit/lib/test_file.py` — local-ignore, read-mode guard, subclass/helper forwarding.

Emulator limitations are gated in tests (with comments) and reflect real-backend behavior:
- fake-gcs drops fixed metadata (content-disposition) on streaming uploads and doesn't persist `cacheControl` — real GCS does both.
- moto doesn't echo custom user metadata via `head_object`.


src/datachain/client/fsspec.py
```diff
@@ -476,20 +490,55 @@ def upload(
...
+        """Whether bytes uploads may use ``pipe_file`` (vs. a streaming open)."""
+        return True
+
+    def _write_kwargs(self, cfg: "WriteConfig", *, streaming: bool) -> dict[str, Any]:
+        """Map a :class:`WriteConfig` to native ``fs.open``/``pipe_file`` kwargs.
+
+        The base implementation forwards only the raw ``extra`` escape hatch;
+        cloud subclasses translate the normalized fields to backend parameters.
+        """
+        return dict(cfg.extra or {})
```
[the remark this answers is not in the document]
shcheklein: Good catch — fixed in 28b41854. The base `_write_kwargs` now maps nothing and raises `NotImplementedError` on `write_options`; only S3 overrides to forward it. This closes the leak on backends without an override (e.g. `HfClient`) and matches the S3-only contract. Added `test_write_options_rejected_unless_backend_supports_it` (covers `HfClient`) and `test_base_default_maps_nothing`.

src/datachain/lib/audio.py
```diff
@@ -174,6 +175,7 @@ def save_audio(
     start: float = 0,
     end: float | None = None,
     client_config: dict | None = None,
+    write_config: "WriteConfig | None" = None,
```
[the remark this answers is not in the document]
shcheklein: Fixed in 1c2a9d9e — documented in the Args/docstring.

src/datachain/lib/video.py
```diff
@@ -610,6 +613,7 @@ def save_video_fragment(
     format: str | None = None,
     client_config: dict | None = None,
     timeout: float | None = None,
+    write_config: "WriteConfig | None" = None,
```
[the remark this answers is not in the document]
shcheklein: Fixed in 1c2a9d9e — documented in the Args/docstring.

src/datachain/client/fsspec.py
```diff
@@ -463,7 +469,15 @@ def upload(
             backends) flushes a multipart part every ``self.blocksize``
             bytes (e.g. 5 MiB on s3fs), so peak RAM is bounded by the
             backend's part size rather than the file size.
+
+        ``write_config`` carries normalized object metadata (content type,
+        content disposition, custom metadata, …) mapped to the backend's
+        native write parameters.
```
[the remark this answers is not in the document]
shcheklein: Fixed in 1c2a9d9e — documented in the Args/docstring.

src/datachain/client/writeconfig.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+@dataclass(frozen=True)
+class WriteConfig:
+    """Normalized, backend-agnostic metadata for object writes.
+
+    Each :class:`~datachain.client.fsspec.Client` maps these fields to its
+    backend's native parameters (S3 ``s3_additional_kwargs``, GCS
+    ``fixed_key_metadata``, Azure ``ContentSettings``). Fields left as ``None``
+    are not sent. ``extra`` is an escape hatch of raw, backend-native kwargs for
```
0x2b3bfa0: > _escape hatch_

Typical term used by Claude Opus 4.8 everywhere. 😄

src/datachain/client/azure.py
```diff
@@ -88,6 +110,91 @@ def info_to_file(self, v: dict[str, Any], path: str) -> File:
...
 
+    def _write_object(
+        self,
+        full_path: str,
+        data: "bytes | bytearray | memoryview | Any",
+        cfg: "WriteConfig",
+        *,
+        overwrite: bool = True,
+    ) -> str | None:
+        # adlfs does not forward content settings on write, so write via the
```
0x2b3bfa0, line 138: Worth adding a `TODO` comment linking to https://github.com/fsspec/adlfs/pull/554?

src/datachain/client/fsspec.py
```diff
@@ -476,19 +493,69 @@ def upload(
...
                 # a natural choice; fall back to 8 MiB otherwise.
                 buf_size = getattr(dst, "blocksize", None) or 8 * 1024 * 1024
-                shutil.copyfileobj(data, dst, length=buf_size)
-        file_info = self.fs.info(full_path, **self._file_info_kwargs())
-        return self.info_to_file(file_info, rel_path)
+                shutil.copyfileobj(src, dst, length=buf_size)
+        return None
+
+    @contextmanager
+    def open_for_write(
```
0x2b3bfa0, line 535: Maybe I'm still a bit sleepy, but, what's the consumer of this context manager? 🤔

0x2b3bfa0:
Left some comments, not sure if constructive.

src/datachain/lib/file.py
```diff
@@ -50,6 +51,24 @@
...
+    write_options: dict[str, Any] | None = None,
+) -> WriteConfig:
+    return WriteConfig(
+        content_type=content_type,
+        content_disposition=content_disposition,
+        cache_control=cache_control,
+        content_encoding=content_encoding,
+        metadata=metadata,
+        extra=write_options,
+    )
```
dreadatour: It looks like it should belongs to `datachain.client.writeconfig`, may be even classmethod of `WriteConfig`.

Also I can not see any advantages of having this helper to be honest, below where we do call `_build_write_config` we can call `WriteConfig` constructor directly.

src/datachain/client/writeconfig.py
```diff
@@ -0,0 +1,43 @@
...
+    backends have no raw write-kwargs passthrough. The local filesystem has no
+    notion of any of these and ignores all fields.
+    """
+
+    content_type: str | None = None
+    content_disposition: str | None = None
+    cache_control: str | None = None
+    content_encoding: str | None = None
+    metadata: Mapping[str, str] | None = None
+    extra: Mapping[str, Any] | None = None
```
dreadatour: It is not 100% clear from docstring what is `extra` and what is the difference between `metadata` and `extra`. Also on other file `extra` is equivalent to `write_options`, and in S3 (where `extra` only make sense) it is `s3_additional_kwargs`. Not a blocker but it would be nice to make it more clear IMO.

src/datachain/lib/file.py
```diff
@@ -606,7 +685,18 @@ def read(self, length: int = -1):
         """Returns file contents."""
         return self.read_bytes(length)
 
-    def save(self, destination: str, client_config: dict | None = None) -> "File":
+    def save(
+        self,
+        destination: str,
+        client_config: dict | None = None,
```
dreadatour, line 673: IMO it would be great to also force `client_config` to be keyword argument, but it might break backward compatibility.

dreadatour (approved):
Looks good to me overall, but some cleanup is still required IMO.
Also I've added couple comments (non-blockers).

src/datachain/client/azure.py
shcheklein: Added in e05bcafb — a TODO in `_write_object` pointing at fsspec/adlfs#554. Once that's in a released adlfs, this SDK path can collapse to `fs.open`/`pipe_file` with a `_write_kwargs` mapping like the other backends.

src/datachain/client/fsspec.py
shcheklein: It's the write-mode path of `File.open` — `client.open_for_write(...)` in `lib/file.py` (the streaming counterpart to `_write_object`, which handles the whole-payload `upload`). Clarified the docstring to say so in e05bcafb.

src/datachain/lib/file.py
shcheklein: Good call — removed `_build_write_config` and construct `WriteConfig(...)` directly at the call sites (e05bcafb). Renaming the field to `write_options` made the helper a pure passthrough, so it's gone.

src/datachain/client/writeconfig.py
shcheklein: Renamed `extra` → `write_options` to match the public kwarg, and clarified the docstring: `metadata` = custom user key/value object metadata; `write_options` = the raw backend-native escape hatch (forwarded into S3 `s3_additional_kwargs`, e.g. `ACL`; rejected on GCS/Azure). e05bcafb.

src/datachain/lib/file.py
shcheklein: Kept `client_config` positional-or-keyword to preserve backward compatibility — it's an existing positional param, so making it keyword-only would break `file.save(dest, cfg)` calls. The new metadata kwargs are all keyword-only. Happy to revisit if you'd prefer the break.

src/datachain/client/fsspec.py
0x2b3bfa0: 🤦🏼 🙏🏼

0x2b3bfa0 (approved):

PR #1848, opened 2026-07-06
Route framework logs to a separate fd when DATACHAIN_INTERNAL_LOG_FD is set

dreadatour:
Follow-up to datachain-ai/studio#13028: keep datachain's internal `INFO`/`DEBUG` out of the customer-facing job log no matter how the job configures logging. See also related Studio PR.

**Known limitation**: A job that reconfigures logging with `dictConfig(disable_existing_loggers=True)` (the default) can still disable these loggers.

Adds `datachain.log_routing`. When `DATACHAIN_INTERNAL_LOG_FD` is set (Studio jobs), the framework loggers (`datachain`, `datachain_saas`, `compute`) are detached from the root logger and every record is written as one JSON line — `{message, time, level, logger}` — to that fd instead of stdout/stderr, so the worker on the other end can tell framework logs apart from user output.

Completely inert without the env var: outside a Studio job the loggers are left untouched and the CLI behaves exactly as before (regression-tested). The level comes from `DATACHAIN_INTERNAL_LOG_LEVEL` (default `INFO`); `-v` still widens the CLI's own console output.

Covers every child we spawn that logs: the CLI self-routes in `main()` (shell jobs and UDF workers), and the parallel-UDF spawn passes the fd through `pass_fds` so it survives `close_fds`.

### DIFF
--- a/src/datachain/log_routing.py
+++ b/src/datachain/log_routing.py
@@ -1,0 +1,112 @@
+"""Route framework (datachain/datachain_saas/compute) logs to a Studio-provided fd.
+
+Inside a SaaS job, framework log records go to the fd from
+DATACHAIN_INTERNAL_LOG_FD as one JSON line each (message, time, level, logger)
+instead of the captured stdout/stderr, so the parent worker can tell framework
+logs from user output.
+"""
+
+import json
+import logging
+import os
+from contextlib import suppress
+from datetime import datetime, timezone
+from io import TextIOWrapper
+
+INTERNAL_LOG_FD_ENV = "DATACHAIN_INTERNAL_LOG_FD"
+INTERNAL_LOG_LEVEL_ENV = "DATACHAIN_INTERNAL_LOG_LEVEL"
+
+_FRAMEWORK_LOGGERS = ("datachain", "datachain_saas", "compute")
+
+_routed = False
+
+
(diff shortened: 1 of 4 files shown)


src/datachain/log_routing.py
```diff
@@ -0,0 +1,112 @@
+"""Route framework (datachain/datachain_saas/compute) logs to a Studio-provided fd.
```
shcheklein: why is this code here in datachain then, not in datachain_saas ? is it only about Studio / datachain saas logs?

dreadatour: I don't want to import any SaaS code from DataChain and same time I don't want to complicate things a lot to be able to import it on-the-fly or define it as plugins or callbacks or something similar.

shcheklein: yep, but it is a bit weird to have this reference (and expose name like datachain_saas) ? and variables like `INTERNAL_LOG_FD_ENV` ....

and hard coded names like `_FRAMEWORK_LOGGERS = ("datachain", "datachain_saas", "compute")` ... 

is it the only option to solve this?

amritghimire (approved):

dreadatour:
Closing in favor of SaaS-only implementation.

PR #1857, opened 2026-07-14
Cap litellm below 1.92 to unbreak Python 3.14

dreadatour:
litellm 1.92.0 is the first release with a native Rust extension and ships only manylinux cp310-cp313 wheels. On Python 3.14 (all OSes) installs fall back to the sdist, which fails to build because its pinned PyO3 0.23.5 supports at most Python 3.13. Cap to 1.91.x, the last pure-python release, until upstream ships cp314 wheels or upgrades PyO3.

See this CI run for details: https://github.com/datachain-ai/datachain/actions/runs/29347557256

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,7 +56,10 @@ dependencies = [
   "dvc-studio-client>=0.23.0,<1",
   "tabulate",
   "websockets",
-  "litellm>=1.83.0,<2",  # >=1.83 excludes the compromised 1.82.7/1.82.8 releases
+  # >=1.83 excludes the compromised 1.82.7/1.82.8 releases;
+  # <1.92 is the last pure-python release — 1.92.0 adds a Rust extension with
+  # no cp314/macOS/Windows wheels and an sdist that can't build on Python 3.14
+  "litellm>=1.83.0,<1.92",
   "tomli;python_version<'3.11'"
 ]
 


pyproject.toml
```diff
@@ -56,7 +56,10 @@ dependencies = [
   "dvc-studio-client>=0.23.0,<1",
   "tabulate",
   "websockets",
-  "litellm>=1.83.0,<2",  # >=1.83 excludes the compromised 1.82.7/1.82.8 releases
+  # >=1.83 excludes the compromised 1.82.7/1.82.8 releases;
```
shcheklein, line 59: - rant mode on

yet another painful-to-maintain library it seems (like *torch and HF) :(

- rant mode off

shcheklein (approved):

pyproject.toml
dreadatour: Yes, it looks like so 😢

pyproject.toml
```diff
@@ -56,7 +56,10 @@ dependencies = [
   "dvc-studio-client>=0.23.0,<1",
   "tabulate",
   "websockets",
-  "litellm>=1.83.0,<2",  # >=1.83 excludes the compromised 1.82.7/1.82.8 releases
```
shcheklein, line 59: can we make it depend on Python version btw?

dreadatour: As far as I understand it is not only about Python version, it is also about platform (linux only).

shcheklein: do they even plan to provide build wheels for other platforms? 🤔

dreadatour: It looks like so, in the issue they were asking for:

```
Please report any install/import problems here, especially for:

macOS arm64 or x86_64
Linux x86_64 or aarch64
Alpine / musllinux containers
Windows x86_64
less common Python runtimes or platforms
```

But we might also want to search for an alternatives, it does not looks like a simple dependency to me according to their plans here: https://docs.litellm.ai/blog/litellm-rust-launch

Not sure yet, worth diving into.

PR #1862, opened 2026-07-16
Fixing flaky yolo test

### LINKED ISSUE #1861: Flaky yolo segment/pose tests: hardcoded contour coords break on ultralytics/opencv bumps
`test_yolo_seg_from_results` (and the pose test with hardcoded x/y lists) asserts exact contour coordinates from `Results.summary()`, which just wraps `cv2.findContours`. Any `ultralytics` or `opencv-python` bump shifts the polygon by 1-2 pixels and breaks the test.

Latest break: CI on `ultralytics 8.4.96` + `opencv-python 5.0.0.93`, golden captured on `8.4.27` + `4.11`. Fix was to re-paste golden lists - will happen again on the next bump.

Proper fix: drop the exact x/y lists, check structure instead (title matches, x/y same non-zero length, all points inside the bbox with small tolerance).

ilongin:
`test_yolo_seg_from_results` asserts exact segment coordinates from `Results.summary()`, which just wraps `cv2.findContours`. Any `ultralytics` or `opencv-python` bump shifts the polygon by 1-2 pixels and breaks the test.

Fix: drop the exact x/y lists in the segment assertions, check structure instead (title matches, x/y same non-zero length, all points inside the bbox with small tolerance).

The pose test also has hardcoded x/y lists but those are deterministic (`int(round(input_keypoint))`), not affected by library bumps, so no change needed there.


shcheklein (approved):

dreadatour (approved):
Thank you for the fix! 🙏
Tests are still broken and also there are few comments from Copilot.

PR #1866, opened 2026-07-17
Follow up to multi function map

### LINKED ISSUE #1865: Multi-signal .map() follow-ups from PR #1807 Copilot review
Two follow-ups from Copilot review of #1807:

1. Bind `BoundSpec` (dc.llm.*) entries in multi-signal `.map()` - currently `_bind_udf_settings` runs on the mapper but not per-entry, so `.map(a=llm.complete(...), b=llm.classify(...))` doesn't work. https://github.com/datachain-ai/datachain/pull/1807#discussion_r3595971997

2. Validate parameter kinds in `_MultiSignalMapper` - positional-only params or `*args`/`**kwargs` currently fail with a confusing `TypeError` instead of a clear `DataChainParamsError`. https://github.com/datachain-ai/datachain/pull/1807#discussion_r3595972036

ilongin:
Follow-ups from the Copilot review on #1807. Closes #1865.

- Bind each `signal_map` entry in `_build_multi_signal_mapper` via `_bind_udf_settings(v, Mapper)` before constructing `_MultiSignalMapper`. Multi-signal `.map(a=llm.complete(...), b=llm.classify(...))` now works - each `BoundSpec` gets turned into its concrete `run(...)` callable, and its input columns come from a new abstract `BoundSpec.input_columns()` method. `_MultiSignalMapper` branches by callable style: bound-spec entries get called positionally with their listed chain columns, plain functions keep the signature-based kwargs + dep-graph path.
- Reject parameter kinds that can't be passed by name (positional-only, `*args`, `**kwargs`) in multi-signal `.map()` at build time with a clear `DataChainParamsError`. Single-signal `.map(func)` still handles them fine because it calls positionally.
- Reject `include_usage=True` in multi-signal `.map()` at build time - the spec produces two outputs and multi-signal only supports one output per entry. Sharper error than the previous "output type 'tuple' is not supported" message.
- A bound spec's input column can now match a sibling entry's name; the spec receives that sibling's computed result instead of a chain column. Same chaining that already worked for plain functions.
- Support Mapper subclass instances in multi-signal `.map(upper=MyMapper(), lower=lambda name: ...)`. `_MultiSignalMapper` detects `UDFBase` entries, inspects `.process` for the signature, invokes `.process(**kwargs)` at runtime, and relays `setup()` / `teardown()` to each Mapper entry in declared order.
- Fix pre-existing bug where multi-signal `.map()` with all-zero-arg entries (or a zero-arg producer feeding sibling consumers) crashed with `SignalResolvingError: cannot resolve signal name 'args'`. Root cause: `UdfSignature.parse` treated `params=[]` the same as `params=None` and fell back to introspecting the wrapper's `process(*args)`. Fix: distinguish `[]` from `None`.
- Fix Studio job label for multi-signal maps: `_MultiSignalMapper.verbose_name` now returns the comma-separated signal names instead of `<unknown>`.


src/datachain/lib/udf.py
```diff
@@ -554,13 +554,35 @@ def __init__(self, signal_map: dict[str, Callable]):
...
+        # column names and are called positionally instead of by kwargs.
+        self._bound_style: set[str] = set()
         # For each function: which of its params come from another
         # function's output (dependencies) vs from an input row column.
         deps: dict[str, set[str]] = {}
         for name, fn in signal_map.items():
-            params = list(inspect.signature(fn).parameters.keys())
-            self._per_func_params[name] = params
-            deps[name] = {p for p in params if p in output_names and p != name}
+            cols = getattr(fn, "__datachain_params__", None)
```
shcheklein: can we do this better? e.g. isinstance or something? this is ugly hack now ... (to expect specific __datachain_params__)

src/datachain/lib/udf.py
```diff
@@ -554,13 +554,35 @@ def __init__(self, signal_map: dict[str, Callable]):
         self._signal_map = signal_map
         output_names = set(signal_map)
         self._per_func_params: dict[str, list[str]] = {}
+        # Functions stamped with `__datachain_params__` (e.g. bound BoundSpec
```
shcheklein: okay, this comment doesn't really explain to me why boud_style is the name, what is does, and also way too technical and complicated ... it would be great to have a comment but a way better onw

src/datachain/lib/udf.py
shcheklein: this is all looks like a bunch of hacks :(

ilongin: Refactored by checking the  instance ... now there is no __datachain_params__ mentions

src/datachain/lib/udf.py
```diff
@@ -554,13 +554,35 @@ def __init__(self, signal_map: dict[str, Callable]):
...
         for name, fn in signal_map.items():
-            params = list(inspect.signature(fn).parameters.keys())
-            self._per_func_params[name] = params
-            deps[name] = {p for p in params if p in output_names and p != name}
+            cols = getattr(fn, "__datachain_params__", None)
+            if isinstance(cols, list):
+                self._per_func_params[name] = list(cols)
+                self._bound_style.add(name)
+                deps[name] = set()
+            else:
```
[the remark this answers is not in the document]
ilongin: Can't silently return a wrong value here: if the sibling output name matches an existing chain column, the query layer rejects the whole `.map()`; if it doesn't exist, `signals_schema.slice(...)` fails at build time

shcheklein (commented):
Please review these findings by AI: https://claude.ai/code/artifact/df07cebf-3506-4c17-8ac4-4a3be10585f9

src/datachain/lib/udf_signature.py
```diff
@@ -5,7 +5,7 @@
 
 from datachain.lib.data_model import DataType, DataTypeNames, is_chain_type
 from datachain.lib.signal_schema import SignalSchema
-from datachain.lib.udf import UDFBase
+from datachain.lib.udf import UDFBase, _reject_var_params
```
shcheklein: let's not import and use protected methods

src/datachain/lib/udf.py
```diff
@@ -582,23 +644,52 @@ def __init__(self, signal_map: dict[str, Callable]):
...
+            if name in self._bound_columns:
+                results[name] = fn(
+                    *[(results[c] if c in results else row_by_name[c]) for c in params]
+                )
+            else:
+                fn_kwargs = {
+                    p: (results[p] if p in results else row_by_name[p]) for p in params
+                }
+                if name in self._mapper_entries:
+                    results[name] = fn.process(**fn_kwargs)
```
shcheklein: looking into this again: can we actually pass params by name to make it uniform?

src/datachain/lib/udf.py
```diff
@@ -71,6 +71,33 @@ class BoundSpec(ABC):
...
+        if allow_var_positional
+        else {inspect.Parameter.VAR_KEYWORD, inspect.Parameter.VAR_POSITIONAL}
+    )
+    var_params = [
+        f"*{p.name}" if p.kind is inspect.Parameter.VAR_POSITIONAL else f"**{p.name}"
+        for p in inspect.signature(func).parameters.values()
+        if p.kind in kinds
+    ]
+    if var_params:
+        raise DataChainParamsError(
```
shcheklein, line 117: I think, this is broken now:

```
def fn(name, **options) -> str:
    return name.upper()

chain.map(fn, params=["name"], output={"res": str})
```

src/datachain/lib/dc/datachain.py
```diff
@@ -1040,10 +1041,27 @@ def _build_multi_signal_mapper(
                 "to a top-level column with a single .map() first."
             )
 
-        multi_mapper = _MultiSignalMapper(signal_map)
+        for k, v in signal_map.items():
+            if isinstance(v, LLMSpec) and v.include_usage:
```
shcheklein: we do very specific check here - looks like a hack (we leak LLMSpec to this generic code)

can we test it in some generic way?

e.g if buy the shape of params or something? and rely only on BoundSpec ?

do we support only a single item per output for each map? so tuples won't work (include usage makes it a tuple) - should we have a generic test for that somewhere?

src/datachain/lib/udf.py
```diff
@@ -549,16 +576,51 @@ class _MultiSignalMapper(Mapper):
...
+        self._bound_columns = bound_columns or {}
         output_names = set(signal_map)
         self._per_func_params: dict[str, list[str]] = {}
+        # Entries that are UDFBase subclass instances (e.g. Mapper subclass);
+        # invoked via .process() and their setup()/teardown() run per worker.
+        self._mapper_entries: set[str] = set()
         # For each function: which of its params come from another
         # function's output (dependencies) vs from an input row column.
         deps: dict[str, set[str]] = {}
         for name, fn in signal_map.items():
```
shcheklein, line 613: check this when params shadow output names:

```
chain.map(a=lambda a: a.upper(), b=lambda name: name)
KeyError: 'a'   # at runtime, per row

chain.map(a=llm.complete("a", "x"), b=lambda name: name)
KeyError: 'a'   # reachable via specs since deps were added for them
```

src/datachain/lib/udf.py
```diff
@@ -582,23 +644,52 @@ def __init__(self, signal_map: dict[str, Callable]):
...
     def hash(self, include_body: bool = True) -> str:
         # cache key must vary with the wrapped functions; the base
         # implementation would hash this class's process method, which is
         # identical across instances.
         parts = [
-            hash_callable(fn, include_body=include_body)
+            hash_callable(
+                fn.process if isinstance(fn, UDFBase) else fn,
+                include_body=include_body,
+            )
```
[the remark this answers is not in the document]
ilongin: This is known and pre-existing limitation of hashing and not related to this PR. If we ever want to fix it it will be in separate issue / PR

src/datachain/lib/udf.py
ilongin: done

src/datachain/lib/udf.py
ilongin: You are right, fixed it

src/datachain/lib/dc/datachain.py
ilongin: Refactored it to include abstract method in BoundSpec  that should return number of outputs.

Multi-signal `.map()` is one output per entry.

- Single-kwarg `.map(pair=fn)` where fn returns `tuple[str, int]`: works, auto-splits into `pair_0`, `pair_1` (routes to single-signal path).
- Multi-kwarg `.map(pair=fn, upper=...)` with a tuple return: rejected at build time with "output type 'tuple' of signal 'pair' is not supported".

The `output_count > 1` reject we just added covers the BoundSpec case (LLMSpec `include_usage`). Plain tuple returns already fail with a clear enough message from `_validate_output`, so no extra test needed unless you want to lock it in.

ilongin: We could though implement multi value / tuple return even in multi signal map ... I will investigate how complex it is .. if simple enough, I can implement in this PR

ilongin: It's not actually clear right now what is the best way to implement it. Having column name as keys in map and then output field with output column names is kind of ambiguous and confusing. We need to brainstorm about this if it's needed in the first place so I would postpone it for now

src/datachain/lib/udf.py
```diff
@@ -71,6 +71,35 @@ class BoundSpec(ABC):
...
+        """Return the chain columns this spec reads, in the order the bound
+        callable expects them. Must be stable across ``bind()``."""
+
+    @property
+    @abstractmethod
+    def output_count(self) -> int:
+        """How many chain columns the bound callable produces per row."""
+
+
+def reject_var_params(
```
shcheklein, line 85: is there a reason to keep it here, no in the (single) place where it is being used?

src/datachain/lib/udf.py
ilongin: Overwriting an existing column via `.map()` is not supported - map only adds new columns, it can't replace one. Single-signal already blocks it with `ValueError: Column already exists`.

Multi-signal had the same limit but no build-time check, so it crashed at runtime with `KeyError: 'a'`. Fixed - now it fails at build time with a clear `DataChainParamsError` asking the user to rename the output or the parameter. Added tests for plain callables and BoundSpec.

src/datachain/lib/udf_signature.py
```diff
@@ -84,7 +91,7 @@ def parse(
             )
 
         udf_params: dict[str, DataType | Any] = {}
-        if params:
+        if params is not None:
```
shcheklein, line 93: why this change?

src/datachain/lib/udf_signature.py
```diff
@@ -68,6 +68,13 @@ def parse(
             cls._func_signature(chain, udf_func)
         )
 
+        if not isinstance(udf_func, AbstractUDF):
```
shcheklein: what is the case for udf_func to be AbstractUDF here?

src/datachain/lib/udf.py
```diff
@@ -549,16 +588,58 @@ class _MultiSignalMapper(Mapper):
...
+        bound_columns: dict[str, list[str]] | None = None,
+    ):
         super().__init__()
         self._signal_map = signal_map
+        self._bound_columns = bound_columns or {}
         output_names = set(signal_map)
         self._per_func_params: dict[str, list[str]] = {}
+        # Entries that are UDFBase subclass instances (e.g. Mapper subclass);
+        # invoked via .process() and their setup()/teardown() run per worker.
+        self._mapper_entries: set[str] = set()
```
shcheklein: I see that we run teardown / setup using _signal_map ? do we still need _mapper_entries ? why?

src/datachain/lib/udf.py
```diff
@@ -71,6 +71,35 @@ class BoundSpec(ABC):
...
     def bind(self, ctx: BindContext) -> Callable: ...
 
+    @abstractmethod
+    def input_columns(self) -> list[str]:
+        """Return the chain columns this spec reads, in the order the bound
+        callable expects them. Must be stable across ``bind()``."""
+
+    @property
+    @abstractmethod
+    def output_count(self) -> int:
```
shcheklein, line 81: we are getting now:

```
def pair(name) -> tuple[str, int]:
    return name, len(name)

chain.map(p=pair, u=lambda name: name.upper())
UdfSignatureError: … output type 'tuple' of signal 'p' is not supported.
Please use DataModel types: BaseModel, int, str, float, …
```

it should be probably "use single map instead" (not like it is not supported) at all

tests/func/test_llm.py
```diff
@@ -62,6 +63,82 @@ def test_map_materializes_typed_columns(fake_llm, test_session):
...
+def test_map_multi_signal_rejects_include_usage(test_session):
+    chain = base(test_session)
+    with pytest.raises(DataChainParamsError, match="produces 2 outputs"):
+        chain.map(
+            scene=llm.complete("text", schema=Scene, include_usage=True),
+            label=llm.complete("text", "x"),
+        )
+
+
+def test_map_multi_signal_spec_reads_sibling_output(fake_llm, test_session):
```
shcheklein, line 126: let's also add a test to cover multiple LLM calls:

```
def test_map_multi_signal_llm_to_llm_chain(fake_llm, test_session):
    fake_llm.text_response = "the summary"
    chain = base(test_session).map(
        summary=llm.complete("text", "summarize"),
        topic=llm.classify("summary", into=["accident", "normal"]),
    )
    rows = chain.to_list("summary", "topic")
    assert all(s == "the summary" for s, _ in rows)
    assert any("the summary" in str(c) for c in fake_llm.calls)
```

src/datachain/lib/udf.py
shcheklein: it is not an existing column overwrite though, no?

src/datachain/lib/udf.py
```diff
@@ -581,24 +662,55 @@ def __init__(self, signal_map: dict[str, Callable]):
...
-                for p in self._per_func_params[name]
-            }
-            results[name] = self._signal_map[name](**fn_kwargs)
+            params = self._per_func_params[name]
+            kw = {p: (results[p] if p in results else row_by_name[p]) for p in params}
+            results[name] = self._callers[name](**kw)
         # Output order follows the user's declared kwarg order, not exec order.
         return tuple(results[name] for name in self._signal_map)
 
+    def setup(self) -> None:
```
shcheklein, line 719: we miss _create / _init?

```
class UsesFrameworkAttrs(dc.Mapper):
    def process(self, name: str) -> str:
        return ",".join(self.signal_names)   # fine in single-signal map

chain.map(sig=UsesFrameworkAttrs(), low=lambda name: name.lower())
AttributeError: 'NoneType' object has no attribute 'to_udf_spec'   # per row
```

src/datachain/lib/udf.py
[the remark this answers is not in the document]
shcheklein: we have though `the _bound_columns routing collision, which this PR introduces` in this case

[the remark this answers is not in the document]
shcheklein:
```
# ColSpec: a minimal conforming BoundSpec; run() = lambda value: f"V:{value}"
.map(x=lambda c1, c2: c1 + c2, a=ColSpec("c1"))   # → ('ab', 'V:a')
.map(x=lambda c1, c2: c1 + c2, a=ColSpec("c2"))   # → ('ab', 'V:b')
udf.hash() is IDENTICAL for both   # different results, same checkpoint key
```

src/datachain/lib/udf_signature.py
ilongin: Needed for the zero-arg multi-signal case, e.g. `.map(x=lambda: "X", y=lambda: "Y")`.

Multi-signal passes `params=multi_mapper.combined_params` to `UdfSignature.parse`. When all entries take zero args that list is `[]`. With `if params:` an empty list is falsy and we fall through to signature inference, which reads `_MultiSignalMapper.process` (`def process(self, *args)`) and ends up with a bogus `args` param. `is not None` keeps `[]` as "user gave us an empty list", so `udf_params = {}` and the zero-arg path works.

src/datachain/lib/udf_signature.py
ilongin: That guard was too broad. It skipped the check for every UDFBase subclass, not just our internal `_MultiSignalMapper` wrapper for which was needed.

Refactored: `reject_var_params` is now called unconditionally. The helper itself picks the check target (`.process` for UDFBase subclasses, the callable itself for plain functions) and the skip cases live in one place. Multi-signal is skipped naturally because it always passes a `params` list (`columns_explicit=True`).

src/datachain/lib/udf.py
ilongin: You are right, it's dead code actually, not used anywhere ... leftover, removing it

src/datachain/lib/udf.py
ilongin: True, improved error handling

tests/func/test_llm.py
ilongin: added

src/datachain/lib/udf.py
ilongin: Nice catch, that was a bug / missing implementation. Fixed it and added tests

src/datachain/lib/udf.py
ilongin: Hm ... you are right I think .. those are 2 different things .. I added explicit check now in lib/udf which should be fine but I will double check this whole thing again

shcheklein:
@ilongin review also supressed comments from Copilot

src/datachain/lib/udf.py
```diff
@@ -71,6 +71,37 @@ class BoundSpec(ABC):
     @abstractmethod
     def bind(self, ctx: BindContext) -> Callable: ...
 
+    @abstractmethod
+    def input_columns(self) -> list[str]:
```
shcheklein, line 75: how is it different from `__datachain_params__` - why in some cases we use one, in other cases another one?

src/datachain/lib/udf.py
```diff
@@ -581,24 +639,97 @@ def __init__(self, signal_map: dict[str, Callable]):
                     seen.add(p)
                     self.combined_params.append(p)
 
+        # Pre-build one uniform kwargs-callable per entry so process() has a
+        # single call path regardless of entry kind.
+        self._callers: dict[str, Callable[..., Any]] = {}
+        for name, fn in signal_map.items():
```
shcheklein, line 649: wrap this into the previous loop?

src/datachain/lib/udf.py
```diff
@@ -549,19 +590,36 @@ class _MultiSignalMapper(Mapper):
...
         # function's output (dependencies) vs from an input row column.
         deps: dict[str, set[str]] = {}
         for name, fn in signal_map.items():
-            params = list(inspect.signature(fn).parameters.keys())
+            if name in self._bound_columns:
+                params = list(self._bound_columns[name])
+            else:
+                params = self._resolve_sig_params(name, fn)
             self._per_func_params[name] = params
             deps[name] = {p for p in params if p in output_names and p != name}
```
shcheklein: double check - is it reachable `p != name`?

shcheklein:
@ilongin there are some unresolved comments still it seems

src/datachain/lib/udf.py
```diff
@@ -581,24 +639,97 @@ def __init__(self, signal_map: dict[str, Callable]):
...
+        return [p.name for p in sig_params]
+
+    def _init(self, sign, params, func):
+        super()._init(sign, params, func)
+        # Give each inner UDFBase entry its own params/output so framework
+        # attrs like self.signal_names, self.output, self.params work.
+        output_values = sign.output_schema.values
+        for name, fn in self._signal_map.items():
+            if isinstance(fn, UDFBase):
+                fn.params = params
```
shcheklein: is it correct? should be some fn-specific params here?

src/datachain/lib/udf.py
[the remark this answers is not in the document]
shcheklein: One more issue found (AI):


> Overriding hash() is the supported way for a class UDF to declare its checkpoint identity — and single-signal map honors it, because there the instance is the UDF. The composite instead calls hash_callable(fn.process) directly, so the same class behaves differently depending on which .map() flavor runs it:

```python
# H(tag): a Mapper whose hash() returns sha256(self.tag)
.map(H("a"), output={"x": str})  # vs H("b") — standalone identities differ ✓
.map(x=H("a"), y=lambda name: name)  # vs H("b") — composite identities IDENTICAL
# the second multi-map run reuses the first's checkpointed output
```

src/datachain/lib/udf.py
ilongin: Yea, we could've used just input_columns, just needed to do some small refactoring .. done

src/datachain/lib/udf.py
ilongin: Yep, not reachable

src/datachain/lib/udf.py
ilongin: Yes, we could make this to accept only those params that are needed .. fixed

src/datachain/lib/udf.py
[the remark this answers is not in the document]
ilongin: fixed both:

- `_bound_columns` routing collision: mixed the routing dict into `_MultiSignalMapper.hash()` so two `.map()` calls that differ only in the routed column now hash differently.
- Mapper `hash()` override bypassed: for inner UDFBase entries we now call `fn.hash(...)` instead of `hash_callable(fn.process, ...)`, so a subclass override is respected (same behavior as single-signal `.map()`).

Added test for both

src/datachain/lib/udf.py
```diff
@@ -71,6 +71,37 @@ class BoundSpec(ABC):
...
+
+
+def reject_var_params(
+    udf_func: "Callable | UDFBase", label: str, *, columns_explicit: bool = False
+) -> None:
+    """Reject user callables with ``*args`` / ``**kwargs`` when we need to
+    infer column routing from the signature. When ``columns_explicit`` is
+    True (caller passed ``params=``), both are safe and pass through. For
+    UDFBase subclasses, the check runs against ``.process``."""
+    if columns_explicit:
```
shcheklein: should we also reject this case:

```
def extract(**row) -> str:
    return row["name"]


chain = dc.read_values(name=["foo"]).map(
    extract,
    params=["name"],
    output={"result": str},
)

chain.to_list("result")
# TypeError: extract() takes 0 positional arguments but 1 was given
``` 

?

src/datachain/lib/udf.py
```diff
@@ -566,26 +622,109 @@ def __init__(self, signal_map: dict[str, Callable]):
...
+        return [p.name for p in sig_params]
+
+    def _init(self, sign, params, func):
+        super()._init(sign, params, func)
+        # Give each inner UDFBase entry its own params/output so framework
+        # attrs like self.signal_names, self.output, self.params work.
+        output_values = sign.output_schema.values
+        for name, fn in self._signal_map.items():
+            if isinstance(fn, UDFBase):
+                fn.params = SignalSchema(
```
shcheklein:
```
class Prefix(dc.Mapper):
    def __init__(self, prefix: str):
        self.prefix = prefix

    def process(self, name: str) -> str:
        return self.prefix + name


chain = dc.read_values(name=["x"]).map(
    prefixed=Prefix("p:"),
    unchanged=lambda name: name,
)

chain.to_list("prefixed")
# AttributeError: 'Prefix' object has no attribute '_func'
``

Standalone mapper instances are initialized through `_create()`. Here, inner mappers receive only direct params and output assignments, so subclasses whose constructor does not call super() never receive _func.

Apparently no super() is supported by datachain. And fails here?

Comment (shcheklein) [inline src/datachain/lib/udf.py]: check also this AI finding:

```
class ReadsParams(dc.Mapper):
    def process(self, prepared: str) -> str:
        typ = self.params.values["prepared"]
        return f"{typ.__name__}:{prepared}"


chain = dc.read_values(name=["x"]).map(
    prepared=lambda name: name.upper(),
    final=ReadsParams(),
)

chain.to_list("final")
# KeyError: "prepared"
```

This comprehension looks only in the outer input schema. Sibling outputs are not present there, even though their values are passed correctly to process().

src/datachain/lib/udf.py
ilongin: Yes, this is an issue - fixed. It's pre-existing on main btw (same raw `TypeError` from Python). Found a similar case while auditing: `def f(*, name)` with explicit `params=["name"]` fails the same way. Same root cause: call site passes values positionally, and a signature with no positional-accepting parameter can't receive them. `reject_var_params` now raises a proper error for both. Test covers both cases.

src/datachain/lib/udf.py
ilongin: Yep, this was the bug .. fixed it and added test. Tried to see if there are some other cases like this, that we didn't covered but so far nothing is found..

ilongin: Confirmed and fixed. `fn.params` was built only from outer input columns, dropping sibling-output params. Fix: pull sibling types from `output_values` too, so `fn.params` matches what flows into `process()`. Test added.

shcheklein (approved):

PR #1867, opened 2026-07-19
fix(models): hydrate nested model collections in UDFs

panDing19:
## Root cause

UDF parameter preparation reconstructed top-level model parameters but passed collection-valued leaves through unchanged. A nested field such as `list[DataModel]` therefore reached the UDF as raw dictionaries. Stream hydration also only followed direct model fields, so `File` values inside model collections did not receive a catalog stream.

The first fix traversed every runtime mapping and collection on every row. That introduced unnecessary collection copies and could recursively walk large or cyclic setup values unrelated to the declared UDF schema.

## Changes

- Convert collection leaves only when their declared annotation requires model hydration or non-string key conversion; preserve ordinary collection objects unchanged.
- Detect model and `File` paths from the declared annotation, with cached and cycle-safe annotation inspection.
- Consolidate stream hydration in `SignalSchema` and use it from both feature conversion and UDF input preparation.
- Traverse nested model/collection/optional annotations only when they can contain `File`, while propagating `download_cb`.
- Avoid traversing arbitrary setup and plain collection values.
- Add regressions for optional dictionaries, nested model collections, plain collection identity, cyclic setup data, and reading a real nested `File` through `map()`.

## Tests

- `pytest -q tests/unit/lib/test_signal_schema.py tests/unit/lib/test_udf.py tests/unit/lib/test_datachain.py` — 495 passed
- `ruff check src/datachain/lib/signal_schema.py src/datachain/lib/udf.py tests/unit/lib/test_signal_schema.py tests/unit/lib/test_udf.py tests/unit/lib/test_datachain.py`
- `ruff format --check src/datachain/lib/signal_schema.py src/datachain/lib/udf.py tests/unit/lib/test_signal_schema.py tests/unit/lib/test_udf.py tests/unit/lib/test_datachain.py`
- `mypy src/datachain/lib/signal_schema.py src/datachain/lib/udf.py --ignore-missing-imports`
- `git diff --check`

## Related

Addresses the performance, setup-recursion, shared traversal, and test-gap findings in the [maintainer review](https://github.com/datachain-ai/datachain/pull/1867#issuecomment-5040057793).

### DIFF
--- a/src/datachain/lib/signal_schema.py
+++ b/src/datachain/lib/signal_schema.py
@@ -615,7 +615,11 @@ class SignalSchema:
                 obj, pos = self._hydrate_model(fr, is_optional, row, pos, label=name)
                 objs.append(obj)
             else:
-                objs.append(row[pos])
+                objs.append(
+                    self._convert_feature_value(
+                        fr_type, row[pos], catalog=None, cache=False
+                    )
+                )
                 pos += 1
         return objs
 
@@ -763,7 +767,7 @@ class SignalSchema:
         self,
         annotation: DataType,
         value: Any,
-        catalog: "Catalog",
+        catalog: "Catalog | None",
         cache: bool,
     ) -> Any:
         """Convert raw DB value into declared annotation if needed."""
@@ -789,7 +793,8 @@ class SignalSchema:
             else:
                 return result
             assert isinstance(obj, BaseModel)
-            SignalSchema._set_file_stream(obj, catalog, cache)
+            if catalog is not None:
+                SignalSchema._set_file_stream(obj, catalog, cache)
             result = obj
         elif origin is list:
             args = get_args(annotation)
--- a/src/datachain/lib/signal_schema.py
+++ b/src/datachain/lib/signal_schema.py
@@ -3,11 +3,12 @@ import hashlib
 import logging
 import math
 import types
+import typing
 import warnings
-from collections.abc import Callable, Iterable, Iterator, Mapping, Sequence
+from collections.abc import Callable, Hashable, Iterable, Iterator, Mapping, Sequence
 from dataclasses import dataclass
 from datetime import datetime
-from functools import cached_property
+from functools import cached_property, lru_cache
 from inspect import isclass
 from typing import (
     IO,
@@ -20,6 +21,7 @@ from typing import (
     get_origin,
 )
 
+from fsspec.callbacks import DEFAULT_CALLBACK, Callback
 from pydantic import BaseModel, Field, ValidationError, create_model
 from sqlalchemy import Cast, asc, cast, desc, nulls_last
 from sqlalchemy.sql.elements import BinaryExpression, Grouping, Label
@@ -615,13 +617,65 @@ class SignalSchema:
                 obj, pos = self._hydrate_model(fr, is_optional, row, pos, label=name)
                 objs.append(obj)
             else:
-                objs.append(
-                    self._convert_feature_value(
-                        fr_type, row[pos], catalog=None, cache=False
+                value = row[pos]
+                if self._requires_feature_conversion(fr_type):
+                    value = self._convert_feature_value(
+                        fr_type, value, catalog=None, cache=False
                     )
-                )
+                objs.append(value)
                 pos += 1
         return objs
+
+    @classmethod
+    def _requires_feature_conversion(cls, annotation: DataType) -> bool:
+        if ModelStore.is_pydantic(annotation):
+            return True
+
+        origin = get_origin(annotation)
+        args = get_args(annotation)
+        if origin in (Union, types.UnionType):
+            inner, has_none = unwrap_optional(annotation)
+            return has_none and cls._requires_feature_conversion(inner)
+        if origin is list:
+            return bool(args) and cls._requires_feature_conversion(args[0])
+        if origin is dict and len(args) == 2:
+            key_type, value_type = args
+            return key_type is not str or cls._requires_feature_conversion(value_type)
+        return False
+
+    @staticmethod
+    def _annotation_contains_type(annotation: Any, target: type) -> bool:
+        return SignalSchema._annotation_contains_type_cached(
+            typing.cast("Hashable", annotation), target
+        )
+
+    @staticmethod
+    @lru_cache
+    def _annotation_contains_type_cached(annotation: Hashable, target: type) -> bool:
+        def contains(current: Any, seen: set[type]) -> bool:
+            if isclass(current) and issubclass(current, target):
+                return True
+
+            origin = get_origin(current)
+            if origin in (Union, types.UnionType, list, dict, tuple, set):
+                return any(
+                    arg is not Ellipsis
+                    and arg is not type(None)
+                    and contains(arg, seen)
+                    for arg in get_args(current)
+                )
+
+            model = ModelStore.to_pydantic(current)
+            if model is None or model in seen:
+                return False
+            seen.add(model)
+            return any(
+                contains(field.annotation, seen)
+                for field in model.model_fields.values()
+                if field.annotation is not None
+            )
+
+        return contains(annotation, set())
 
     @staticmethod
     def _all_values_none(value: Any) -> bool:
@@ -830,17 +884,50 @@ class SignalSchema:
 
     @staticmethod
     def _set_file_stream(
-        obj: BaseModel, catalog: "Catalog", cache: bool = False
+        obj: Any,
+        catalog: "Catalog",
+        cache: bool = False,
+        download_cb: Callback = DEFAULT_CALLBACK,
+        annotation: DataType | None = None,
+        _seen: set[int] | None = None,
     ) -> None:
+        seen = _seen if _seen is not None else set()
+        if id(obj) in seen:
+            return
+        seen.add(id(obj))
+
         if isinstance(obj, File):
-            obj._set_stream(catalog, caching_enabled=cache)
-        for field, finfo in type(obj).model_fields.items():
-            # Unwrap Optional so a File under Optional[...] still gets its stream.
-            inner, _ = unwrap_optional(finfo.annotation)
-            if ModelStore.is_pydantic(inner):
-                value = getattr(obj, field)
-                if value is not None:
-                    SignalSchema._set_file_stream(value, catalog, cache)
+            obj._set_stream(catalog, caching_enabled=cache, download_cb=download_cb)
+
+        if isinstance(obj, BaseModel):
+            for field, finfo in type(obj).model_fields.items():
+                field_annotation = finfo.annotation
+                if (
+                    field_annotation is not None
+                    and SignalSchema._annotation_contains_type(field_annotation, File)
+                ):
+                    SignalSchema._set_file_stream(
+                        getattr(obj, field),
+                        catalog,
+                        cache,
+                        download_cb,
+                        field_annotation,
+                        seen,
+                    )
+        elif isinstance(obj, Mapping):
+            args = get_args(annotation) if annotation is not None else ()
+            value_annotation = args[1] if len(args) == 2 else None
+            for value in obj.values():
+                SignalSchema._set_file_stream(
+                    value, catalog, cache, download_cb, value_annotation, seen
+                )
+        elif isinstance(obj, (list, tuple, set)):
+            args = get_args(annotation) if annotation is not None else ()
+            item_annotation = args[0] if args else None
+            for value in obj:
+                SignalSchema._set_file_stream(
+                    value, catalog, cache, download_cb, item_annotation, seen
+                )
 
     def get_column_type(self, col_name: str, with_subtree: bool = False) -> DataType:
         """
--- a/src/datachain/lib/signal_schema.py
+++ b/src/datachain/lib/signal_schema.py
@@ -3,7 +3,6 @@ import hashlib
 import logging
 import math
 import types
-import typing
 import warnings
 from collections.abc import Callable, Hashable, Iterable, Iterator, Mapping, Sequence
 from dataclasses import dataclass
@@ -20,6 +19,7 @@ from typing import (
     get_args,
     get_origin,
 )
+from typing import cast as typing_cast
 
 from fsspec.callbacks import DEFAULT_CALLBACK, Callback
 from pydantic import BaseModel, Field, ValidationError, create_model
@@ -618,7 +618,7 @@ class SignalSchema:
                 objs.append(obj)
             else:
                 value = row[pos]
-                if self._requires_feature_conversion(fr_type):
+                if self._row_conversion_required[name]:
                     value = self._convert_feature_value(
                         fr_type, value, catalog=None, cache=False
                     )
@@ -626,8 +626,15 @@ class SignalSchema:
                 pos += 1
         return objs
 
+    @cached_property
+    def _row_conversion_required(self) -> dict[str, bool]:
+        return {
+            name: self._requires_row_conversion(annotation)
+            for name, annotation in self.values.items()
+        }
+
     @classmethod
-    def _requires_feature_conversion(cls, annotation: DataType) -> bool:
+    def _requires_row_conversion(cls, annotation: DataType) -> bool:
         if ModelStore.is_pydantic(annotation):
             return True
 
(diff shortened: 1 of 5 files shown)


shcheklein:
> [!NOTE]
> This review was **AI-generated** (Claude Code) — posting for review and follow-up fixes. Please verify claims before acting on them; benchmark numbers come from a local `timeit` micro-benchmark (script at the bottom), absolute values will vary — the ratios are the point.

## Findings

### 1. Hot-path performance regression (major)

Both changed functions run **per row** (`_parse_row` → `row_to_objs` + `_set_stream_recursive` per object, `src/datachain/lib/udf.py:357-365`). Micro-benchmark of before/after:

| Case | Before | After |
|---|---|---|
| `row_to_objs`, leaf `list[float]` (1536-dim embedding) | ~0 µs/row | **~480 µs/row** + full list copy |
| `_set_stream_recursive`, model/list with 1536-float array | 0.2 µs/row | **~345 µs/row** |
| `_set_stream_recursive`, 10k-entry dict **setup value** | 0.2 µs/row | **~2,300 µs/row** |
| `_set_stream_recursive`, plain `File` | 0.8 µs/row | 2.5 µs/row (fine) |
| scalar leaf (`str`) | ~0 | 0.33 µs/row (fine) |

A `map()` over 1M rows with an embedding param now spends **~8 minutes** rebuilding lists plus **~6 minutes** traversing them, doing nothing. This regresses even for users who never touch nested models — any hydrated model with a `list[float]` field now gets its array walked element-by-element, where the old code skipped non-`DataModel` fields entirely.

### 2. Setup values are now deep-traversed every row (major)

`row_to_objs` appends **setup values** (arbitrary user objects: vocab dicts, config lists, pydantic settings) into `obj_row`, and `_parse_row` runs `_set_stream_recursive` over all of them. Previously a no-op for anything that isn't a `DataModel`; now:

- a static 10k-entry dict costs ~2.3 ms/row, every row (see table);
- a **self-referential** structure (`d["self"] = d`, or any cyclic object graph) raises `RecursionError` — verified. New failure mode for previously-working pipelines.

### 3. Same bug remains in the `to_iter()`/`collect()` path (design)

The PR fixes stream attachment for UDF *inputs* only. `row_to_features` (used by `to_iter()`, `src/datachain/lib/dc/datachain.py:1735`) relies on `SignalSchema._set_file_stream`, which recurses only into *pydantic-typed fields* — `ModelStore.is_pydantic(list[File])` is `False`, so Files inside a model's `pages: list[File]` field still come out of `to_iter()` with no catalog and fail on `.open()`. The codebase now has **two divergent recursive stream-setters** (`UDFBase._set_stream_recursive` vs `SignalSchema._set_file_stream`) with different traversal semantics and different `download_cb` support.

### 4. Silent behavior changes for existing UDFs

- `dict[int, ...]`-typed params: keys were raw JSON strings (`"1"`), now `json.loads`-converted (`1`).
- Collection-of-model params: raw dicts → model instances. Any UDF written defensively against the old behavior (`item["count"]`) now breaks.

Arguably both are fixes (they align UDF params with `to_iter()`), but they change what existing UDFs receive.

### 5. Test gaps

- No end-to-end test that actually **reads** a File from a nested collection inside a `map()` with a real catalog — the unit test's `catalog = object()` won't catch signature/semantics drift in `_set_stream`. Prefetch depends on the same wiring (`_prefetch` bails out silently when `_catalog` is None), so a miss here also silently disables prefetching.
- The dict branch (`dict[str, MyFr]`) and Optional-unwrap branch (`Optional[list[MyFr]]`) of the new leaf path are untested through `map()`.
- No batch UDF variant (`BatchMapper` shares `_prepare_row`, so it should just work — worth pinning).

### 6. Minor

- `_convert_feature_value(..., catalog=None, cache=False)` — `cache` is meaningless when `catalog is None`; consider giving `catalog` a `None` default instead of threading dead args.
- PR body says "Allow model hydration without a catalog during row conversion," but `row_to_objs`'s `_hydrate_model` call already ran without a catalog on main — only `_convert_feature_value` needed the change. Cosmetic.

## Recommendations

1. **Schema-driven gating — decide once per param, not once per element per row.** The schema is static:
   - In `SignalSchema`, precompute per leaf whether its annotation tree can contain a pydantic model or a non-str dict key; only those need `_convert_feature_value`, otherwise keep the old passthrough.
   - In `_parse_row`, only run `_set_stream_recursive` on params whose declared type can contain a `File`. This also fixes finding 2 for free: setup params are declared as `str` in `slice()`, so they'd never be traversed (alternatively, skip setup-param positions explicitly).
2. **Consolidate the two stream-traversal helpers** into one shared implementation (with `download_cb` support) used by both the UDF path and `row_to_features` — or at minimum file a follow-up issue for the `to_iter()` asymmetry.
3. Add the missing tests from finding 5 (real-catalog read of a File in a nested collection, dict/Optional leaf params, batch variant).
4. Changelog/release-notes callout for the behavior changes in finding 4.

<details>
<summary>Benchmark script (reproducible)</summary>

```python
"""Micro-benchmark of PR #1867 hot-path changes (simulated on main)."""

import timeit
from collections.abc import Mapping

# --- 1. row_to_objs leaf conversion: old (passthrough) vs new (_convert_feature_value)
schema = SignalSchema({"emb": list[float]})
emb = [float(i) for i in range(1536)]  # typical embedding column

old = timeit.timeit(lambda: emb, number=10_000)
new = timeit.timeit(
    lambda: schema._convert_feature_value(list[float], emb, None, False),
    number=10_000,
)
print(f"row_to_objs leaf list[float][1536]: old {old*100:.1f}us/row  new {new*100:.1f}us/row")

s = timeit.timeit(
    lambda: schema._convert_feature_value(str, "hello", None, False), number=100_000
)
print(f"row_to_objs leaf str:               new {s*10:.2f}us/row (old ~0)")


# --- 2. _set_stream_recursive: old vs new traversal on the same inputs
def set_stream_old(obj):
    from datachain.lib.file import DataModel

    if isinstance(obj, File):
        pass  # _set_stream: 3 attr assignments
    if isinstance(obj, DataModel):
        for field_name in type(obj).model_fields:
            field_value = getattr(obj, field_name, None)
            if isinstance(field_value, DataModel):
                set_stream_old(field_value)


def set_stream_new(obj):
    if isinstance(obj, File):
        pass
    if isinstance(obj, BaseModel):
        for field_name in type(obj).model_fields:
            set_stream_new(getattr(obj, field_name, None))
    elif isinstance(obj, Mapping):
        for value in obj.values():
            set_stream_new(value)
    elif isinstance(obj, (list, tuple, set)):
        for value in obj:
            set_stream_new(value)


for label, value in [
    ("list[float][1536]", emb),
    ("File()", File(path="x")),
    ("dict vocab (10k entries)", {str(i): i for i in range(10_000)}),
]:
    o = timeit.timeit(lambda v=value: set_stream_old(v), number=1_000)
    n = timeit.timeit(lambda v=value: set_stream_new(v), number=1_000)
    print(f"_set_stream_recursive {label}: old {o*1000:.1f}us/row  new {n*1000:.1f}us/row")

# --- 3. cycle: what the new traversal does on a self-referential setup value
d: dict = {}
d["self"] = d
try:
    set_stream_new(d)
    print("cycle: ok")
except RecursionError:
    print("cycle: RecursionError")
```

</details>

shcheklein:
@panDing19 thanks! please check the review I generated with AI. It seems we need to address most of it in some way.

shcheklein (changes requested):
check the comments above please

panDing19:
Addressed the review findings in `0ae9f497`:\n\n- schema-gated conversion preserves plain collections and avoids the embedding-list copy regression;\n- UDF hydration runs only for annotations that can contain `File`, so arbitrary/cyclic setup values are not traversed;\n- `SignalSchema` now owns the shared annotation-aware stream traversal used by UDFs and `row_to_features()`;\n- nested callbacks are propagated;\n- added optional-dict, real nested-file read, plain-collection identity, and cyclic-setup regressions.\n\nValidation after rebasing onto current `upstream/main`: 495 affected tests passed, Ruff check/format passed, mypy passed, and `git diff --check` passed.

panDing19:
CI follow-up: lint, benchmark, pre-commit, and the complete DataChain matrix all passed on the updated commit. The five failing example jobs are credential-only failures: four `multimodal` jobs report missing `OPENAI_API_KEY`, and `llm_and_nlp` reports missing `ANTHROPIC_API_KEY` / `HF_TOKEN`. These are unrelated to the changed modules and reproduce the credential limitation already present in the prior run.

panDing19:
Synced the branch with current `main` (`e840dda3`) and revalidated the affected scope: 495 tests passed; Ruff check/format, mypy, and `git diff --check` all passed. The review findings remain addressed in `0ae9f497`. @shcheklein, please re-review when convenient.

panDing19:
Synced with current `main` in `aa8aaa59`. Local validation remains green: 495 affected tests passed; Ruff check/format, mypy, static security scan, and `git diff --check` passed.\n\nThe refreshed core matrix passed on every platform/Python combination except Ubuntu Python 3.14, where all 5,171 functional tests passed but the import-time guard measured 801 ms against an 800 ms limit (`tests/test_import_time.py`). An isolated same-environment A/B run (10 fresh imports each) showed no PR regression: upstream `main` median 333.726 ms / min 329.624 ms; PR median 333.461 ms / min 330.349 ms. This is a 1 ms threshold flake. I attempted to rerun only that job, but fork contributors lack the required repository permission (HTTP 403). Could a maintainer rerun job 90184367038?\n\nThe five example failures remain the known missing-credential jobs (`OPENAI_API_KEY`, `ANTHROPIC_API_KEY`, and `HF_TOKEN`), unrelated to this diff.

panDing19:
Synced with current `main` in `6c34292d` and revalidated the affected scope locally: 495 tests passed; Ruff check/format, mypy, static security scan, and `git diff --check` passed.

The refreshed core CI matrix passed on 9/10 platform/Python combinations. Ubuntu Python 3.12 passed all 5,171 functional tests but the import-time guard measured 824 ms against its 800 ms limit (`tests/test_import_time.py`): https://github.com/datachain-ai/datachain/actions/runs/30423971998/job/90486278188. This is the same threshold-only flake that previously landed on Ubuntu Python 3.14; that Python 3.14 job now passes. I attempted to rerun the new failed job, but GitHub rejected the rerun for this fork contributor. Could a maintainer rerun job 90486278188?

The five allowed-to-fail example jobs remain the known missing-credential failures (`OPENAI_API_KEY`, `ANTHROPIC_API_KEY`, and `HF_TOKEN`), unrelated to this diff.

src/datachain/lib/udf.py
```diff
@@ -385,23 +385,29 @@ def _parse_row(
         assert self.params
         row = [row_dict[p] for p in self.params.to_udf_spec()]
         obj_row = self.params.row_to_objs(row)
-        for obj in obj_row:
-            self._set_stream_recursive(obj, catalog, cache, download_cb)
+        for obj, annotation in zip(obj_row, self.params.values.values(), strict=True):
+            if self.params._annotation_contains_type(annotation, File):
```
shcheklein: Can it be a VideoFile for example here?

src/datachain/lib/signal_schema.py
```diff
@@ -615,10 +617,66 @@ def row_to_objs(self, row: Sequence[Any]) -> list[Any]:
                 obj, pos = self._hydrate_model(fr, is_optional, row, pos, label=name)
                 objs.append(obj)
             else:
-                objs.append(row[pos])
+                value = row[pos]
+                if self._requires_feature_conversion(fr_type):
```
shcheklein: Should this check also be cached?

src/datachain/lib/signal_schema.py
```diff
@@ -615,10 +617,66 @@ def row_to_objs(self, row: Sequence[Any]) -> list[Any]:
...
+            return has_none and cls._requires_feature_conversion(inner)
+        if origin is list:
+            return bool(args) and cls._requires_feature_conversion(args[0])
+        if origin is dict and len(args) == 2:
+            key_type, value_type = args
+            return key_type is not str or cls._requires_feature_conversion(value_type)
+        return False
+
+    @staticmethod
+    def _annotation_contains_type(annotation: Any, target: type) -> bool:
```
shcheklein: Inline? Plus we need a better name here since it seems this is only about dicts?

shcheklein: What is the point in having a separate method that just wraps the second one?

src/datachain/lib/signal_schema.py
```diff
@@ -825,17 +884,50 @@ def _convert_feature_value(
...
     def _set_file_stream(
-        obj: BaseModel, catalog: "Catalog", cache: bool = False
+        obj: Any,
+        catalog: "Catalog",
+        cache: bool = False,
+        download_cb: Callback = DEFAULT_CALLBACK,
+        annotation: DataType | None = None,
+        _seen: set[int] | None = None,
     ) -> None:
+        seen = _seen if _seen is not None else set()
```
shcheklein, line 895: Could you explain again, why do we need to have seen set?

tests/unit/lib/test_datachain.py
```diff
@@ -911,6 +911,70 @@ class _TestFr(BaseModel):
         assert x.my_name == test_fr.my_name
 
 
+def test_map_hydrates_models_in_nested_collection_param(test_session):
```
shcheklein: What does it mean? What are we testing for?

src/datachain/lib/udf.py
panDing19: Yes. `_annotation_contains_type(annotation, File)` uses `issubclass`, so `VideoFile` and other `File` subclasses match. I kept the base-class check so every supported file subtype gets stream attachment.

src/datachain/lib/signal_schema.py
panDing19: Yes. Addressed in c784cd3a: the conversion decision is now computed once per schema in the cached `_row_conversion_required` mapping, rather than re-walking the annotation for every row.

src/datachain/lib/signal_schema.py
panDing19: Addressed in c784cd3a. I renamed the predicate to `_requires_row_conversion` because it covers both nested model hydration and non-string dict-key conversion; the per-schema cached mapping now makes the call site a direct lookup.

panDing19: Agreed. Removed the UDF wrapper in c784cd3a; `_parse_row` now calls the shared `SignalSchema._set_file_stream` implementation directly.

src/datachain/lib/signal_schema.py
panDing19: The set prevents infinite recursion when a hydrated object graph contains a cycle, and avoids revisiting a shared object reachable through multiple fields. This is defensive at the traversal boundary because pydantic models can be constructed with cyclic references even when their declared schema is finite; ordinary acyclic inputs only add one identity lookup per visited object.

tests/unit/lib/test_datachain.py
panDing19: It verifies the reported bug end to end: a DB-backed nested `list[MyFr]` leaf passed via `params=["collection.items"]` must reach the map function as `MyFr` instances rather than raw dicts. Renamed it to `test_map_converts_nested_collection_items_to_models` in c784cd3a to state that contract explicitly.

panDing19:
Addressed the current-head review in `c784cd3a`: cached the row-conversion decision once per schema, renamed the predicate to describe both model and dict-key conversion, removed the redundant UDF stream wrapper, and clarified the nested-model test name. I also replied inline about `File` subclasses and cycle protection.\n\nValidation: 495 affected tests passed; Ruff check/format on all changed files passed; mypy on both changed source modules passed; `git diff --check` passed.

panDing19:
CI is complete on `c784cd3a`: lint, benchmarks, pre-commit, patch coverage, and all 10 DataChain platform/Python jobs passed. The five remaining example failures are the same credential-only jobs: four multimodal jobs require `OPENAI_API_KEY`, and the `llm_and_nlp` job requires `ANTHROPIC_API_KEY` / `HF_TOKEN`; their logs show those environment variables are empty. These failures are unrelated to the changed modules. The branch is clean and the current-head review feedback has been addressed; ready for re-review.

panDing19:
Synced the branch with current `main` in `25bedd0c` and revalidated the affected plus newly merged listing scope locally: 572 tests passed; Ruff check/format, mypy, static security scan, and `git diff --check` passed.

The refreshed core matrix passed on 9/10 platform/Python combinations. Ubuntu Python 3.14 passed 5,176 functional tests but the import-time guard measured 851 ms against its 800 ms limit: https://github.com/datachain-ai/datachain/actions/runs/30786069756/job/91599733580. An isolated local run passed all three attempts (336/335/332 ms), and this PR has previously shown the same threshold flake moving between Python jobs while the remaining core matrix passes. Fork permissions still do not allow me to rerun the job. Could a maintainer rerun job 91599733580?

The five example failures remain the known missing-credential jobs (`OPENAI_API_KEY`, `ANTHROPIC_API_KEY`, and `HF_TOKEN`), unrelated to this diff.

panDing19:
Synced the branch with current `main` in `8352f2a7`. Local validation remains green for the PR scope: 495 affected tests passed; Ruff check/format, mypy, static security scan, and `git diff --check` passed. The newly merged upstream unit/storage-pattern scope also passed (96 tests).

The refreshed core CI matrix passed on 9/10 platform/Python combinations. Ubuntu Python 3.12 passed all 5,200 functional tests but the import-time guard measured 832 ms against its 800 ms limit: https://github.com/datachain-ai/datachain/actions/runs/30880523254/job/91900712830. This is the same threshold-only flake previously seen on this PR moving among Ubuntu Python versions; Ubuntu 3.10, 3.11, 3.13, and 3.14 all pass on the current head. I attempted to rerun the failed job, but GitHub reports that the job cannot be rerun from this fork contributor account. Could a maintainer rerun job 91900712830?

The five example failures remain the known missing-credential jobs (`OPENAI_API_KEY`, `ANTHROPIC_API_KEY`, and `HF_TOKEN`), unrelated to this diff.

shcheklein (approved):
Failures are unrelated, there will be a follow with some improvements to move faster on this.

PR #1871, opened 2026-07-21
fix: preserve interactive sessions after exceptions

### LINKED ISSUE #1157: Calling not existing method breaks chain
### Description

```python
>>> import datachain as dc
>>> meta = dc.read_json("gs://datachain-demo/dogs-and-cats/*json", column="meta", anon=True).persist()
>>>
>>> meta.no_existing_method()  # HERE
AttributeError: 'DataChain' object has no attribute 'no_existing_method'

>>> # THe chain got broken - no methods works
>>> meta.show()
...
  File "/Users/dmitry/venv/py312/lib/python3.12/site-packages/sqlalchemy/sql/schema.py", line 892, in _autoload
    conn_insp.reflect_table(
  File "/Users/dmitry/venv/py312/lib/python3.12/site-packages/sqlalchemy/engine/reflection.py", line 1548, in reflect_table
    raise exc.NoSuchTableError(table_name)
sqlalchemy.exc.NoSuchTableError: ds_local_local_session_global_4810f6_be5ce5_1_0_0
```

### Version Info

```Text
0.19.3.dev5+g75f06ea5
Python 3.12.10
```

panDing19:
## Root cause

DataChain installs `Session.except_hook` as the process exception hook. The hook finalized the active job and removed temporary datasets for every uncaught exception. That is correct when a script terminates, but the standard Python REPL invokes the hook and then continues running. A simple typo after `.persist()` therefore deleted the persisted table while the chain was still usable, causing the next operation to fail with `TableMissingError`.

## Changes

- Detect the standard interactive interpreter through `sys.ps1`.
- Delegate interactive exceptions to the original exception hook without finalizing the job or cleaning the session.
- Add a regression that verifies the traceback hook is called, the job stays running, and persisted values remain readable.

Non-interactive exceptions keep the existing failure finalization and cleanup behavior.

## Tests

- `python -m pytest tests/unit/test_job_management.py tests/unit/test_session.py tests/func/test_session.py -q` (27 passed)
- `ruff check src/datachain/query/session.py tests/unit/test_job_management.py`
- `ruff format --check src/datachain/query/session.py tests/unit/test_job_management.py`
- `mypy src/datachain/query/session.py tests/unit/test_job_management.py`
- Real `python -i` smoke test: an uncaught missing-method `AttributeError` is displayed and the persisted chain still returns `[1, 2]`
- `git diff --check`

Closes #1157

### DIFF
--- a/src/datachain/query/session.py
+++ b/src/datachain/query/session.py
@@ -340,6 +340,11 @@ class Session:
 
     @staticmethod
     def except_hook(exc_type, exc_value, exc_traceback):
+        if hasattr(sys, "ps1"):
+            if Session.ORIGINAL_EXCEPT_HOOK:
+                Session.ORIGINAL_EXCEPT_HOOK(exc_type, exc_value, exc_traceback)
+            return
+
         if Session.GLOBAL_SESSION_CTX:
             # Handle KeyboardInterrupt specially - mark as canceled and exit with
             # signal code
--- a/src/datachain/query/session.py
+++ b/src/datachain/query/session.py
@@ -340,9 +340,9 @@ class Session:
 
     @staticmethod
     def except_hook(exc_type, exc_value, exc_traceback):
-        if hasattr(sys, "ps1"):
-            if Session.ORIGINAL_EXCEPT_HOOK:
-                Session.ORIGINAL_EXCEPT_HOOK(exc_type, exc_value, exc_traceback)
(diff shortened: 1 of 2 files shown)


shcheklein:
@panDing19 please review the comment (Copilot review)

shcheklein (approved):

PR #1877, opened 2026-07-22
fix: verify registered models against the stored spec on deserialization

### LINKED ISSUE #1872: Custom model name collisions: same-named generated models shadow each other across datasets
### Description

Schema deserialization resolves custom types by **name + version only**. When two datasets contain generated models that share a name but have different shapes (e.g. both were created by `explode()` on a column named `json`, or by `read_parquet()` from files that both have an `items` list-of-struct column), only one of them can be read correctly per process — whichever is loaded first registers its model in `ModelStore` and silently shadows the other.

### Repro

```python
import datachain as dc

c1 = dc.read_values(json=[{"a": 1}]).explode("json")    # generates JsonExplodedModel {a}
c2 = dc.read_values(json=[{"b": "x"}]).explode("json")  # generates JsonExplodedModel {b}
c1.save("e1")
c2.save("e2")

dc.read_dataset("e2").to_values("json_expl")  # works
dc.read_dataset("e1").to_values("json_expl")  # KeyError: 'json_expl__b'
```

In a **fresh process** the winner flips with read order:

```python
import datachain as dc

dc.read_dataset("e1").to_values("json_expl")  # works
dc.read_dataset("e2").to_values("json_expl")  # KeyError: 'json_expl__a'
```

Depending on how the shapes overlap, the failure is a `KeyError` (missing DB column), a Pydantic `ValidationError: Extra inputs are not permitted`, or — if the shapes share field names — a silent misread.

### Root cause

`SignalSchema._deserialize_custom_type` (src/datachain/lib/signal_schema.py) has the complete field spec available inline in the dataset's serialized schema (`custom_types[type_name].fields`), but before using it, it short-circuits:

```python
if fr := ModelStore.get(model_name, target_version):
    return fr
```

Any already-registered model with a matching `name@version` is returned unconditionally — no comparison against the stored spec, even though `CustomType` already carries a `partial_fingerprint` that could detect the mismatch. All generated models are `@v1`, so name+version always tie. Since `ModelStore` is process-global, the first same-named model to be registered (via deserialization, schema inference, or a user-defined class) wins for the rest of the process.

Name collisions are easy to produce because inferred-schema model names derive from column names:

- `explode()` → `f"{col.title()}ExplodedModel"` — everyone's `json` column becomes `JsonExplodedModel`
- structs in lists/columns → `ArrowDataModel_{column}` — two parquet files with an `items` column of different shapes collide (common names: `items`, `data`, `entries`, ...)

Note: the data and each dataset's stored schema are intact — this is read-time shadowing, not persistent corruption. Workarounds today: `explode(model_name=...)`, or unique column names.

### Suggested fix

In `_deserialize_custom_type`, when `type_name` is present in `custom_types`, validate the registered model against the stored spec (fingerprint or field-by-field) before returning it; on mismatch, prefer recreating the model from the stored spec (possibly under a disambiguated runtime name) instead of returning the registered one. The ingredients (`fields`, `partial_fingerprint`) are already serialized.

### Related

Found while reviewing #1860, which fixes the *within-one-schema* variant of the collision (all struct-in-list models previously shared the single name `ArrowDataModel_`, corrupting the serialized schema of any dataset with two differently-shaped list-of-struct columns). The cross-dataset shadowing described here is independent of that fix and cannot be solved by naming alone.

### Version Info

```text
main @ 0a764bfd (reproduced 2026-07-20)
Python 3.13
```

rahul188:
## Problem

`SignalSchema._deserialize_custom_type` resolves custom types by `name@version` alone:

```python
if fr := ModelStore.get(model_name, target_version):
    return fr
```

Since `ModelStore` is process-global and all generated models are `@v1`, two datasets whose inferred models share a name (both `explode()`d from a column named `json`, two parquet files with an `items` list-of-struct column, …) shadow each other: whichever model registers first is used to read every same-named dataset for the rest of the process. Depending on how the shapes overlap that surfaces as a `KeyError` (missing DB column), a Pydantic `ValidationError`, or a silent misread. Repro from #1872:

```python
c1 = dc.read_values(json=[{"a": 1}]).explode("json")
c2 = dc.read_values(json=[{"b": "x"}]).explode("json")
c1.save("e1"); c2.save("e2")
dc.read_dataset("e2").to_values("json_expl")  # works
dc.read_dataset("e1").to_values("json_expl")  # KeyError: 'json_expl__b'
```

## Fix

Following the approach sketched in the issue:

- A registered model resolves a custom type only when its field spec matches the spec stored in the dataset schema. The check reserializes the registered model with the existing `_serialize_custom_model` machinery and compares `fields` per custom type, so a divergence in a *nested* model (identical top-level field strings, different `Inner@v1` shapes) is caught too.
- On mismatch, the model is recreated from the stored spec under a spec-fingerprinted alias (`{name}_{fingerprint[:10]}`, the same shape `to_partial` uses for partials) instead of clobbering the registered model's name in `ModelStore`. Repeated reads of the same dataset hit the alias cache and reuse the recreated class.
- When there is no registered model, behavior is unchanged (recreate under the stored name and register).

The fingerprint covers the transitive closure of the type's spec (fields, bases, hidden fields of every reachable custom type), so two colliding specs that differ only in a nested type or base still get distinct aliases.

## Testing

- `test_deserialize_same_name_different_shape_not_shadowed` — the cross-dataset collision, plus re-read identity (same class object on repeated deserialization of both specs)
- `test_deserialize_nested_shape_divergence_not_shadowed` — identical top-level specs, diverging nested model
- `test_deserialize_prefers_registered_model_when_spec_matches` — round-trip still returns the registered class itself, not a recreation
- `test_explode_same_column_name_across_datasets` — the issue's end-to-end repro (`explode` → `save` → `read_dataset` both ways)

All four were run against unmodified `main` first: the three collision tests fail there (the e2e test with `KeyError: 'json_expl__a'`), and the registered-model test passes on both, confirming no regression of round-trip identity. Full `tests/unit/lib/test_signal_schema.py` (120) and `tests/unit/lib/test_datachain.py` (349) pass; `ruff check`, `ruff format --check`, and `mypy` are clean on the changed files.

Fixes #1872

### DIFF
--- a/src/datachain/lib/signal_schema.py
+++ b/src/datachain/lib/signal_schema.py
@@ -2,6 +2,7 @@ import copy
 import hashlib
 import logging
 import math
+import re
 import types
 import warnings
 from collections.abc import Callable, Iterable, Iterator, Mapping, Sequence
@@ -416,12 +417,57 @@ class SignalSchema:
         return subtypes
 
     @staticmethod
+    def _model_matches_custom_type(
+        fr: type[BaseModel], type_name: str, custom_types: dict[str, Any]
+    ) -> bool:
+        """Check that a registered model's field spec equals the spec stored for
+        ``type_name``, including every nested custom type."""
+        if ModelStore.get_name(fr) != type_name:
+            return False
+        reserialized: dict[str, Any] = {}
+        try:
+            SignalSchema._serialize_custom_model(type_name, fr, reserialized)
+            for name, data in reserialized.items():
+                if name not in custom_types:
+                    return False
+                stored = CustomType.deserialize(custom_types[name], name)
+                current = CustomType.deserialize(data, name)
+                if current.fields != stored.fields:
+                    return False
+        except ValidationError:
+            return False
+        return True
+
+    @staticmethod
+    def _custom_type_fingerprint(type_name: str, custom_types: dict[str, Any]) -> str:
+        """Deterministic fingerprint of a stored custom type spec, covering every
+        nested custom type reachable from it."""
+        spec: dict[str, Any] = {}
+        pending = [type_name]
+        while pending:
+            name = pending.pop()
+            if name in spec or name not in custom_types:
+                continue
+            try:
+                ct = CustomType.deserialize(custom_types[name], name)
+            except ValidationError:
+                continue
+            spec[name] = ct.model_dump(exclude={"schema_version"}, exclude_none=True)
+            for field_type in ct.fields.values():
+                pending.extend(re.findall(r"[A-Za-z_]\w*(?:@v\d+)?", field_type))
+        payload = json.dumps(spec, sort_keys=True, separators=(",", ":"))
+        return hashlib.sha256(payload.encode("utf-8")).hexdigest()
+
+    @staticmethod
     def _deserialize_custom_type(
         type_name: str, custom_types: dict[str, Any]
     ) -> type | None:
         """Given a type name like MyType@v1 gets a type from ModelStore or recreates
         it based on the information from the custom types dict that includes fields and
-        bases."""
+        bases. A registered model resolves the type only when it matches the stored
+        spec; a same-named model with a different shape is recreated from the spec
+        under a spec-fingerprinted alias, so reads are not shadowed by whichever
+        same-named model happened to register first."""
         model_name, target_version = ModelStore.parse_name_version(type_name)
 
         if type_name in custom_types:
@@ -433,7 +479,15 @@ class SignalSchema:
                 ) from exc
 
             if fr := ModelStore.get(model_name, target_version):
-                return fr
+                if SignalSchema._model_matches_custom_type(fr, type_name, custom_types):
+                    return fr
+                fingerprint = SignalSchema._custom_type_fingerprint(
+                    type_name, custom_types
+                )
+                alias = f"{model_name}_{fingerprint[:10]}"
+                if cached := ModelStore.get(alias, target_version):
+                    return cached
+                type_name = f"{alias}@v{target_version}" if target_version else alias
 
             fields = {
                 field_name: SignalSchema._resolve_type(field_type_str, custom_types)
--- a/tests/unit/lib/test_datachain.py
+++ b/tests/unit/lib/test_datachain.py
@@ -1715,6 +1715,25 @@ def test_explode_list_with_null_items(test_session):
     assert chain.to_values("json_expl.email") == [["user@example.com", None]]
 
 
+def test_explode_same_column_name_across_datasets(test_session):
+    dc.read_values(json=[{"a": 1}], session=test_session).explode("json").save(
+        "expl_collide_a"
+    )
+    dc.read_values(json=[{"b": "x"}], session=test_session).explode("json").save(
+        "expl_collide_b"
+    )
+
+    values_b = dc.read_dataset("expl_collide_b", session=test_session).to_values(
+        "json_expl"
+    )
+    values_a = dc.read_dataset("expl_collide_a", session=test_session).to_values(
+        "json_expl"
+    )
+
+    assert values_b[0].b == "x"
+    assert values_a[0].a == 1
+
+
 def test_explode_raises_on_wrong_column_type(test_session):
     chain = dc.read_values(f1=features, session=test_session)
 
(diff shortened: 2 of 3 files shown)


rahul188:
Hi 👋 Gentle nudge on this PR — it's **mergeable and CI is passing**, just waiting on a review. Would really appreciate a look whenever you have bandwidth, and I'm glad to make any changes you'd like. Thanks for your time and for maintaining this project! 🙏

shcheklein (commented):
**AI-generated review (Codex), posted as part of the requested review of #1969 and related PRs.**

I checked the registry-collision fix at `e2063a3`. Two gaps are detailed inline: a rejected `DataModel` can be selected again as the synthetic model's base, and the compatibility predicate mutates the global registry while inspecting nested classes. The first is an incomplete fix for the existing collision problem; the second is a new side effect of validating registry hits.

This overlaps directly with #1969. Reuse its `register_pydantic=False` serialization path and its collision/cycle safeguards for one side-effect-free comparison traversal, while retaining the schema-v1 policy here. The stored-spec fingerprint serves a different purpose (distinguishing synthetic cache entries), so it should remain separate from live-class compatibility; `compute_model_fingerprint` is not a drop-in replacement because it fingerprints a field selection and live defaults.

The existing 120 signal-schema tests and the new explode/read-back test pass locally. Two additional unit tests and an actual save/read-back test reproduce the issues below with Python 3.13.3, Pydantic 2.13.5, and SQLite. I did not run ClickHouse or change the working branch.

src/datachain/lib/signal_schema.py
```diff
@@ -433,7 +479,15 @@ def _deserialize_custom_type(
...
-                return fr
+                if SignalSchema._model_matches_custom_type(fr, type_name, custom_types):
+                    return fr
+                fingerprint = SignalSchema._custom_type_fingerprint(
+                    type_name, custom_types
+                )
+                alias = f"{model_name}_{fingerprint[:10]}"
+                if cached := ModelStore.get(alias, target_version):
+                    return cached
+                type_name = f"{alias}@v{target_version}" if target_version else alias
```
shcheklein, line 490: **AI-generated review (Codex) — [P2] Do not inherit the incompatible class when rebuilding its alias.**

Changing `type_name` avoids overwriting the registered name, but the later `ct.bases` loop still picks that same rejected class as `base_model`: `_get_bases` includes the model itself as its first entry. Pydantic then inherits fields that are absent from the saved schema.

Confirmed with two `create_model("ReviewRecord", __base__=dc.DataModel, ...)` classes: save the first with `a: int = 7`, then save a same-named replacement with `b: str = "new"`. Reading the first dataset here raises `KeyError: 'item__b'`. At the schema level the synthesized alias has fields `{a, b}`, although the stored fields are only `{a}`.

This remains a gap in the collision fix, rather than a newly introduced collision case. Exclude the rejected class from fallback inheritance and ensure the chosen base cannot reintroduce incompatible fields. Add a save/read-back case using real serialized `DataModel` metadata; the hand-written collision fixtures use `bases=[]`, which bypasses this path.

src/datachain/lib/signal_schema.py
```diff
@@ -415,13 +416,58 @@ def _split_subtypes(type_name: str) -> list[str]:
...
+    def _model_matches_custom_type(
+        fr: type[BaseModel], type_name: str, custom_types: dict[str, Any]
+    ) -> bool:
+        """Check that a registered model's field spec equals the spec stored for
+        ``type_name``, including every nested custom type."""
+        if ModelStore.get_name(fr) != type_name:
+            return False
+        reserialized: dict[str, Any] = {}
+        try:
+            SignalSchema._serialize_custom_model(type_name, fr, reserialized)
```
shcheklein, line 429: **AI-generated review (Codex) — [P2] Make the compatibility check leave ModelStore unchanged.**

`_serialize_custom_model` calls `_serialize_type`/`_type_to_str`, which register nested Pydantic classes. Merely checking an older parent therefore overwrites a same-named newer child in the process-global registry, even when this comparison returns `True`.

Reproduced: serialize a `ReviewParent` containing `ReviewChild(value: int)`, then register and serialize a new `ReviewChild(value: str)`. Calling this predicate on the unchanged parent returns `True` but replaces the registered string child with the integer child. Deserializing the string child's own schema then returns a synthetic alias instead of the already-registered original class. An unrelated parent read has changed subsequent identity resolution.

#1969 already adds `register_pydantic=False` through this serializer for precisely this kind of inspection. Reuse that path instead of introducing a second comparison with different side effects. Add a regression asserting both that the registry is unchanged and that the independently registered child still resolves to its original class after the parent is inspected.

PR #1882, opened 2026-07-23
Document fine-grained team permissions (groups and rules)

amritghimire:
Studio team permissions are getting the RBAC treatment (groups + rules, tracked in datachain-ai/studio#12737), so the team collaboration guide needed to catch up — the old version still said team roles give blanket access to everything, which isn't true anymore.

While updating it, the page name stopped making sense. Most of it had become security material that nobody would go looking for under "Team Collaboration", and it had grown to 400 lines covering three unrelated topics. So this also splits it up (per @shcheklein's feedback below).

**The docs content:**
- Reworked the Roles section around the grant-only model. Admins still bypass everything, but for everyone else access to a namespace/project/dataset comes only from an explicit rule, a group grant, or owning it (creating a namespace). No fallback to the team role.
- Privilege tables now show what a `read` vs `write` grant allows instead of Viewer/Editor/Admin columns. The team-management table stays role-based since that's still governed by the role.
- New section documenting groups and rules, plus a short namespace/project/dataset primer that links out to the namespaces guide.
- Added screenshots of the new Permissions UI.

**The structure:** `team-collaboration.md` is now `docs/studio/user-guide/teams/`, following the same directory + `index.md` pattern `jobs/` and `git-connections/` already use:

| Page | Holds |
| --- | --- |
| Overview | Making a team, inviting people, tour of the workspace |
| Security & Permissions | Roles, how access is resolved, privilege tables, groups and rules |
| Team Settings | Name, credentials, collaborators and seats, SSO, plan |

Two side effects of the split:
- The GitLab and SSO sections were mostly repeating `git-connections/custom-gitlab-server.md` and `authentication/single-sign-on.md`, so they're a sentence and a link now.
- The old `/studio/user-guide/team-collaboration/` URL is public and indexed, so this adds `mkdocs-redirects` to point it at the new overview instead of letting it 404.

Checked with `mkdocs build --strict` (which fails on any dead link) and `mkdocs serve` — every cross-page anchor resolves, the screenshots load at their new depth, and the old URL redirects.


shcheklein:
Let's think about better name / structure (see other products). Probably we need now Studio -> Security & Permissions (or something like that). Team Collaboration doesn't saying anything to me.

Settings part of that section can go into a different section.

Or do Studio -> Team (some better name here?) > [Security, Settings ... etc]

amritghimire:
> Let's think about better name / structure (see other products). Probably we need now Studio -> Security & Permissions (or something like that). Team Collaboration doesn't saying anything to me.
> 
> Settings part of that section can go into a different section.
> 
> Or do Studio -> Team (some better name here?) > [Security, Settings ... etc]

restructured it. 
Went with following structure:

```
☁️ Studio
└ User Guide
   ├ ...
   ├ Git Connections
   ├ Teams
   │   ├ Overview                teams/index.md
   │   ├ Security & Permissions  teams/permissions.md
   │   └ Team Settings           teams/settings.md
   ├ Authentication
   └ Troubleshooting
```


- **Overview** - making a team, inviting people, tour of the workspace
- **Security & Permissions** - roles, who can see what, groups and rules
- **Team Settings** - name, credentials, seats, SSO, plan

Also made the old /studio/user-guide/team-collaboration/ URL redirect to the new overview so it doesn't 404.

docs/studio/user-guide/teams/permissions.md
```diff
@@ -0,0 +1,258 @@
+# Security & Permissions
+
+Access inside a team comes from two things: each collaborator's **team role**,
+and the **fine-grained permissions** that grant `read` or `write` on individual
+namespaces, projects, and datasets.
+
+## Roles
+
+Every collaborator has one of three team roles. The role controls team-level
+capabilities, but, except for admins, it does **not** by itself grant access to
```
shcheklein: it also doesn't restrict, right? it is important to mention

shcheklein: also, except Admins, right?

docs/studio/user-guide/teams/permissions.md
```diff
@@ -0,0 +1,258 @@
...
+- **`Admins`** - Have full access to everything in the team. Admins can see and
+  modify every namespace, project, and dataset regardless of permission rules,
+  invite and remove collaborators, manage groups and rules, manage team settings,
+  configure cloud credentials, and manage billing.
+- **`Editors`** - Can create resources, run jobs, and work with queries and
+  experiment projects. On datasets, they can edit or delete a dataset and its
+  files where they have a `write` [grant](#fine-grained-permissions) on it (or on
+  the namespace/project that contains it). They cannot modify team settings,
+  manage collaborators, or manage permissions.
+- **`Viewers`** - Have read-only access. They can explore jobs and experiments,
```
shcheklein: can they still edit dataset if they have a grant on it?

docs/studio/user-guide/teams/permissions.md
```diff
@@ -0,0 +1,258 @@
...
+  and any dataset they have a `read` [grant](#fine-grained-permissions) on. They
+  cannot create or modify resources, and cannot change team settings.
+
+Fine-grained grants only gate datasets and the namespaces and projects that
+contain them (see [How access is resolved](#how-access-is-resolved)), an Editor
+or Viewer with no matching grant sees no datasets. Jobs, queries, experiments,
+and storage are governed by the team role itself, not by grants.
+
+DataChain Studio does not have the concept of an `Owner` role. The user who
```
shcheklein: we don't need back-references, historic references, explanations of what we don't have (review carefully everything) ... except obvious beneficial cases (very rare)

docs/studio/user-guide/teams/permissions.md
```diff
@@ -0,0 +1,258 @@
...
+- **Rules cascade downward.** A rule on a namespace also covers every project and
+  dataset inside it; a rule on a project covers every dataset inside it.
+- **Creating a namespace grants you `write` on it.** The creator of a namespace
+  automatically gets `write` on that namespace (and, by cascade, on everything
+  inside it). Creating a project or dataset inside an existing namespace does not
+  add a new grant on its own.
+
+!!! note
+
+    Existing teams keep working after this feature rolls out. Each team starts
```
shcheklein: We don't need this (sounds more like a plug from release notes or something)

docs/studio/user-guide/teams/permissions.md
```diff
@@ -0,0 +1,258 @@
...
+
+Fine-grained grants govern **datasets** (and, by cascade, the namespaces and
+projects that contain them); the datasets table below shows what a `read` versus
+`write` [grant](#fine-grained-permissions) allows. **Jobs, queries, experiments,
+and storage** are governed by the **team role** instead, their tables show what
+each role can do. Admins can perform every action regardless of grant or role.
+
+### Privileges for datasets
+
+| Feature                     | Read | Write |
```
shcheklein: why Admin is missing here (and don't later, in similar tables)

docs/studio/user-guide/teams/permissions.md
```diff
@@ -0,0 +1,258 @@
...
+| Configure cloud credentials        | No     | No     | Yes   |
+| Manage GitLab server connections   | No     | No     | Yes   |
+| Configure Single Sign-on (SSO)     | No     | No     | Yes   |
+| Manage team plan and billing       | No     | No     | Yes   |
+| Delete a team                      | No     | No     | Yes   |
+
+## Fine-grained permissions
+
+Team roles are coarse, an Editor can write across everything they're granted, a
+Viewer can read it. **Permissions** let admins be specific: sort collaborators
```
shcheklein: let's remove the first sentences - complicated and not new information

shcheklein: Persmissions to my mind replace (except for Admins) roles - they don't complement them. They are the default way to deal with datasets

docs/studio/user-guide/teams/permissions.md
```diff
@@ -0,0 +1,258 @@
...
+namespaces, projects, and datasets.
+
+You'll find Permissions under **Collaborators & Permissions** in the
+[team settings](settings.md) page, below the Collaborators list. This area is
+**admin-only**, other collaborators see "Only team admins can manage
+permissions."
+
+!!! note
+
+    Non-admins only reach namespaces, projects, and datasets through permissions.
```
shcheklein: this should pretty much somewhere at the start, explained that this Roles don't really cover datasets

docs/studio/user-guide/teams/permissions.md
amritghimire: > it also doesn't restrict, right? it is important to mention

 It does restrict, actually. The role caps what kind of action you can take anywhere in the team, a Viewer can only ever read, no matter what they've been granted because they cannot open new tab or run new jobs and so on. So both things are true: the role doesn't get you to a dataset, but it does limit what you can do once you're there.


> also, except Admins, right?

 Yep. Admins skip both checks, they reach everything and rules don't apply to them. Called out in the third paragraph now.

docs/studio/user-guide/teams/permissions.md
amritghimire: No. The role check runs at the endpoint before grants are looked at, so a Viewer gets rejected before the grant matters. The write grant sits dormant until you make them an Editor. Added that to the Viewers bullet, since it's the obvious question to have.

docs/studio/user-guide/teams/permissions.md
amritghimire: Removed,

amritghimire: reviewed other places too,

docs/studio/user-guide/teams/permissions.md
amritghimire: Added. It was missing because that table's columns are grants (read/write) while the others are roles, so Admin isn't really the same kind of column, but that distinction only made sense to me because I'd just written it. Now there's an Admin column like everywhere else, plus a line above the table explaining that write actions need the Editor role on top of the grant.

docs/studio/user-guide/teams/permissions.md
amritghimire: removed

amritghimire: Somewhat agreed, and I've rewritten the page to match that part. Permissions are the only thing that gets you to a dataset; roles contribute nothing there. The page now leads with that.

But they don't fully replace roles, because the code still checks the role first. The dataset mutations are gated too, not just job creation: rename, delete version, and delete dataset (dqlapp/schema/dataset.py:661,760,812) each carry require_datachain_team("write") and an allowed_objects(..., "write") check. Both, stacked, on the same mutation.


So a Viewer with a write grant can't write, and that's enforced in several places like running jobs deliberately rather than by accident. The page documents it as it behaves today: permissions decide what you reach, the role caps what you can do with it.

 If the intent is that a grant should win outright, that's a code change across those endpoints and worth its own issue, happy to make the changes.

docs/studio/user-guide/teams/permissions.md
amritghimire: Moved — it's the opening line of the page now.

docs/studio/user-guide/teams/permissions.md
shcheklein: I thought we designed it to be decoupled from Roles, no? (besides obvious side effects - e.g. not able to run a job). But if I want to edit a dataset (while I can't run a job), I think I should be able to if I have a grant

amritghimire: The obvious side effect like running a job will make it confusing if I say they can edit a dataset or something like that, but we can do that, thats just a matter of switch on decorator in backend and probably in frontend too.

shcheklein: > The obvious side effect like running a job will make it confusing

not sure I understand? what would be the confusing flow?

amritghimire: That you cannot edit dataset somewhere (like through job), but can through edit dataset feature. That edit dataset is basically just changing dataset description to be honest, you cannot edit to make new version or such

shcheklein: you can't run a job, right? if you are a viewer ... so it will be clear that the whole mechanism (jobs) is closed for you (e.g. because we don't give you a way to run any compute, etc)

shcheklein: you can probably delete datasets also, no?

amritghimire: Yes, that we can make, but for now, the api is gated on write, so lets keep it. If we want, we can loosen those later

amritghimire:
@shcheklein PTAL, I would like to merge it along with release

shcheklein (approved):

docs/studio/user-guide/teams/permissions.md
shcheklein: to be clear - I just don't like the dependency of anything related to RBAR to roles and other way around - it automatically complicates everything. And requires (here is an example of that) complicated explanations that will be hard to understand by anyone.

PR #1887, opened 2026-07-26
Support account in az:// URIs (az://container@account/...)

### LINKED ISSUE #1859: Support account in az:// URIs
### Description

`az://container` doesn't include storage account, so it has to be defined outside `client_config={"account_name": "myaccount"}` or env variables AZURE_STORAGE_ACCOUNT_NAME=myaccount or AZURE_STORAGE_CONNECTION_STRING or CLI param `--account-name myaccount`. This means ambiguity, a single account per session and no way to reference two accounts in one run.

We need a single string that identifies **both** the container and the account.

The "traditional" approach used in Spark/Delta is using URL with a different schema `abfss://container@account/...` (absff makes me wince every time 🤢)

**Fix:** Supporting account in az prefix like `az://container@account/...`.

Open question: Potentially we can also support Spark's way of doing this and accept abfss:// as an alias. Personally I'd need a real justification to bring in a prefix that ugly.

PS: THis problem does not exist in S3/GCS since buckets are global.

dreadatour:
Implements #1859: embed the Azure storage account in the URI — `az://container@account/path` — so a single string identifies both the container and the account, and URIs pointing to different accounts can coexist in one run. Supersedes #1798, which threaded `--account-name` through the knowledge-skill scripts instead.

## Design

The account is parsed once, at the `AzureClient` boundary, and **stays in the client name**: `client.name`, `client.uri`, and `File.source` keep `container@account`, so saved datasets round-trip the account with no side-channel config. adlfs already tolerates `az://container@account/path` in every path operation (`_strip_protocol` folds the netloc back into `container/path`), so the change is confined to:

- `AzureClient.__init__` — parse the netloc, inject `account_name` into fs kwargs (URI wins over `client_config`), keep `self.container` for container-level SDK calls (`_fetch_flat`)
- `AzureClient.from_source` — preserve `@account` (base `_strip_protocol` would drop it)
- `AzureClient.bucket_status` — accept `container@account`, enabling the anonymous-access probe without an extra flag

Listing dataset names need nothing: `@` is already sanitized injectively (`_x40`) and round-trips.

## Knowledge skill

- `source_to_https` builds `https://account.blob.core.windows.net/container` from the `container@account` netloc (env `AZURE_STORAGE_ACCOUNT_NAME` as fallback). The old code read `az://account/container/`, which is inverted relative to how DataChain actually lists Azure — fixed.
- `bucket_overview` passes the account to `fsspec.filesystem("az", ...)`, strips the bare container (not the logical netloc) from adlfs entry names, and with `--anon` builds the chain on a session with `client_config={"anon": True}` so sampled File rows stay readable during enrichment. SKILL.md documents both.
- `bucket_scan` needed no changes — the account rides through in the URI.

## Guard: conflicting connection string

adlfs gives a connection string precedence over `account_name`, which would silently read account A while `File.source` says B. If the URI embeds an account and an explicit or env connection string names a *different* `AccountName`, the client now raises. A matching connection string works as before (this is also what the azurite functional test exercises). Documented in the remotes guide: account keys / SAS tokens / connection strings are account-specific, so mixing accounts in one run needs per-account-valid credentials (Azure AD or anonymous).

## Not included (deliberately)

- `abfss://` alias — open question in #1859, needs its own justification.
- The listing-error JSON handling and empty-bucket guards discussed in #1798 — valid but orthogonal to this change.

Closes #1859

<sup>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sup>

### DIFF
--- a/src/datachain/skill/knowledge/scripts/bucket_overview.py
+++ b/src/datachain/skill/knowledge/scripts/bucket_overview.py
@@ -32,6 +32,10 @@ def _open(uri: str, anon: bool):
         kw: dict[str, str | bool] = {}
         if anon:
             kw = {"token": "anon"} if scheme == "gs" else {"anon": True}
+        if scheme == "az":
+            account = urlparse(uri).netloc.partition("@")[2]
+            if account:
+                kw["account_name"] = account
         return fsspec.filesystem(scheme, **kw), path, scheme
     raw = uri.removeprefix("file://") if scheme == "file" else uri
     return (
@@ -65,7 +69,9 @@ def _make_file(scheme: str, netloc: str, entry: dict):
 
     name = entry["name"]
     if scheme in REMOTE:
-        path = name.removeprefix(netloc + "/") if netloc else name
+        # adlfs names entries by bare container, without the "@account" part.
+        prefix = netloc.partition("@")[0] if scheme == "az" else netloc
+        path = name.removeprefix(prefix + "/") if prefix else name
         source = f"{scheme}://{netloc}/" if netloc else f"{scheme}://"
     else:
         path = name
@@ -82,7 +88,11 @@ def bucket_overview(
     netloc = urlparse(uri).netloc if scheme in REMOTE else ""
     files = [_make_file(scheme, netloc, e) for e in _sample(fs, root, limit)]
     name = name or f"overview_{_slug(uri)}_{int(time.time())}"
-    return dc.read_values(file=files).save(name)
+    session = None
+    if anon and scheme in REMOTE:
+        # Keep the saved File rows readable during later enrichment.
+        session = dc.Session.get(client_config={"anon": True})
+    return dc.read_values(file=files, session=session).save(name)
 
 
 def _to_bucket_json(ds, uri: str, anon: bool, dataset_name: str, output: str) -> None:
--- a/tests/unit/test_skill_knowledge_scripts.py
+++ b/tests/unit/test_skill_knowledge_scripts.py
@@ -5,6 +5,7 @@ import sys
 from datetime import datetime, timezone
 from pathlib import Path
 from types import SimpleNamespace
+from unittest.mock import MagicMock, call
 
 # Insert the scripts directory so bare imports work (matches runtime behavior).
 SCRIPTS_DIR = str(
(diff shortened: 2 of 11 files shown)


src/datachain/skill/knowledge/scripts/bucket_overview.py
```diff
@@ -32,6 +32,10 @@ def _open(uri: str, anon: bool):
         kw: dict[str, str | bool] = {}
         if anon:
             kw = {"token": "anon"} if scheme == "gs" else {"anon": True}
+        if scheme == "az":
+            account = urlparse(uri).netloc.partition("@")[2]
+            if account:
+                kw["account_name"] = account
         return fsspec.filesystem(scheme, **kw), path, scheme
```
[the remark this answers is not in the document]
dreadatour: adlfs strips the @account part itself: _ls passes the input through _strip_protocol/split_path, which fold container@account/... into container/... (the netloc parses as username+host). Verified against adlfs directly — split_path("container@account/dir/blob.txt") returns ('container', 'dir/blob.txt'). The fs is constructed with account_name, so listing targets the right account and the bare container.

tests/unit/test_skill_knowledge_scripts.py
```diff
@@ -397,6 +421,76 @@ def test_source_to_https_empty_returns_none():
     assert source_to_https("") is None
 
 
+# ---------------------------------------------------------------------------
```
shcheklein: needed?

shcheklein (commented):
- Do we still support regular fsspec URIs? do we support / respect Azure env variables?

(I think we don't want to create completely custom way of accessing things)

- Can we check that `@` in Azure is not reserved in some way to pass user name of some sort in some cases (ask to do full research on Azure URLs, custom deployments, etc, etc)

- Ask to find all discussion on account name on fsspec, rclone forums - to double check on some stuff we are missing here potentially

- How is it now compatible with older datasets where File doesn't have account name in them?

tests/unit/test_skill_knowledge_scripts.py
dreadatour: Not needed, but it is the same as in this file, so I prefer to keep it here for consistency.
I am going to clean up comments like this in https://github.com/datachain-ai/datachain/pull/1888

dreadatour: Or I'll merge https://github.com/datachain-ai/datachain/pull/1888 and then update this branch, any way, this was a right question to ask 👍

dreadatour: Merged, removed.

amritghimire (commented):
Tests seems to be failing cc. @dreadatour

PR #1889, opened 2026-07-27
Route internal metastore queries through the overridable select seams

amritghimire:
Some methods in `AbstractDBMetastore` build their queries with bare `select(...)` and raw `self._<table>` refs instead of the `_<table>_select()` seams the rest of the class uses. Doesn't really matter for SQLite, but any subclass that overrides those seams (e.g. to scope rows per-tenant) ends up copying the whole method body just to get its filtering in - the bare select never goes through the seam.

So this just routes those queries through the seams they should've been using all along:

- `remove_namespace` / `remove_project` / `remove_dataset_version` - the `SELECT ... FOR UPDATE` locks and the remaining-versions count go through the `_*_select` seams now
- `link_dataset_version_to_job` - updates `dataset_version.job_id` via `_datasets_versions_update()` instead of poking the table directly
- `get_direct_dataset_dependencies` - the dataset/namespace join sources come from two new `_dependency_datasets_source` / `_dependency_namespaces_source` props (default to the raw tables), so a subclass can hand back a filtered subquery instead. Same trick `get_dataset_dependency_nodes` already pulls with `_datasets_select().subquery()`.

Nothing changes for SQLite - each seam returns the same statement the bare `select(...)` did, so it's a pure internal refactor.

### DIFF
--- a/src/datachain/data_storage/metastore.py
+++ b/src/datachain/data_storage/metastore.py
@@ -1080,7 +1080,9 @@ class AbstractDBMetastore(AbstractMetastore):
             # sneak in between the count and the delete below. No-op on
             # SQLite (writes are already serialized at the database level).
             self.db.execute(
-                select(n.c.id).where(n.c.id == namespace_id).with_for_update()
+                self._namespaces_select(n.c.id)
+                .where(n.c.id == namespace_id)
+                .with_for_update()
             )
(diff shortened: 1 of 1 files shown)


dreadatour (approved):
Looks good to me 🔥 Thank you for cleaning this up 🙏

Issue #1891, opened 2026-07-30
read_storage - path issue


dmpetrov:
### Description

Claude code found an issue:
```
  2. Found a real datachain bug in your dev build (~/src/datachain, commit 1955f8e5): any read_storage()
  prefix or .glob() pattern containing _ returns 0 rows. Root cause: patterns are escaped via
  sql_escape_like() (_ → \_) but emitted as LIKE without an ESCAPE '\' clause — e.g.
  data_storage/warehouse.py:864 — so SQLite treats the backslash literally and nothing matches. Repro:
  dc.read_storage("s3://robolake/droid/droid_100/") → 0, while raw SQLite GLOB 'droid/droid_100/*' on the same
  listing table → matches. The L1 script works around it with startswith filters. Happy to write the fix +
  test in the datachain repo if you want.
```

### Version Info

```Text
0.59.4.dev1+g1955f8e5b
Python 3.12.11
```

shcheklein:
Fixed by #1892, #1893 and #1895 (plus datachain-ai/studio#13128 for the ClickHouse side). All merged to `main`.

The cause turned out to be different from the diagnosis in the report, so it's worth recording what was actually wrong — it was three separate bugs:

**The 0-row reads were glob anchoring, not `_`** (#1892). Listing rows store paths relative to the listing root, and `apply_glob_filter()` only prepended that subpath when the pattern had no `/`. So `read_storage("s3://bucket/droid/droid_*/*.txt")` filtered with `GLOB 'droid_*/*.txt'` against rows like `droid/droid_100/a.txt` and matched nothing. `_` is incidental here — `droidX*/*.txt` failed identically. It stayed hidden because a pattern starting with `*` still works: SQLite's `GLOB *` crosses `/`, so the leading wildcard absorbs the missing prefix.

**`LIKE` without `ESCAPE` was real, but in different queries** (#1895). `sql_escape_like()` escapes `%`/`_` with a backslash, and SQLite has no default escape character, so the clause has to be declared too. `select_node_fields_by_parent_path_tar()` escaped without declaring (0 rows for any parent containing `_`), and `get_node_by_path()` did neither, so a nonexistent `dir_2/` resolved via a sibling `dirX2/` — `datachain ls` printed nothing and exited 0 instead of erroring. Both now go through `AbstractWarehouse.prefix_match()`, which ClickHouse overrides because it can't parse `ESCAPE` before 26.6.

[...]

PR #1892, opened 2026-08-01
fix: anchor glob patterns at the listing root

shcheklein:
Part of #1891 — see below on why this doesn't close it.

Listing rows store paths relative to the listing root. `apply_glob_filter()` only prepended `list_path` when the glob had no `/`, so a multi-segment glob under a non-root prefix was evaluated at the wrong root and matched nothing.

It went unnoticed because a pattern starting with `*` still works: SQLite's `GLOB *` crosses `/`, so the leading wildcard absorbs the missing prefix. Only a literal first segment breaks. Nothing over-matched either way — `apply_glob_filter()` calls `ls(list_path)` first, so results stay confined to the requested subtree.

Always anchor the converted glob at `list_path`. Converting before anchoring also preserves the existing zero-directory behavior of a leading `**`.

## Relationship to #1891

The issue attributes the 0-row reads to `sql_escape_like()` emitting `LIKE` without an `ESCAPE` clause. That mechanism is real and is fixed in #1895, but it isn't what the reporter hit: `read_storage()` filters via `GLOB`, where `_` is an ordinary character, and I could not reproduce the plain-prefix repro from the issue (`s3://…/droid_100/`) on local FS or moto-S3, fresh or reused listing.

What does reproduce is the glob half, and `_` is incidental there — `droidX*/*.txt` fails identically. So this PR fixes the reported symptom and #1895 fixes the mechanism the issue names. Marked "Part of" rather than "Fixes" so merging this alone doesn't auto-close the issue while the `LIKE`/`ESCAPE` half is still open.

## Tests

- `test_get_listing_reuses_broader_listing_for_glob_base` — `read_storage()` splits the pattern off and looks the base up; pins that the lookup reuses the broader listing and returns the subpath under it.
- `test_glob_with_literal_first_segment_under_broader_listing` — the case that actually regresses.
- `test_globstar_under_broader_listing` — passes unanchored too, so it guards the globstar conversion rather than the anchoring.
- `test_read_storage_glob_with_dir_component` — the reported cloud case (`dogs/oth*/*`).

The first matters more than it looks: that subpath is what makes anchoring observable. Without reuse it comes back empty, the read builds its own listing scoped to the glob's base, and the bug cannot appear — so a local glob test alone would pass against unfixed code. Cloud is unaffected, since paths there are always bucket-relative.

### DIFF
--- a/src/datachain/lib/dc/storage_pattern.py
+++ b/src/datachain/lib/dc/storage_pattern.py
@@ -243,11 +243,8 @@ def apply_glob_filter(
 
     chain = ls(dc, list_path, recursive=use_recursive, column=column)
 
-    if list_path and "/" not in pattern:
-        filter_pattern = f"{list_path.rstrip('/')}/{pattern}"
-    else:
-        filter_pattern = pattern
-
-    glob_pattern = convert_globstar_to_glob(filter_pattern)
+    glob_pattern = convert_globstar_to_glob(pattern)
+    if subpath := list_path.strip("/"):
+        glob_pattern = f"{subpath}/{glob_pattern}"
 
     return chain.filter(Column(f"{column}.path").glob(glob_pattern))
--- a/tests/func/test_datachain.py
+++ b/tests/func/test_datachain.py
@@ -85,6 +85,12 @@ def test_read_storage_glob(cloud_test_catalog):
     assert chain.count() == 3
 
 
+def test_read_storage_glob_with_dir_component(cloud_test_catalog):
+    ctc = cloud_test_catalog
+    chain = dc.read_storage(f"{ctc.src_uri}/dogs/oth*/*", session=ctc.session)
+    assert chain.count() == 1
+
+
 def test_read_storage_as_image(cloud_test_catalog):
     ctc = cloud_test_catalog
     chain = dc.read_storage(ctc.src_uri, session=ctc.session, type="image")
--- a/tests/func/test_storage_pattern.py
+++ b/tests/func/test_storage_pattern.py
@@ -202,6 +202,22 @@ def test_no_pattern_behavior_unchanged(tmp_dir):
     assert files == {"temp1.tmp", "temp2.tmp", "log1.log", "logfile"}
 
 
+def test_glob_with_literal_first_segment_under_broader_listing(tmp_dir):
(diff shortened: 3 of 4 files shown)


shcheklein:
Heads-up on d1fc1a71: after 7492251 the local test passed both with and without the source fix, so it wasn't guarding the anchoring any more.

`convert_globstar_to_glob("**/*.mp3")` collapses to `*.mp3`, and since SQLite's `GLOB *` crosses `/`, that pattern still matches once `ls()` has narrowed the chain to the subtree — with or without the prefix. Only patterns with a **literal first segment** actually regress.

I put a `mus*/rock/*` case back alongside the globstar one. Verified: the test now fails on `main` and passes on the branch. The cloud test was unaffected and still fails on `main` for both s3 variants, so the PR was never fully unguarded — just the local reuse path.

amritghimire (approved):
LGTM, but in convert_globstar_to_glob, the globstar feels like some brand name, if we can change it

shcheklein:
All three addressed.

**Comments in tests** — removed, and `AGENT.md` is stricter than I'd been treating it ("name the test so its intent is obvious, and skip the docstring/comment"). Both flagged comments also tripped the change-narrative tells it warns about — "which is what gives them", "Without reuse it comes back empty", plus a cross-reference to another test. Rather than delete and leave one test covering two pattern shapes, I split it so each name carries the intent:

- `test_glob_with_literal_first_segment_under_broader_listing` — the shape that actually regresses
- `test_globstar_under_broader_listing` — passes unanchored too, so it guards the globstar conversion

The `get_listing` docstring is gone as well. Zero comments remain in anything this PR adds.

**`Fixes #1891`** — you're right, and it's now `Part of #1891`. The issue attributes the 0-row reads to `sql_escape_like()` emitting `LIKE` without `ESCAPE`; that mechanism is real and is fixed in #1895 with tests, but it isn't what the reporter hit — `read_storage()` filters via `GLOB`, where `_` is an ordinary character, and the plain-prefix repro from the issue doesn't reproduce at all (local FS or moto-S3, fresh or reused listing). What reproduces is the glob half, where `_` is incidental: `droidX*/*.txt` fails identically.

So this PR fixes the reported symptom and #1895 fixes the named mechanism. #1895 now carries `Closes #1891`, so merging this one alone won't auto-close the issue while the `LIKE`/`ESCAPE` half is outstanding.

PR #1898, opened 2026-08-06
fix(gc): collect requeued dataset tombstones

yathomasi:
## Summary
Allow GC to delete finalized tombstones requeued for metadata removal, supporting Studio’s version-entry deletion flow.

## Changes
- Treat pending metadata deletion as authoritative for tombstones
- Collect `REMOVED` versions with `pending_metadata_drop`
- Preserve concurrent deletion requests during finalization
- Add race, retry, and override coverage

Related to datachain-ai/studio#13109


shcheklein:
@yathomasi was it really broken before? I thought GC is covering this already (REMOVING state + True bool for removing meta)

src/datachain/catalog/catalog.py
```diff
@@ -1245,34 +1235,27 @@ def _remove_versions(
...
             try:
                 v = dataset.get_version(version)
-                if v.status == DatasetStatus.REMOVED:
+                if v.pending_metadata_drop and v.status in (
+                    DatasetStatus.REMOVING,
+                    DatasetStatus.REMOVED,
+                ):
+                    keep = False
+                elif v.status == DatasetStatus.REMOVED:
                     continue
```
yathomasi, line 1270: We simply used to skip if the dataset version is in `REMOVED` state, now we need further metadata drop even it's aleady in `REMOVED` state.

yathomasi:
> was it really broken before? I thought GC is covering this already (REMOVING state + True bool for removing meta)

Not really broken for current state. This is for further deletion required for datachain-ai/studio#13115. I have marked the part as well.

src/datachain/catalog/catalog.py
```diff
@@ -1193,20 +1193,10 @@ def _claim_and_remove(
...
-        so the version row stays as a tombstone and GC ignores it.
-
-        With ``keep_metadata=False`` the version row is deleted. This
-        path also works when the version is already REMOVED: the guarded
-        UPDATE flips REMOVED to REMOVING and the version row is deleted,
-        which lets a wipe finish removing a previous tombstone.
-
-        Returns True if this call actually changed state, False if
-        another caller had already claimed it.
+        """Claim the version (guarded UPDATE), drop its rows table, then flip to
```
ilongin: Why did we shrink these comments so much, specially regarding the fact this PR extends the behavior

src/datachain/catalog/catalog.py
```diff
@@ -1245,34 +1235,27 @@ def _remove_versions(
...
-          keep-metadata remove (retries rows-table drop, flips status to
-          REMOVED).
-        - REMOVED: finalized tombstone, nothing to do.
-        - Anything else: wipe (incomplete/failed/stale versions to clean).
-
-        When given explicitly, honors the caller's intent per version, with
-        a downgrade to wipe for cases that can't keep metadata (internal
-        datasets, non-COMPLETE/REMOVING/REMOVED versions) - mirrors
-        ``catalog.remove_dataset`` so batch callers don't have to filter.
+        """Bulk remove versions. A tombstone with ``pending_metadata_drop`` set
```
ilongin: This comment has been shrinked as well

src/datachain/catalog/catalog.py
yathomasi: While touching it I felt it would be nice to shorten it. 

I can simply update on the existing one if that would be better.

src/datachain/data_storage/metastore.py
```diff
@@ -1969,6 +1969,11 @@ def get_dataset_versions_to_clean(
                             ),
                         ),
                     ),
+                    # Re-queued tombstones have no running job to wait for.
+                    and_(
+                        dv.c.status == DatasetStatus.REMOVED,
+                        dv.c.pending_metadata_drop.is_(True),
```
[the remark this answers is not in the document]
yathomasi: already fixed and pushed in 19b5d00a

tests/unit/test_dataset_status_management.py
```diff
@@ -1115,6 +1115,98 @@ def test_gc_skips_finalized_tombstones(test_session, dataset_complete):
...
+        catalog.get_dataset(name)
+
+
+def test_keep_metadata_remove_clears_pending_metadata_drop(
+    test_session, dataset_complete
+):
+    catalog = test_session.catalog
+    version = dataset_complete.latest_version
+
+    catalog.remove_dataset_version(dataset_complete, version, keep_metadata=True)
```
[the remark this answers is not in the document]
yathomasi: Removed the test — the flag can never be set on this path, so the assertion can't fail; `test_remove_keep_metadata_lands_on_removed` already covers it.

ilongin (approved):

PR #1901, opened 2026-08-12
fix: handle param overlap between model and its dotted child

### LINKED ISSUE #1883: IndexError when a map reads both a parent object and a nested field of it
### Description

The code that builds Python objects from row values crashes when the UDF input list contains both a parent column and one of its dotted children. Repro on main:

```python
import datachain as dc

class Doc(BaseModel):
    body: str

dc.read_values(doc=[Doc(body="hello")]).map(
    lambda body, doc: body.upper() + doc.body,
    params=["doc.body", "doc"],
    output={"res": str},
).to_list()
# IndexError: list index out of range
```

The row-unpacking step loses its position bookkeeping over the parent/child overlap and reads past the end of the row.

- Reproduces on main with hand-written `params=["doc.body", "doc"]`.
- Multi-signal `.map()` with an llm spec + sibling function on the parent object hits the same crash (spec contributes the dotted name silently), so users can reach it without writing `params=`.
- Fix belongs in the row-unpacking layer, not in `.map()` argument building.

### Version Info

```Text
0.54.3.dev6+g4a23aef23
Python 3.12.4
```

ilongin:
Fixes #1883.

When `.map(params=[...])` (or `to_iter(...)`) asked for both a model and one of its dotted children, e.g. `params=["fr.name", "fr"]`, we hit `IndexError` when building the UDF input row.

Root cause: `to_udf_spec` dedupes shared columns (dict keyed by db-flat name), so the row has one copy of the shared leaf. `row_to_objs` walked a linear `pos` counter and expected two copies. `row_to_features` has the same shape - not user-visible today because `db_signals` doesn't dedupe, but same fragile pattern.

Both now build a `{db_name -> row_index}` map and look up each value's columns by name.

- `row_to_objs`: positions built from `to_udf_spec()`.
- `row_to_features`: positions built from `db_signals()` (dupes still work - dict overwrite is a no-op since both indices hold the same column value).
- New helper `_sub_row_for_model` assembles the per-model sub-row, handling the `_type_tag` sentinel for nested `Optional[Model]`.

Tests: unit tests for `row_to_objs` and `row_to_features` covering scalar/nested/optional overlap, plus e2e coverage via `.map()` and `to_iter`.


src/datachain/lib/signal_schema.py
```diff
@@ -604,21 +604,45 @@ def to_udf_spec(self) -> dict[str, type[SQLType] | SQLType]:
 
     def row_to_objs(self, row: Sequence[Any]) -> list[Any]:
         self._init_setup_values()
+        positions = {name: i for i, name in enumerate(self.to_udf_spec())}
```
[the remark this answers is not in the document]
ilongin: fixed

shcheklein (approved):

src/datachain/lib/signal_schema.py
```diff
@@ -626,26 +626,36 @@ def to_udf_spec(self) -> dict[str, type[SQLType] | SQLType]:
                 res[db_name] = self._db_leaf_sql_type(path, type_)
         return res
 
+    @cached_property
+    def _udf_row_positions(self) -> dict[str, int]:
```
shcheklein, line 630: AI generated as an attempt to simplify this:

> _parse_row() already has a RowDict keyed by these flattened DB names, so converting it to a list and then reconstructing name→position caches seems avoidable. Could we pass that mapping directly to row_to_objs() and gather model subrows by their ordered DB column names? For sequence compatibility, row_to_objs() could normalize sequences with dict(zip(self.to_udf_spec(), row, strict=True)).
This would remove the four position caches and _model_row_positions(), while leaving row_to_features() and _hydrate_model() unchanged. Feature rows retain duplicate columns and already handle parent/child overlap correctly; the new to_iter/row_to_features overlap tests also pass on main.

shcheklein (commented):
Please check if we can simplify the implementation

shcheklein (changes requested):

src/datachain/lib/signal_schema.py
ilongin: I think it's good now, I simplified it as suggested in this AI comment. Now we are working with dicts all the time so no need for all those caching to know what is the key of some value since we had only lists of values....this reduced number of lines changed as well.

src/datachain/lib/signal_schema.py
```diff
@@ -755,6 +756,32 @@ def _file_bearing_fields(model: type[BaseModel]) -> tuple[str, ...]:
             _FILE_BEARING_FIELDS[model] = fields
         return fields
 
+    @staticmethod
+    def _model_row_values(
+        name: str,
+        fr: type[BaseModel],
```
dreadatour: What is `fr`? May be rename this variable into `model_type`? I know it is internal method, but it is har to understand what it is doing.

src/datachain/lib/signal_schema.py
```diff
@@ -626,26 +626,27 @@ def to_udf_spec(self) -> dict[str, type[SQLType] | SQLType]:
                 res[db_name] = self._db_leaf_sql_type(path, type_)
         return res
 
-    def row_to_objs(self, row: Sequence[Any]) -> list[Any]:
+    def row_to_objs(self, row: Sequence[Any] | Mapping[str, Any]) -> list[Any]:
```
dreadatour: Only one caller of `row_to_objs` — from `src/datachain/lib/udf.py`:

```python
    def _parse_row(
        self, row_dict: RowDict, catalog: "Catalog", cache: bool, download_cb: Callback
    ) -> list[Any]:
        assert self.params
        obj_row = self.params.row_to_objs(row_dict)
        self.params.set_file_streams(obj_row, catalog, cache, download_cb)
        return obj_row
```

Why do we need to extend this type here: `row: Sequence[Any] | Mapping[str, Any]` instead of using `RowDict`? Why do we leave `Sequence[Any]` here?

src/datachain/lib/signal_schema.py
```diff
@@ -626,26 +626,27 @@ def to_udf_spec(self) -> dict[str, type[SQLType] | SQLType]:
...
         self._init_setup_values()
+        if not isinstance(row, Mapping):
+            row = dict(zip(self.to_udf_spec(), row, strict=True))
```
dreadatour: How can it be not `Mapping`? See my comment above.

src/datachain/lib/signal_schema.py
```diff
@@ -626,26 +626,27 @@ def to_udf_spec(self) -> dict[str, type[SQLType] | SQLType]:
...
 
         objs: list[Any] = []
-        pos = 0
         for name, fr_type in self.values.items():
             inner_type, is_optional = unwrap_optional(fr_type)
             if self.setup_values and name in self.setup_values:
                 objs.append(self.setup_values.get(name))
             elif (fr := ModelStore.to_pydantic(inner_type)) is not None:
-                obj, pos = self._hydrate_model(fr, is_optional, row, pos, label=name)
+                sub = self._model_row_values(name, fr, is_optional, row)
```
dreadatour: What is `sub` here? Why not `row` as in other caller and in `_model_row_values`?

It is mixed with `row` from input param, but we can use `row_dict` as input param as in caller (`_parse_row`).

src/datachain/lib/signal_schema.py
```diff
@@ -755,6 +756,32 @@ def _file_bearing_fields(model: type[BaseModel]) -> tuple[str, ...]:
...
+        name: str,
+        fr: type[BaseModel],
+        is_optional: bool,
+        row: Mapping[str, Any],
+    ) -> list[Any]:
+        """Gather a model's flattened values in hydration order.
+
+        For example, model ``fr`` reads ``fr__name`` and ``fr__deep__value``
+        from the row and returns their values in the model's field order.
+        """
```
dreadatour: I think we can have a better docstring here, it is not really clear. I mean what is `hydration` order and why it is here (does it belongs to caller and why this method needs to know about it?).

Also example confirms it — docstring is a bit cryptic, so we need an example, to make it clear.

dreadatour (approved):
Couple comments about readability and code style from me, otherwise I have no questions. Looks good to me.

src/datachain/lib/signal_schema.py
ilongin: Not sure honestly, maybe AI slop that I overlooked :D .. returned to RowDict.

src/datachain/lib/signal_schema.py
ilongin: fixed

src/datachain/lib/signal_schema.py
ilongin: fixed

src/datachain/lib/signal_schema.py
ilongin: Fixed

shcheklein (approved):

PR #1911, opened 2026-08-16
fix(models): route type-tree walking through one decomposer

shcheklein:
Follow-up to #1867. Three bugs and a per-row performance cliff, all tracing to the same cause: five hand-written type-tree walkers that each recognised a different set of collection spellings. #1867 added two of them.

**Up front, so the diff isn't misread: this is net +111 lines of production code, not a reduction.** The shared decomposer and the two new caches cost more than the duplicated branches they replace. The win is behavioural correctness and per-row cost, not line count.

## The cause

| walker | origin test | `Sequence`/`Mapping` | `tuple` |
|---|---|---|---|
| `python_to_sql` | structural — `issubclass(list, origin)` | ✅ | ✅ |
| `is_chain_type` | `orig is list` / `orig is dict` | ❌ | ❌ |
| `_convert_feature_value` | `origin is list` / `origin is dict` | ❌ | ❌ |
| `_requires_row_conversion` (#1867) | `origin is list` / `origin is dict` | ❌ | ❌ |
| `_annotation_contains_type` (#1867) | `origin in (…, list, dict, tuple, set)` | ❌ | ✅ |

`python_to_sql` had it right: `issubclass(list, origin)` asks "would a list satisfy this origin?", which is what makes `Sequence` and `Mapping` work. That test is extracted into `annotation_parts()` and friends in `data_model.py`, and **all five walkers in the data path now share it**.

`type_to_str` keeps its own `origin is list` check deliberately — it is a serialiser and must emit the literal spelling of a type, so `Sequence[X]` must not be written out as `list[X]`.

## Bugs fixed

**`.setup()`-provided files lost their catalog** — regression from #1867, live on `main` today. Setup params are typed `str` as a placeholder because their value comes from a user callable evaluated lazily on a worker, so no annotation can describe it. The gate asked `_annotation_contains_type(str, File)`, got `False`, and skipped them.

```python
(
    dc.read_storage(DATA_URI)
    .setup(ref=lambda: dc.File(path="reference.txt", source=DATA_URI))
    .map(n=lambda ref: len(ref.read()), output=int)
    .to_values("n")
)
# main (prefetch on, the default):  RuntimeError: cannot prefetch file because catalog is not setup
# main (prefetch=0):                RuntimeError: Cannot open file: catalog is not set
# this PR:                          [9, 9]
```

`_file_stream_params` now records such params as *unknown* and inspects the value at runtime. Arbitrary and cyclic setup containers stay untraversed, and bare setup collections remain untraversed exactly as before #1867.

**Dict signals whose key type wasn't literally `str` crashed.** Keys are stored as strings, so only a key type that *cannot itself be a string* needs decoding — `str | None` and `Any` do not.

```python
class TagDoc(dc.DataModel):
    tags: dict[str | None, int]

dc.read_values(doc=[TagDoc(tags={"foo": 1})]).map(
    n=lambda tags: sum(tags.values()), params=["doc.tags"], output=int
).to_values("n")
# main:     json.JSONDecodeError: Unexpected character found when decoding 'false'
# this PR:  [1]
```

Fixed in the converter rather than the gate, because `to_iter()`/`to_list()` reach the same code and crash on `main` too. One change repairs both readers.

**`tuple`, `Sequence` and `Mapping` were half-recognised.** `tuple[Model, ...]` params arrived as raw dicts; files inside `Sequence[File]` / `Mapping[str, File]` fields never got a stream. Mapping *keys* are now traversed too. `set` is deliberately excluded throughout, matching `python_to_sql`, which cannot store it.

## Performance

The annotation walk leaves the per-row path entirely. Reachability is cached per **model class** — a small, stable key space — rather than per annotation:

| | main | this PR |
|---|---|---|
| annotation checks / row (model + 50 files) | 411 | **0** |
| `_parse_row`, model w/ `list[File]` ×50 | ~288 µs | **~240 µs** |
| `_parse_row`, `file: File` (typical) | ~26.5 µs | ~25.5 µs |
| `lru_cache` @ 400 distinct annotations | 2.79 µs/lookup, **0% hit rate** | n/a (uncached) |

The old `lru_cache(maxsize=128)` was keyed on arbitrary annotations — an unbounded key space — so past 128 the hit rate collapsed to zero and every call re-walked the tree.

## Tests

3071 unit tests pass; `ruff check` and `ruff format` clean. New coverage: the decomposer helpers directly (including abstract origins and the deliberate `set` exclusion); setup files and setup models with file collections; bare and cyclic setup values still untraversed; dict key decoding across `str`, `str | None`, `Any`, `int`, `int | None` through both `row_to_objs` and `row_to_features`; `tuple[Model, ...]` hydration end to end; `Sequence`/`Mapping` stream attachment; mapping-key traversal.

The two setup tests fail against `main` and pass here.

### DIFF
--- a/src/datachain/lib/data_model.py
+++ b/src/datachain/lib/data_model.py
@@ -128,6 +128,82 @@ def compute_model_fingerprint(
     return hashlib.sha256(json_str.encode("utf-8")).hexdigest()
 
 
+def _origin_accepts(container: type, t: Any) -> bool:
+    """Whether `container` would satisfy the origin of `t`.
+
+    Note the direction: `issubclass(dict, origin)`, not `origin is dict`, so
+    abstract spellings like `Mapping[str, X]` are recognised too. Some classes
+    refuse `issubclass` outright -- TypedDicts and non-method protocols raise --
+    and for those the answer is simply no.
+    """
+    orig = get_origin(t)
+    if not inspect.isclass(orig):
+        return False
+    try:
+        return issubclass(container, orig)
+    except TypeError:
+        return False
+
+
+def is_mapping_annotation(t: Any) -> bool:
+    """Whether `t` is a dict-like type, so `Mapping[str, X]` counts too.
+
+    Origins loose enough to accept a list as well -- `Collection`, `Iterable` --
+    are not mappings. Excluding them keeps this predicate mutually exclusive with
+    `is_sequence_annotation`, so callers cannot disagree by testing in a different
+    order, and matches `python_to_sql`, which maps those to `Array`.
+    """
+    return _origin_accepts(dict, t) and not _origin_accepts(list, t)
+
+
+def is_sequence_annotation(t: Any) -> bool:
+    """Whether `t` is a list- or tuple-like type; `Sequence[X]` counts too.
+
+    `set` is excluded because it is not a `DataType` and `python_to_sql` has no
+    column type for it.
+    """
+    return _origin_accepts(list, t) or _origin_accepts(tuple, t)
+
+
+def is_tuple_annotation(t: Any) -> bool:
+    """Whether `t` is genuinely a tuple, whose shape must be restored on read.
+
+    Note the direction: `issubclass(orig, tuple)`, not the reversed form used by
+    the container predicates. `Sequence[X]` is satisfiable by a tuple but is not
+    one, and rebuilding an ordinary list as a tuple would be wrong.
+    """
+    orig = get_origin(t)
+    return inspect.isclass(orig) and issubclass(orig, tuple)
+
+
+def annotation_parts(t: Any) -> tuple[Any, ...]:
+    """Components of a union or collection, `()` for a leaf.
+
+    `list[X]` -> `(X,)`, `dict[K, V]` -> `(K, V)`, `X | None` -> `(X,)`.
+    """
+    if get_origin(t) in (Union, types.UnionType):
+        return tuple(a for a in get_args(t) if a is not type(None))
+    if is_mapping_annotation(t) or is_sequence_annotation(t):
+        return tuple(a for a in get_args(t) if a is not Ellipsis)
+    return ()
+
+
+def key_needs_json_decode(t: Any) -> bool:
+    """Whether mapping keys of type `t` are JSON-encoded in the DB.
+
+    Keys are stored as strings, so `dict[int, ...]` reads back as `{"1": ...}`
+    and needs decoding, but a key type that can be a plain string must not --
+    `"foo"` is not valid JSON.
+    """
+    if t is str or t is Any:
+        return False
+    if get_origin(t) in (Union, types.UnionType):
+        # one arm that can be a plain string makes decoding unsafe for all of them
+        return all(key_needs_json_decode(p) for p in annotation_parts(t))
+    # a collection key is never a plain string, whatever its components are
+    return True
+
+
 def unwrap_optional(t: Any) -> tuple[Any, bool]:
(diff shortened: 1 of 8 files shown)


src/datachain/lib/data_model.py
```diff
@@ -128,6 +128,52 @@ def _fingerprint_tree(
...
+    return ()
+
+
+def key_needs_json_decode(t: Any) -> bool:
+    """Whether mapping keys of type `t` are JSON-encoded in the DB.
+
+    Keys are stored as strings, so `dict[int, ...]` reads back as `{"1": ...}`
+    and needs decoding, but a key type that can be a plain string must not --
+    `"foo"` is not valid JSON.
+    """
```
dreadatour, line 197: > Keys are stored as strings

Not a subject for this PR, but in general. If I am not mistaken, it might be an edge case here, if dict have both `"1"` and a string and `1` and an int keys, e.g.:

```
{"1": "foo", 1: "bar"}
```

will results as

```
{"1": "bar"}
```

in the DB?

dreadatour (approved):
Looks good to me 👍

dmpetrov (approved):
LG

src/datachain/lib/data_model.py
shcheklein: Good catch, and it's worse than an ambiguity on read — the loss happens at **write** time:

```
stored JSON  :  {"1": "foo", "1": "bar"}     # json.dumps emits a duplicate key
parsed back  :  {'1': 'bar'}                 # last wins, "foo" is gone
```

Confirmed end to end: `M(d={"1": "foo", 1: "bar"})` with `d: dict[str | int, str]` flattens to `[{'1': 'foo', 1: 'bar'}]`, serializes to duplicate keys, and reads back as `{'1': 'bar'}`. So no decoding rule on this PR's side could recover it — the value is destroyed before the row is written, which is exactly why it's separate from the key-decode work here.

Agreed it's out of scope. Filed as https://github.com/datachain-ai/datachain/issues/1914 with the repro and a suggested direction (reject key annotations that admit both `str` and non-`str` at schema-build time, so it fails loudly instead of silently).

src/datachain/lib/signal_schema.py
```diff
@@ -638,33 +656,31 @@ def _requires_row_conversion(cls, annotation: DataType) -> bool:
...
+        if is_mapping_annotation(annotation):
+            key_type, value_type = get_args(annotation)
+            return key_needs_json_decode(key_type) or cls._requires_row_conversion(
+                value_type
+            )
+        if is_sequence_annotation(annotation):
+            return any(
+                cls._requires_row_conversion(part)
+                for part in annotation_parts(annotation)
+            )
```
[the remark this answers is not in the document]
shcheklein: Already fixed — this comment targets `690ad41d`; the change landed in `b6c34ad6`. On current head:

```
_requires_row_conversion(tuple[int, ...])  =  True
row_to_objs(tuple[int, ...])               =  (1, 2)
row_to_objs(tuple[str, ...])               =  ('a', 'b')
```

The fix is exactly what you suggested — tuple reconstruction is now a conversion requirement in its own right, independent of whether the elements need hydration:

```python
if is_sequence_annotation(annotation):
    # a declared tuple always needs rebuilding: the DB hands back a list
    if is_tuple_annotation(annotation):
        return bool(parts)
    return any(cls._requires_row_conversion(part) for part in parts)
```

Pinned by `test_row_to_objs_restores_tuple_shape_without_conversion`. The related lines 891 and 911 are covered by the same change, plus positional conversion for fixed tuples (`test_row_to_objs_converts_fixed_tuple_positionally`).

PR #1925, opened 2026-08-18
fix: include class-based UDF instance state in hash

### LINKED ISSUE #1903: Agg/Map-er constructor state is not part of the UDF hash - 2nd call returns the 1st call's rows.
### Description

A class-based UDF is hashed by its code, its parameter schema and its output schema. Its instance state is not included, so two instances that differ only in constructor arguments hash identically and the second one is served the first one's rows. No error, no warning, no version bump.

```python
import datachain as dc

class Out(BaseModel):
    key: int
    n: int

class CountAbove(dc.Aggregator):
    def __init__(self, limit: int):
        self.limit = limit                       # not part of the hash
    def process(self, key, value):
        yield Out(key=key[0], n=sum(1 for v in value if v > self.limit))

src = dc.read_values(key=[1]*5 + [2]*5, value=[1,2,3,4,5,1,2,3,4,5])
for limit in (0, 3, 0):
    rows = src.agg(CountAbove(limit), partition_by="key",
                   params=["key", "value"], output={"o": Out}).to_list("o.key", "o.n")
    print(limit, sorted(rows))

# limit=0 -> [(1, 5), (2, 5)]     correct
# limit=3 -> [(1, 5), (2, 5)]     WRONG, expected [(1, 2), (2, 2)]
# limit=0 -> [(1, 5), (2, 5)]     correct
```

`DATACHAIN_IGNORE_CHECKPOINTS=1` fixes this case.

**It is not only checkpoints.** With `DATACHAIN_IGNORE_CHECKPOINTS=1` set, the same collision still
returns stale rows once the aggregation feeds a merge and a second aggregation:

```python
class Final(BaseModel):
    key: int
    total: int

class Sum(dc.Aggregator):
    def process(self, key, n):
        yield Final(key=key[0], total=sum(n))

other = dc.read_values(key=[1, 2], tag=["a", "b"])
for limit in (0, 3, 0):
    agg = src.agg(CountAbove(limit), partition_by="key", params=["key", "value"], output={"o": Out})
    joined = agg.merge(other, on="o.key", right_on="key", inner=True)
    rows = joined.agg(Sum(), partition_by="o.key", params=["o.key", "o.n"],
                      output={"f": Final}).to_list("f.key", "f.total")
    print(limit, sorted(rows))

# limit=0 -> [(1, 5), (2, 5)]     correct
# limit=3 -> [(1, 5), (2, 5)]     WRONG, expected [(1, 2), (2, 2)]  (checkpoints disabled)
# limit=0 -> [(1, 5), (2, 5)]     correct
```

### Root cause

`AbstractUDF.hash()` (`src/datachain/lib/udf.py:235-257`) hashes `self._func` or `self.process`,
`self.params` and `self.output`. `UDFStep.hash_inputs()` (`src/datachain/query/dataset.py:647-656`)
adds only the partition columns and the generator flag. Nothing observes the instance.

### Impact

Wrong answers that look plausible, because they are real rows for a different question. Also, any
query benchmark that leaves checkpoints on is timing a cache hit: in our harness the affected
queries reported 0.00x s against an honest 3.3-4.0 s.

### Suggested fix

Include the instance's state in the hash. Cheapest version: hash the cloudpickle payload of the UDF
instance, which is already produced for distributed execution. Stricter version: require class-based
UDFs to declare their parameters. A UDF whose behaviour depends on constructor arguments must not
collide with a differently-constructed one.

### Version Info

```Text

```

ilongin:
Fixes #1903.

Class-based UDFs (`Mapper` / `Generator` / `Aggregator` subclasses) hashed to the same value regardless of constructor args, so the second `.agg(CountAbove(3), ...)` call was served the first `.agg(CountAbove(0), ...)`'s cached rows. No error, no version bump, wrong numbers.

Changes:

- `UDFBase.__new__` now hashes the constructor arguments (positional + keyword, defaults applied) before `__init__` runs, and stashes the result on the instance as `_constructor_identity_hash`.
- `UDFBase.hash()` mixes that value in for class-based UDFs so two instances that differ only in constructor args no longer collide.
- Extension point: override `identity_hash()`. When you do, `__new__` skips the automatic path and the override owns the complete cache identity. Use it when constructor args contain callables or opaque objects that DataChain cannot hash on its own.
- `ArrowGenerator` and `HFGenerator` override `identity_hash()`. Both include `output_schema` and normalize their own opaque args (`ParseOptions`, `CsvFileFormat`, `Partitioning`, HF dataset handles). Each computes the digest once in `__init__` and returns the cached string, so repeated calls stay stable.
- Unsupported values (opaque objects, closures, custom container subclasses) fall back to a per-instance random UUID: stable within one instance, distinct across instances. Cache reuse is disabled for that instance; correctness wins over reuse.
- Docs: `docs/references/udf.md` documents `identity_hash()` and when to override it; `docs/guide/checkpoints.md` and `docs/guide/python-engine.md` explain how constructor args affect checkpoint reuse.
- Tests: unit tests cover per-instance stability, cross-instance distinctness, pydantic-class args, container-subclass handling, and Arrow / HF hash stability with an opaque kwarg; a func test reproduces the exact issue scenario.

Function-based UDFs (plain lambdas and `def` functions passed to `.map()`) are untouched. The constructor-hash path only runs for class-based UDFs.


src/datachain/lib/udf.py
```diff
@@ -226,6 +226,30 @@ def process(self, file) -> list[float]:
...
+            # when __init__ uses *args instead of a named `self` parameter.
+            bound = inspect.signature(instance.__init__).bind(*args, **kwargs)
+        except TypeError:
+            # Pickle allocates an empty instance and restores its attributes
+            # afterwards, without calling __init__ or supplying its arguments.
+            return instance
+
+        bound.apply_defaults()
+        arguments = cls._constructor_hash_args(dict(bound.arguments))
+        instance._constructor_state_hash = hash_value(arguments)
```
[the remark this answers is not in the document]
ilongin: This is a known and intentional limitation. Scope of the hash is the constructor arguments the user passed in, plus the process() bytecode. Anything beyond that is "rabbit hole" we don't want to go in I think ...

src/datachain/hash_utils.py
```diff
@@ -16,6 +18,48 @@
...
+            "breaks caching (full recompute every run).",
+            stacklevel=2,
+        )
+    return value
+
+
+def hash_value(value) -> str:
+    """Hash a value after normalizing unordered containers."""
+    normalized = normalize_hash_value(value)
+    return hashlib.sha256(repr(normalized).encode("utf-8")).hexdigest()
```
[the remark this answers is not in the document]
ilongin: Valid comment, applied the fix

src/datachain/lib/arrow.py
```diff
@@ -75,6 +99,26 @@ def __init__(
         self.parse_options = kwargs.pop("parse_options", None)
         self.kwargs = kwargs
 
+    @classmethod
+    def _constructor_hash_args(cls, arguments):
+        # output_schema is a dynamically-created pydantic class with a random
```
amritghimire: do we need long docstring even for internal methods?

ilongin: Reduced comment

src/datachain/hash_utils.py
```diff
@@ -16,6 +16,43 @@
...
+    value_type = type(value)
+
+    if value is None:
+        return ("none",)
+    if value_type in (bool, int, float, str, bytes):
+        return (value_type.__name__, value)
+    if value_type is dict:
+        items = (
+            (normalize_hash_value(k), normalize_hash_value(v)) for k, v in value.items()
+        )
```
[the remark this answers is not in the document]
ilongin: already fixed

docs/guide/python-engine.md
```diff
@@ -151,6 +151,40 @@ class ImageEncoder(Mapper):
...
+import json
+from datachain.lib.udf import Mapper
+
+class Tokenize(Mapper):
+    def __init__(self, tokenizer, tokenizer_id: str, max_length: int):
+        self.tokenizer = tokenizer
+        self.tokenizer_id = tokenizer_id
+        self.max_length = max_length
+
+    def state_hash(self) -> str:
```
shcheklein: Q: is it really a "state" hash? does it change during the operation?

docs/guide/python-engine.md
```diff
@@ -151,6 +151,40 @@ class ImageEncoder(Mapper):
         del self.model
 ```
 
+### Caching class-based operations
```
shcheklein: Q: do we document class based approach at all? where?

shcheklein: I wonder what terminology we use there ...

shcheklein: "operations" or something else

docs/references/llm.md
```diff
@@ -139,7 +139,8 @@ Reliability is layered:
 Materialized `llm.*` columns are cached and versioned, so re-running a chain reads
 the stored result instead of re-calling the model; the cache invalidates when any
 output-affecting input changes (model, prompt, schema, the input column, `type`,
-params, ...).
+params, ...). Custom objects in parameters receive a per-operation identity, safely
```
shcheklein: this not clear tbh ... in the user facing docs ... why is that a problem, why we do this, etc

src/datachain/lib/arrow.py
```diff
@@ -75,6 +98,25 @@ def __init__(
         self.parse_options = kwargs.pop("parse_options", None)
         self.kwargs = kwargs
 
+    @classmethod
+    def _constructor_hash_args(cls, arguments):
+        arguments = arguments.copy()
+        generated_output_schema = arguments.pop("_generated_output_schema")
+        if generated_output_schema:
```
shcheklein: is it really a problem to hash it again?

docs/guide/python-engine.md
ilongin: Yea ...  "state" reads like something that changes at runtime. Renamed the API to `identity_hash()` (and `_constructor_identity_hash` for the internal field), and reworded the paragraph so we talk about a stable per-instance identity rather than state.

docs/guide/python-engine.md
ilongin: Class-based UDFs are documented right above this in `## Class-Based Lifecycle` (line 136), and referenced from `docs/guide/index.md`, `docs/references/index.md`, and `docs/guide/best-practices.md`. Terminology across those places is a mix of "class-based lifecycle", "class-based operations", and "class-based Mapper".

Since this subsection sits inside `## Class-Based Lifecycle`, the "class-based operations" in the heading is redundant. And it's not really about caching in general - it's about the identity used for cache keying. Renamed the heading to `### Cache identity` and updated the anchor in `checkpoints.md`.

docs/references/llm.md
ilongin: Right, that sentence was strange. Rewrote it to name the concrete case (callables / Python objects passed as parameters), say why (can't hash them deterministically), and give the takeaway (reuse the same operation instance to hit the cache).

src/datachain/lib/arrow.py
ilongin: The pop was the quick fix to land the constructor-hash change: pydantic classes trip the primitive-only hasher (warn + random uuid, cache reuse disabled), and the shape is already captured elsewhere via the output signal schema, so dropping it here was safe.

The smell is that it was already duplicated in `hf.py`, only covers the `output_schema` arg name, and doesn't help user-authored UDFs. Cleanup:

- Added a small walker in `_hash_constructor_args` that swaps any pydantic class in the args tree for its `SignalSchema._serialize_type` output.
- Made `dict_to_data_model`'s auto-generated name deterministic (shape-based sha256, was random uuid4) so same-shape calls produce identically-named classes and hash equal downstream.
- Dropped both `output_schema` pops (`arrow.py`, `hf.py`) and the `_generated_output_schema` flag.

docs/guide/checkpoints.md
```diff
@@ -202,3 +202,4 @@ dc.read_storage("gs://datachain-demo/dogs-and-cats/", anon=True).map(
 - **Script path matters:** DataChain links runs by the script's absolute path. Moving the script breaks checkpoint linking.
 - **Threading/multiprocessing:** Checkpoints are automatically disabled when Python threading or multiprocessing is detected. DataChain's built-in `parallel` setting for Python operations is not affected.
 - **Unhashable callables:** Built-in functions (`len`, `str`), C extensions, and `Mock` objects produce a different hash on each run, so checkpoints using these as Python operations will always recompute. Use regular `def` functions or lambdas instead.
+- **Class UDF state:** Constructor callables and custom objects recompute by default. Override [`identity_hash()`](python-engine.md#cache-identity) only when you can provide a complete stable identity.
```
shcheklein: This is still not clear at all.

Class UDF state - need a link to Class UDF

> Constructor callables and custom objects

what are those, what is the difference?

> only when you can provide a complete stable identity

why only then? what does it mean?

---------------

Can we please stop using AI for this and review and edit manually? to make it actually human readable?

docs/guide/python-engine.md
```diff
@@ -151,6 +151,40 @@ class ImageEncoder(Mapper):
         del self.model
 ```
 
+### Cache identity
```
shcheklein: I don't think this document is good place for this tbh? This is niche and complicated topic ... while this is guide is a general overview, right?

src/datachain/lib/udf.py
```diff
@@ -274,11 +305,49 @@ def process(self, file) -> list[float]:
...
     prefetch: int = 0
+    # Set in __new__ from bound constructor args (see hash()); default keeps
+    # mypy happy and acts as a safe fallback for instances that skip __new__.
+    _constructor_identity_hash: str = ""
     # Class-level default so subclasses that skip super().__init__() still
     # have this attribute; _MultiSignalMapper sets params/output per entry
     # but doesn't call _init(), so _func would otherwise be missing.
     _func: Callable | None = None
 
+    def __new__(cls, *args, **kwargs):
```
shcheklein, line 322: what is happening here? some cryptic / hard to understand stuff - either we need explanation and good justification for this or can we simplify all of this?

src/datachain/lib/udf.py
```diff
@@ -304,6 +373,25 @@ def hash(self, include_body: bool = True) -> str:
             self.params.hash() if self.params else "",
             self.output.hash(),
         ]
+        # For class-based UDFs, mix in constructor state so two instances that
+        # differ only in constructor args don't collide.
+        if self._func is None:
+            identity_hash = self.identity_hash()
```
shcheklein: very complicated - like 10+ lines jsut to check the value

src/datachain/lib/udf.py
```diff
@@ -596,6 +684,12 @@ class _MultiSignalMapper(Mapper):
     irrelevant. Cycles raise ``ValueError`` at construction time.
     """
 
+    @classmethod
+    def _constructor_hash_args(cls, arguments):
+        # hash() is fully overridden below; skip the base state hash so we
```
shcheklein: what is happening here?

src/datachain/lib/udf.py
ilongin: We need `__new__` here because class UDFs define their own `__init__` and usually do not call `super().__init__()`. It is therefore the only common point where DataChain can see the original constructor call before user initialization runs. Hashing `self.__dict__` later would not be equivalent: constructor arguments may be transformed or discarded, and instance attributes may change while the UDF runs.

I simplified this code in response to the comment. `__new__` now only creates the instance and stores the result of a small `_constructor_identity()` helper. The signature binding, defaults, and fallback are isolated in that helper. I also removed `_constructor_hash_args` and its confusing empty override in `_MultiSignalMapper`. Internal Arrow/HF generators now use the same public `identity_hash()` override as other class UDFs.

DataChain hashes constructor arguments automatically, unless the class overrides `identity_hash()`, in which case that method owns the complete cache identity.

src/datachain/lib/udf.py
ilongin: Refactored

src/datachain/lib/udf.py
ilongin: I already removed this method completely when refactoring for one of the previous comment

docs/guide/python-engine.md
ilongin: I moved the detailed cache identity section from the general Python operations guide to the UDF reference.

docs/guide/checkpoints.md
ilongin: I rewrote this to explain when the behavior applies and why DataChain recomputes. I also linked "class-based operation" to its guide and moved the advanced identity details to the UDF reference.

src/datachain/lib/data_model.py
```diff
@@ -329,9 +328,11 @@ def _model_fields_by_aliases(cls) -> dict[str, tuple[str, FieldInfo]]:
                     field_info[str(alias)] = (_name, field)
             return field_info
 
-    # Generate random unique name if not provided
+    # Derive a stable shape-based name if none provided, so two calls with the
+    # same fields produce classes that hash equal downstream.
     if not name:
-        name = f"DataModel_{uuid.uuid4().hex[:8]}"
+        shape_repr = repr([(k, str(v[0])) for k, v in sorted(fields.items())])
```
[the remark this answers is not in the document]
ilongin: Sorting is needed to have consistent hash with caller of this method. Order of arguments is not important.

src/datachain/lib/udf.py
```diff
@@ -223,6 +224,36 @@ def prefetch(self) -> int:
         return self.inner.prefetch
 
 
+def _normalize_pydantic_in_args(value: Any) -> Any:
+    """Replace pydantic classes anywhere in the args tree with a stable
+    signal-schema description so hash_value can normalize the result."""
+    if isinstance(value, type) and issubclass(value, BaseModel):
+        custom_types: dict[str, Any] = {}
+        type_name = SignalSchema._serialize_type(value, custom_types)
+        return {"__pydantic_class__": type_name, "__custom_types__": custom_types}
```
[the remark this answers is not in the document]
ilongin: this is known limitation. User can implement identity_hash() if needed

src/datachain/lib/data_model.py
dreadatour, line 334: Order is important 🤔

```python
from datachain.lib.data_model import dict_to_data_model
from datachain.lib.model_store import ModelStore

first = dict_to_data_model("", {"a": int, "b": int})
schema = SignalSchema({"row": first})
saved = schema.serialize()

second = dict_to_data_model("", {"b": int, "a": int})
ModelStore.register(second)
restored = SignalSchema.deserialize(saved)

print("Same model name:", first.__name__ == second.__name__)
print("Before:", schema.row_to_objs([1, 2])[0].model_dump())
print("After: ", restored.row_to_objs([1, 2])[0].model_dump())
```

Output:
```
Same model name: True
Before: {'a': 1, 'b': 2}
After:  {'b': 1, 'a': 2}
```

src/datachain/lib/udf.py
```diff
@@ -223,6 +226,42 @@ def prefetch(self) -> int:
...
+    if isinstance(value, dict):
+        return {k: _normalize_pydantic_in_args(v) for k, v in value.items()}
```
dreadatour, line 237: Not sure if it is really important, but if we want to be strict, we should think about it 🤔

```python
from datachain import Mapper

class Config(dict):
    def __init__(self, answer):
        self.answer = answer

class ReadConfig(Mapper):
    def __init__(self, config):
        self.config = config

    def process(self, x: int) -> int:
        return self.config.answer

first = ReadConfig(Config(1))
second = ReadConfig(Config(2))

print("Same identity:", first.identity_hash() == second.identity_hash())
print("Results:", first.process(0), second.process(0))
```

Output:
```
Same identity: True
Results: 1 2
```

dreadatour (commented):
I've read through the code and it looks good to me overall, but I have this feeling it is a bit overcomplicated 😅. Same time I lean to approve this PR.

Couple findings from Codex below, some are edge cases, one (with order) IMO is important.

---

```python
from pyarrow.dataset import CsvFileFormat
from datachain.lib.arrow import ArrowGenerator

reader = ArrowGenerator(format=CsvFileFormat())
before = reader.identity_hash()

print("Before:", type(reader.kwargs["format"]).__name__)

reader._process_record(
    {"value": 1},
    File(source="file:///tmp", path="example.csv"),
    0, None, False,
)

print("After:", reader.kwargs["format"])
print("Same hash:", before == reader.identity_hash())
```

Output:
```
Before: CsvFileFormat
After: csv
Same hash: False
```

docs/guide/python-engine.md
```diff
@@ -168,6 +168,10 @@ class ImageEncoder(Mapper):
         del self.model
 ```
 
+Constructor arguments are included in checkpoint identity. For constructors that
```
shcheklein, line 171: it is not clear here at all .. what Constructor? what checkpoint identity ... etc

docs/guide/checkpoints.md
```diff
@@ -208,3 +208,4 @@ dc.read_storage("gs://datachain-demo/dogs-and-cats/", anon=True).map(
 - **Script path matters:** DataChain links runs by the script's absolute path. Moving the script breaks checkpoint linking.
 - **Threading/multiprocessing:** Checkpoints are automatically disabled when Python threading or multiprocessing is detected. DataChain's built-in `parallel` setting for Python operations is not affected.
 - **Unhashable callables:** Built-in functions (`len`, `str`), C extensions, and `Mock` objects produce a different hash on each run, so checkpoints using these as Python operations will always recompute. Use regular `def` functions or lambdas instead.
+- **Class-based operation arguments:** Constructor arguments are part of a [class-based operation's](python-engine.md#class-based-lifecycle) checkpoint identity. For objects such as tokenizers or model clients, DataChain cannot know whether two instances behave identically, so it recomputes instead of risking an incorrect checkpoint result. If the object has a stable name or version, see [custom cache identity](../references/udf.md#cache-identity).
```
shcheklein, line 211: I don't think it is clear here what Constructor is

docs/references/llm.md
```diff
@@ -141,7 +141,10 @@ Reliability is layered:
 Materialized `llm.*` columns are cached and versioned, so re-running a chain reads
 the stored result instead of re-calling the model; the cache invalidates when any
 output-affecting input changes (model, prompt, schema, the input column, `type`,
-params, ...).
+params, ...). Callables and custom Python objects passed as parameters can't be
```
shcheklein, line 144: again, what are we really trying to say here?

docs/references/udf.md
```diff
@@ -9,6 +9,41 @@ Any Python function works as an operation. The classes below are useful to imple
 operation where a plain function is insufficient, such as when additional `setup()` or `teardown()`
 steps need to happen before or after the processing function runs.
 
+## Cache identity
+
+DataChain hashes class operation code, schemas, and constructor arguments. Primitive
```
shcheklein, line 14: is only about class? title is generic - Cache identity

src/datachain/lib/udf.py
```diff
@@ -274,11 +313,52 @@ def process(self, file) -> list[float]:
...
+        )
+        return instance
+
+    @classmethod
+    def _constructor_identity(cls, init, args, kwargs) -> str:
+        """Hash arguments passed to a class UDF constructor."""
+        try:
+            # bound-method signature so `self` is already stripped; correct even
+            # when __init__ uses *args instead of a named `self` parameter.
+            bound = inspect.signature(init).bind(*args, **kwargs)
```
shcheklein, line 340: are there some possible side effect of bounding it here?

src/datachain/lib/udf.py
```diff
@@ -274,11 +313,52 @@ def process(self, file) -> list[float]:
...
+        except TypeError:
+            # Pickle or a subclass __new__ override lands here; seed a valid
+            # random hash so .hash() stays well-formed.
+            return hashlib.sha256(uuid4().bytes).hexdigest()
+
+        bound.apply_defaults()
+        return _hash_constructor_args(
+            dict(bound.arguments),
+            warn_on_unsupported=cls.hash is UDFBase.hash,
```
shcheklein, line 349: why this?

src/datachain/lib/udf.py
```diff
@@ -274,11 +313,52 @@ def process(self, file) -> list[float]:
...
+            # bound-method signature so `self` is already stripped; correct even
+            # when __init__ uses *args instead of a named `self` parameter.
+            bound = inspect.signature(init).bind(*args, **kwargs)
+        except TypeError:
+            # Pickle or a subclass __new__ override lands here; seed a valid
+            # random hash so .hash() stays well-formed.
+            return hashlib.sha256(uuid4().bytes).hexdigest()
+
+        bound.apply_defaults()
+        return _hash_constructor_args(
```
shcheklein, line 347: can it run for a long time? e.g. some binary model passed - will it be hashing it? is it a problem?

are side effects possible - e.g. we are instantiating some args too early?

src/datachain/lib/udf.py
ilongin: Yes, this was a bug. Fixed and added tests

src/datachain/lib/data_model.py
ilongin: Yea .. I went back and forth with this sorting since i wanted to preserve cache i.e. not to invalidate cache when just order of input dict is change but you are right, this does mean we could have a bug since 2 same chains with different orders could produce same key in our model storage and we could reuse wrong class on read. Removed it and added test

src/datachain/lib/udf.py
ilongin: `bind()` does not call `__init__` or evaluate arguments. It only maps the already-created positional and keyword arguments to their constructor parameter names, so no normal user-code side effects are expected.

src/datachain/lib/udf.py
ilongin: Arbitrary model objects are rejected when encountered and are not serialized or traversed internally. Large surrounding containers or supported values such as `bytes`, lists, and dictionaries can still take time proportional to their size. Arguments are already evaluated before Python enters `__new__`, and this code only binds their existing values to constructor parameter names.

src/datachain/lib/udf.py
ilongin: It was to not show wrong warning but when looking at this better it's not really needed, i.e. it can be simplify so I removed this condition and changed the warning to state only what we know: the automatic constructor identity was randomized, regardless of how an overridden hash() uses it.

docs/references/llm.md
ilongin: I updated comment, hope it's more clear now. We are saying that if arguments passed to llm.complete(), llm.classify(), llm.score(), or llm.embed() contain values DataChain cannot hash safely, such as a custom client object, rebuilding the chain will rerun the model instead of reusing its cache.

docs/references/llm.md
```diff
@@ -141,7 +141,10 @@ Reliability is layered:
...
+hashed deterministically, so each freshly constructed `llm.*` call is treated as a
+new operation and re-runs the model. Reuse the same operation instance to hit the
+cache.
```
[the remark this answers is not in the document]
ilongin: Fixed comemnt

docs/guide/python-engine.md
ilongin: Reworded this in plain terms and updated the example to show ImageEncoder("ViT-B-32"), making it clear which values are included in the checkpoint hash

docs/guide/checkpoints.md
ilongin: Replaced “constructor arguments” and “checkpoint identity” with a direct explanation of how values passed to class-based operations affect checkpoint reuse.

docs/references/udf.md
ilongin: Yes ..., renamed to Cache identity of class-based operations so the scope is explicit. Updated the two cross-links in python-engine.md and checkpoints.md to the new anchor.

src/datachain/lib/udf.py
```diff
@@ -274,11 +313,52 @@ def process(self, file) -> list[float]:
...
+    def __new__(cls, *args, **kwargs):
+        instance = super().__new__(cls)
+
+        # An identity_hash() override owns the complete cache identity.
+        if cls.identity_hash is not UDFBase.identity_hash:
+            return instance
+
+        instance._constructor_identity_hash = cls._constructor_identity(
+            instance.__init__, args, kwargs
+        )
```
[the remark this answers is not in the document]
ilongin: fixed

PR #1959, opened 2026-08-28
Report Studio's terminal websocket errors instead of reconnecting silently

dreadatour:
`datachain job logs <id>` against a team name the server does not accept printed nothing but a reconnect countdown and never said why — 15 attempts with backoff capped at 60s, so roughly ten minutes of silence before "Failed to reconnect". In practice you interrupt it first:

```
DATACHAIN_STUDIO_TEAM=wrong-team datachain job logs a23466fb-f11a-4fbd-82ef-35feca494faf
>>>> WebSocket closed, reconnecting in 8s...^CError: Operation cancelled by the user
```

Studio does report the problem over the log-streaming websocket. The client dropped it, in two independent places.

**Refused handshake.** For an unknown team, a team you cannot access, or an invalid token, Studio refuses the websocket handshake with HTTP 403. `websockets` raises a handshake exception, which `tail_job_logs` caught in a blanket `except (WebSocketException, OSError)` and turned into a `logger.debug` plus a bare `return` — indistinguishable from a normal end-of-stream.

**Error frame.** For a malformed job id, or a job you cannot access, Studio accepts the connection, sends `{"message": "Job ID is incorrect or not found"}`, and closes with code 4404. That frame was yielded into a branch chain in `show_logs_from_client` that only handles `log_blobs` / `logs` / `job`, so it fell through every branch and was discarded.

## What changed

- `tail_job_logs` raises `DataChainError` when a refused handshake carries 401 or 403, naming the team. Every other failure — a dropped connection, a 5xx from a gateway mid-deploy — still returns and lets the caller reconnect, so `job run` keeps its REST-polling fallback.
- `show_logs_from_client` raises the server's own message when it sees the terminal `{"message": ...}` frame.
- The reconnect banner is now cleared before each connection attempt instead of when the first message arrives, and also when the backoff sleep is interrupted. A terminal error or a Ctrl-C no longer prints on top of a stale `reconnecting in Ns...` countdown.
- Backoff moved into `_wait_before_reconnect`, next to the existing `_print_reconnect_msg` and `_clear_line` helpers. Behavior is identical — the sleep still uses the pre-increment retry count and the debug line still reports the same attempt number.

The error message names the team *and* the token as possible causes, because Studio deliberately answers the same way whether a team does not exist or you simply cannot reach it. The client cannot tell those apart and should not pretend to.

<sub>🤖 Co-authored by Claude Code and ChatGPT</sub>

### DIFF
--- a/src/datachain/remote/studio.py
+++ b/src/datachain/remote/studio.py
@@ -327,9 +327,9 @@ class StudioClient:
                         logger.error("Error receiving websocket message: %s", e)
                         break
         except (websockets.exceptions.WebSocketException, OSError) as e:
-            # Studio rejects the handshake with 403 for an unknown team, an
-            # invalid token, or a missing job_id; retrying cannot fix those.
-            # Any other status (a 5xx from a gateway mid-deploy) is transient.
+            # An auth refusal (401, 403) is terminal: the same request will be
+            # refused again. Any other status (a 5xx from a gateway mid-deploy)
+            # is worth another attempt.
             response = getattr(e, "response", None)
             status = getattr(response, "status_code", None) or getattr(
                 e, "status_code", None
--- a/src/datachain/remote/studio.py
+++ b/src/datachain/remote/studio.py
@@ -327,6 +327,19 @@ class StudioClient:
                         logger.error("Error receiving websocket message: %s", e)
                         break
         except (websockets.exceptions.WebSocketException, OSError) as e:
+            # Studio rejects the handshake with 403 for an unknown team, an
+            # invalid token, or a missing job_id; retrying cannot fix those.
+            # Any other status (a 5xx from a gateway mid-deploy) is transient.
+            response = getattr(e, "response", None)
+            status = getattr(response, "status_code", None) or getattr(
+                e, "status_code", None
+            )
+            if status in (401, 403):
+                raise DataChainError(
+                    f"Studio refused the log stream connection (HTTP {status})."
+                    f" Check the team name ({self.team}) and your token."
+                ) from e
+
             logger.debug("WebSocket connection failed: %s", e)
             return
 
--- a/src/datachain/studio.py
+++ b/src/datachain/studio.py
@@ -476,6 +476,27 @@ def _process_logs_message(
     return received, last_log_id
 
 
+async def _wait_before_reconnect(retry_count: int) -> str:
+    sleep_sec = min(
+        RECONNECT_BACKOFF_BASE_SEC * 2**retry_count,
+        RECONNECT_BACKOFF_MAX_SEC,
+    ) + random.uniform(0, 1)  # noqa: S311
+    logger.debug(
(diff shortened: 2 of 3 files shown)


src/datachain/remote/studio.py
```diff
@@ -327,6 +327,19 @@ async def tail_job_logs(
...
                         break
         except (websockets.exceptions.WebSocketException, OSError) as e:
+            # Studio rejects the handshake with 403 for an unknown team, an
+            # invalid token, or a missing job_id; retrying cannot fix those.
+            # Any other status (a 5xx from a gateway mid-deploy) is transient.
+            response = getattr(e, "response", None)
+            status = getattr(response, "status_code", None) or getattr(
+                e, "status_code", None
+            )
+            if status in (401, 403):
```
shcheklein, line 337: why 401 then?

dreadatour: `HTTP 401` status is `Unauthorized`, `403` is `Forbidden`.

shcheklein: comment above I think doesn't mention it .. so, it it really returned on not?

src/datachain/studio.py
```diff
@@ -476,6 +476,27 @@ def _process_logs_message(
     return received, last_log_id
 
 
+async def _wait_before_reconnect(retry_count: int) -> str:
```
dreadatour, line 479: Note: move out from `show_logs_from_client` to avoid `# noqa: PLR0915`.

src/datachain/remote/studio.py
dreadatour: Yes, comment does not mentioned that, because I was thinking it is too obvious to be commented (`401 Unauthorized` one). Studio does not returns it for now but it is in HTTP standards and in case of `Unauthorized` error returned (possible in future) we should not retry connections, so I've decided to add it here too.

dreadatour: I would like to keep it here, I can also add a comment about `401`. WDYT?

dreadatour: Updated the comment.

shcheklein: yep, absolutely fine - this was to double check

tests/test_cli_studio.py
```diff
@@ -1751,6 +1751,173 @@ async def mock_tail_job_logs(jid, no_follow=False):
...
+    from websockets.datastructures import Headers
+    from websockets.http11 import Response
+
+    def mock_connect(url, additional_headers):
+        raise websockets.exceptions.InvalidStatus(
+            Response(503, "Service Unavailable", Headers())
+        )
+
+    mocker.patch("datachain.remote.studio.websockets.connect", side_effect=mock_connect)
+    mocker.patch("datachain.studio.RECONNECT_MAX_ATTEMPTS", 0)
```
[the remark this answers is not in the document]
dreadatour: Good finding! Fixed.

amritghimire (approved):

PR #1963, opened 2026-08-29
fix: keep the element type of a variadic tuple

shcheklein:
`tuple[int, ...]` is stored as `Array(JSON)` rather than `Array(Int64)`. `...` is
a real object — `Ellipsis` — so it turns up as an entry in the type's arguments:

```
tuple[int, ...]   get_args -> (int, Ellipsis)
tuple[int, int]   get_args -> (int, int)
```

`_list_to_array` saw `int` and `Ellipsis`, concluded "mixed", and fell back to
JSON — the right answer for `tuple[int, str]` and the wrong one here. It is read
as variable length only where it means that: **tuple origin, two arguments,
`Ellipsis` second**. Anywhere else it stays unresolvable, so `list[int, ...]` is
not quietly accepted as `list[int]`.

| annotation | before | after |
| --- | --- | --- |
| `tuple[int, ...]` | `Array(JSON)` | `Array(Int64)` |
| `tuple[int \| None, ...]` | `Array(JSON)` | `Array(Nullable(Int64))` |
| `tuple[int, str]` | `Array(JSON)` | unchanged — genuinely mixed |
| `list[int, ...]` | `Array(JSON)` | unchanged — not a valid annotation |

## Element mappings this makes reachable

Honouring the element type exposed two mappings that were harmless only while
everything landed in `Array(JSON)`:

- an enum mapped to the **builtin `str`**, which is not a column type at all, so
  `tuple[IntEnum, ...].to_dict()` raised
- a `Literal` mapped to `String` whatever it held, so `tuple[Literal[1, 2], ...]`
  carried integers in a string column

Both now follow the values, compared by exact type:

| annotation | maps to |
| --- | --- |
| `IntEnum`, `Literal[1, 2]`, `Literal[1] \| Literal[2]` | `Int64` |
| str-mixin `Enum`, `StrEnum`, `Literal["a", "b"]` | `String` |
| float-backed `Enum` | `Float` |
| `Literal[True, False]` | `Boolean` |
| `Literal["a", None]` | `String` — the `None` is nullability's business |
| `Literal[None]` | `String` — only ever holds NULL |
| `Literal[1, True]`, `Literal[1, "a"]`, plain `Enum` | **refused** |

`Literal[1] | Literal[2]` holds the same values as `Literal[1, 2]` and now
reaches the same type. Mixed categories are refused rather than mapped to JSON:
JSON cannot tell a stored `"1"` from the number, so one arm would come back as
the other's type. A plain enum raises rather than returning the builtin `str` —
that is what lets an array fall back to JSON, as before.

## Checkpoints

Changing which SQL type an annotation maps to leaves the **logical** schema hash
untouched, so a checkpoint written before this would be reused after it and its
rows read back as the wrong thing — `(1, 2)` as `("1", "2")`, under metadata
claiming integers.

A fingerprint of the flattened physical types (`SQLType.to_dict()`, sorted keys,
nested `item_type` and `dc_nullable` included) now goes into the UDF hash, from
which both the completed and partial checkpoint keys derive. It is mixed in at
`UDFAdapter.hash` rather than `UDFBase.hash`, because `_MultiSignalMapper`
overrides `hash()` and a user subclass may too — a multi-output map carried
neither protection otherwise. The recorded hashes in `test_query_steps_hash` move
for that reason.

## Compatibility

- **Old data still reads.** A column written as `Array(JSON)` holds `["1","2"]`;
  the read path accepts that and hydrates `(1, 2)`.
- **Comparisons across the boundary do not hold.** Same value, same declared
  schema, different stored text, so `merge(on=...)`, `distinct`, `subtract` and
  grouping disagree between a dataset written before and one written after.
  Nothing marks which codec a column was written with — hence the fingerprint.
- `Literal[1, "a"]` and plain enums are refused where they previously resolved to
  `String`. That mapping stored one arm as the other's type; failing is louder.

Premise for the array codec work in
[#1968](https://github.com/datachain-ai/datachain/issues/1968): while a variadic
scalar tuple is accidentally `Array(JSON)`, the JSON element rule cannot be made
uniform without that accident driving it.

## Stack

1. **this PR** — variadic tuple element type, element mappings, physical fingerprint
2. #1978 — canonical array conversion, codec version
3. #1976 — nullable element spellings, nullable-collection fallback
4. #1977 — optional model elements

### DIFF
--- a/src/datachain/lib/convert/python_to_sql.py
+++ b/src/datachain/lib/convert/python_to_sql.py
@@ -26,6 +26,16 @@ from datachain.sql.types import (
     String,
 )
 
+
+class UnstorableTypeError(TypeError):
+    """No column type holds these values faithfully.
+
+    Distinct from a type simply not resolving: a heterogeneous tuple falls back
+    to JSON, but a refusal must not, or the values it refuses would be stored
+    lossily by the fallback instead.
+    """
+
+
 PYTHON_TO_SQL = {
     int: Int64,
     str: String,
@@ -114,13 +124,26 @@ def _values_to_sql(values) -> Any:
     Values of more than one storage category have no single column type between
     them and are carried as JSON.
     """
-    # An enum member stores as its value, whether it arrived as the member or as
-    # a Literal holding one.
-    kinds = {
-        type(value.value if isinstance(value, Enum) else value)
-        for value in values
-        if value is not None
-    }
+    kinds: set[type] = set()
+    saw_member = saw_raw = False
+    for value in values:
+        if value is None:
+            continue
+        if isinstance(value, Enum):
+            saw_member = True
+            if not isinstance(value, (bool, int, float, str)):
+                # A plain enum member is not its value and nothing converts it
+                # through .value on the way to the column.
+                raise UnstorableTypeError(f"Cannot store enum member {value!r}")
+            kinds.add(type(value.value))
+        else:
+            saw_raw = True
+            kinds.add(type(value))
+
+    if saw_member and saw_raw:
+        # Stored, IntKind.ONE and 1 are the same value; reading cannot tell which
+        # was written, so the member would come back as the raw one.
+        raise UnstorableTypeError("Cannot store enum members beside their raw values")
     if not kinds:
         # Nothing but None, so the column only ever holds NULL and any type
         # would do; String is what it mapped to before.
@@ -129,10 +152,12 @@ def _values_to_sql(values) -> Any:
         # No column type holds both faithfully, and JSON does not either: it
         # cannot tell a stored "1" from the number, so refuse rather than
         # corrupt one of the arms.
-        raise TypeError(f"Cannot resolve values {sorted(map(str, kinds))} to one type")
+        raise UnstorableTypeError(
+            f"Cannot resolve values {sorted(map(str, kinds))} to one type"
+        )
     sql_type = PYTHON_TO_SQL.get(kinds.pop())
     if sql_type is None:
-        raise TypeError("Cannot resolve these values to a column type")
+        raise UnstorableTypeError("Cannot resolve these values to a column type")
     return sql_type
 
 
@@ -166,6 +191,10 @@ def list_of_args_to_type(args) -> SQLType:
             next_type = python_to_sql(next_arg)
             if next_type != first_type:
                 return JSON()
+        except UnstorableTypeError:
+            # Refused on purpose; JSON would store it lossily rather than not
+            # at all, which is what the refusal is for.
+            raise
         except TypeError:
             return JSON()
     return first_type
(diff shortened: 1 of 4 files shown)


src/datachain/lib/convert/python_to_sql.py
```diff
@@ -73,18 +83,109 @@ def python_to_sql(typ):  # noqa: PLR0911
...
+    By the exact type of each value, so a bool is not taken for an int, and
+    ignoring None -- whether the column is nullable is decided separately.
+    Values of more than one storage category have no single column type between
+    them and are carried as JSON.
+    """
+    kinds: set[type] = set()
+    member_values: set = set()
+    raw_values: set = set()
+    for value in values:
+        if value is None:
```
ilongin, line 140: Dropping `None` here loses element nullability. `tuple[Literal[1, None], ...]` currently maps to `Array(Int64)` instead of `Array(Nullable(Int64))`, so ClickHouse cannot store the valid `None` value. Could we preserve whether the literal contains `None` and mark the array item type nullable?

src/datachain/lib/convert/python_to_sql.py
```diff
@@ -73,18 +83,109 @@ def python_to_sql(typ):  # noqa: PLR0911
...
         raise TypeError(f"Cannot resolve type '{typ}' for flattening features")
+    if get_origin(typ) is tuple and len(args) == 2 and args[1] is Ellipsis:
+        # tuple[T, ...] is a homogeneous tuple: the Ellipsis marks variable
+        # length rather than naming a second element type. Anywhere else it is
+        # not valid, and stays unresolvable.
+        args = args[:1]
+    if not args:
+        raise TypeError(f"Cannot resolve type '{typ}' for flattening features")
     args0 = args[0]
     if ModelStore.is_pydantic(args0):
```
ilongin, line 190: This early return makes tuple handling order-dependent. `tuple[Model,Literal[1, "1"]]` returns `Array(JSON)`, while the reversed tuple raises `UnstorableTypeError`.

tests/unit/lib/test_python_to_sql.py
```diff
@@ -70,3 +71,162 @@ def test_pep_604_union_syntax():
...
+    ids=[
+        "variadic-int",
+        "variadic-str",
+        "fixed-same",
+        "fixed-mixed",
+        "nested-variadic",
+        "list-unchanged",
+    ],
+)
+def test_variadic_tuple_keeps_its_element_type(annotation, expected):
```
ilongin, line 95: The tests verify the SQL type mapping, but none writes and reads an actual `tuple[int, ...]` to DB. Could we add one functional round-trip test with a tuple inside a Pydantic model? This would cover conversion to the array column and hydration back to a tuple.

PR #1964, opened 2026-08-29
fix: normalize models inside collections before they reach the warehouse

shcheklein:
The same model is stored two different ways depending on the container it sits in.

```python
class Item(dc.DataModel):
    n: int

class InList(dc.DataModel):
    items: list[Item]

class InTuple(dc.DataModel):
    items: tuple[Item, ...]
```

What lands in the column:

```
list[Item]        [{"n":1}]        an object
tuple[Item, ...]  ["{\"n\":1}"]    a string holding JSON
```

## Why

A model inside a `list` is replaced by its dump before it reaches storage. Inside a tuple, a mapping value, or a UDF's returned collection it is not — it arrives live and the warehouse converts it there instead, by different rules.

Two places decide this and neither did it completely:

- `_flatten_fields_values` chose from the **runtime type** of the value and knew only `list` and `dict`. Its `dict` branch converted a value that *was* a model, but not one that was a list of them.
- `flatten_value`, which UDF outputs go through, converted nothing unless the whole annotation was a model — so a UDF returning `list[Item]` handed over live models.

Both now share one normalizer that walks by **declared type**.

## What that buys, beyond consistency

| in a `tuple[Model, ...]` | before | after |
|---|---|---|
| numpy field | **fails to save** | works |
| redacting `when_used="json"` serializer | persists `'***'`, original lost | stores the real value |
| `Decimal` | `"1.20"` | `1.2`, same as a list |
| `Path` field | works | fails, as in every other shape |

The serializer row is the one worth reading twice: a serializer written to hide a value in an API response was deciding what went into the database, and the original was unrecoverable.

## Deciding from the annotation

- A type that cannot hold a model is returned untouched, so `list[int]` is never copied — and this holds per element, so `tuple[Item, list[int]]` converts the model and leaves the list alone.
- `Optional` is unwrapped before arguments are read, so `list[Item] | None` and `None | list[Item]` behave identically.
- Elements are converted one at a time rather than from whatever the first one is, which `tuple[Item, int]` needs.
- An erased annotation — `Any`, `object`, a bare container — says nothing about its contents, so it is walked rather than assumed model-free.

## Compatibility

`tuple[Model, ...]` columns written before this will not compare equal to ones written after — `filter`, `distinct`, `subtract`, `merge`. **Reads are unaffected**: verified by writing such a column on `main` and reading it here, where both spellings load as the declared tuple.

## Relationship to #1943

Independent — neither needs the other. But together they close two of that PR's strict `xfail`s: a colliding mapping inside a live model reaches `model_dump(mode="json")`, which merges the keys before anything can inspect them. Normalizing first means it arrives as a plain dict and the existing check sees both keys. UDF-returned `list[Model]` and `dict[str, Model]` are covered too. Whichever lands second should flip those markers.

### DIFF
--- a/src/datachain/lib/convert/flatten.py
+++ b/src/datachain/lib/convert/flatten.py
@@ -1,5 +1,5 @@
 from collections.abc import Generator, Iterator
-from typing import Any, NamedTuple
+from typing import Any, NamedTuple, get_args, get_origin
 
 from pydantic import BaseModel
 
@@ -72,7 +72,7 @@ def flatten_value(value: Any, anno: Any) -> tuple:
             # Non-Optional model None (outer-merge pad): per-leaf placeholders.
             return tuple(_emit_absent(kind.inner))
         return flatten(value)
-    return (value,)
+    return (normalize_models(value, anno),)
 
 
 def flatten_list(obj_list: list[BaseModel]) -> tuple:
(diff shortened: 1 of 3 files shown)


tests/unit/lib/test_signal_schema.py
```diff
@@ -1580,6 +1580,70 @@ def test_signal_resolving_type_error_with_unpicklable_field():
...
+    "annotation,value,expected",
+    [
+        (list[_Boxed], [_Boxed(n=1)], [{"n": 1}]),
+        (dict[str, _Boxed], {"k": _Boxed(n=1)}, {"k": {"n": 1}}),
+        (tuple[_Boxed, ...], (_Boxed(n=1),), [{"n": 1}]),
+        (list[int], [1, 2], [1, 2]),
+    ],
+    ids=["list", "dict", "tuple", "no-models"],
+)
+def test_flatten_value_normalizes_udf_output(annotation, value, expected):
```
ilongin, line 1634: This bypasses the new normalizer for a single tuple-typed output. For `output={"items": tuple[Item, ...]}`, a result `(Item(n=1), Item(n=2))` is returned unchanged instead of becoming `([{"n": 1}, {"n": 2}],)`. The models remain live and the tuple can be mistaken for two output columns.

ilongin, line 1634: Can we also add a functional test that saves and reads a real UDF result?

src/datachain/lib/convert/flatten.py
```diff
@@ -83,12 +83,65 @@ def flatten_list(obj_list: list[BaseModel]) -> tuple:
...
-    assert isinstance(value, list)
-    if value and ModelStore.is_pydantic(type(value[0])):
-        return [val.model_dump() for val in value]
-    if value and isinstance(value[0], list):
-        return [_flatten_list_field(v) for v in value]
+_ERASED = (Any, object, None)
+_BARE_CONTAINERS = (dict, list, tuple, set, frozenset)
+
+
+def _may_hold_model(annotation: Any) -> bool:
```
ilongin, line 90: Does this adds another recursive annotation walk for every row?

PR #1969, opened 2026-08-31
Ilongin/1910 preserve pydantic class identity

### LINKED ISSUE #1910: Nested pydantic models lose their class identity in another process
### Description

A dataset row containing a plain `pydantic.BaseModel` reads back, in a process that did not write it, as a **different class** rebuilt from the stored schema. `isinstance` fails and passing the value into anything annotated with that type raises a `ValidationError` that contradicts itself.

```python
# models.py

class Thresholds(BaseModel):
    name: str
    limit: float

class Scenario(BaseModel):
    key: int
    thresholds: Thresholds
```

```python
# write.py  (process 1)
import datachain as dc
from models import Scenario, Thresholds

dc.read_values(key=[1, 2]).map(
    s=lambda key: Scenario(key=key, thresholds=Thresholds(name="v1", limit=0.5)),
    output={"s": Scenario},
).save("nested_demo")
```

```python
# read.py  (process 2)
import datachain as dc
from models import Scenario, Thresholds

row = dc.read_dataset("nested_demo").to_list("s")[0][0]
print(type(row.thresholds))                    # <class 'datachain.lib.signal_schema.Thresholds'>
print(isinstance(row.thresholds, Thresholds))  # False
Scenario(key=99, thresholds=row.thresholds)    # ValidationError
```

```
ValidationError: 1 validation error for Scenario
thresholds
  Input should be a valid dictionary or instance of Thresholds
  [type=model_type, input_value=Thresholds(name='v1', limit=0.5), input_type=Thresholds]
```

Write and read in **one** process and it prints `<class 'models.Thresholds'>` / `True`. Subclassing `dc.DataModel` also works. Plain `BaseModel` in a fresh process always fails

### Root cause

`SignalSchema._deserialize_custom_type` (`src/datachain/lib/signal_schema.py:419-460`) returns
`ModelStore.get(name, version)` if it hits, and otherwise **rebuilds** the class from the stored field
spec. `DataModel` registers itself in `ModelStore` at class definition
(`data_model.py:__pydantic_init_subclass__`); a plain `BaseModel` never registers, so a fresh process
always misses and rebuilds. In the writing process the lookup hits, which is why the bug is invisible
locally and only appears once producing and consuming steps are separate jobs, the normal case in
production.

The rebuilt class is then registered under the bare `name@version`, which is also the mechanism behind
#1872 (same-named models shadowing each other across datasets).

### Suggested fix

Synthesizing a class should be the last resort, not the first. In `_deserialize_custom_type`, before
rebuilding:

1. `ModelStore.get(name, version)` as today, but keyed additionally by a structural fingerprint so
   different shapes cannot collide (this also closes #1872);
2. **resolve the already-loaded class**: `sys.modules.get(module)` -> `getattr(module, name)`, accept
   if the field spec matches. The module is already persisted: `_get_bases` stores the full MRO with
   `__module__`, so `bases[0]` is `["Thresholds", "models", null]` in existing datasets. No schema
   change, no migration, works on data written by current versions;
3. otherwise scan the live `BaseModel.__subclasses__()` closure by name + field spec (covers
   notebooks, `__main__`, re-exports);
4. only then rebuild, and warn once when a same-named class with a different shape is loaded in this
   process, instead of silently returning a class that will not match.

Important: resolve only from classes **already imported in this process**. Do not import a module path
taken from dataset metadata: datasets are shared, and that would be code execution from data. No user
action is required either way, since using the type at all means the user has already imported it.

Two follow-ups in the same area: user validators will start running on read once the real class is
resolved (correct, but a behaviour change worth a release note), and there is currently no
cross-process test for this. A two-process fixture (write in one, read in another) over
`BaseModel`/`DataModel` x nested/list/optional would have caught it.

### Version Info

```Text

```

ilongin:
## Summary

Preserve the identity of plain Pydantic models when reading saved datasets in a new
process.

Previously, only `DataModel` subclasses were registered automatically. A plain
`BaseModel` was reconstructed from its stored schema even when the original class was
already imported. The reconstructed class had the same fields but different Python
identity, which broke `isinstance()` checks and nested Pydantic validation.

Fixes #1910.

## Changes

- Reuse an already-imported plain Pydantic class when its module, name, and stored
  schema match.
- Compare nested models through DataChain's existing schema serialization path.
- Keep existing `ModelStore` behavior, including compatibility with schema-v1
  datasets.
- Resolve classes only from `sys.modules`; stored metadata never triggers an import or
  module `__getattr__`.
- Fall back to the existing synthetic model when the original class is unavailable or
  does not match the stored schema.
- Document automatic resolution and synthetic fallback.

## Tests

- Added a cross-process dataset test covering direct, list, and optional nested
  Pydantic models, `isinstance()` behavior, and reader-side validation.
- Added parametrized unit coverage for imported model reuse across nested annotation
  forms.
- Added unit coverage for root and nested schema mismatches, schema-v1/v2
  `ModelStore` compatibility, side-effect-free comparison, and safe module lookup.


src/datachain/lib/signal_schema.py
```diff
@@ -456,7 +578,13 @@ def _deserialize_custom_type(
                     f"cannot deserialize custom type '{type_name}': {exc}"
                 ) from exc
 
+            expected_shape_hash = _stored_shape_hash(type_name, custom_types)
             if fr := ModelStore.get(model_name, target_version):
+                if _class_matches_shape(fr, expected_shape_hash):
+                    return fr
+                _warn_shape_drift(fr, type_name)
```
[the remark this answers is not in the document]
ilongin: Fixed after refactoring

src/datachain/lib/signal_schema.py
```diff
@@ -212,6 +213,101 @@ def __reduce__(self):
...
+        name = ModelStore.get_name(model)
+        if name in models:
+            return
+        models[name] = {
+            "fields": {
+                field_name: type_to_str(info.annotation, register_pydantic=False)
+                for field_name, info in model.model_fields.items()
+            },
+            "bases": [list(base) for base in SignalSchema._get_bases(model)],
+        }
```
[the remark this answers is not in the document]
ilongin: Fixed by refactoring

src/datachain/lib/signal_schema.py
```diff
@@ -456,7 +578,13 @@ def _deserialize_custom_type(
...
                 ) from exc
 
+            expected_shape_hash = _stored_shape_hash(type_name, custom_types)
             if fr := ModelStore.get(model_name, target_version):
+                if _class_matches_shape(fr, expected_shape_hash):
+                    return fr
+                _warn_shape_drift(fr, type_name)
+
+            if fr := _resolve_from_sys_modules(ct, type_name, expected_shape_hash):
                 return fr
```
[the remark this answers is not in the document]
ilongin: Skipping this. If user imports on the top of the file as expected, everything will work just fine. Code would ge too complex if we would now go into solving this and there is no actual need.

tests/func/test_pydantic_class_identity.py
shcheklein, line 1: do we really need the whole new file for a single test?

ilongin: Moved into test_model_store_rebuild.py and dropped the standalone file

src/datachain/lib/signal_schema.py
```diff
@@ -237,6 +238,49 @@ def deserialize(cls, data: dict[str, Any], type_name: str) -> "CustomType":
...
+            stored_type = CustomType.deserialize(custom_types[name], name)
+        except ValidationError as exc:
+            raise SignalSchemaError(
+                f"cannot deserialize custom type '{name}': {exc}"
+            ) from exc
+        if CustomType.deserialize(data, name) != stored_type:
+            matches = False
+            break
+    if matches:
+        return candidate
```
shcheklein, line 298: **AI-generated review by OpenAI Codex — [P1] Preserve stored field values when reusing models with aliases**

Returning the matching imported class here exposes a read regression for ordinary Pydantic field aliases, even when the class definition has not changed. DataChain stores the Python field name (`value`), but `_hydrate_model()` calls `fr(**j)` with that name. With Pydantic's default alias configuration, the original class expects `externalValue` instead: a required field raises `ValidationError`, while a field with a default silently replaces the saved value with that default.

Reproduce in a fresh directory using the default local SQLite backend. Create these three files, then run `python write.py` followed by `python read.py` as separate processes:

```python
# models.py
from pydantic import BaseModel, Field

class AliasedModel(BaseModel):
    value: int = Field(0, alias="externalValue")
```

```python
# write.py
import datachain as dc
from models import AliasedModel

dc.read_values(
    settings={"prefetch": False},
    item=[AliasedModel(externalValue=7)],
).save("pr1969_alias_repro")
```

```python
# read.py
import datachain as dc
from models import AliasedModel

row = dc.read_dataset("pr1969_alias_repro").to_list("item")[0][0]
assert row.value == 7, repr(row)
```

At `f079f7b`, the assertion fails with `AliasedModel(value=0)`. Changing the field to `Field(alias="externalValue")` makes the read fail with a missing-`externalValue` validation error instead. Both read-back cases pass with the PR's base version of `signal_schema.py` (`c83d0216`), which reconstructs a model using the stored field names.

Please keep the structural compatibility check, and make hydration consume persisted dictionaries as field-name data. One tested approach is `model_validate(data, by_alias=False, by_name=True)`; both `_hydrate_model()` and `_convert_feature_value()` have constructor calls to consider, including nested/container cases. Replacing those two calls in an isolated experiment passed the existing 164 tests plus the two alias regressions. Please add permanent cross-process coverage for both required aliases and aliases with defaults.

Verified locally with Python 3.13, Pydantic 2.13.5, and SQLite; ClickHouse was not rerun locally.

src/datachain/lib/signal_schema.py
```diff
@@ -237,6 +238,49 @@ def deserialize(cls, data: dict[str, Any], type_name: str) -> "CustomType":
...
+    module = sys.modules.get(module_name)
+    if module is None:
+        return None
+    candidate = vars(module).get(class_name)
+    if candidate is None or not ModelStore.is_pydantic(candidate):
+        return None
+    serialized: dict[str, Any] = {}
+    SignalSchema._serialize_custom_model(
+        ct.name, candidate, serialized, register_pydantic=False
+    )
```
shcheklein: **AI-generated review by OpenAI Codex — [P2] Reach synthetic fallback when the imported definition has become recursive**

This call serializes the entire candidate before comparing any fields. `_serialize_custom_model()` only writes `custom_types[version_name]` after serializing the model's children, so its existing `if version_name in custom_types` check does not stop a cycle. A previously supported, acyclic saved schema becomes unreadable merely because the imported class now contains a recursive field: comparison raises `RecursionError` instead of rejecting the changed class and rebuilding from the stored schema.

Standalone reproduction:

```python
from pydantic import BaseModel, Field
from datachain.lib.model_store import ModelStore

class Node(BaseModel):
    value: int

saved = SignalSchema({"item": Node}).serialize()
ModelStore.store.pop("Node", None)  # Simulate a fresh reader process.

class Node(BaseModel):
    value: int
    children: list["Node"] = Field(default_factory=list)

restored = SignalSchema.deserialize(saved)
assert restored.values["item"] is not Node
assert restored.row_to_objs([7])[0].value == 7
```

At `f079f7b`, `SignalSchema.deserialize(saved)` raises `RecursionError` through `_serialize_custom_model()` / `_serialize_type()`. The same script passes with the base module (`c83d0216`). The dataset being read is not recursive; only the incompatible candidate is, so supporting recursive dataset storage is not a prerequisite for handling this case correctly.

Please make compatibility checking cycle-safe and let an unsupported cyclic candidate fall back to the stored definition. Add a regression for an acyclic saved model whose imported replacement introduces a recursive child. The fix should retain nested compatibility checking, rather than accepting the candidate by module/name alone.

Verified locally with Python 3.13 and Pydantic 2.13.5; this reproduction exercises schema resolution without a database.

src/datachain/lib/signal_schema.py
```diff
@@ -237,6 +238,49 @@ def deserialize(cls, data: dict[str, Any], type_name: str) -> "CustomType":
...
+        if name not in custom_types:
+            matches = False
+            break
+        try:
+            stored_type = CustomType.deserialize(custom_types[name], name)
+        except ValidationError as exc:
+            raise SignalSchemaError(
+                f"cannot deserialize custom type '{name}': {exc}"
+            ) from exc
+        if CustomType.deserialize(data, name) != stored_type:
```
shcheklein, line 294: **AI-generated review by OpenAI Codex — [P2] Do not let serialization by short name skip an incompatible nested class**

This comparison can report a match even though a nested field's type changed. `_serialize_custom_model()` deduplicates using `ModelStore.get_name()`, which omits the module. If an imported parent contains two distinct classes named `Leaf`, serialization records the first and skips the second entirely. The serialized parent still refers to `Leaf` for both fields, so changing only the second class is invisible to this comparison.

Standalone reproduction, representing classes from two different modules:

```python
from pydantic import create_model
from datachain.lib.model_store import ModelStore

Left = create_model("Leaf", __module__="left_models", value=(int, ...))
Right = create_model("Leaf", __module__="right_models", value=(int, ...))
Outer = create_model(
    "Outer", __module__=__name__, left=(Left, ...), right=(Right, ...)
)
saved = SignalSchema({"item": Outer}).serialize()
for name in ("Outer", "Leaf"):
    ModelStore.store.pop(name, None)  # Simulate a fresh reader process.

Right = create_model("Leaf", __module__="right_models", value=(str, ...))
Outer = create_model(
    "Outer", __module__=__name__, left=(Left, ...), right=(Right, ...)
)
restored = SignalSchema.deserialize(saved)
assert restored.values["item"] is not Outer  # Fails: changed class was accepted.
assert restored.row_to_objs([1, 2])[0].right.value == 2
```

On `f079f7b`, the first assertion fails: the current, incompatible `Outer` is reused. Without that assertion, reading the stored integer raises a validation error for `right.value`. I also reproduced this through actual `.save()` / `read_dataset(...).to_list()` in separate processes: the writer's two `Leaf.value` fields are `int`, and the reader changes only the right-hand one to `str`.

Both the standalone and cross-process read-back cases pass with the base module (`c83d0216`), which reconstructs the saved integer fields. This is therefore a false-positive introduced by using the existing name-deduplicating serializer as proof that every reachable candidate class is compatible; it does not require fixing the broader `ModelStore` collision problem in this PR.

Please ensure every distinct reachable candidate class is checked. At minimum, treat ambiguous collisions between distinct classes sharing a serialized name as a non-match and use the existing synthetic fallback. A short-name cache hit is not sufficient evidence that the second class has already been checked.

Verified locally with Python 3.13, Pydantic 2.13.5, and SQLite; ClickHouse was not rerun locally.

shcheklein (commented):
**AI-generated review by OpenAI Codex, posted at the user's request.**

Re-reviewed all five changed files at `f079f7b8ad830d9d1f8c984d82ed62b0efa19855`, with the requirement that automatic class reuse must check compatibility with the stored schema, including nested models. Reusing the existing serializer is a reasonable simplification; lookup by module/name alone would not satisfy that requirement.

Three confirmed issues are detailed inline: alias hydration can silently replace saved values, a recursive replacement class prevents synthetic fallback, and distinct nested classes with the same name can hide an incompatible schema change.

Validation: the existing signal-schema and model-rebuild suites pass (164 tests). Five added regression tests covering these three issues fail on the PR and pass with the base version of the changed production module (`c83d0216`). Local validation used Python 3.13, Pydantic 2.13.5, and SQLite; I did not rerun ClickHouse locally.

src/datachain/lib/signal_schema.py
ilongin: Confirmed with the repro - real regression from switching to the imported class. Switched `_hydrate_model` and `_convert_feature_value` to `model_validate(..., by_alias=False, by_name=True)`, which is exactly the "accept the Python field name even when aliases exist" contract DataChain's storage assumes. Added a cross-process regression test in `test_model_store_rebuild.py` covering both a defaulted alias and a required alias.

src/datachain/lib/signal_schema.py
ilongin: This is a bug, verified with the repro. Wrapped the `_serialize_custom_model` call in `_resolve_from_sys_modules` in a `try/except RecursionError` so a cyclic candidate falls through to the synthetic rebuild path instead of blowing the stack, with a warning for parity with the shape-mismatch case. The serializer itself is unchanged  - the recursion is only a problem when we're using it as a comparability check on an imported class we don't control, not when serializing an acyclic tree for storage. Added a regression test with an acyclic stored schema and a recursive imported replacement.

src/datachain/lib/signal_schema.py
ilongin: Fixed by detecting when distinct nested Pydantic classes collapse to the same serialized name during the existing comparison traversal. In that ambiguous case, we now reject the imported parent and fall back to the stored schema. Added a regression test covering the example.

shcheklein:
**AI-generated review by OpenAI Codex, posted at the user's request.**

Reviewed commit `4799ea07cd482719c0df4aa8632069587f7b267b`. The earlier three findings have been addressed, but I recommend holding the merge for two new hydration regressions.

1. **[P1] Models serialized with aliases can silently lose values.**

   With `ConfigDict(serialize_by_alias=True)`, models inside lists/dictionaries are stored using alias keys, such as `{"externalValue": 7}`. The new `by_alias=False` rejects those keys: a defaulted field reads back as **`0` instead of `7`**, while a required field raises `ValidationError`. This also affects lists nested inside another model, and reproduces in same-process save/read-back.

   Flattened model fields and JSON container contents currently use different representations; hydration must accommodate both. Simply enabling aliases everywhere also fails when an alias equals another field's name: a model with `left: int = Field(alias="right")` and `right: int` validates `{"left": 1, "right": 2}` as `left=2, right=2` when both alias and name validation are enabled. See [container hydration](https://github.com/datachain-ai/datachain/blob/4799ea07cd482719c0df4aa8632069587f7b267b/src/datachain/lib/signal_schema.py#L1054) and [model hydration](https://github.com/datachain-ai/datachain/blob/4799ea07cd482719c0df4aa8632069587f7b267b/src/datachain/lib/signal_schema.py#L923).

2. **[P2] Top-level `RootModel` read-back now breaks.**

   For `class Scalar(RootModel[int])`, the previous `Scalar(**{"root": 7})` worked. The replacement `Scalar.model_validate({"root": 7}, ...)` fails because validation expects the root value, `7`. This affects existing same-process usage too. Preserve root-model input semantics when changing [this constructor call](https://github.com/datachain-ai/datachain/blob/4799ea07cd482719c0df4aa8632069587f7b267b/src/datachain/lib/signal_schema.py#L923).

**Simplification suggestions**

I would retain nested compatibility checking and reuse of the existing serializer, then make these targeted changes:

- **Detect cycles using the existing tracking map.** A class already in `serialized_models` but absent from completed `custom_types` is still being visited. Reject it immediately instead of recursing until `RecursionError`. This reuses state already introduced for collisions. [Tracking code](https://github.com/datachain-ai/datachain/blob/4799ea07cd482719c0df4aa8632069587f7b267b/src/datachain/lib/signal_schema.py#L434).
- **Emit the fallback warning once.** Collect the incompatibility reason, then use one warning/return path. The resolver currently duplicates warning construction and combines exception handling with a mutable `matches` flag. [Resolver](https://github.com/datachain-ai/datachain/blob/4799ea07cd482719c0df4aa8632069587f7b267b/src/datachain/lib/signal_schema.py#L246).
- **Keep the original known-base regression test unchanged.** Its `MyType3` is a `DataModel`, which the new resolver explicitly excludes. I verified that the original test still passes; replacing it with a function-local `HiddenChild` is unnecessary.
- **Share the subprocess test harness and model definitions.** One temporary `models.py`, imported by writer and reader, would reduce duplicated scripts and directly exercise the external-module scenario. Current subprocess definitions live in `__main__`.

**Other missing pieces**

- Document that restored classes run reader-side validators. Matching stored structure does not establish compatibility of validators, defaults, or constraints; reads can transform or reject saved values.
- Document the import-order caveat: import models before the first dataset read. An existing synthetic `ModelStore` entry takes precedence over a subsequently imported class. This can remain a documented limitation without expanding this PR's implementation.
- Add permanent coverage for the regressions above, including defaulted/required aliases across lists, dictionaries, and nested containers, on both supported backends.

**Validation:** the existing `tests/unit/lib/test_signal_schema.py` and `tests/func/test_model_store_rebuild.py` suites pass **181 tests**. The seven additional save/read-back cases below all fail on this PR and all pass when only `signal_schema.py` is replaced with the base version from `4aa5bf706e79a943da350d665e299ed89485f505`. Tested with Python 3.13.3, Pydantic 2.13.5, and SQLite. ClickHouse was not run locally. Two additional RootModel compositions also failed on the base and are excluded from these findings.

<details>
<summary>Seven regression cases (place in tests/func/ to use the existing test_session fixture)</summary>

```python
import pytest
from pydantic import BaseModel, ConfigDict, Field, RootModel

import datachain as dc


def test_root_model_readback(test_session):
    class Scalar(RootModel[int]):
        pass

    dc.read_values(
        session=test_session,
        settings={"prefetch": False},
        item=[Scalar(7)],
    ).save("review_root_model")
    restored = dc.read_dataset(
        "review_root_model", session=test_session
    ).to_list("item")[0][0]
    assert restored.root == 7


@pytest.mark.parametrize("container", ["list", "dict", "nested_list"])
@pytest.mark.parametrize("required", [False, True])
def test_serialized_aliases_readback(test_session, container, required):
    class Aliased(BaseModel):
        model_config = ConfigDict(serialize_by_alias=True)
        value: int = Field(... if required else 0, alias="externalValue")

    class Wrapper(BaseModel):
        values: list[Aliased]

    obj = Aliased(externalValue=7)
    item = {
        "list": [obj],
        "dict": {"a": obj},
        "nested_list": Wrapper(values=[obj]),
    }[container]
    dc.read_values(
        session=test_session,
        settings={"prefetch": False},
        item=[item],
    ).save("review_alias_model")
    restored = dc.read_dataset(
        "review_alias_model", session=test_session
    ).to_list("item")[0][0]
    if container == "list":
        assert restored[0].value == 7
    elif container == "dict":
        assert restored["a"].value == 7
    else:
        assert restored.values[0].value == 7
```

</details>

shcheklein (approved):

shcheklein:
@ilongin review / fix tests

shcheklein (commented):
**AI-generated re-review (Codex), posted at the user's request.**

Reviewed all five changed files at `4799ea0`, plus the related changes in #1947, #1877, #1964, and the merged annotation-walker refactor #1911. The head is unchanged from the [previous detailed review](https://github.com/datachain-ai/datachain/pull/1969#issuecomment-5787175948). I recommend addressing the read regressions before merging.

Current findings:

- **[P1, previously reported and reverified] Aliased model values inside containers can be silently replaced by defaults.** All six earlier cases still fail: defaulted/required alias × list/dictionary/nested list. `serialize_by_alias=True` writes alias keys, while the new hydration flags reject them.
- **[P2, previously reported and reverified] Top-level `RootModel[int]` read-back fails.** The new `model_validate` call receives `{"root": 7}` where this model expects `7`.
- **[P2, additional finding] Equality after annotation erasure is not proof of nested compatibility.** A changed child inside `Sequence[Child]` is accepted because `type_to_str` writes only `Sequence` and never records/compares `Child`. Actual saved values then fail validation. Reproduction and a narrow fallback recommendation are inline.
- **[P2, current CI failure] The recursive-model fixture does not exercise the intended path on Python 3.10.** The [Ubuntu job](https://github.com/datachain-ai/datachain/actions/runs/35735573512/job/106771818153) fails this newly added test: it sees an unresolved child name and a generic mismatch warning, rather than the asserted recursive-definition warning. Fix the fixture and rerun the supported-version checks.

Related-PR gaps:

- **[#1877 review](https://github.com/datachain-ai/datachain/pull/1877#pullrequestreview-5298956516):** verified and commented on two issues: synthetic fallback inherits the incompatible registered `DataModel`, bringing back fields absent from storage; and its comparison registers nested classes, changing later identity resolution. The real-class inheritance case is not covered by the hand-built schemas with empty `bases`.
- **#1947:** this is an alternative to #1969 for the same issue, not another layer to merge. Keep one resolver. #1969's `vars(module).get(...)` and actual cross-process dataset tests are preferable pieces to retain; #1947's extra subclass scan and second recursive comparison should not be added alongside them.
- **#1964:** its collection normalizer is the right place to coordinate how model dictionaries are written, but it currently calls `model_dump()` with the model's alias policy. Merging it does not establish the field-name-only input contract assumed by #1969. Future writes and already-saved alias-key data both need an explicit read/write agreement.

Concrete simplification and reuse plan:

| Existing piece | Reuse / boundary |
| --- | --- |
| `_serialize_custom_model` plus `register_pydantic=False` | One inspection traversal shared with #1877. Keep the relevant schema-v1 rules at the caller; do not blindly replace its legacy policy with full v2 metadata equality. Reject comparisons whose serialization omitted reachable model information. |
| `serialized_models` and completed `custom_types` | Reuse their visiting/completed distinction to detect cycles immediately, instead of exhausting the Python stack. Keep distinct-class name collisions as a separate rejection. |
| `annotation_parts`, `is_sequence_annotation`, `is_mapping_annotation` from #1911 | Reuse these to identify supported nested annotations and detect erased edges. Keep persisted type spelling stable; do not casually normalize old `Sequence` metadata to `list`. |
| Existing two hydration entry points | Centralize the persisted-value policy while respecting flattened fields, JSON containers, and `RootModel` inputs. Simply setting `by_alias=True` everywhere is unsafe when an alias equals another field name. Avoid separate ad-hoc fixes at each constructor. |
| `normalize_models` proposed in #1964 | Share the collection-writing boundary rather than adding another model/container encoder here. Test compatibility with legacy data before claiming normalization resolves the alias regression. |
| #1877's stored-spec fingerprint | Use for synthetic cache identity only. It is not another live-class match hash, and the existing selection/default fingerprint has a different contract. |
| `run_test_subprocess`, `wait_for_test_subprocess`, `catalog_tmpfile` | One parametrized writer/reader harness and a temporary shared `models.py`. This reduces duplicate environment/script setup and tests the original external-module scenario instead of only duplicated `__main__` definitions. |

Within this PR, keep one warning/fallback exit and restore the original known-base test unless a behavior change requires replacing it. Its original `DataModel` case is outside the new plain-model resolver. Treat the hydration correction as an independently testable prerequisite to the small identity lookup change. There is no need to bring back parallel hash walkers or add another global cache.

Missing permanent coverage: aliases with defaults and required aliases across direct/container inputs; root models; an imported nested type drifting under an abstract collection; registry state and identity before/after comparison; real `DataModel` base metadata in collision tests; and actual recursive annotations on Python 3.10. Import-before-first-read and reader-side validator execution also need documentation, as discussed in the earlier review.

Validation: the existing signal-schema/model-rebuild suites pass **181 tests** on Python 3.13.3/Pydantic 2.13.5/SQLite. The seven previously reported save/read-back cases plus the additional Sequence case fail on this head and all **8 pass** when only `signal_schema.py` is replaced with the base version `4aa5bf7`. For the Sequence comparison, the base's expected unresolved-type warning is allowed; it reconstructs the stored value as raw dictionaries, whereas this head selects the incompatible live class and raises `ValidationError`. Related #1877 has 120 existing schema tests plus its explode/read-back regression passing, while the three additional checks described above fail. ClickHouse and the full version matrix were not rerun locally; I inspected the current CI failure. All local review files are isolated under `/tmp`; working branches are unchanged.

src/datachain/lib/signal_schema.py
```diff
@@ -238,6 +239,73 @@ def deserialize(cls, data: dict[str, Any], type_name: str) -> "CustomType":
...
+            matches = False
+            break
+    if matches:
+        return candidate
```
shcheklein, line 298: **AI-generated review (Codex) — [P2] Reject compatibility checks that erased nested model types.**

`type_to_str` recognizes literal list/dict/tuple origins, but serializes `collections.abc.Sequence[Child]` as just `Sequence` and does not include `Child` in `subtypes`. Consequently this comparison accepts an imported parent even when its nested child's field changed from `int` to `str`: both serialized parents contain `{"items": "Sequence"}`, and there is no child entry to compare.

I reproduced this through `.save()` / `read_dataset(...).to_list()`: write `Outer(items=[Child(value=7)])` with `items: Sequence[Child]`; clear the two model registrations to simulate a fresh reader; expose same-named imported classes with `Child.value: str`. On this head the incompatible `Outer` is selected and reading raises `ValidationError` at `items.0.value`. With the base module, the expected unresolved-Sequence warning is emitted and the synthetic model preserves `items == [{"value": 7}]`. The existing annotation tests exercise concrete containers only.

Please keep the stored-schema check, but treat erased reachable type information as non-comparable and fall back rather than accepting equal incomplete serializations. The annotation helpers from #1911 can identify these edges without adding another independent list/dict-only walker. Add this saved-dataset regression before extending the automatic-reuse guarantee to abstract collection annotations.

tests/unit/lib/test_signal_schema.py
```diff
@@ -1911,3 +1940,203 @@ def test_enrich_expr_types_unknown_column():
...
+
+
+def test_deserialize_falls_back_when_imported_model_became_recursive(monkeypatch):
+    monkeypatch.setattr(ModelStore, "store", {})
+    stored = create_model("RecursiveNodeDrift", __module__=__name__, value=(int, ...))
+    serialized = SignalSchema({"x": stored}).serialize()
+
+    class RecursiveNodeDrift(BaseModel):
+        value: int
+        children: list["RecursiveNodeDrift"] = Field(default_factory=list)
```
shcheklein, line 2089: **AI-generated review (Codex) — [P2] Resolve the recursive fixture on Python 3.10 before asserting its fallback.**

The current [Python 3.10 Ubuntu CI job](https://github.com/datachain-ai/datachain/actions/runs/35735573512/job/106771818153) fails this new test with Pydantic 2.13.5. Its warnings are `Unable to determine name of type 'RecursiveNodeDrift'` and the generic stored-schema mismatch warning; neither matches `recursive definition`. The child annotation reaches serialization as an unresolved name, so this environment never tests the cycle-handling branch. The 3.10 jobs are currently red on all three operating systems.

Make the fixture's self-reference explicitly resolve/rebuild and assert the child argument is the actual class before exercising fallback, then rerun the supported-version matrix. Broadening the warning regex alone would make the test pass without covering the recursion regression. The same test passes locally on Python 3.13, so that result alone misses this gap.

src/datachain/lib/signal_schema.py
ilongin: Fixed by detecting parameterized collection annotations whose type arguments are lost during schema serialization. These candidates are now treated as non-comparable and fall back to the stored synthetic model. Added a regression test for Sequence[Child] with a drifted child schema.

tests/unit/lib/test_signal_schema.py
ilongin: fixed

PR #1985, opened 2026-09-09
Split the core skill into SKILL.md and SDK.md

dreadatour:
Split the `datachain-core` skill into `SKILL.md` and `SDK.md` so the SDK rules can be served outside a local checkout.

## Why

The core skill mixes two kinds of content. Most of it is SDK rules: UDF signatures, settings, delta semantics, the API reference, pipeline templates. A handful of lines depend on a local checkout with the knowledge skill installed beside it: read `dc-knowledge/index.md` first, check `dc-knowledge/buckets/` for the `anon` flag, run `datachain bucket status`, never write under `dc-knowledge/`.

Studio wants to hand the SDK rules to agents, so they write correct scripts. Those agents have no shell and no `dc-knowledge/` directory, so every local-only line is a dead pointer at best and a wrong instruction at worst (an agent taking the bucket-status line literally stops and asks for credentials Studio already holds). Serving the file as is and filtering those lines in Studio would be brittle against edits here. Splitting at the source keeps one copy of the rules.

## What changed

- `src/datachain/skill/core/SDK.md` (new): the pre-generation checklist, Section 1 dataset reuse, the CAST quick reference, and Sections 2 to 8, moved verbatim. The only edits are the removed local-only lines: the bucket-access checklist item and the two `dc-knowledge/` steps in Section 1, whose third step now stands alone as "Reuse over rebuild". No `{skill_dir}` placeholders, no `dc-knowledge/` paths.
- `src/datachain/skill/core/SKILL.md`: now a short wrapper. Same frontmatter, tells the agent to read `{skill_dir}/SDK.md` in full, keeps the scope and CAST hand-off paragraphs, and holds the local-only steps under "Before writing any pipeline code".
- `src/datachain/skill/knowledge/SKILL.md`: the Mode B pointer reads `{core_skill_dir}/SDK.md` directly instead of going through the wrapper.
- `src/datachain/cli/commands/skill.py`: two installer gaps that the split made visible. Installing `knowledge` alone left its pointer into `core` dangling, so `knowledge` now declares `core` as a dependency and `datachain skill install --skills knowledge` installs both. And the installer only substituted `{skill_dir}`, so `{knowledge_skill_dir}` in the core skill was always written out literally; it now resolves `{<name>_skill_dir}` for every skill name to the absolute install path, in `SKILL.md` and in the generated command files.

## What did not change

- `SDK.md` needs no installer support beyond the above. The skill directory is copied with `copytree`, so it lands next to `SKILL.md`; the package data pattern `**/*.md` already includes it.
- Every rule keeps its wording and section numbers, so cross-references like "See §2 Rule 2" still resolve.

<sub>🤖 Co-authored by Claude Code and Codex</sub>


src/datachain/cli/commands/skill.py
```diff
@@ -6,6 +6,7 @@
 from typing import TypedDict
 
 SKILLS = ("core", "knowledge", "jobs")
+SKILL_DEPENDENCIES = {"knowledge": ("core",)}
```
ilongin, line 9: Will this break the dependency if a user runs `datachain skill install knowledge` and later `datachain skill uninstall core`? It seems that `knowledge` would remain installed with a broken reference to `core/SDK.md`. Should we refuse to remove `core` while an installed skill depends on it, and add a regression test?

src/datachain/skill/core/SDK.md
```diff
@@ -0,0 +1,799 @@
...
+**Naming convention:**
+
+```
+l1_<source>_<descriptor>      # Container — listings, headers, sidecar metadata
+l2_<source>_<descriptor>      # Asset — extracted/reshaped raw data
+l3_<source>_<descriptor>      # Sense — model-derived signals
+<descriptor>                  # Task — no prefix
+```
+
+The `l1_` / `l2_` / `l3_` prefix is enough; do NOT add layer-type infixes like `_container_`. Cap at 30 chars; the full doctrine is in `knowledge/CAST.md` §3.
```
ilongin: Since `SDK.md` should work standalone, should we remove the `knowledge/CAST.md` reference? That file may not exist when Studio serves only the SDK rules.

src/datachain/cli/commands/skill.py
```diff
@@ -166,28 +172,27 @@ def install_skills(skills: str | None, target: str, local: bool) -> int:
...
                 cmd_dest = commands_dir / f"datachain-{skill_name}{command_ext}"
-                skill_dir_resolved = str(dest.resolve())
                 if command_ext == ".mdc":
                     content = _transform_cursor_mdc(skill_md)
                 elif command_ext == ".instructions.md":
                     content = _transform_copilot_instructions(skill_md)
                 else:
                     content = skill_md.read_text()
-                cmd_dest.write_text(content.replace("{skill_dir}", skill_dir_resolved))
+                cmd_dest.write_text(_resolve_placeholders(content, skills_dir, dest))
```
ilongin, line 198: Will Cursor and Copilot automatically load `SDK.md` when their generated instruction file only contains the `SKILL.md` wrapper? Previously the generated file contained all SDK rules, while it now only points to another file. Should we include the `SDK.md` content in generated instruction files for targets that expect self-contained rules, and test for a known SDK rule?

shcheklein (approved):

src/datachain/cli/commands/skill.py
dreadatour: Good catch, thank you! Fixed.

src/datachain/skill/core/SDK.md
dreadatour: Good catch, thank you! Fixed.

src/datachain/cli/commands/skill.py
dreadatour: Good catch, thank you! Fixed.

ilongin (approved):

PR #1992, opened 2026-09-17
fix(jobs): store params and metrics as JSON values, not as JSON text

shcheklein:
## What this changes

`create_job()` and `update_job()` `json.dumps()` a job's `params` and `metrics` before the
insert, so a JSON column is handed the JSON text of the dict rather than the dict. This
passes the values instead.

## Nothing changes on SQLite

The sqlite3 adapters serialize dicts with this package's own json and store a string as it
is, so both shapes land as the same text. Measured, writing the dict (this branch) against
writing its JSON text (main):

| metrics value | this branch | main |
|---|---|---|
| `{"rows": 10}` | `{"rows":10}` | `{"rows":10}` |
| datetime | `{"at":"2026-01-01T00:00:00"}` | same |
| UUID | `{"id":"00000000-…-0001"}` | same |
| bytes | `TypeError: reject_bytes is on` | same |
| NaN | `{"loss":NaN}` | same |

Byte for byte identical, coercions included. Reads are unaffected — the column still holds
JSON text and `Job.parse()` still decodes it — so an existing database opens as before, and
a database written by this branch opens on an older version.

## Why it is worth changing anyway

A JSON column reached through SQLAlchemy's bind path, rather than through the sqlite3
adapters, encodes that string a second time and stores a quoted string. Anything reading the
same row without going through `Job.parse()` then gets a `str` where a dict is expected.
That is what a PostgreSQL-backed subclass of `AbstractDBMetastore` sees, and it is a real
bug there: in DataChain Studio one such row fails a whole page of the jobs API.

Handling PostgreSQL's read shape stays in that subclass, where it already is — this side has
no reason to know about it.

`test_job_management.py`, `test_metastore.py`, `test_session.py` and
`test_dataset_status_management.py` pass (117).


src/datachain/data_storage/metastore.py
```diff
@@ -2563,8 +2563,8 @@ def update_job(
             values["error_stack"] = error_stack
         if finished_at is not None:
             values["finished_at"] = finished_at
-        if metrics:
-            values["metrics"] = json.dumps(metrics)
+        if metrics is not None:
+            values["metrics"] = metrics
```
shcheklein, line 2567: **[P2] Normalize dict subclasses before handing them to sqlite3**

The sqlite3 adapter is registered for the exact `dict` type; it does not apply to subclasses. Removing `json.dumps()` therefore breaks inputs accepted by the annotated API and by the previous implementation. The same problem affects `params=params or {}` in `create_job()` at line 2503.

I ran both the parent revision's methods and this revision's methods against SQLite:

| Call/input | Parent revision | This revision |
| --- | --- | --- |
| `create_job(..., params=OrderedDict(key="value"))` | round-trips | `ProgrammingError: Error binding parameter 11: type 'collections.OrderedDict' is not supported` |
| `update_job(..., metrics=Counter(rows=3))` | round-trips | `ProgrammingError: Error binding parameter 1: type 'Counter' is not supported` |
| `update_job(..., metrics=defaultdict(int, rows=3))` | round-trips | `ProgrammingError: Error binding parameter 1: type 'collections.defaultdict' is not supported` |

Plain dictionaries continue to work, so the existing 117 passing tests do not catch this. Please normalize provided params and metrics to plain dictionaries before binding (while retaining the `None`/empty-dict distinction for updates), and add parametrized regression tests for creation and updating with dict subclasses. This preserves the native JSON-object fix for PostgreSQL.

shcheklein (commented):
Reviewed together with datachain-ai/studio#13271. The existing four unit-test files named in this PR pass (117 tests). Additional before/after probes found the SQLite regression below; those same probes confirmed datetime/UUID round-trips and the intended clearing of previously nonempty metrics with `{}`.

shcheklein:
Closing: the premise was wrong, and the reviews here are what showed it.

Storing these columns as JSON values cannot work, because a metric can be a NaN or an infinity - `dc.metrics.set()` takes any float - and no JSON column can hold one. Postgres has no syntax for them and rejects the row; jsonb cannot represent them at all. The double encoding this PR removed is what carries them, since ujson writes `NaN` and reads it back as a float.

So this package is already right: params and metrics are JSON text, and the backend that holds those columns is the one that has to declare them as text. That is being fixed in datachain-ai/studio#13271, where the columns become `text` with a migration, instead of here.

Thanks for the dict-subclass catch - `OrderedDict`, `Counter` and `defaultdict` all failing to bind was real, and it disappears with the change reverted.

PR #1993, opened 2026-09-17
Ilongin/13207 trace dataset usage

ilongin:
Part of datachain-ai/studio#13207.

Lets the metastore record when a dataset version was last used.

- Adds `record_dataset_version_access(dataset, version)` on `AbstractMetastore`. Default is a no-op so `SQLiteMetastore` and other implementations don't need changes.
- Calls it from `QueryStep` when a step runs against a dataset version, so any query touching a version marks it as accessed.
- The SaaS PG metastore (in Studio) overrides it to actually write the timestamp.


src/datachain/query/dataset.py
```diff
@@ -310,6 +310,10 @@ def apply(self) -> "StepResult":
         def q(*columns):
             return sqlalchemy.select(*columns)
 
+        self.catalog.metastore.record_dataset_version_access(
```
shcheklein, line 313: are there are other potential places (besides export)? 

place AI found. I wonder if there is a more reliable way to detect this (intercept CH select queries? or something?)

The access hook: [QueryStep.apply()](https://github.com/datachain-ai/datachain/blob/58c81dbda556a9b3624e8c248500598a304d19df/src/datachain/query/dataset.py#L309-L317). The common caller is [DatasetQuery.apply_steps()](https://github.com/datachain-ai/datachain/blob/58c81dbda556a9b3624e8c248500598a304d19df/src/datachain/query/dataset.py#L2787-L2789).

Missed storage-browser reads: [Studio calls read_listing_dataset()](https://github.com/datachain-ai/studio/blob/5ea0ff883fc3522a14644b47759cbf39ff0d442d/backend/dqlapp/views/filebrowser.py#L56-L63), which uses [ReadOnlyQueryStep.apply()](https://github.com/datachain-ai/datachain/blob/58c81dbda556a9b3624e8c248500598a304d19df/src/datachain/lib/dc/listings.py#L23-L51). That override never calls the hook.

Delta reads: [Compare old and new source versions](https://github.com/datachain-ai/datachain/blob/58c81dbda556a9b3624e8c248500598a304d19df/src/datachain/delta.py#L149-L163), then [merge with the previous output](https://github.com/datachain-ai/datachain/blob/58c81dbda556a9b3624e8c248500598a304d19df/src/datachain/delta.py#L458-L469). These execute ordinary dataset queries.

Retries: [Error-row retries and missing-row retries](https://github.com/datachain-ai/datachain/blob/58c81dbda556a9b3624e8c248500598a304d19df/src/datachain/delta.py#L181-L201). Also, [delta_retry explicitly disables save-checkpoint reuse](https://github.com/datachain-ai/datachain/blob/58c81dbda556a9b3624e8c248500598a304d19df/src/datachain/lib/dc/datachain.py#L789-L797), so retries cannot silently return the cached output.

Reuse without a row read: [Completed save-checkpoint reuse](https://github.com/datachain-ai/datachain/blob/58c81dbda556a9b3624e8c248500598a304d19df/src/datachain/lib/dc/datachain.py#L846-L860) and [no-change delta reuse](https://github.com/datachain-ai/datachain/blob/58c81dbda556a9b3624e8c248500598a304d19df/src/datachain/lib/dc/datachain.py#L928-L939). Both return a dataset reference without touching its access timestamp.

Partial UDF continuation: [The _continue_udf() branch](https://github.com/datachain-ai/datachain/blob/58c81dbda556a9b3624e8c248500598a304d19df/src/datachain/query/dataset.py#L1270-L1275). The starting-step hook runs before reaching this branch.

Export bypass: [Catalog.export_dataset_table() directly calls the warehouse](https://github.com/datachain-ai/datachain/blob/58c81dbda556a9b3624e8c248500598a304d19df/src/datachain/catalog/catalog.py#L1779-L1794). In comparison, [Studio’s export task explicitly records access](https://github.com/datachain-ai/studio/blob/5ea0ff883fc3522a14644b47759cbf39ff0d442d/backend/dqlapp/tasks/datasets.py#L121-L125).

Pagination: [extract() applies steps once, then starts pagination](https://github.com/datachain-ai/datachain/blob/58c81dbda556a9b3624e8c248500598a304d19df/src/datachain/query/dataset.py#L2900-L2909). [Subsequent page queries](https://github.com/datachain-ai/datachain/blob/58c81dbda556a9b3624e8c248500598a304d19df/src/datachain/data_storage/warehouse.py#L263-L283) do not refresh access.

Maintenance that refreshes recency: [fix_version_previews executes a normal DatasetQuery](https://github.com/datachain-ai/studio/blob/5ea0ff883fc3522a14644b47759cbf39ff0d442d/backend/dqlapp/management/commands/fix_version_previews.py#L42-L48).

Recording failures are swallowed: [API/network failures](https://github.com/datachain-ai/studio/blob/5ea0ff883fc3522a14644b47759cbf39ff0d442d/datachain_saas/src/datachain_saas/shared/metastore.py#L990-L1003) and [PostgreSQL failures](https://github.com/datachain-ai/studio/blob/5ea0ff883fc3522a14644b47759cbf39ff0d442d/datachain_saas/src/datachain_saas/postgresql.py#L2284-L2291).

PR #1994, opened 2026-09-17
feat(studio): surface and document what a compute cluster is

shcheklein:
Follow-up to Studio [#13276](https://github.com/datachain-ai/studio/pull/13276), deployed. `GET /api/datachain/clusters/` reports each cluster's `cloud_region`, `instance_type`, `compute_class`, `disk_size`, `active_workers` and `busy_workers` alongside its `id`. This puts them to use, and documents what was already there.

## StudioClient

`ClusterListData` was `list[dict[str, Any]]` — the fields existed but nothing said so. It is now `list[ClusterData]`, a `TypedDict` naming each one, required and nullable: Studio reports every field on every cluster, and what varies is whether a cluster configures a value. `get_clusters()` and `get_jobs()` gained docstrings — neither had one — including what `include_steps` fills in.

## `datachain job clusters`

```
+------------+--------------+----------+------------------+-----------+-----------------+-----------------+----------------+-------------------+--------------+
| ID         | Name         | Status   | Cloud Provider   | Region    | Instance Type   | Compute Class   | Disk Request   | Busy/Active/Max   | Is Default   |
+============+==============+==========+==================+===========+=================+=================+================+===================+==============+
| k3f9x2mq7a | prod-cluster | ACTIVE   | AWS              | us-west-2 | m5.xlarge       | gpu             | 100Gi          | 2/4/8             | True         |
+------------+--------------+----------+------------------+-----------+-----------------+-----------------+----------------+-------------------+--------------+
```

Identified by `id`. Unset fields render `-` through tabulate's `missingval`, so a real `0` or `false` survives.

## `datachain job ls --json`

`job clusters` gained `--json` here and `job ls` had none. Both are format switches only — the other flags still decide what is fetched, so `--json --extended --status failed --limit 50` means what it reads like.

`--extended` itself was documented: it adds the cluster and per-stage timings, which had one line between them. Also a bug fix — `--cluster` takes a cluster's **name**, and the example passed an id.

## The jobs skill is removed

Asking what it provided that the CLI and MCP do not came back almost empty. Its `--clusters` was byte-identical to `job clusters --json`; logs are `job logs`; MCP's `get_job` returns more about a single job than the skill ever did. The one real gap was a machine-readable job *list* — MCP cannot list jobs at all, and `job ls` had no `--json` — which the line above closes. What remained was a local staleness cache and a set of interpretation rules, and those rules now live in the command docs where every reader gets them.

`datachain skill install` offers `core` and `knowledge`, otherwise untouched. `RETIRED_SKILLS` keeps `jobs` removable: install and uninstall both clear a copy left by an earlier release, because a stale skill directory keeps instructing whichever agent reads it.

## Fields dropped

**`job_quota`** was the namespace quota on Kubernetes Jobs, and a worker is one of those — so it capped workers, the same ceiling the cluster reports live as `max_workers`, which is the one Studio actually enforces against. One number in two columns, with the confusing name on the copy nothing checks. Gone from both sides ([studio#13281](https://github.com/datachain-ai/studio/pull/13281)).

**A cluster has one `id`.** Studio collapsed its numeric id and its uuid onto a single short generated string in [studio#13287](https://github.com/datachain-ai/studio/pull/13287), so `ClusterData`, the table and the docs all use `id`, and a job joins to a cluster on `compute_cluster_id`. Nothing is deprecated and nothing is dual-written.

## What review changed

Several of these were wrong in ways that looked right:

- **`job_quota` was not jobs per worker.** The first draft had the skill treating per-job cost as an upper bound because of the name.
- **`compute_class` is not spot vs on-demand.** It is a node class such as `Performance` or `gpu`. Keying a rate on it picks the wrong price while appearing to work.
- **`disk_size` cannot price storage.** It is requested temporary storage, not allocated capacity.
- **Stage timings were capped at 200 jobs**, and past that the output still claimed `enriched: true`.
- **There is no per-job endpoint.** `get_jobs(job_id=...)` is the list endpoint filtered by id, so the skill's enrichment loop re-fetched what it already had.
- **A configured quota of `0` rendered as unknown**, because `or "-"` swallowed it.
- **Jobs were matched to clusters by name**, which mis-attributes a job when a retired cluster's name is reused.

## Test plan

- `tests/test_cli_studio.py` — the `job clusters` columns, unset fields, a false flag against `missingval`, `--json` on both commands, and that `--json` leaves `--extended` in charge.
- `tests/unit/test_cli_skill.py` — retired-skill cleanup on install, on uninstall-all, and by name, each seeded from a previous installation; all three fail without the fix.
- `pytest tests/test_cli_studio.py tests/unit/test_cli_skill.py` — 106 passed. ruff clean, `mkdocs build --strict` clean, mypy clean on 406 files with `--extra dev`.


src/datachain/skill/jobs/scripts/jobs.py
```diff
@@ -151,7 +192,7 @@ def _enrich_job(client, job: dict) -> dict:
     if not job_id:
         return job
     try:
-        response = client.get_jobs(job_id=job_id)
+        response = client.get_jobs(job_id=job_id, include_steps=True)
```
shcheklein: **[P2] Fetch stage timings for all returned jobs**

The new timing fields depend on this per-job request, but `cmd_fetch` calls `_enrich_job` for at most `ENRICH_LIMIT` (200) terminal jobs. I reproduced `--fetch --enrich` with 201 completed jobs and the default limit of 500: all 201 jobs are returned, only 200 have queue/run timings, and the JSON reports `enriched: true`, `truncated: false`. There is a stderr warning, but the prescribed index format does not retain the partial-enrichment state. Subsequent Queue/Run totals therefore omit jobs whose timings the server does have.

The list endpoint already accepts `include_steps`, so requesting `get_jobs(limit=limit, include_steps=enrich)` would retrieve stages for the full result in one call and avoid the cap and per-job requests. Otherwise, preserve explicit enrichment coverage in the output/index and qualify aggregates. Please cover more than 200 completed jobs in a regression test.

src/datachain/skill/jobs/SKILL.md
```diff
@@ -89,16 +103,25 @@ Duration cells contain plain seconds strings like `"9000s"`. Parse the integer b
...
-3. Compute per job: `duration_seconds / 3600 × rate × workers`. Group by user/day/cluster as requested.
-4. Present as a table: User | Compute-hours | Est. cost (@$X/hr × N workers)
+1. Establish the hourly rate per cluster from its Clusters row — instance type and region are what a cloud price list is keyed on. State the rate you used and where it came from. If the row's instance type is `—`, or you have no price for it, ask: "What is the hourly rate in $/hr for <cluster> (<instance type> in <region>)?"
+2. Nothing in the index says whether a cluster runs on spot or on-demand capacity — Compute Class is a node class (e.g. `Performance`, `gpu`), not a purchase model. Spot can cost a fraction of on-demand, so when the answer turns on it, ask rather than assume: "Is <cluster> running spot or on-demand capacity?"
+3. If Workers column is all `—` → ask: "How many workers per job?" or compute single-worker cost and note it.
+4. Compute per job: `duration_seconds / 3600 × rate × workers`. Group by user/day/cluster as requested.
+5. Disk is the worker's disk request, not a separate billed volume — fold it in only if the user asks for storage cost, and say the rate is per-GB-month.
```
shcheklein: **[P2] Do not derive storage cost from `disk_size`**

This correctly says Disk is a request, but then permits folding it into a per-GB-month storage estimate. Studio reads `resources.requests.ephemeral-storage`; actual volume sizes come from separate `ephemeralStorage.size` and `localStorage.size` settings. A `1Gi` request can coexist with `20Gi` or `500Gi` of allocated volume capacity without `disk_size` changing, so this value cannot establish the billable storage quantity.

When storage cost is requested, require the actual volume capacity/type and applicable rate, or state that the cluster response lacks enough information. Please also clarify the CLI's Disk description. The corresponding API-description issue is [commented on Studio #13281](https://github.com/datachain-ai/studio/pull/13281#discussion_r4042975049).

src/datachain/skill/jobs/SKILL.md
```diff
@@ -89,16 +104,25 @@ Duration cells contain plain seconds strings like `"9000s"`. Parse the integer b
 - Example: filter rows for user "alice" in the last 7 days, sum all Duration values → total seconds → divide by 3600 for hours.
 - If all Duration cells are `—` (enriched: false) → say: "Duration data requires enrichment. Re-fetch with: `python3 {skill_dir}/scripts/jobs.py --fetch --enrich`" and offer to do so.
 
+### Stage breakdown — where the time went
+The Queue and Run columns split a job's wall clock: `waiting` before it started, `running_query` doing the work. The remainder is setup.
+- "How long do jobs wait?" → sum or average the Queue column. A queue that rivals Run means the cluster is at its worker cap, not that jobs are slow.
```
shcheklein: **[P2] Do not diagnose worker saturation from queue duration alone**

Queue time comparable to runtime establishes that waiting accounts for much of the elapsed time; it does not establish that the cluster reached its worker limit. Scheduling interruptions can also extend the wait, and this index contains no historical capacity measurement establishing the cause. The instruction would make the agent give an unsupported explanation and potentially recommend unnecessary capacity.

Suggested wording: ‘Compare queue time with runtime to show where time was spent. Check cluster capacity or diagnostics before explaining why the job waited.’

shcheklein (commented):
Documentation feedback on the changes in this PR:

- **Make the new cluster output section easier to read.** The first example is 199 characters wide and is followed by column definitions, a full JSON payload, and another JSON example. Start with the list-and-select workflow (`datachain job clusters`, then `datachain job run analysis.py --cluster prod-cluster`), use a clearly labeled compact output excerpt, and keep one full JSON example for scripting.
- **Simplify the new field definitions.** Suggested wording: **Instance Type:** worker machine type or family; **Compute Class:** worker class, such as Performance or gpu; **Disk Request:** requested temporary storage per worker (allocated capacity may differ); **Job Quota:** configured worker limit. Keep the pricing limitations together in one short note. This preserves the useful distinctions without repeating the earlier review discussion.
- **Correct the new extended-output example and stage explanation.** The table has six header cells but five border sections. Capture actual output or label a simplified excerpt. Remove the claimed fixed stage order: files are downloaded before dependency installation, as the example itself shows. Say that only available timings are shown and missing timings are unknown.
- **Keep the newly added stage-answer instructions focused on the user's result.** Queue duration shows time spent waiting; it does not by itself establish worker saturation. Missing timings also do not establish that enrichment was skipped. Fetch timings when needed, report unavailable data and aggregation coverage plainly, and keep helper-script commands in the agent's execution instructions. The two inline comments give the specific corrections.

src/datachain/skill/jobs/SKILL.md
```diff
@@ -89,16 +104,25 @@ Duration cells contain plain seconds strings like `"9000s"`. Parse the integer b
...
+- "Why is this job slow?" → compare Queue, Run, and `Duration − Queue − Run` (setup: dependency installs, file downloads, warehouse wake-up).
+- All `—` → the index is not enriched. Say so and offer: `python3 {skill_dir}/scripts/jobs.py --fetch --enrich`.
```
shcheklein: **[P2] Check timing availability instead of equating missing values with missing enrichment**

`--enrich` can still return no completed stage timings: `_stage_seconds` omits stages without both timestamps, and some jobs have no recorded steps. This branch would incorrectly tell users their index is not enriched and offer a re-fetch that cannot recover unrecorded data. It also makes users manage an internal script to get an ordinary analytics answer.

Check `enriched` explicitly. If timings were not requested, fetch them as part of answering; if they were requested but remain unavailable, say ‘Stage timings are unavailable for these jobs’ and state coverage when aggregating. Keep the script command in the agent's execution instructions.

src/datachain/skill/jobs/scripts/jobs.py
shcheklein: Fixed in 30458d2c — took the list-endpoint route you suggested. `cmd_fetch` now calls `get_jobs(limit=limit, include_steps=enrich)`, and `_enrich_job` no longer asks for or merges `steps` at all, so stages come for every returned job in one call with no per-job request and no ENRICH_LIMIT cap.

Regression test added with `ENRICH_LIMIT + 1` completed jobs (`TestFetchStageTimings::test_every_job_is_timed_past_the_enrich_limit`); it asserts all 201 carry queue/run timings and that the list call asked for steps. I checked it fails against the old shape before keeping it. A second test covers the non-enriched path.

src/datachain/skill/jobs/SKILL.md
shcheklein: Fixed in 30458d2c. The skill no longer permits a storage estimate from it: step 5 now says the cluster data does not carry the billable capacity and to ask for the volume size, type and rate instead.

Renamed to "Disk Request" in the CLI table and the index Clusters table, described as temporary storage a worker asks for and explicitly not the size of its disk. Studio's API description fixed on #13281 (e5f386821) for your linked comment.

src/datachain/skill/jobs/SKILL.md
shcheklein: Agreed — fixed in 7ba75e95, close to your wording. The rule now reads that queue time shows *where* the time went, not why; that a long queue does not by itself establish the cluster hit its worker limit, since the index holds no capacity history and scheduling can delay a job that had workers to spare; and to check the cluster's capacity before explaining the wait. Added an explicit "never recommend more workers on queue time alone", since that was the costly end of the wrong conclusion.

src/datachain/skill/jobs/SKILL.md
shcheklein: Fixed in 7ba75e95. The rule now checks the `enriched` flag rather than the cells: if `false`, the agent re-runs Step 2 with `--enrich` itself as part of answering; if `true` and cells are still `—`, it says "Stage timings are unavailable for these jobs" and states coverage (how many of how many) when aggregating. The script command stays in the agent's execution instructions only.

The Duration rule one section up had the same fault — missing cells read as "not enriched", plus a script for the user to run — so I fixed it the same way rather than leave two contradictory rules.

shcheklein:
All four documentation points addressed in `14b1d259`.

**1. Cluster output section easier to read.** The page now opens on the workflow you described:

```bash
datachain job clusters
datachain job run analysis.py --cluster prod-cluster
```

The 199-character table is replaced by a labelled excerpt (UUID, Name, Status, Region, Instance Type, Busy/Active/Max) that names the columns it leaves out. One JSON example remains, under "All fields", for scripting — the duplicate `--json` example is gone.

**2. Field definitions simplified** to your wording: Instance Type "worker machine type or family"; Compute Class "worker class, such as Performance or gpu"; Disk Request "requested temporary storage per worker (allocated capacity may differ)"; Job Quota "configured worker limit". The pricing limitations are in one note at the end instead of trailing each definition, and I trimmed the same repetition out of the skill.

**3. Extended-output example and stage explanation.** You were right that it was malformed — six header cells over five border sections, because I hand-drew it. Both that table and the clusters one are now captured from the command itself, so they cannot be miscounted or drift. Dropped the fixed-order claim, which the example contradicted (files download before dependencies install), and it now says only stages with a recorded time are listed, anything missing is unknown rather than zero, and a stage still going reads `running`.

**4. Stage-answer instructions** — fixed in `7ba75e95`, replied on both inline threads.

src/datachain/skill/jobs/scripts/jobs.py
```diff
@@ -178,30 +222,24 @@ def cmd_fetch(days: int, limit: int, enrich: bool):  # noqa: C901
...
                 if c.get("name"):
-                    clusters_by_id[c["name"]] = c["name"]
+                    clusters_by_key[c["name"]] = c["name"]
     except Exception:  # noqa: BLE001, S110
         pass
 
-    response = client.get_jobs(limit=limit)
+    # include_steps fills every returned job's stages in this one call. Asking per
+    # job instead would cap them at ENRICH_LIMIT and leave later jobs untimed.
+    response = client.get_jobs(limit=limit, include_steps=enrich)
```
ilongin: Now that the list request uses `include_steps=True` and appears to return these job fields, do we still need a separate API request for every terminal job here? If the detail endpoint does provide something unavailable in the list response, could we document which field requires it? Otherwise, could we remove the per-job enrichment loop?

src/datachain/remote/studio.py
```diff
@@ -25,7 +26,34 @@
...
+
+    id: int
+    uuid: str
+    name: str
+    status: str
+    cloud_provider: str
+    cloud_credentials: str | None
+    is_active: bool
+    default: bool
+    max_workers: int
```
ilongin, line 45: Can these worker counts be `null`? `_cluster_workers()` handles `None`, but they are typed as `int` here. If Studio can omit them, should these be `int | None`?

src/datachain/skill/jobs/scripts/jobs.py
```diff
@@ -178,30 +222,24 @@ def cmd_fetch(days: int, limit: int, enrich: bool):  # noqa: C901
     now = datetime.now(tz=timezone.utc)
     cutoff = now - timedelta(days=days)
 
-    # Fetch clusters for name reference
-    clusters_by_id = {}
+    # Fetch clusters for name reference, keyed by uuid and by name.
+    clusters_by_key: dict[Any, Any] = {}
```
ilongin: Could this use a precise key/value type instead of `dict[Any, Any]`? It looks like the current values are cluster names and the keys are UUIDs or names.

docs/commands/job/clusters.md
```diff
@@ -5,20 +5,96 @@ List compute clusters in Studio.
...
+usage: datachain job clusters [-h] [-v] [-q] [--team TEAM] [--json]
 ```
 
 ## Description
 
-This command lists compute clusters available in Studio. You can specify a team to list clusters for. The command provides information about the compute resources available for running jobs.
+This command lists the compute clusters your team can run jobs on, with the machine
+each one provisions and how busy it is. Use it to pick a cluster for
+[`datachain job run --cluster`](run.md), to check spare capacity before submitting,
+or to work out what a job costs.
```
ilongin: Do we even provide enough info for user to calculate cost?

src/datachain/studio.py
```diff
@@ -814,27 +816,42 @@ def show_job_logs(job_id: str, team_name: str | None):
...
             "Status": cluster.get("status"),
             "Cloud Provider": cluster.get("cloud_provider"),
-            "Cloud Credentials": cluster.get("cloud_credentials"),
-            "Is Active": cluster.get("is_active"),
+            "Region": cluster.get("cloud_region") or "-",
+            "Instance Type": cluster.get("instance_type") or "-",
+            "Compute Class": cluster.get("compute_class") or "-",
+            "Disk Request": cluster.get("disk_size") or "-",
+            "Busy/Active/Max": _cluster_workers(cluster),
+            "Job Quota": cluster.get("job_quota") or "-",
```
shcheklein: **[P2] Reuse `tabulate`'s missing-value handling so zero stays zero**

`cluster.get("job_quota") or "-"` renders a configured quota of `0` as unknown. I reproduced this with a cluster whose counts and quota are all zero: Busy/Active/Max is `0/0/0`, but Job Quota is `-`.

Keep nullable values as `None` in the row and pass `missingval="-"` to the existing `tabulate.tabulate(...)` call. That removes the repeated `or "-"` expressions and preserves `0` and `False`. Please cover `job_quota=0` alongside `None`; the current unset-fields test does not actually assert the rendered quota cell.

tests/unit/test_skill_jobs_scripts.py
```diff
@@ -109,3 +112,155 @@ def test_matches_graph_utils(self, tmp_path):
...
+        """
+        import jobs as jobs_module
+
+        total = 250
+        listed = [
+            {
+                "id": f"job-{i}",
+                "name": f"job-{i}",
+                "status": "COMPLETE",
+                "created_at": "2026-09-16T00:00:00Z",
```
shcheklein: **[P2] Make the new fetch tests independent of today's date**

Both fetch tests use jobs created on 2026-09-16 but call `cmd_fetch(days=30)` against the real clock. Once those fixtures age out, the first test gets zero jobs and the second indexes an empty list. I ran these two tests with `jobs.datetime.now()` fixed at 2026-10-17 and reproduced both failures (`0 == 250` and `IndexError`).

Freeze the fetcher's clock at a known point near these fixtures, or derive their timestamps from a controlled current time. Keep the date window meaningful rather than enlarging it to postpone the failure.

src/datachain/skill/jobs/scripts/jobs.py
```diff
@@ -129,79 +159,58 @@ def cmd_clusters():
...
-                }
-                clusters_list.append(entry)
-                if c.get("id"):
-                    clusters_by_id[c["id"]] = c.get("name", c["id"])
+                clusters_list.append(dict(c))
+                if c.get("uuid"):
+                    clusters_by_key[c["uuid"]] = c.get("name") or c["uuid"]
                 if c.get("name"):
-                    clusters_by_id[c["name"]] = c["name"]
+                    clusters_by_key[c["name"]] = c["name"]
```
shcheklein: **Cleanup: reuse the job response's cluster name**

Studio's `JobOutput` already requires `compute_cluster_name`, and `_get_job_output` supplies it through `cluster_name_for(job)`, including the retired-cluster lookup and UUID fallback. The CLI already reads this field directly.

Can this path also use `j.get("compute_cluster_name")`? That lets us remove `clusters_by_key`, the UUID/name indexing here, and the `cluster_key`/`looked_up`/`isinstance` fallback block below. Keep fetching clusters for the Clusters output, but let Studio resolve the job's recorded cluster identity once.

tests/unit/test_skill_jobs_scripts.py
```diff
@@ -109,3 +112,155 @@ def test_matches_graph_utils(self, tmp_path):
...
+            }
+            for i in range(total)
+        ]
+
+        class Response:
+            def __init__(self, data):
+                self.ok, self.data, self.message = True, data, ""
+
+        class FakeClient:
+            list_calls: ClassVar[list[bool]] = []
```
shcheklein: **Cleanup: reuse `Response` and the existing mock fixture**

The two new tests each define another `Response` and `FakeClient`, including custom call tracking. We already have `datachain.remote.studio.Response` and pytest's `mocker` fixture. A patched `StudioClient` returning real `Response(data, ok=True, message="", status=200)` objects would remove both fake implementations; `get_jobs.assert_called_once_with(limit=500, include_steps=True)` verifies the one-call behavior directly.

Keep the 250-job regression case and the unenriched case, with shared setup for the client and controlled clock.

src/datachain/remote/studio.py
```diff
@@ -521,6 +549,21 @@ def get_jobs(
...
+        `error_message`, `metrics`, and the cluster it ran on as
+        `compute_cluster_id`, `compute_cluster_uuid` (joins to a cluster's `uuid`)
+        and `compute_cluster_name`.
+
+        `include_steps` fills `steps` with the job's stages - `waiting`,
+        `requesting_workers`, `preparation`, `virtualenv`, `downloading_files`,
+        `dw_wake_up`, `running_query` - each with a `status`, `started_at` and
+        `finished_at`, which is how long a job spent queued as against running.
+        Without it `steps` is null; with it, a stage missing from the list means the
+        job never reached it, and one with no `finished_at` was still in it.
```
shcheklein: **Documentation: make the new timing descriptions agree**

This still says a missing stage was never reached, and a missing end timestamp means the job was still in it. Missing historical data and jobs that stop without closing a stage make those conclusions unsafe. The new skill guidance now handles those cases correctly, but its `stage_note` also retains the ‘never reached’ wording.

The new CLI paragraph has the opposite mismatch: ‘Only the stages with a recorded time are listed’ is not what `_format_stages` does; it also lists stages as `-` or `running`, as the existing CLI test asserts.

Keep the API description short: ‘With `include_steps`, returns recorded stages and their timestamps. Missing timestamps mean the duration is unavailable.’ In the CLI page, explain `running` and `-`; in the skill note, use ‘timing unavailable’ without inferring which stage a job reached.

docs/commands/job/clusters.md
```diff
@@ -31,9 +107,19 @@ datachain job clusters
...
+```
 
 ## Notes
 
-* The command shows all compute clusters available to your team
-* Clusters represent the compute resources where your jobs can run
-* Use the `--team` option to view clusters for a different team
+* **Working out a cost.** This tells you *what machine* a job ran on - its instance
+  type, region and provider, and how many workers - which is what you need to look a
+  rate up. It does not give you the rate, and two things it does not report change it:
```
shcheklein: **Documentation: keep the cost note to what this command establishes**

The command reports current cluster worker counts, so ‘what machine a job ran on … and how many workers’ reads as per-job usage that this output does not supply. `job ls --extended` supplies stages and the cluster name, but no per-job worker count or billable duration. Also, `instance_type` can be a machine family, as the field definition above correctly says.

A shorter and more accurate note would be: ‘Provider, region, and machine information can help identify a rate. Estimating a job's cost also requires its resource usage and the applicable compute and storage rates.’ This avoids suggesting that these two tables alone supply all the inputs.

shcheklein (commented):
Cleanup review of the changed files at 5a194cc. The per-job request loop is gone and the public field definitions are clearer. I ran the two focused test files: **92 passed**. Additional probes reproduced a zero quota rendered as unknown and both new fetch tests failing with the clock set to 2026-10-17; details and reuse opportunities are inline.

A few smaller cuts within this diff:

- Once the name lookup is removed, pass the cluster response through directly (`clusters_list = cr.data or []`). These paths only serialize it, so `dict(c)` and `list(response.data or [])` copies add no useful behavior.
- The new field inventories overlap across `ClusterData`, the script docstring, and the skill's ‘What Studio reports’ section. Keep `ClusterData` as the cluster shape and document the script's derived job fields once; the skill needs the available capabilities and interpretation rules.
- Shorten the new queue/storage guidance to the operational rules. The scheduling explanation and the `1Gi`/`20Gi`/`500Gi` example repeat review rationale. For storage: ‘Storage cost requires allocated capacity, storage type, and rate; Disk Request is insufficient.’
- Remove comments explaining the type annotation and the obvious JSON print. `_cluster_workers` can also be inlined now that it is just one formatting expression. Give `_stage_seconds` the precise return type `dict[str, int]`.
- Update the new test naming/docstring: `test_studio_clusters_shows_what_a_worker_costs` still says ‘how many jobs share it,’ although this change establishes that the quota is a worker limit.
- Refresh the PR title/body around the final behavior. It still describes `NotRequired`, `?/?/8`, and `_cluster_entry` tests that are no longer present.

The companion Studio diff is small and its revised field descriptions look clear; I did not find a further substantive simplification there.

src/datachain/skill/jobs/scripts/jobs.py
shcheklein: Removed in 5a194cc0 — and you were more right than the question assumed: there is no separate detail endpoint. `get_jobs(job_id=...)` is the same list endpoint filtered by id, and both go through one `_get_job_output`, so the loop re-fetched fields the list had already returned.

I verified `workers`, `duration`, `python_version` and `cluster` all populate with `--enrich` off, and that both paths now make exactly one request. Gone with the loop: `ENRICH_LIMIT`, its 200-job cap, the stderr warning, `TERMINAL_STATUSES` and the N+1. `--enrich` now just means "ask for stage timings" and costs one request either way — the skill's instructions said one call per terminal job, which is no longer true, so that's corrected too.

src/datachain/remote/studio.py
shcheklein: Good catch — the types were right and the code was wrong. Studio declares all three counts non-nullable and always sends them, so the `None` branch was unreachable; it was a leftover from an earlier draft that tried to tolerate a Studio too old to report them. Removed in 5a194cc0, and the formatting is now inline in the row (851d7854).

src/datachain/skill/jobs/scripts/jobs.py
shcheklein: Fixed as `dict[str, str]` in 5a194cc0 — then the map went away entirely in 851d7854. shcheklein pointed out Studio already resolves the job's cluster name (`cluster_name_for`, retired clusters and uuid fallback included), so the script reads `compute_cluster_name` off the response and the uuid/name index, the `isinstance` guard and the fallback chain are all gone.

docs/commands/job/clusters.md
shcheklein: Fair — we don't, and the page said we did. It no longer claims you can work out a job's cost from it. Per shcheklein's wording the note is now: "Provider, region and machine information can help identify a rate. Estimating a job's cost also requires its resource usage and the applicable compute and storage rates, none of which this command reports." (851d7854)

The gap is real and worth naming: Studio does not report whether a cluster is billed spot or on-demand, and `disk_size` is a request rather than allocated capacity. Ivan is aware; exposing the purchase model is the one field that would make cost attribution work end to end.

src/datachain/studio.py
shcheklein: Fixed in 851d7854, the way you suggested. The rows keep `None` and `tabulate(..., missingval="-")` draws the dash, so `0` and `False` survive and the repeated `or "-"` is gone.

Reproduced your case first: quota `0` rendered as `-` next to `0/0/0` counts. New `test_studio_clusters_zero_is_not_unknown` asserts the quota cell is `0`, and the unset-fields test now asserts that cell is `-` — you were right that it never checked it. Both read the cell by locating the column in the header rather than substring-matching.

tests/unit/test_skill_jobs_scripts.py
shcheklein: Fixed in 851d7854. The jobs now take their timestamps from the fetcher's own clock (`jobs.datetime.now()`), so fixtures and the date window move together no matter when the suite runs — the window stays 30 days.

I reproduced your probe before and after: with the clock at 2026-10-17 the old tests failed exactly as you describe, and the new ones pass with it set as far out as 2027-06.

src/datachain/skill/jobs/scripts/jobs.py
shcheklein: Done in 851d7854. Checked `cluster_name_for` first — it queries `all_objects` so retired clusters resolve, and falls back to the uuid string — so the script's own index could not do better. It reads `j.get("compute_cluster_name")` now, and `clusters_by_key`, the uuid/name indexing and the `cluster_key`/`looked_up`/`isinstance` block are gone. Clusters are still fetched for the Clusters output, passed straight through as `cr.data or []` with no copy.

tests/unit/test_skill_jobs_scripts.py
shcheklein: Done in 851d7854. Both fakes are gone: a `studio_jobs` fixture patches `StudioClient` through `mocker` and serves real `Response(data, ok=True, message="", status=200)` objects, and the one-call behaviour is asserted with `get_jobs.assert_called_once_with(limit=500, include_steps=True)` as you suggested. An `a_job()` helper builds the fixtures so both tests share the setup, and both cases are kept — the 250-job regression and the unenriched one.

src/datachain/remote/studio.py
shcheklein: All three fixed in 851d7854, and you were right that the CLI paragraph was simply wrong about its own behaviour.

- API docstring, your wording: "With `include_steps`, `steps` holds the job's recorded stages and their timestamps... Missing timestamps mean the duration is unavailable." No inference about which stage a job reached.
- CLI page: "A stage still going reads `running`, and one whose timing is unavailable reads `-` - never `0s`", which is what `_format_stages` does and what the existing test asserts.
- `stage_note`: "or when that timing is unavailable", replacing "never reached that stage". The skill's stage list says a job carries the ones Studio recorded a timing for, rather than inferring how far it got.

docs/commands/job/clusters.md
shcheklein: Applied your wording verbatim in 851d7854: "Provider, region and machine information can help identify a rate. Estimating a job's cost also requires its resource usage and the applicable compute and storage rates, none of which this command reports."

You were right that the old note implied these two tables supply everything — `job ls --extended` gives the cluster and stages but no per-job worker count or billable duration, and my "how many workers" phrasing was reading the cluster's current counts as per-job usage. The field definition's "machine type or family" stands; the note no longer restates it.

src/datachain/skill/jobs/SKILL.md
```diff
@@ -87,18 +102,27 @@ Read `dc-knowledge/jobs/index.md` and answer the user's question.
...
 - Overall: `failed_count / total_jobs * 100` from frontmatter.
 - Per user or per day: count rows matching Status = `failed` in the table.
 
 ### Price estimation
 When the user asks for cost:
-1. If hourly rate unknown → ask: "What is the instance hourly rate in $/hr? (e.g. `3.20` for $3.20/hr)"
-2. If Workers column is all `—` → ask: "How many workers per job?" or compute single-worker cost and note it.
-3. Compute per job: `duration_seconds / 3600 × rate × workers`. Group by user/day/cluster as requested.
-4. Present as a table: User | Compute-hours | Est. cost (@$X/hr × N workers)
+1. Establish the hourly rate per cluster from its Clusters row — instance type and region are what a cloud price list is keyed on. State the rate you used and where it came from. If the row's instance type is `—`, or you have no price for it, ask: "What is the hourly rate in $/hr for <cluster> (<instance type> in <region>)?"
```
shcheklein: **[P2] Reuse the fetched UUIDs when matching jobs to cluster metadata**

The new instructions introduce UUID-based identity, but the prescribed saved index omits both cluster `uuid` and job `cluster_uuid`. Step 3 reads that index, and this pricing rule then looks up the Clusters row by the remaining name. Studio allows a retired cluster's name to be reused, while the clusters endpoint excludes retired clusters. An older job on cluster A named `prod` can therefore pick up replacement cluster B's machine and region.

I checked that the fetch output preserves this case correctly: the job and current cluster have the same name but different UUIDs. Reuse those existing IDs for the metadata join. For an answer based on a cached index, either retrieve the JSON before doing that lookup or retain the IDs as machine-readable metadata. If the job's UUID is absent from the cluster list, report the metadata as unavailable. Keep the API-resolved name for display.

src/datachain/skill/jobs/SKILL.md
```diff
@@ -87,18 +102,27 @@ Read `dc-knowledge/jobs/index.md` and answer the user's question.
...
+
+### Stage breakdown — where the time went
+The Queue and Run columns split a job's wall clock: waiting before it started, running the query. The remainder is setup.
+- "How long do jobs wait?" → sum or average the Queue column. It shows where the time went, not why. Check cluster capacity or diagnostics before explaining a wait, and never recommend more workers on queue time alone.
+- "Why is this job slow?" → compare Queue, Run, and `Duration − Queue − Run` (setup: dependency installs, file downloads, warehouse wake-up).
+- Missing timings: check `enriched` in the frontmatter, not the cells. If it is `false`, re-run Step 2 with `--enrich` yourself as part of answering. If it is `true` and the cells are still `—`, the timings were never recorded for those jobs — say "Stage timings are unavailable for these jobs", and state the coverage (how many of how many) whenever you aggregate.
+- For a finer split than Queue/Run, use the `stages` map from Step 2's output.
```
shcheklein: **Reuse the fetch path when a cached index lacks the requested stage detail**

A fresh index makes Step 1 skip Step 2, but the saved format retains only Queue and Run. In a later session, a question about download or dependency-install time therefore has no `stages` map to read from ‘Step 2's output.’ The `enriched: true` check above does not solve this: it records that stages were fetched, not that all of them were saved.

Please make this new instruction explicit: when the question needs a stage absent from the cached table, run the existing fetch with `--enrich` and answer from its `stages` map. That reuses the current fetcher and keeps the index compact.

tests/unit/test_skill_jobs_scripts.py
```diff
@@ -109,3 +114,134 @@ def test_matches_graph_utils(self, tmp_path):
...
+        client.get_jobs.assert_called_once_with(limit=500, include_steps=True)
+        assert len(out["jobs"]) == total
+        assert all(j["queue_seconds"] == 30 for j in out["jobs"])
+        assert all(j["run_seconds"] == 1080 for j in out["jobs"])
+
+    def test_stages_are_left_out_without_enrich(self, capsys, studio_jobs):
+        import jobs as jobs_module
+
+        yesterday = jobs_module.datetime.now(tz=timezone.utc) - timedelta(days=1)
+        client = studio_jobs([a_job(1, yesterday, stages=False)])
```
shcheklein: **Reuse this test to cover the remaining fetch contract**

The updated unenriched test verifies duration, but `a_job()` still omits `workers`, `compute_cluster_name`, and `compute_cluster_uuid`, and the client fixture always returns an empty cluster list. As a result, these tests would pass if the newly simplified name lookup, the new UUID field, or the cluster passthrough were lost.

Extend this existing case with a non-default worker count, a named/identified cluster, and a cluster payload. Assert the worker count, name/UUID, and unchanged cluster data alongside duration. Reusing the existing cluster test payload as a fixture would cover the PR's data contract without another fake client or another parallel set of tests.

shcheklein (commented):
Follow-up cleanup review at 851d785: the previous runtime/test findings are addressed. **93 focused tests pass**, and the two fetch tests also pass with the clock moved to June 2027.

The remaining useful changes concern reusing the information already fetched: stable cluster IDs for metadata matching, full stage details when the cached table is insufficient, and extending the existing fixtures to exercise the new fields. Those are inline.

Two minor simplifications in the changed text/tests:

- The new `a_job` helper can use timezone-aware `.isoformat()` directly. `_parse_dt` already accepts `+00:00`, and existing parser tests cover both that form and `Z`; the four `.replace("+00:00", "Z")` calls add no coverage.
- Make the new index placeholder for Job Quota explicitly use a dash **only for null**, matching the corrected CLI. `<job_quota or —>` still reads as a truthiness fallback and conflicts with the null-only section rule.

The main Python implementation is now small enough that further general-purpose helper extraction would add little value.

src/datachain/skill/jobs/SKILL.md
shcheklein: Fixed in 64187d38 — a real mispricing, not just a looseness. The index now carries both: a `UUID` column on the Clusters table and `Cluster UUID` on the Jobs table, and the pricing rule matches on those, never on the name, with the reason stated inline so it is not "tidied" away later.

Where a job's Cluster UUID matches no Clusters row, the rule says the cluster is retired or gone and its details are unavailable — it does not fall back to the name. The API-resolved name is still what gets displayed, as you suggested.

The index-wide cell rule is now "write `—` when, and only when, the value is null", so this stays legible next to the null-only placeholders.

src/datachain/skill/jobs/SKILL.md
shcheklein: Right — a fresh index skips Step 2, and the saved table only ever held Queue and Run, so "use the stages map from Step 2's output" pointed at something that may not exist in that session. And as you say, `enriched: true` records that stages were fetched, not that they were all saved.

The rule is now explicit (64187d38): for a stage the table does not hold, re-run Step 2 with `--enrich` and answer from the `stages` map in its output. That keeps the index compact and reuses the existing fetcher.

tests/unit/test_skill_jobs_scripts.py
shcheklein: Fixed in 64187d38, and you were right that the tests were hollow there. `a_job()` now carries `workers`, `python_version`, `compute_cluster_name` and `compute_cluster_uuid`, and the fixture serves a cluster payload. The unenriched case asserts the worker count, python version, resolved cluster name, cluster uuid, and that `out["clusters"]` is the payload unchanged.

I checked they bite: reverting the name lookup, nulling the uuid, or dropping the cluster passthrough each fails this one test. No new fake client — it reuses the `studio_jobs` fixture, with the cluster payload as a module constant.

Also took the two from your summary: `a_job` uses timezone-aware `.isoformat()` directly (`_parse_dt` accepts `+00:00`, and the parser tests already cover both forms), and the Job Quota placeholder is null-only now.

src/datachain/cli/commands/skill.py
```diff
@@ -5,7 +5,7 @@
 from pathlib import Path
 from typing import TypedDict
 
-SKILLS = ("core", "knowledge", "jobs")
+SKILLS = ("core", "knowledge")
```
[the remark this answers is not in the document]
shcheklein: Confirmed again against 29d45a2 with an isolated home directory. I seeded `.cursor/skills/jobs/SKILL.md` and the generated `.cursor/rules/datachain-jobs.mdc`, then ran install-all followed by uninstall-all: both returned success and both old files remained. Explicit removal (`datachain skill uninstall jobs --target cursor`) is rejected as an unknown skill.

Reusing the existing target layouts for legacy cleanup would cover both the skill directory and its generated rule/command. An upgrade regression test should seed a previous installation before invoking the current installer/uninstaller; the fresh-install tests cannot catch this.

docs/commands/job/ls.md
```diff
@@ -5,24 +5,61 @@ List jobs in Studio.
...
++--------------------------------------+--------+----------+----------------------+--------------+--------------+---------------------------------+
+```
+
+A job passes through some of: waiting in queue, requesting workers, preparation, installing dependencies, downloading files, waking up the data warehouse, and running the query. A stage still going reads `running`, and one whose timing is unavailable reads `-` - never `0s`.
+
+Comparing time queued against time running the query is how you tell a slow job from one that sat waiting for a worker.
+
+## All fields
+
+`--json` prints everything Studio holds about each job - the cluster it ran on and that cluster's UUID, the exit code and error message, workers, Python version, requirements, metrics. It changes the output format only, so the other flags still apply: add `--extended` for the stages, and `--status`, `--limit` and `--team` filter it as usual.
```
shcheklein: **Optional cleanup: explain JSON once, in terms of what the reader can do**

This paragraph and the two new Notes bullets repeat the same distinction, while “everything Studio holds” is broader than the response this command exposes. A shorter version would be:

> `--json` returns the full job details as JSON. Add `--extended` for stage timestamps; `--status`, `--limit`, and `--team` still apply.

Keep the example and the useful UUID-matching guidance below. The Notes addition can then be just the link to `job logs`. This keeps the new documentation focused without repeating the response fields or implementation rationale.

src/datachain/studio.py
```diff
@@ -814,32 +823,43 @@ def show_job_logs(job_id: str, team_name: str | None):
...
+                f"{cluster['busy_workers']}/{cluster['active_workers']}"
+                f"/{cluster['max_workers']}"
+            ),
+            "Is Default": cluster["default"],
         }
         for cluster in clusters
     ]
 
-    print(tabulate.tabulate(rows, headers="keys", tablefmt="grid"))
+    # missingval renders the unset fields, so a quota of 0 stays 0.
```
shcheklein: **Optional cleanup: remove the comments left over from the pricing implementation**

This comment still refers to a quota, but that column has now been removed. The JSON branch also says “what a cost calculation reads,” and the new `get_clusters()` docstring ends with “for anything pricing a cluster.” These are general listing/serialization operations; the cost limitation already has a home in the command docs.

Delete the stale quota/JSON comments, keep `get_clusters()` focused on the returned clusters and UUID relationship, and rename `CLUSTER_WITH_PRICING` to reflect cluster metadata. No new helper is needed.

shcheklein (commented):
Reviewed the full current diff again, including the removed skill, CLI/API contract, docs and tests. The shared StudioClient calls and existing CLI now cover the functionality without another fetcher or cache; I would keep that structure.

The remaining functional issue is the upgrade/uninstall path already reported at https://github.com/datachain-ai/datachain/pull/1994#discussion_r4051962547; I reproduced it and am replying there rather than opening a duplicate. I also left a compatibility comment on studio#13281 about existing clients reading the removed numeric id.

Two small cleanup suggestions are inline. Please also rewrite the PR description around the final change: its table still includes Job Quota, and the test plan still cites the deleted jobs-script tests and a zero-quota case. The review history can stay in the discussion.

Validation at 29d45a2: 101 existing targeted tests passed; five additional temporary probes passed for empty JSON, error exit/stdout behavior and combined filters; mypy passed on all seven changed Python files; strict docs build and diff whitespace check passed.

src/datachain/cli/commands/skill.py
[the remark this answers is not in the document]
shcheklein: Fixed in a06c4f3f. A `RETIRED_SKILLS` tuple names skills that shipped once and no longer do. Uninstall accepts those names and sweeps them on an unqualified run; install clears them too, so whichever command an upgrading user reaches for, the stale directory and its generated rule/command both go. They stay out of the installable set and out of `skill list`.

Thanks for reproducing it concretely — the seeded-install detail is what made the gap obvious, and the regression tests do exactly that: seed `.cursor/skills/jobs/SKILL.md` plus `.cursor/rules/datachain-jobs.mdc`, then run install-all, uninstall-all, and `uninstall jobs` as three cases. I checked all three fail against the previous commit and pass now; the fresh-install tests stay green and could not have caught this.

docs/commands/job/ls.md
shcheklein: Taken as written: "`--json` returns the full job details as JSON. Add `--extended` for stage timestamps; `--status`, `--limit` and `--team` still apply." The example and the UUID-matching paragraph stay, and the Notes bullet is now just the `job logs` link. You were right that "everything Studio holds" claimed more than the endpoint exposes.

src/datachain/studio.py
shcheklein: All removed in a06c4f3f — the missingval comment that still mentioned a quota, the "what a cost calculation reads" line on the JSON branch, and the pricing clause on `get_clusters()`, which now just says what it returns and that `uuid` is what a job's `compute_cluster_uuid` points at. `CLUSTER_WITH_PRICING` is now `CLUSTER`. No new helper.

shcheklein (commented):
Full review repeated from the base through 977a34e. No new blocking correctness issue found.

The retired-skill fix closes the previous finding. I exercised the base revision's installer followed by the current installer, uninstall-all, and explicit uninstall-jobs, across all five targets in both local and global mode: all 30 cases removed the old skill and its generated command/rule while preserving unrelated skills. `_remove_skill` is a useful reuse of the existing removal behavior; I would keep the present structure.

The CLI and companion API agree, including keeping numeric id in the response while displaying UUID in the new table. The remaining inline comments are low-priority documentation consistency points.

The PR description still needs the cleanup requested in the previous review: its table includes Job Quota, its test plan cites the deleted jobs-script suite and a zero-quota case, and the retained original reasoning says Studio stops returning id. Keep the body focused on the final behavior and current verification.

Validation: 105 targeted tests passed; five temporary JSON probes passed for empty/error responses and combined flags; all 30 upgrade cases passed; mypy passed on the seven changed Python files; strict docs build and diff whitespace check passed. All current DataChain CI checks are green.

src/datachain/cli/parser/skill.py
```diff
@@ -72,7 +72,7 @@ def add_skill_parser(subparsers, parent_parser) -> None:
         default=None,
         metavar="SKILLS",
         help=(
-            "Comma-separated skill names to uninstall: core, knowledge, jobs "
+            "Comma-separated skill names to uninstall: core, knowledge "
```
shcheklein: **[P3] Keep uninstall help and errors aligned with the accepted skill names**

`uninstall jobs` now works, but this help text still lists only `core, knowledge`. The invalid-name error in `uninstall_skills()` also builds its “Valid skills” list from `SKILLS`, although validation accepts `SKILLS + RETIRED_SKILLS`. Someone trying to remove an old installation therefore gets incomplete guidance.

Use the same removable-name set for uninstall validation, help and error text, including `jobs`. It can remain absent from install help and `skill list`.

docs/commands/job/clusters.md
```diff
@@ -5,20 +5,89 @@ List compute clusters in Studio.
...
+| `Region` | Where the cluster runs |
+| `Instance Type` | Worker machine type or family |
+| `Compute Class` | Worker class, such as `Performance` or `gpu` |
+| `Disk Request` | Requested temporary storage per worker (allocated capacity may differ) |
+| `Busy/Active/Max` | Workers running jobs / started / allowed |
+| `Is Default` | The cluster a job runs on when `--cluster` is omitted |
+
+A `-` means the cluster does not set that field. It never means zero.
+
+## All fields
```
shcheklein: **Optional documentation polish: label this as JSON output and mark the sample as abbreviated**

Studio now retains `id`, and the CLI deliberately preserves it in `--json`; I verified that against the companion schema. This sample omits it, so “All fields” no longer describes the example.

Rename the section to “JSON output” and introduce an abbreviated example. That keeps the deprecated field out of the new user guidance without implying that the displayed object is the complete response. No serialization change is needed.

src/datachain/cli/parser/skill.py
shcheklein: Fixed in cf91122a. Both now use the removable set: the help reads "core, knowledge, jobs" and `uninstall_skills()` builds its "Valid skills" list from `SKILLS + RETIRED_SKILLS`. Install help and `skill list` still show only the installable two, as you suggested.

Added `test_uninstall_error_names_the_retired_skill` so the message keeps naming it — the person who needs that guidance is exactly the one being told the name does not exist.

docs/commands/job/clusters.md
shcheklein: Both done — the label in cf91122a, the heading in 4bdb302a. It is "JSON output" now, and the intro says the example is abbreviated because the response also carries a deprecated `id` that new code should ignore in favour of `uuid`. You are right that "All fields" stopped being true the moment the sample dropped it.

Renamed the same section on the `job ls` page too, so the two pages match; that one is not abbreviated and does not claim to be.

src/datachain/cli/commands/skill.py
```diff
@@ -113,6 +116,63 @@ def _transform_copilot_instructions(skill_md_path: Path) -> str:
...
+    """
+    try:
+        frontmatter = (skill_dest / "SKILL.md").read_text()
+    except OSError:
+        return False
+    return bool(
+        re.search(
+            rf"^name:\s*datachain-{re.escape(name)}\s*$", frontmatter, re.MULTILINE
+        )
+    )
```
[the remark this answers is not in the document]
shcheklein: Right — and it was my own fix that opened it. I named the variable frontmatter but passed the whole file.

Reproduced: a third-party skill whose body shows our frontmatter in a fenced yaml block gets claimed and wiped by skill install. Worth noting an indented example escaped it only because ^name: will not match the indent — luck, not a rule.

Fixed in 273f549a. A _frontmatter helper takes the leading --- block and the check reads only that. Verified four cases end to end: our name in a fenced block, bare in the body, and a file with no frontmatter at all are all kept; a genuinely datachain-owned skill is still swept. Dropping the block parse fails the new regression test.

amritghimire (approved):

